PR #3, opened 2020-12-14
Enable opting-out of form submissions

seanpdoyle:
Similar to [Disabling Turbo Drive on Specific Links][turbo-drive-false],
guard intercepting form subissions with a similar check for
`[data-turbo-drive="false"]` on the submitter, `<form>` element, or an
ancestor.

Similarly, *do not intercept* when the submitter or form element
specifies a method that isn't an HTTP verb (via
[formmethod="dialog"][formmethod] attribute or [method="dialog"][method]
attribute, respectively).

In practice, HTTP submitting `<form>` elements will only ever declare
`method="get"` or `method="post"`. The one [exceptional case is
`method="dialog"`][method-dialog], which we check for as a special case.

Testing
---

Since we're covering behavior for the `<dialog>` element _and_ running
the suite in both Chrome and Firefox, the test suite needs to polyfill
support for the `<dialog>` element.

[turbo-drive-false]: https://turbo.hotwire.dev/handbook/drive#disabling-turbo-drive-on-specific-links
[formmethod]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formmethod
[method]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-method
[method-dialog]: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-2

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -121,7 +121,7 @@ export class Session implements NavigatorDelegate {
   // Link click observer delegate
 
   willFollowLinkToLocation(link: Element, location: Location) {
-    return this.linkIsVisitable(link)
+    return this.elementIsNavigable(link)
       && this.locationIsVisitable(location)
       && this.applicationAllowsFollowingLinkToLocation(link, location)
   }
@@ -151,8 +151,8 @@ export class Session implements NavigatorDelegate {
 
   // Form submit observer delegate
 
-  willSubmitForm(form: HTMLFormElement, submitter?: HTMLElement) {
-    return true
+  willSubmitForm(form: HTMLFormElement, submitter?: HTMLElement): boolean {
+    return this.elementIsNavigable(form) && this.elementIsNavigable(submitter)
   }
 
   formSubmitted(form: HTMLFormElement, submitter?: HTMLElement) {
@@ -246,8 +246,8 @@ export class Session implements NavigatorDelegate {
     return isAction(action) ? action : "advance"
   }
 
-  linkIsVisitable(link: Element) {
-    const container = link.closest("[data-turbo]")
+  elementIsNavigable(element?: Element) {
+    const container = element?.closest("[data-turbo]")
     if (container) {
       return container.getAttribute("data-turbo") != "false"
     } else {
--- a/src/tests/fixtures/form.html
+++ b/src/tests/fixtures/form.html
@@ -5,6 +5,21 @@
     <title>Form</title>
     <script src="/src/tests/fixtures/turbo.es2017-umd.js" data-turbo-track="reload"></script>
     <script src="/src/tests/fixtures/test.js"></script>
+    <script type="module">
+      import dialogPolyfill from "https://cdn.skypack.dev/dialog-polyfill"
+
+     addEventListener("click", ({ target }) => {
+        if (target instanceof HTMLElement) {
+          const id = target?.getAttribute("data-open")
+          const dialog = document.querySelector(`dialog#${id}`)
+
+          if (dialog && dialog.tagName == "DIALOG") {
+            dialogPolyfill.registerDialog(dialog)
+            dialog.showModal()
+          }
+        }
+      })
+    </script>
   </head>
   <body>
     <div id="standard">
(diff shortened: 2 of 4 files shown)


domchristie:
Would this be a good place to incorporate the changes from https://github.com/turbolinks/turbolinks/pull/495/files?

I've not manually tested Turbo Drive with non-fetch form submissions, but from the look of `Session#start`, the initial `replace` will still break the "Are you sure…?"  refresh behaviour after rendering a form submission: https://github.com/hotwired/turbo/blob/c52d87c9426ebbbb735e9e8947b3cbdfde8a2a10/src/core/drive/history.ts#L35

After a traditional POST, calling `history.replaceState` will replace the POST entry with a GET, and therefore refreshing won't trigger the confirm dialog. The fix was to store the initial restoration identifier as a property on the `History` instance rather than replacing the browser's history state.

I'd be happy to port these changes over if there's interest?

ChrisZou:
My I ask when would this be merged? 😂😂 Quite a few issues are depending on this to be fixed. It's preventing me to use hotwire to do so many things.

src/tests/fixtures/form.html
```diff
@@ -8,7 +8,17 @@
     <script type="module">
       import dialogPolyfill from "https://cdn.skypack.dev/dialog-polyfill"
 
-      window.dialogPolyfill = dialogPolyfill
+     addEventListener("click", ({ target }) => {
```
seanpdoyle, line 11: @sstephenson I was unable to bridge the gap between loading the dialog polyfill and executing arbitrary JS from the test TypeScript file.

I think this will pass the suite, but I'm open to suggestions on how to dynamically pull in the polyfill and apply it from TS.

src/core/session.ts
```diff
@@ -246,8 +249,8 @@ export class Session implements NavigatorDelegate {
     return isAction(action) ? action : "advance"
   }
 
-  linkIsVisitable(link: Element) {
-    const container = link.closest("[data-turbo]")
+  turboIsEnabled(element: Element) {
```
sstephenson, line 252: Let's call this `elementIsNavigable`.

sstephenson: If we change the signature to `element?: Element` we can avoid the `submitter` conditional expression in `willSubmitForm`.

src/core/session.ts
```diff
@@ -151,8 +152,10 @@ export class Session implements NavigatorDelegate {
 
   // Form submit observer delegate
 
-  willSubmitForm(form: HTMLFormElement, submitter?: HTMLElement) {
-    return true
+  willSubmitForm(form: HTMLFormElement, submitter?: HTMLElement): boolean {
+    const methodIsHttp = !!fetchMethodFromString(submitter?.getAttribute("formmethod") || form.method)
```
sstephenson, line 156: I wonder if this condition should just be moved into FormSubmitObserver? I think Turbo should probably just outright ignore any form with `formmethod="dialog"`.

sstephenson (changes requested):

src/core/session.ts
seanpdoyle: I've moved the HTTP method conditional up into the FormSubmitObserver.

> I think Turbo should probably just outright ignore submit events from any form with formmethod="dialog".

I think there is utility in passing through `submit` events. Consider an example:

```html
<dialog>
  <form method="post" action="/comments">
    <textarea name="body"></textarea>

    <button type="submit">Save</button>
    <button type="submit" formmethod="dialog">Discard</button>
  </form>
</dialog>
```

Despite a button containing an element with `[formmethod]`, we'd still want to submit the `<form>` if the `<button type="submit">Save</button>` is clicked.

sstephenson: Ah, yep, you're right 👍

What do you think about comparing the value against `"dialog"` instead of searching for it in the list of HTTP methods? Are there any other special methods we ought to account for?

seanpdoyle: I did my best to explain it in the commit message, but I could elaborate more:

> In practice, `<form>` elements will only ever declare method="get" or
method="post", but given the addition of method="dialog" and the
fact that the fetchMethodFromString and the FetchMethod enum type
exist, we might as well be careful and future-proof.

Does that feel pre-mature?

sstephenson: Yeah, I think it's OK to take [the form submission spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-2) at its word for now:

> 23. If the value of method is dialog then jump to the submit dialog steps.
>
> Otherwise, select the appropriate row in the table below based on the value of scheme as given by the first cell of each row. Then, select the appropriate cell on that row based on the value of method as given in the first cell of each column. Then, jump to the steps named in that cell and defined below the table.
> […]

seanpdoyle: Done!

sstephenson (approved):

jon-sully:
Just adding in case other folks come around, I believe contrary to the initial description of this PR, the necessary attribute on the `<form>` needs to be `data-turbo="false"`, not `data-turbo-drive="false"`. 

In Rails parlance, 

```ruby
form_with model: @xyz, url: etc, data: { turbo: false }
```

Which is great because `data: { 'turbo-drive' => false }` doesn't look nearly as slick 😜

PR #6, opened 2020-12-14
Proposal: <turbo-frame> elements progressively enhance target attribute

seanpdoyle:
Since `<turbo-frame>` elements are a spiritual success to `<iframe>`
elements, there is an opportunity to build atop a concept with existing
history and understanding.

Replace the `data-turbo-frame` attribute with the `target` attribute,
progressively enhancing both [`<form target="...">`][form-target] and
[`<a target="...">`][a-target].

Instead of introducing a new `data-turbo-frame` attribute, layer on
additional behavior to a well understood attribute that determines which
`<iframe>` element on the page to submit the `<form>` element or
navigate the `<a>` element within.

In both cases, the `_top` value is a special case value:

> the topmost browsing context (the "highest" context that’s an ancestor
> of the current one). If no ancestors, behaves as _self.

This commit also modifies the check for `[data-turbo-frame="top"]` to
instead be a check for `[target="_top"]`.

To set the `target` attribute on [an `<a>` constructed with
`link_to`][link_to], pass it directly as `target:`.

To set the `target` attribute on [a `<form>` constructed with
`form_with`][form_with], pass it as part of the `html: { target: ... }`
option.

To set the `target` attribute on [a `<form>` constructed with
`button_to`][button_to], pass it as part of the `form: { target: ... }`
option.

[link_to]: https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to
[form_with]: https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/FormHelper.html#method-i-form_with
[button_to]: https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-button_to

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -118,7 +118,7 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
   }
 
   private findFrameElement(element: Element) {
-    const id = element.getAttribute("data-turbo-frame")
+    const id = element.getAttribute("target")
     return getFrameElementById(id) ?? this.element
   }
 
@@ -184,7 +184,7 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
   }
 
   private shouldInterceptNavigation(element: Element) {
-    const id = element.getAttribute("data-turbo-frame") || this.element.getAttribute("target")
+    const id = element.getAttribute("target") || this.element.getAttribute("target")
 
     if (!this.enabled || id == "_top") {
       return false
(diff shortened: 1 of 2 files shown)


darylbarnes:
I like the idea of progressively enhancing the web platform in this way. We shouldn't create new attributes for concepts that already exist when using custom elements. Instead we progressively enhance the existing where possible with new/improved functionality.

inopinatus:
I don't think this is compatible with existing `target` attributes. They have a specific and existing meaning, and it's one _that has not been thrown away by Hotwire_. The usual behaviour may well be expected for most of the rest of any given page. Appropriating `target` for our own use seems to me like an intrusion.

PR #20, opened 2020-12-23
Add a `streamElement` property to the `turbo:before-stream-render` event

nbgoodall:
This adds a `streamElement` property to the detail of the `turbo:before-stream-render`. Feel free to do with it as you wish!

Personally I'm using this for intercepting the event to smoothly transition between the frames, e.g.:

```
document.addEventListener('turbo:before-stream-render', function(event) {
  event.preventDefault()

  /* Transition the old frame away ... */

  event.detail.streamElement.performAction()
})
```

Thank you for all your work! 🙏

### DIFF
--- a/src/elements/stream_element.ts
+++ b/src/elements/stream_element.ts
@@ -75,13 +75,7 @@ export class StreamElement extends HTMLElement {
   }
 
   private get beforeRenderEvent() {
-    let properties: CustomEventInit = { bubbles: true, cancelable: true }
-
-    if (this.action !== 'remove') {
-      properties.detail = { newTemplate: this.templateContent }
-    }
-
-    return new CustomEvent("turbo:before-stream-render", properties)
+    return new CustomEvent("turbo:before-stream-render", { bubbles: true, cancelable: true, detail: { streamElement: this }})
   }
 }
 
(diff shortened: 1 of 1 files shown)


panda-madness:
Wouldn't it be better to just pass in a reference to the entire `StreamElement`? That way we get access to every aspect of the stream action, not just the template

src/elements/stream_element.ts
```diff
@@ -75,7 +75,13 @@ export class StreamElement extends HTMLElement {
   }
 
   private get beforeRenderEvent() {
-    return new CustomEvent("turbo:before-stream-render", { bubbles: true, cancelable: true })
+    let properties: CustomEventInit = { bubbles: true, cancelable: true }
+
+    if (this.action !== 'remove') {
+      properties.detail = { newTemplate: this.templateContent }
```
panda-madness, line 81: Something like `properties.detail = { streamElement: this }` would be more flexible.

nbgoodall: Great idea, thank you!

sstephenson:
Thanks for the patch! You can already access the stream element using the `event.target` property, so I don't think we need to pass anything along in the details object 👍

pySilver:
Looks like `performAction` is no longer available in API

PR #28, opened 2020-12-23
Pausable rendering

domchristie:
Just like `turbo:before-visit`, this pull request makes the `turbo:before-render` event cancellable with `preventDefault`. Additionally it adds a `render` method to the event to let the render process resolve (using Promises). This effectively makes rendering pausable; useful for waiting for exit animations, and _could_ open the door to custom renderers e.g. DOM diffing. A basic example might look like:

```js
addEventListener('turbo:before-render', async function (event) {
  event.preventDefault()
  await animateOut()
  event.detail.render()
})
```

Here's a demo of some basic animation possibilities: https://laser-crawling-chanter.glitch.me/ ([View source](https://glitch.com/edit/#!/laser-crawling-chanter?path=turn.js%3A1%3A0))

This pull request also adds the Visit's `action` to the event detail, making it possible to change the behaviour depending on whether it's an `advance` or `restore`.

If there's interest, I could do similar features for streams/frames + add documentation.

There are a few old Turbolinks issues that make reference to something like this:
- turbolinks/turbolinks#68
- turbolinks/turbolinks#70
- turbolinks/turbolinks#153
- turbolinks/turbolinks#184
- turbolinks/turbolinks#293
- turbolinks/turbolinks#408

### DIFF
--- a/src/core/drive/renderer.ts
+++ b/src/core/drive/renderer.ts
@@ -1,7 +1,7 @@
 export type RenderCallback = () => void
 
 export interface RenderDelegate {
-  viewWillRender(newBody: HTMLBodyElement): void
+  applicationAllowsImmediateRendering(newBody: HTMLBodyElement, render:() => void): boolean
   viewRendered(newBody: HTMLBodyElement): void
   viewInvalidated(): void
 }
@@ -10,10 +10,15 @@ export abstract class Renderer {
   abstract delegate: RenderDelegate
   abstract newBody: HTMLBodyElement
 
-  renderView(callback: RenderCallback) {
-    this.delegate.viewWillRender(this.newBody)
-    callback()
-    this.delegate.viewRendered(this.newBody)
+  renderView(callback: RenderCallback):Promise<void> {
+    return new Promise<void>((resolve) => {
+      if (this.delegate.applicationAllowsImmediateRendering(this.newBody, resolve)) {
+        resolve()
+      }
+    }).then(() => {
+      callback()
+      this.delegate.viewRendered(this.newBody)
+    })
   }
 
   invalidateView() {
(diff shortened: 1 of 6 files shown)


src/core/drive/renderer.ts
```diff
@@ -10,10 +10,15 @@ export abstract class Renderer {
   abstract delegate: RenderDelegate
   abstract newBody: HTMLBodyElement
 
-  renderView(callback: RenderCallback) {
-    this.delegate.viewWillRender(this.newBody)
-    callback()
-    this.delegate.viewRendered(this.newBody)
+  renderView(callback: RenderCallback):Promise<void> {
+    return new Promise((resolve:() => void) => {
```
nbgoodall, line 14: @domchristie I believe this should be `return new Promise<void>((resolve) => {`, which would resolve the error you mentioned.

domchristie: Thanks! 🙏

rainerborene:
This is the only thing stopping me from using `turbo` at my company. Just waiting for @sstephenson bless on this PR :sweat_smile:

daviddeutsch:
Apologies for the additional bump, but this has been in waiting since January and would be an amazing addition.

weaverryan:
@domchristie i don’t know the status of who would merge things at the moment, but, to help, perhaps you could resolve the conflicts?

And, while we’re waiting, is there any decent workaround?

Thanks!

daviddeutsch:
Seems like this is orphaned after the recent changes at basecamp.

Pinging @dhh !

dhh:
Happy to see this carried forward, if someone else wants to pickup the PR.

kirillplatonov:
@dhh The feature looks great. I will prepare an updated PR for it.

PR #36, opened 2020-12-25
introduce eslint

43081j:
was having a read through the sources and figured this might be useful while i was in there.

if not, no worries. tried to align the config with whats in the source, too. to avoid code changes

does raise the question also: you have a fair amount of empty methods, a few which are never used outside tests too. might be worth figuring out if all those are necessary


43081j:
rebased and fixed new lint errors in case this ever gets picked up

src/elements/stream_element.ts
```diff
@@ -97,7 +98,11 @@ export class StreamElement extends HTMLElement {
   }
 
   private get targetElementsById() {
-    const element = this.ownerDocument?.getElementById(this.target!)
+    if (!this.target) {
+      return [];
+    }
+
+    const element = this.ownerDocument?.getElementById(this.target)
```
dhh, line 105: This seems like a regression in both readability and style. Is there a way to avoid this?

43081j: we could allow non-null assertions if its likely they'll be common in the codebase.

IMO we should avoid using them so there's less chance of someone forgetting to do a null check.

another option is to disallow them but add an ignore comment to this particular line and the one below. there's only two so its probably better to make those the exception rather than turn the rule off

dhh:
We now have a linter running.

PR #38, opened 2020-12-25
add jsdoc to exported components

43081j:
again, was trying to get an understanding of this code base so had been documenting my understanding.

here's some jsdoc if its of any use. if its something you'd like to merge just let me know any rewording you want etc.

i know this project doesn't aim to be consumed by JS much, but the element jsdoc especially is useful for tooling which hints at the available properties/attrs of custom elements.

on that note, you may want to change the stream element to have identical 'mirrored' properties to its attributes just like the frame element does. so it can be configured by attributes _and_ by properties (and can then be documented).

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -9,34 +9,88 @@ const session = new Session
 const { navigator } = session
 export { navigator }
 
+/**
+ * Starts the main session.
+ * This initialises any necessary observers such as those to monitor
+ * link interactions.
+ */
 export function start() {
   session.start()
 }
 
+/**
+ * Registers an adapter for the main session.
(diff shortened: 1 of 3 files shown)


ChrisZou (approved):

43081j:
bump.

this and #36 were done quite some time ago now. would be good if someone can get them merged or let me know a reason why not. the repo is looking pretty active so not sure how these two vanished under the radar

43081j:
rebased onto master in case you ever want this still

43081j:
rebased and updated according to pr feedback in case anyone still wants this (its been here a while now :D)

Issue #22, opened 2020-12-23
Error: Form responses must redirect to another location


cj:
Hi,

I am using https://viewcomponent.org/ and I am handling redirects for forms using stimulus doing the following:

```javascript
  async turboSubmit(event) {
    const { success } = event.detail

    if (success && this.redirectValue) {
      Turbo.visit(this.redirectValue)
    } else {
      this.enableSubmit()
    }
  }
```

But it currently throws this error `Error: Form responses must redirect to another location`, which is due to the following code https://github.com/hotwired/turbo/blob/aae160bed764ca7322be1e3a9e00366787d96f7f/src/core/drive/form_submission.ts#L103, is there a reason this is needed?

Cheers!

cj:
So I ended up creating a patch for now to get it working on my end https://github.com/cj/app-template/blob/main/patches/%40hotwired%2Bturbo%2B7.0.0-beta.1.patch, but it lead to what might be another bug with `Turbo.visit`:

![CleanShot 2020-12-22 at 20 51 06](https://user-images.githubusercontent.com/1819/102953676-79d9cf00-4497-11eb-938c-b16b87411474.gif)

This is where I am calling `Turbo.visit` https://github.com/cj/app-template/blob/main/app/components/form_component/base.js#L121

cj:
So I think ^ is a bug due to the fact I was using render plain https://github.com/cj/app-template/blob/main/app/controllers/test_controller.rb#L11.  When you hit the back button, the full html of `/login` is there but Turbo is not picking that up.

cj:
ah so the `Turbo.visit` is not a bug, it was because I had render plain https://github.com/cj/app-template/blob/main/app/controllers/test_controller.rb#L11!

danjac:
I'm running into this issue as well with simple form submissions that re-render the form (i.e. the HTML page) with errors when server validation fails.  Is there a reason for this behavior? Is it documented?

sstephenson:
We intend to allow HTTP 422 responses for form submissions inside frames, but it didn’t make the cut for the initial beta release.

For now, you can try sending a Turbo Streams response that re-renders your form with errors.

sstephenson:
Browsers record form submissions in history along with the verb. In a standard "postback" scenario, if you reload the page after submission, or navigate away and then back, you'll see a dialog box like this:

[...]

PR #39, opened 2020-12-26
Render 400-500 status response HTML

seanpdoyle:
Related to https://github.com/hotwired/turbo-rails/issues/12
Related to https://github.com/hotwired/turbo-rails/issues/34

Typically, Turbo expects each FormSubmission request to result in a
redirect to a new Location.

When a FormSubmission request fails with an HTTP Status code between
400-499 or 500-599 (e.g. [unprocessable entity][422]), render the
response HTML. This commit brings the same behavior to `<turbo-frame>`
elements.

[422]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -99,7 +99,9 @@ export class FormSubmission {
   }
 
   requestSucceededWithResponse(request: FetchRequest, response: FetchResponse) {
-    if (this.requestMustRedirect(request) && !response.redirected) {
+    if (response.clientError || response.serverError) {
+      this.delegate.formSubmissionFailedWithResponse(this, response)
+    } else if (this.requestMustRedirect(request) && !response.redirected) {
       const error = new Error("Form responses must redirect to another location")
       this.delegate.formSubmissionErrored(this, error)
     } else {
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -3,6 +3,7 @@ import { FetchResponse } from "../../http/fetch_response"
 import { FormSubmission } from "./form_submission"
 import { Locatable, Location } from "../location"
 import { Visit, VisitDelegate, VisitOptions } from "./visit"
+import { Snapshot } from "./snapshot"
 
 export type NavigatorDelegate = VisitDelegate & {
   allowsVisitingLocation(location: Location): boolean
@@ -87,8 +88,15 @@ export class Navigator {
     }
   }
 
-  formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
-    console.error("Form submission failed", formSubmission, fetchResponse)
+  async formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
+    const responseHTML = await fetchResponse.responseHTML
+
+    if (responseHTML) {
+      debugger
+      const snapshot = Snapshot.fromHTMLString(responseHTML)
+      this.view.render({ snapshot }, () => {})
+      this.view.clearSnapshotCache()
+    }
   }
 
   formSubmissionErrored(formSubmission: FormSubmission, error: Error) {
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -109,7 +109,7 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
   }
 
   formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
-
+    this.element.controller.loadResponse(fetchResponse)
   }
 
   formSubmissionErrored(formSubmission: FormSubmission, error: Error) {
--- a/src/http/fetch_response.ts
+++ b/src/http/fetch_response.ts
@@ -15,6 +15,14 @@ export class FetchResponse {
     return !this.succeeded
   }
 
+  get clientError() {
+    return this.statusCode >= 400 && this.statusCode <= 499
+  }
+
+  get serverError() {
+    return this.statusCode >= 500 && this.statusCode <= 599
+  }
+
   get redirected() {
     return this.response.redirected
   }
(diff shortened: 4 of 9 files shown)


src/core/drive/form_submission.ts
```diff
@@ -99,7 +99,9 @@ export class FormSubmission {
   }
 
   requestSucceededWithResponse(request: FetchRequest, response: FetchResponse) {
-    if (this.requestMustRedirect(request) && !response.redirected) {
+    if (response.unprocessableEntity) {
```
seanpdoyle, line 102: Is this the only special case status we want to support?

src/core/drive/navigator.ts
```diff
@@ -87,8 +87,12 @@ export class Navigator {
...
   }
 
-  formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
-    console.error("Form submission failed", formSubmission, fetchResponse)
+  async formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
+    const responseHTML = await fetchResponse.responseHTML
+
+    this.view.clearSnapshotCache()
+
+    this.proposeVisit(this.location, { response: { statusCode: fetchResponse.statusCode, responseHTML } })
```
seanpdoyle, line 95: Is clearing the Snapshot cache and storing the invalid response HTML at the current location a good idea here?

Intrepidd:
Note : I realised that form submits sent by turbo don't have the `Turbo-Frame`header. This header helps telling the backend this is a frame request. Rails for instance doesn't render the layout if this is a frame request. Should we send this header now that we accept a frame response back ?

src/core/drive/form_submission.ts
dylanjha: It might make sense to cover all `4xx` errors, not only `422`. 422 is pretty specific to rails scaffolding conventions where as `4xx` is more widely used for forms that have errors.

src/core/drive/form_submission.ts
```diff
@@ -99,7 +99,9 @@ export class FormSubmission {
...
+      this.delegate.formSubmissionFailedWithResponse(this, response)
+    } else if (this.requestMustRedirect(request) && !response.redirected) {
       const error = new Error("Form responses must redirect to another location")
```
santib, line 105: What is the technical reason why we need to have this error trigger when the response is not a redirect?

src/core/drive/form_submission.ts
edariedl: `422` is problematic also for Rails. Default `HTML` response for invalid form is `200`, `422` is usually used only for `JSON` responses. I know it can be changed, but support for default behaviour would be nicer :).

Example from rails scaffold generator:
```ruby
  # POST /rooms
  # POST /rooms.json
  def create
    @room = Room.new(room_params)

    respond_to do |format|
      if @room.save
        format.html { redirect_to @room, notice: 'Room was successfully created.' }
        format.json { render :show, status: :created, location: @room }
      else
        format.html { render :new }
        format.json { render json: @room.errors, status: :unprocessable_entity }
      end
    end
  end
```

src/core/drive/navigator.ts
seanpdoyle: @sstephenson is there an existing alternative to `this.proposeVisit` that can inject the `responseHTML` into the page?

src/core/drive/navigator.ts
```diff
@@ -87,8 +87,12 @@ export class Navigator {
...
   }
 
-  formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
-    console.error("Form submission failed", formSubmission, fetchResponse)
+  async formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
+    const responseHTML = await fetchResponse.responseHTML
+
+    this.view.clearSnapshotCache()
+
+    this.proposeVisit(this.location, { action: "replace", response: { statusCode: fetchResponse.statusCode, responseHTML } })
```
sstephenson, line 95: We don't actually want to go through the visit machinery here, since history isn't changing and we don't need the adapter to be involved. Instead we can just create a Snapshot and pass it to View#render:

```ts
const snapshot = Snapshot.fromHTMLString(responseHTML)
this.view.render({ snapshot }, () => {})
this.view.clearSnapshotCache()
```

(The second argument to render is a callback which Visit uses to change the scroll position afterwards. Ideally we would just scroll to the top of the page here.)

src/core/drive/form_submission.ts
sstephenson: https://github.com/hotwired/turbo/issues/22#issuecomment-750336078

sstephenson (changes requested):
Nice work. I agree with your choice to expand this to cover all 4xx and 5xx responses.

src/core/drive/form_submission.ts
sstephenson: > What is the technical reason why we need to have this error triggered when the response is not a redirect?

https://github.com/hotwired/turbo/issues/22#issuecomment-750336078

src/core/drive/navigator.ts
seanpdoyle: Done. Thank you for the guidance.

sstephenson (approved):

src/core/drive/navigator.ts
```diff
@@ -87,8 +88,15 @@ export class Navigator {
     }
   }
 
-  formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
-    console.error("Form submission failed", formSubmission, fetchResponse)
+  async formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
+    const responseHTML = await fetchResponse.responseHTML
+
+    if (responseHTML) {
+      debugger
```
javan, line 95: 👻

seanpdoyle: 🤠

seanpdoyle: https://github.com/hotwired/turbo/commit/18b755f6210a270ad6d9eedaf1195a474e82d9cb!

Intrepidd:
For long forms, page is not scrolled to the top after submit, do you think this should be fixed natively in turbo ?

PR #40, opened 2020-12-26
Unobtrusive JavaScript polyfill

seanpdoyle:
Closes https://github.com/hotwired/turbo-rails/issues/33.
Closes https://github.com/hotwired/turbo-rails/issues/37.

When loaded alongside [@rails/ujs][], start it via `Rails.start()`, and
then declare the [`turbo:`-prefixed events][turbo-events] that
correspond to their `ajax:`-prefixed counterparts.

[@rails/ujs]: https://github.com/rails/rails/tree/v6.1.0/actionview/app/assets/javascripts
[turbo-events]: https://turbo.hotwire.dev/reference/events


src/tests/fixtures/unobtrusive_javascript.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Unobtrustive JavaScript</title>
+    <script type="module">
+      import "/dist/turbo.es2017-umd.js"
```
seanpdoyle, line 7: This is currently limited to a Functional test HTML fixture file. I struggled to declare the code within this `<script>` tag in an ESM/UMD module, and then optionally import it. Unfortunately, I was unable to figure it out.

If anyone could provide guidance, I'd prefer to declare something like:

```js
export function startUjs(Rails: any) {
  const { delegate, disableElement, enableElement } = Rails

  delegate(document, Rails.linkDisableSelector,   "turbo:before-cache", enableElement)
  delegate(document, Rails.buttonDisableSelector, "turbo:before-cache", enableElement)
  delegate(document, Rails.buttonDisableSelector, "turbo:submit-end", enableElement)

  delegate(document, Rails.formSubmitSelector, "turbo:submit-start", disableElement)
  delegate(document, Rails.formSubmitSelector, "turbo:submit-end", enableElement)
  delegate(document, Rails.formSubmitSelector, "turbo:before-cache", enableElement)
}

// ...
import Rails from "..."
import { startUjs } from "@hotwire/turbo/ujs"

startUjs(Rails)
```

sstephenson:
Don't think this belongs in Turbo itself.

rainerborene:
@sstephenson Totally agree. `rails-ujs` is a past thing. For those who are interest, here's how you can accomplish almost the same thing using Stimulus and Hotwire: https://gist.github.com/rainerborene/644da1be59ecb4f6c4c9637abf2622e0

PR #49, opened 2020-12-28
Modify SubmitEvent polyfill to read `type` property

seanpdoyle:
Follow-up to https://github.com/hotwired/turbo/pull/1#discussion_r546728950

Replaces the `SubmitEvent` polyfill's `type` attribute access with
[`type property access][type].

[type]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement#Properties


seanpdoyle:
The original code for the polyfill was written before we had a functional test suite covering Form Submissions.

Now that we have a suite to cover this behavior, would it be possible to run it in CI against multiple browsers via a tool like Browser Stack? That way, we can have some confidence in the behavior applying correctly in polyfilled browser environments.

Issue #48, opened 2020-12-28
turbo_frame_tag on Table element


michelson:
Hi , I'm trying Turbo , simple crud, table list:

example code: 

```erb
<table>
  <thead>
    <tr>
      <th>Name</th>
      <th>Key</th>
      <th>User</th>
      <th colspan="3"></th>
    </tr>
  </thead>
  <tbody>
    <%= turbo_frame_tag "rooms" do %>
      <% @rooms.each do |room| %>
        <%= render room %>
      <% end %>
    <% end %>
  </tbody>
</table>
```

but the frame tag is rendered separate from the list , why ?

note the red line is the frame tag border style:
![Screen Cast 2020-12-28 at 2 06 33 AM](https://user-images.githubusercontent.com/11976/103190821-5c4ea000-48b1-11eb-813d-16fe0f294d84.gif)

michelson:
I can confirm that is working properly if I change the element to `<div>`

Intrepidd:
Table éléments only accept a set of elements as children. It's not valid html to put a div inside a table or a tr directly.

I've had the same problem, my solution was to fake a table by using divs and the correct display attributes such as table-row or table-cell. 


This is not ideal though, maybe we could have a way of telling turbo that the whole tr / tbody should be a frame via a data attribute or something, having forms inside tables that update live on change is a useful pattern so it's a shame that it's a hassle to implement with turbo.

inopinatus:
Only table elements are [welcome inside tables](https://html.spec.whatwg.org/multipage/tables.html).

You can use streams to target replacement of individual ids though.

michelson:
I would like to understand why this does not work on tbody. 

> I can confirm that is working properly if I change the element to <div>
 by that I meant that the turbo frame works properly on divs, I was not referring to replace the tbody with div tho

inopinatus:
`<turbo-frame>` is not a table element, so your browser hoists it out of the table to the nearest adjacency where it is permitted. So the above is invalid HTML.

If you want to replace the tbody, then instead give the tbody an id attribute, and use a turbo streams response to target that.

ayb:
I understand that HTML has constraints over what types of elements are allowed within tables, but the big picture idea here has merit.

i.e. it would be very useful for turbo to be able to replace table rows.

[...]

Issue #34, opened 2020-12-24
Is there a way to programmatically invoke a Turbo Stream?


panda-madness:
Ditto on the issue title. It would be nice to have a programmatic way to invoke a stream URL, same way we currently can trigger a visit with `Turbo.visit()`.

`Turbo.stream('/some/url/here').then(r => console.log('maybe even return a promise'))`.

thetutlage:
Yup, it will open up a room for lot of custom features. One is to have infinite scroll. I detect the page end using Javascript and then load more items

danjac:
You can use visit() to fetch a stream, unfortunately it pushes the URL to your history. I've found axios.get() does not work - Turbo doesn't detect the incoming stream. What about plain fetch() ?

danjac:
I haven't found any obvious way to trigger a stream using fetch(). Of course the server can return a valid stream, but Turbo just ignores it. The only way seems to be through a WebSocket, or by a form or link submission. You can try hacking a form submit with requestSubmit() (plain submit() won't work) but this feels hacky compared to a simple API call.

sstephenson:
There's no such thing as a "stream URL," but there are a few ways to work with Turbo Streams programmatically.

If you have a WebSocket or EventSource object, you can use `Turbo.connectStreamSource(object)` to process stream messages automatically as they come in. (You can actually pass any object which responds to `addEventListener("message", ...)` and dispatches `message` events.)

If you have a stream message in a string of HTML that you received from a fetch request, you can call `Turbo.renderStreamMessage(html)`. Or you can just manually insert the HTML into your document! [That's what Turbo does internally](https://github.com/hotwired/turbo/blob/18b755f6210a270ad6d9eedaf1195a474e82d9cb/src/core/session.ts#L85-L87).

zealot128:
For reference, @sstephenson hint about using Turbo.renderStreamMessage, I was able to use fetch together with Turbo to fetch outside of Turbo's realm and pipe the response to Turbo, which handled the containing turbo-stream actions without problem. Important to note, that you need to pass headers Accept to the fetch, so the controller can respond with a ``turbo_stream`` response format.

```javascript
 fetch(href, {
      headers: {
        Accept: "text/vnd.turbo-stream.html",
      },
    })
      .then(r => r.text())
      .then(html => Turbo.renderStreamMessage(html))
//  optionally reflect the url    .then(_ => history.replaceState(history.state, "", href))
```

tonywok:
[...]

PR #52, opened 2020-12-29
Restrict Accept: turbo-stream to Form Submissions

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/139

Since Turbo Stream responses over HTTP are in response to user
interactions, and the conventional way to trigger destructive user
interactions is through form submissions, this commit limits the
`StreamObserver` _only_ inject the [Accept][] header containing the
`text/html; turbo-stream` content type during form submissions.

Replace the [`turbo:before-fetch-request` event listener][turbo-event]
with a `turbo:submit-start` event listener.

From a Rails perspective, this enables an interesting pattern. Consider
the following controller:

```ruby
class MessagesController < ApplicationController
  def create
    @message = Message.new(params.require(:message).permit(:body))

    if @message.save
      redirect_to messages_url
    else
      render :new
    end
  end
end
```

By omitting a [respond_to][] block in the controller's action, the
controller offloads the rendering responsibility to the view layer,
enabling the presence or absence of a template to be the deciding factor
on the type of response:

```html+erb
<%# app/views/messages/new.html.erb %>
<%= turbo_frame_tag @message do %>
  <%= render partial: "form", locals: { message: @message } %>
<% end %>

<%# app/views/messages/new.turbo_stream.erb %>
<%= turbo_stream.replace dom_id(@message, :form) do %>
  <%= render partial: "form", locals: { message: @message }, formats: :html %>
<% end %>
```

Prior to this change, `GET` requests initiated by Turbo Drive
navigations were submitting requests with the content type, resulting in
requests to the `messages#new` being handled by the
`messages/new.turbo_stream.erb` template, instead of the fully-formed
HTML responses provided by the `messages/new.html.erb` template.

[Accept]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept
[turbo-event]: https://turbo.hotwire.dev/reference/events
[respond_to]: https://edgeapi.rubyonrails.org/classes/ActionController/MimeResponds.html#method-i-respond_to

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -2,6 +2,7 @@ import { FetchRequest, FetchMethod, fetchMethodFromString, FetchRequestHeaders }
 import { FetchResponse } from "../../http/fetch_response"
 import { expandURL } from "../url"
 import { dispatch } from "../../util"
+import { StreamMessage } from "../streams/stream_message"
 
 export interface FormSubmissionDelegate {
   formSubmissionStarted(formSubmission: FormSubmission): void
@@ -110,6 +111,7 @@ export class FormSubmission {
       if (token) {
         headers["X-CSRF-Token"] = token
       }
+      headers["Accept"] = StreamMessage.contentType
     }
     return headers
   }
--- a/src/core/streams/stream_message.ts
+++ b/src/core/streams/stream_message.ts
@@ -1,6 +1,7 @@
 import { StreamElement } from "../../elements/stream_element"
 
 export class StreamMessage {
+  static readonly contentType = "text/vnd.turbo-stream.html"
   readonly templateElement = document.createElement("template")
 
   static wrap(message: StreamMessage | string) {
--- a/src/observers/stream_observer.ts
+++ b/src/observers/stream_observer.ts
@@ -1,4 +1,3 @@
-import { FetchRequestOptions } from "../http/fetch_request"
 import { FetchResponse } from "../http/fetch_response"
 import { StreamMessage } from "../core/streams/stream_message"
 import { StreamSource } from "../core/types"
@@ -19,7 +18,6 @@ export class StreamObserver {
   start() {
     if (!this.started) {
       this.started = true
-      addEventListener("turbo:before-fetch-request", this.prepareFetchRequest, true)
       addEventListener("turbo:before-fetch-response", this.inspectFetchResponse, false)
     }
   }
@@ -27,7 +25,6 @@ export class StreamObserver {
   stop() {
     if (this.started) {
       this.started = false
-      removeEventListener("turbo:before-fetch-request", this.prepareFetchRequest, true)
       removeEventListener("turbo:before-fetch-response", this.inspectFetchResponse, false)
     }
   }
@@ -49,14 +46,6 @@ export class StreamObserver {
   streamSourceIsConnected(source: StreamSource) {
     return this.sources.has(source)
   }
-
-  prepareFetchRequest = <EventListener>((event: CustomEvent) => {
-    const fetchOptions: FetchRequestOptions = event.detail?.fetchOptions
-    if (fetchOptions) {
-      const { headers } = fetchOptions
-      headers.Accept = [ "text/vnd.turbo-stream.html", headers.Accept ].join(", ")
-    }
-  })
 
   inspectFetchResponse = <EventListener>((event: CustomEvent) => {
     const response = fetchResponseFromEvent(event)
@@ -93,5 +82,5 @@ function fetchResponseFromEvent(event: CustomEvent) {
 
 function fetchResponseIsStream(response: FetchResponse) {
   const contentType = response.contentType ?? ""
-  return /^text\/vnd\.turbo-stream\.html\b/.test(contentType)
+  return contentType.startsWith(StreamMessage.contentType)
 }


inopinatus:
There are plenty of reasons to send a turbo stream response to a GET. Replacing a <tbody> is already a case in point in these issue reports, and I can think of more besides, from menus to form errors to dashboard updates.

Observe also the request for programmatic access in https://github.com/hotwired/turbo/issues/34. I suspect there will be _more_ uses for turbo stream responses in the future, not fewer.

seanpdoyle:
> There are plenty of reasons to send a turbo stream response to a GET. 

This change set proposes that _navigational_ GET requests should omit the Turbo Steam content type from the headers. The `turbo:before-fetch-request` event listener does not apply to _all_ fetch requests. 

> Replacing a `<tbody>` is already a case in point in these issue reports, and I can think of more besides, from menus to form errors to dashboard updates.

I'm not familiar with the tbody example. Could you share a link? To me, these use cases seem to by good matches for GET-powered turbo frames.


> Observe also the request for programmatic access in https://github.com/hotwired/turbo/issues/34. I suspect there will be _more_ uses for turbo stream responses in the future, not fewer.

If there were a `Turbo.stream` programmatic interface introduced in the future, this changeset wouldn't prevent those requests from submitting with the Turbo Stream content type header.

inopinatus:
> This change set proposes that navigational GET requests should omit the Turbo Steam content type
A dry read through the diff seemed to indicate that turbo frame GETs would also be omitted? since they only deliver `turbo:before-fetch-request`. 

> I'm not familiar with the tbody example. Could you share a link?
https://github.com/hotwired/turbo/issues/48 here, but also repeated by others in private spaces, and it's a well-known issue besides when working with component-style custom elements.

> these use cases seem to by good matches for GET-powered turbo frames
In general, the opposite is so; autonomous custom elements are phrasing content, so they're not permitted in tables, or other contexts where they'd have obvious potential otherwise e.g. `<option>` lists, and don't enclose multiple areas of the page, thus making them doubly inappropriate, unless being misused as an ersatz programmatic stream fetch trigger.

This would also be instructing other folks how to go about their business, which some may consider anathematic.

Another possibility could be merging the turbo-frame element, or its constituent parts, into the prototypes of custom extensions to existing table elements, which the DOM should hopefully then permit.

> If there were a Turbo.stream programmatic interface introduced in the future, this changeset wouldn't prevent those requests from submitting with the Turbo Stream content type header.

I was rather expecting to simply fire a `turbo:before-fetch-request` from my own stuff to maintain a loose coupling, unless/until a Turbo.stream comes along, or possibly use a Hey.com-style controller<>frame whammy for programmatic visits.   With this change, I'd be looking instead at using an inappropriate `turbo:submit-start`, talking directly to Turbo internal APIs, or using a POST for things that are not actually POSTs.

If the intention is to identify Turbo Drive GETs contextually to shave a line or two of format negotiation from controller methods, then I'd definitely suggest circling around on the signalling mechanisms and consider how to elegantly achieve just that, rather than carving it as negative space from other parts.

kaspth (approved):
Looks good! Already reviewed the enctype changes, so this is pretty simple really.

src/core/drive/form_submission.ts
```diff
@@ -110,6 +110,7 @@ export class FormSubmission {
       if (token) {
         headers["X-CSRF-Token"] = token
       }
+      headers["Accept"] = "text/vnd.turbo-stream.html"
```
kaspth, line 113: Actually, now that this is no longer in the StreamObserver, do we want to read this from a constant like `Turbo.Stream.contentType` that we import?

seanpdoyle: @kaspth is there a way to transform a string constant into the StreamObserver's regular expression?

```js
return /^text\/vnd\.turbo-stream\.html\b/.test(contentType)
```

seanpdoyle: Is that now a `contentType.startsWith?(OUR_CONSTANT)` check, or is it more nuanced than that?

kaspth: Oof, I don't know, but maybe we could do:

```ruby
export default contentType() {
  return "text/vnd.turbo-stream.html"
}

export default contentTypeCheck() {
  return /^text\/vnd\.turbo-stream\.html\b/
}
```

from a single file that both could import. When they're together that way the duplication doesn't matter.

sstephenson: I think it'd be fine to switch to `String#startsWith()`. Not as precise as `/\b/` but I don't think it matters here.

Menduist:
Sorry to bump this, but if the goal is to remove the turbo-stream header from navigational GETs, shouldn't the forms with `method="get" action=""` include the turbo-stream accept, since they're not navigational per se?

Example use case:
```html
<form method="get" action="">
  <input type="search" name="s">
  <button type="submit">Search</button>
</form>
<ul id="results">[Whatever]</ul>
```

Server side (pseudocode) :
```
if headers["Accept"].contains("text/vnd.turbo-stream.html"):
  sendTurboStream(action="replace", id="results", value=computeResults())
else:
  sendFullPage(computeFullPage())
```
The form could be a POST, but then you wouldn't have the `?s=xx` in the URL, which is needed in some cases (eg, to "save the state")
And in the current version, with GET, the full page must always be sent, which is a bit wasteful

feliperaul:
@seanpdoyle @dhh Given the number of issues and questions regarding Turbo stop handling GET requests (see https://github.com/kaminari/kaminari/pull/1067#issuecomment-988054409), I think it would be great if this could be reconsidered.

I too find it immensely useful to have a link trigger a GET request for a `.turbo_stream.erb` response capable of aggregating multiple turbo_stream responses. This is the biggest feature missing from the good `.js.erb` days IMHO. 

What about if regular links included the turbo Accept header, but only if they were annotated with `data: { turbo_method: :get }`, now that turbo handles the `data-method` attribute?

dhh:
I'd be happy to see a way where we can force a GET to process turbo streams. I think the data-turbo-method=get is an interesting approach. Could also go with a specific one like data-turbo-stream=true. Thoughts @seanpdoyle?

feliperaul:
Happy that you agree that there should be a way of achieving this. 

My vote is for `data: { turbo_method: :get }`, because it keeps it consistent (it's the option I'd try intuitively without looking at any documentation) and avoids creating another data attribute that won't be present on the others `data-turbo-method`s that will accept turbo-streams.

cpanderson:
Further to using a specific one is there currently, or could there be, support for adding this to a link_to helper? Maybe with an easy option like "turbo_stream: true"? In this case setting data: {turbo_method: :get} on a request that is :get by default might make it less confusing. Or would this be outside the scope for a link_to helper?

feliperaul:
@cpanderson Yes, I think the `data: { turbo_method: :get }` could be used with `link_to`, `button_to`, or forms. 

@seanpdoyle Sean, first of all, thank you for the huge contribution on Turbo. I'm learning a lot from https://github.com/thoughtbot/hotwire-example-template (for the ones that don't know, there's a different branch for every tutorial there). 

Sorry to bump this, but did you have any time to consider this proposal? By the number of votes on DHH's comment above, it seems this would be a greatly appreciated addition!

seanpdoyle:
> Could also go with a specific one like `data-turbo-stream=true`

Is writing to the request headers something that applications could opt-into themselves using events that are already available?

```javascript
addEventListener("submit", ({ target: formElement, submitter }) => {
  const forcesTurboStream = submitter?.hasAttribute("data-turbo-stream") || target.hasAttribute("data-turbo-stream")

  if (forcesTurboStream) {
    formElement.addEventListener("turbo:submit-start", ({ detail: { formSubmission }) => {
     formSubmission.fetchRequest.headers["Accept"] = "text/vnd.turbo-stream.html"
    }, { once: true })
  }
})
```

bilogic:
> > Could also go with a specific one like `data-turbo-stream=true`
> 
> Is writing to the request headers something that applications could opt-into themselves using events that are already available?
> 
> ```js
> addEventListener("submit", ({ target: formElement, submitter }) => {
>   const forcesTurboStream = submitter?.hasAttribute("data-turbo-stream") || target.hasAttribute("data-turbo-stream")
> 
>   if (forcesTurboStream) {
>     formElement.addEventListener("turbo:submit-start", ({ detail: { formSubmission }) => {
>      formSubmission.fetchRequest.headers["Accept"] = "text/vnd.turbo-stream.html"
>     }, { once: true })
>   }
> })
> ```

Does this mean the above needs to be written for every form that needs it? I'm doing a `GET` `filter=search_term`

PR #53, opened 2020-12-29
Support Lazy-loading `<turbo-frame>` contents

seanpdoyle:
Add support for declaring `[loading="lazy"]` (in the style of
[`iframe[loading="lazy"]`][iframe-loading] and
[`img[loading="lazy"]`][img-loading]). When both `src` and
`loading="lazy"` are declared on the frame, navigation will be deferred
until the element is visible in the document.

As an example, consider a `<details>`/`<summary>` pairing:

```html
<details>
  <summary>Lazy load</summary>

  <turbo-frame id="frame" src="/lazy-loaded.html" loading="lazy"></turbo-frame>
</details>
```

On initial load, the frame's contents will be empty. When the
`<summary>` element is clicked and the contents are visible, the
`<turbo-frame>` element will navigate to the URL specified in the `src`
element.

By default, `<turbo-frame>` elements that omit the `loading` attribute
will be treated as if they've declared `loading="eager"`.

The attribute (and corresponding `TurboFrameElement.loading` property
accessors) provide a means of controlling the loading strategy
declaratively without additional application-side JavaScript, while
still providing a means for application-side JavaScript to intervene if
necessary.

If visibility-based intersection is insufficient for an application's
need, there are various ways to control the loading strategy.

For example, consider an optimization to pre-load the frame when the
`<details>` element's disclosure `<summary>` button is hovered or
focused. First, consider HTML annotated with Stimulus attributes:

```html
<details data-controller="lazy-loader">
  <summary data-action="hover->lazy-loader#prefetch focus->lazy-loader#prefetch">Lazy load</summary>

  <turbo-frame id="frame" src="/lazy-loaded.html" loading="lazy" data-lazy-loader-target="frame"></turbo-frame>
</details>
```

Next, consider the corresponding `lazy-loader` Stimulus controller:

```js
import { Controller } from "stimulus"

  static targets = [ "frame" ]

  prefetch() {
    this.frameTarget.loading = "eager"
  }
}
```

[img-loading]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Img#attr-loading
[iframe-loading]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-loading

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -19,24 +19,19 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
     this.element = element
     this.linkInterceptor = new LinkInterceptor(this, this.element)
     this.formInterceptor = new FormInterceptor(this, this.element)
-    this.intersectionObserver = new IntersectionObserver((entries) => {
-      const { src } = this.element
-
-      if (src) {
-        entries.forEach(({ isIntersecting }) => {
-          if (isIntersecting) {
-            this.element.loaded = this.visit(src)
-            this.unobserveIntersections()
-          }
-        })
-      }
+    this.intersectionObserver = new IntersectionObserver(entries => {
+      entries.forEach(({ isIntersecting }) => {
+        if (isIntersecting) {
+          this.loadSourceURL()
+        }
+      })
     })
   }
 
   connect() {
     this.linkInterceptor.start()
     this.formInterceptor.start()
-    if (this.element.loading == FrameLoadingStyle.lazy) {
+    if (this.loadingStyle == FrameLoadingStyle.lazy) {
       this.observeIntersections()
     }
   }
@@ -45,6 +40,28 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
     this.linkInterceptor.stop()
     this.formInterceptor.stop()
     this.unobserveIntersections()
+  }
+
+  sourceURLChanged() {
+    if (this.loadingStyle == FrameLoadingStyle.eager) {
+      this.loadSourceURL()
+    }
+  }
+
+  loadingStyleChanged() {
+    if (this.loadingStyle == FrameLoadingStyle.lazy) {
+      this.observeIntersections()
+    } else {
+      this.unobserveIntersections()
+      this.loadSourceURL()
+    }
+  }
+
+  loadSourceURL() {
+    if (this.isActive && this.sourceURL) {
+      this.element.loaded = this.visit(this.sourceURL)
+      this.unobserveIntersections()
+    }
   }
 
   shouldInterceptLinkClick(element: Element, url: string) {
@@ -244,6 +261,18 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
 
   get enabled() {
     return !this.element.disabled
+  }
+
+  get sourceURL() {
+    return this.element.src
+  }
+
+  get loadingStyle() {
+    return this.element.loading
+  }
+
+  get isActive() {
+    return this.element.isActive
   }
 }
 
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -64,6 +64,21 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
     }
   }
 
+  async visit(url: Locatable) {
+    const location = Location.wrap(url)
+    const request = new FetchRequest(this, FetchMethod.get, location)
+
+    return new Promise<void>(resolve => {
+      this.resolveVisitPromise = () => {
+        this.resolveVisitPromise = () => {}
+        resolve()
+      }
+      request.perform()
+    })
+  }
+
+  // Link interceptor delegate
+
   shouldInterceptLinkClick(element: Element, url: string) {
     return this.shouldInterceptNavigation(element)
   }
@@ -71,6 +86,8 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
   linkClickIntercepted(element: Element, url: string) {
     this.navigateFrame(element, url)
   }
+
+  // Form interceptor delegate
 
   shouldInterceptFormSubmission(element: HTMLFormElement) {
     return this.shouldInterceptNavigation(element)
@@ -89,18 +106,7 @@ export class FrameController implements FetchRequestDelegate, FormInterceptorDel
     }
   }
 
-  async visit(url: Locatable) {
-    const location = Location.wrap(url)
-    const request = new FetchRequest(this, FetchMethod.get, location)
-
-    return new Promise<void>(resolve => {
-      this.resolveVisitPromise = () => {
-        this.resolveVisitPromise = () => {}
-        resolve()
-      }
-      request.perform()
-    })
-  }
+  // Fetch request delegate
 
   additionalHeadersForRequest(request: FetchRequest) {
     return { "Turbo-Frame": this.id }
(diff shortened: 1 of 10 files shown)


src/tests/functional/navigation_tests.ts
```diff
@@ -115,6 +115,24 @@ export class NavigationTests extends TurboDriveTestCase {
...
   }
+
+  async "test lazy loading within a details element"() {
+    await this.nextBeat
+
+    const frameContents = "#loading-lazy turbo-frame h2"
+    this.assert.notOk(await this.hasSelector(frameContents))
+
+    await this.clickSelector("#loading-lazy summary")
+    await this.nextBeat
```
seanpdoyle, line 126: @sstephenson this test is currently failing. I suspect the "next beat" after the button click waits for the `<details>` element to expand, instead of waiting for the `<turbo-frame>` contents to get loaded.

Is that a correct assumption? When I test this out locally, the frame lazily loads as expected. I think the test failure has to do with the test harness and assertions.

inopinatus:
I very much like how hey.com manages the lazy loading for details/summary-driven dropdowns, especially the graceful leaning on browser baseline behaviour through interception of activation events on elements with a `data-turbo-frame` attribute. It's so effortlessly elegant, that I'd already assumed Hey's method is the gold standard.

More generally; by allowing an external controller to mediate the decision, turbo frames can participate in any variant of lazy-loading UX, rather than just this one example, and this might be desirable.

swanson:
Just wanted to share some "prior art" that it looks like GitHub is using. You can see the same pattern for the `Author`, `Label`, `Milestone`, etc filters on the `Issues` index page.

https://github.com/github/details-menu-element#deferred-loading
https://github.com/github/include-fragment-element/

One detail that I thought was very nice: the usage of `hover` on the details disclosure to trigger the server fetch to shave off a few milliseconds of the "Loading" state being visible. I know there are similar patterns for using hover in "Turbo-esque" setups (e.g. https://instant.page).

inopinatus:
@swanson yep, there's more than one kinda of lazy loading. I think turbo frames is better off providing one of the puzzle pieces, rather than trying to supply an entire UI component behaviour.

In terms of UI latency, I have pure CSS drop-downs that use focus-within in preference to details/summary (although I like both approaches, and don't discount `<dialog>`, or even an old-school checkbox hack), because it triggers on mousedown rather than up. Preloading the content on _hover_ is a terrific idea for progressive enhancement, thanks for highlighting it!

The particular use of intersectionObserver, which I think is the cool part of this PR, would (if extracted) make a neat _documentation example_ of how turbo frames can be wired up for lazy loads.  To me that'd feel more open for extension, and more inviting of novelty, than hard-wiring one behaviour and saying "you have two modes to choose from".

seanpdoyle:
> To me that'd feel more open for extension, and more inviting of novelty, than hard-wiring one behaviour and saying "you have two modes to choose from".

The two values ("eager" and "lazy") reflect the possible loading modes that [ elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img) and [<iframe> elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#Attributes) do/will support. The intention behind supporting them as turbo-frame attributes was to provide a general case, declarative way to defer the HTTP request. By default, the current behavior is to eager load all frames immediately.

This diff will require some changes to support it, but one improvement would be to monitor changes to the `loading` attribute. That way, consumer applications could declare elements with loading="lazy" initially, and the listen for whatever events are appropriate in their JS (e.g. a hover on a `<summary>` element, an intersection observer of their own, etc), and then set the attribute to loading="eager" to trigger the load at that time.

Does that sound flexible and extensible enough to serve as a puzzle piece?

swanson:
Some way to programmatically trigger (likely from a Stimulus controller) the server fetch to happen would be great. I'm not sure I would personally want visibility-based automatic triggering -- it felt kind of backwards to me: content that is below the fold is what I want to lazy load, but I don't want to wait until it becomes visible; I want it to load as soon as possible.

inopinatus:
> Does that sound flexible and extensible enough to serve as a puzzle piece?

Not really - it sounds like the kind of tight coupling that Turbo and Stimulus are trying to avoid. Maybe it's my Ruby roots showing, but I don't think one component should be interfering with the attributes of another in order to force behaviour.

Most of the Hotwire ecosystem signals their collaborators via events, and leans on existing browser-level capabilities besides, and nowhere is this more elegantly expressed that how hey.com relies on _overhearing a toggle_, and then _sending a simple click_, to drive the lazy-loading of frame content.

Without events for loose coupling, it becomes much harder to scale up, e.g. for other components listening to the same.

Turbo frames doesn't seem to have been pitched as a replacement/alternative for iframes, so I think any comparison is a problematic equivalence. From an application perspective, they have different applicability, granularity, and capability; the custom element itself does not extend HTMLIFrameElement, doesn't try to implement any of the other iframe behaviours, and when we want an iframe we can still have one.  Given a guiding principle of Hotwire being "for the HTML we already have", grafting specific UI behaviours from other elements into page-level infrastructure seems like crossing a line.

(It doesn't help that the [spec for lazy loading](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#lazy-loading-attributes) is still a sea of red caveats and XXX/FIXME markings)

In other words, this still sounds to me like behaviour best driven by events emitted from a Stimulus controller. Hopefully, one that is then composable/re-usable for _other_ event-driven uses of intersectionObserver.

kylekeesling:
Perhaps I'm not fully understanding some of the other points made above, but I can see a lot of value in this PR as proposed. I have a number of screens in my app that present lists that allow users to drill down (via detail/summary tags) into information if they want/need, but it doesn't make sense to pre-render all of that content because in all likelihood most of that data may never actually need be presented to the user.

Having a concise way to only load the content upon expansion on the elements could very useful in improving the initial loads of these pages.

dixpac:
For me this is a great addition.
Before `turbo-frame`, we had stimulus controller (`content_loader`) that had basically same functionality.

Sure, we can implement this without `loading: lazy`, but it will require stimulus controller + more bloat on the HTML side.
This addition will make those "use-cases" beautifully simple.

jonathanbruno:
This is a good one. as @dixpac said it could be done with `content_loader`, but we need to handle some layout stuff to make it work properly, having this feature out-of-the-box on turbo-frame tags would be awesome.

dixpac:
From my perspective this PR buys Porsche 911 for the price of VW Golf. I hope it will get merged, since it will clean our codebase a lot. But, I may be subjective 🤷🏼

src/core/frames/frame_controller.ts
```diff
@@ -165,6 +159,19 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
...
   // Private
 
+  private async visit(url: Locatable) {
+    const location = Location.wrap(url)
+    const request = new FetchRequest(this, FetchMethod.get, location)
+
+    return new Promise<void>(resolve => {
+      this.resolveVisitPromise = () => {
+        this.resolveVisitPromise = () => {}
+        resolve()
```
seanpdoyle, line 169: I've seen this pattern crop up across several projects. What's makes this nesting beneficial?

sstephenson (approved):
Wonderful start on this, @seanpdoyle. 👏 I went ahead and made some changes:

* To match `` behavior, preserve the `loading` attribute when loading is complete.
* To match `HTMLImageElement#complete`, expose a `FrameElement#complete` property.
* Keep loading logic out of `FrameElement` by delegating attribute changes to the controller.
* Pull the intersection observer machinery out into `AppearanceObserver`.

firedev:
How do I mark a lazy loaded frame as "stale" so it reloads the next time it is displayed?

Issue #42, opened 2020-12-26
Turbo Drive breaks skip links


cwgw:
Navigating to same-page fragments results in a new visit to the current page and fails to set the sequential focus navigation starting point to the target fragment.

```html
<a href="#main-content">Skip to main content</a>
```

The above should allow users to move the focus starting point to the element with id `main-content`. But, after Turbo Drive steps in and reloads the page content, the focus starting point is instead reset to the top of the page.

You can see this in action at [turbo.hotwire.dev](https://turbo.hotwire.dev). Try tabbing through the page using the skip link and you'll end up in a loop.

I think this behavior is another result of turbolinks/turbolinks#75. I know you can use `data-turbo="false"` to opt-out, and that changing the behavior may be deceptively complex (esp. when accounting for native implementations), but the current behavior seems out of step as Turbo seems to embrace an "it just works" experience.

---

The following workaround from [this comment](https://github.com/turbolinks/turbolinks/issues/75#issuecomment-445325162) is good enough for me in the meantime:

```js
document.addEventListener('turbolinks:click', function (event) {
  var anchorElement = event.target
  var isSamePageAnchor = (
    anchorElement.hash &&
    anchorElement.origin === window.location.origin &&
    anchorElement.pathname === window.location.pathname
  )
  
  if (isSamePageAnchor) {
    Turbolinks.controller.pushHistoryWithLocationAndRestorationIdentifier(
      event.data.url,
      Turbolinks.uuid()
    )
    event.preventDefault()
  }
})
```

...but without access to something like `pushHistoryWithLocationAndRestorationIdentifier`, I'm not sure how to keep my history in order. Is there a Turbo Drive analogue to the method? Or an approved method of manually adding to the session history?

seanpdoyle:
@cwgw I've opened https://github.com/hotwired/turbo/pull/57 to add test coverage to ensure the correct behavior, but I'm unable to reproduce the issue.

Could you comment on that PR to help me craft the correct HTML and test coverage to fail the test?

cwgw:
@seanpdoyle Thanks for your response. I'm happy to assist with test coverage.

Just to clarify a bit: The issue is not with window location or scroll position but with focus navigation. The [html spec](https://html.spec.whatwg.org/multipage/interaction.html#sequential-focus-navigation:sequential-focus-navigation-starting-point) says that…

> User agents are required to set the sequential focus navigation starting point to the target element when navigating to a fragment.

[...]

PR #57, opened 2020-12-29
Skip links: focus and scroll restoration

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/42

Navigate with a skip link within our Functional test suite, assert that
the correct element is scrolled to, the page's Location path and hash
are correct, and that the initial tab stop occurs after the skipped-to
content.

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -84,7 +84,7 @@ export class Visit implements FetchRequestDelegate {
     this.location = location
     this.isSamePage = (
       getAnchor(location) != null &&
-      getRequestURL(location) === getRequestURL(new URL(window.location.href))
+      getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation)
     )
 
     this.restorationIdentifier = restorationIdentifier || uuid()
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -310,8 +310,9 @@ export class Visit implements FetchRequestDelegate {
   }
 
   scrollToAnchor() {
-    if (getAnchor(this.location) != null) {
-      this.view.scrollToAnchor(getAnchor(this.location))
+    const anchor = getAnchor(this.location)
+    if (anchor != null) {
+      this.view.scrollToAnchor(anchor)
       return true
     }
   }
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -2,7 +2,7 @@ import { Adapter } from "../native/adapter"
 import { FetchMethod, FetchRequest, FetchRequestDelegate } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { History } from "./history"
-import { getAnchor } from "../url"
+import { getAnchor, getRequestURL } from "../url"
 import { PageSnapshot } from "./page_snapshot"
 import { Action } from "../types"
 import { uuid } from "../../util"
@@ -70,6 +70,7 @@ export class Visit implements FetchRequestDelegate {
   frame?: number
   historyChanged = false
   location: URL
+  isSamePage: boolean
   redirectedToLocation?: URL
   request?: FetchRequest
   response?: VisitResponse
@@ -81,6 +82,11 @@ export class Visit implements FetchRequestDelegate {
   constructor(delegate: VisitDelegate, location: URL, restorationIdentifier: string | undefined, options: Partial<VisitOptions> = {}) {
     this.delegate = delegate
     this.location = location
+    this.isSamePage = (
+      getAnchor(location) != null &&
+      getRequestURL(location) === getRequestURL(new URL(window.location.href))
+    )
+
     this.restorationIdentifier = restorationIdentifier || uuid()
 
     const { action, historyChanged, referrer, snapshotHTML, response } = { ...defaultOptions, ...options }
@@ -234,10 +240,14 @@ export class Visit implements FetchRequestDelegate {
       const isPreview = this.shouldIssueRequest()
       this.render(async () => {
         this.cacheSnapshot()
-        await this.view.renderPage(snapshot, isPreview)
-        this.adapter.visitRendered(this)
-        if (!isPreview) {
-          this.complete()
+        if (this.isSamePage) {
+          this.adapter.visitRendered(this)
+        } else {
+          await this.view.renderPage(snapshot, isPreview)
+          this.adapter.visitRendered(this)
+          if (!isPreview) {
+            this.complete()
+          }
         }
       })
     }
@@ -248,6 +258,15 @@ export class Visit implements FetchRequestDelegate {
       this.location = this.redirectedToLocation
       this.history.replace(this.redirectedToLocation, this.restorationIdentifier)
       this.followedRedirect = true
+    }
+  }
+
+  goToSamePageAnchor() {
+    if (this.isSamePage) {
+      this.render(async () => {
+        this.cacheSnapshot()
+        this.adapter.visitRendered(this)
+      })
     }
   }
 
@@ -346,9 +365,13 @@ export class Visit implements FetchRequestDelegate {
   }
 
   shouldIssueRequest() {
-    return this.action == "restore"
-      ? !this.hasCachedSnapshot()
-      : true
+    if (this.action == "restore") {
+      return !this.hasCachedSnapshot()
+    } else if (this.isSamePage) {
+      return false
+    } else {
+      return true
+    }
   }
 
   cacheSnapshot() {
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -2,7 +2,7 @@ import { Adapter } from "../native/adapter"
 import { FetchMethod, FetchRequest, FetchRequestDelegate } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { History } from "./history"
-import { getAnchor, getRequestURL } from "../url"
+import { getAnchor } from "../url"
 import { PageSnapshot } from "./page_snapshot"
 import { Action } from "../types"
 import { uuid } from "../../util"
@@ -39,7 +39,8 @@ export type VisitOptions = {
   historyChanged: boolean,
   referrer?: URL,
   snapshotHTML?: string,
-  response?: VisitResponse
+  response?: VisitResponse,
+  restorationIdentifier?: string,
 }
 
 const defaultOptions: VisitOptions = {
@@ -70,7 +71,6 @@ export class Visit implements FetchRequestDelegate {
   frame?: number
   historyChanged = false
   location: URL
-  isSamePage: boolean
   redirectedToLocation?: URL
   request?: FetchRequest
   response?: VisitResponse
@@ -82,11 +82,6 @@ export class Visit implements FetchRequestDelegate {
   constructor(delegate: VisitDelegate, location: URL, restorationIdentifier: string | undefined, options: Partial<VisitOptions> = {}) {
     this.delegate = delegate
     this.location = location
-    this.isSamePage = (
-      getAnchor(location) != null &&
-      getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation)
-    )
-
     this.restorationIdentifier = restorationIdentifier || uuid()
 
     const { action, historyChanged, referrer, snapshotHTML, response } = { ...defaultOptions, ...options }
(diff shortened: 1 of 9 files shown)


cwgw:
I think your HTML is fine for covering #42. But for more true-to-life markup, you'd want have stuff between the skip link and the target element. Maybe move `id="main"` to an element after the navigation:

```html
<a href="#main">Skip Link</a>

<section>
  <h1>Navigation</h1>
  <p><a id="same-origin-unannotated-link" href="/src/tests/fixtures/one.html">Same-origin unannotated link</a></p>
  <p><a id="same-origin-replace-link" href="/src/tests/fixtures/one.html" data-turbo-action="replace">Same-origin data-turbo-action=replace link</a></p>
  <p><a id="same-origin-false-link" href="/src/tests/fixtures/one.html" data-turbo="false">Same-origin data-turbo=false link</a></p>
  <p data-turbo="false"><a id="same-origin-unannotated-link-inside-false-container" href="/src/tests/fixtures/one.html">Same-origin unannotated link inside data-turbo=false container</a></p>
  <p data-turbo="false"><a id="same-origin-true-link-inside-false-container" href="/src/tests/fixtures/one.html" data-turbo="true">Same-origin data-turbo=true link inside data-turbo=false container</a></p>
  <p><a id="same-origin-anchored-link" href="/src/tests/fixtures/one.html#element-id">Same-origin anchored link</a></p>
  <p><a id="same-origin-anchored-link-named" href="/src/tests/fixtures/one.html#named-anchor">Same-origin link to named anchor</a></p>
  <p><a id="cross-origin-unannotated-link" href="about:blank">Cross-origin unannotated link</a></p>
  <p><a id="same-origin-targeted-link" href="/src/tests/fixtures/one.html" target="_blank">Same-origin targeted link</a></p>
  <p><a id="same-origin-download-link" href="/src/tests/fixtures/one.html" download="one.html">Same-origin download link</a></p>
  <svg width="600" height="100" viewbox="-300 -50 600 100"><text><a id="same-origin-link-inside-svg-element" href="/src/tests/fixtures/one.html">Same-origin link inside SVG element</a></text></svg>
  <svg width="600" height="100" viewbox="-300 -50 600 100"><text><a id="cross-origin-link-inside-svg-element" href="about:blank">Cross-origin link inside SVG element</a></text></svg>
</section>

<main id="main">
  <a href="about:blank">A focusable element within #main</a>
</main>
```

src/tests/functional/navigation_tests.ts
```diff
@@ -116,6 +116,22 @@ export class NavigationTests extends TurboDriveTestCase {
...
+    await this.nextBeat
+
+    this.assert.equal(await this.pathname, "/src/tests/fixtures/navigation.html")
+    this.assert.equal(await this.hash, "#main")
+    this.assert.ok(await this.isScrolledToSelector("#main"))
+
+    await this.remote.pressKeys('\uE004') // TAB
+    const activeElement = await this.remote.getActiveElement()
+    const skippedLink = await this.querySelector("#ignored-link")
+    const firstLinkWithinMain = await this.querySelector("#main a:first-of-type")
```
seanpdoyle, line 131: @sstephenson I was hoping that https://github.com/hotwired/turbo/commit/5525c21288edc222de0a05c72538aa7f898abd0c would resolve this issue, but it hasn't.

sstephenson:
I performed the test steps manually and things seem to behave as expected. When I tab to the skip link, press return, and then tab again, the "Same-origin unannotated link" has focus. But the test case doesn't pass for me.

src/tests/functional/navigation_tests.ts
```diff
@@ -116,6 +116,22 @@ export class NavigationTests extends TurboDriveTestCase {
...
+
+  async "test skip link with hash-only path"() {
+    await this.clickSelector('a[href="#main"]')
+    await this.nextBeat
+
+    this.assert.equal(await this.pathname, "/src/tests/fixtures/navigation.html")
+    this.assert.equal(await this.hash, "#main")
+    this.assert.ok(await this.isScrolledToSelector("#main"))
+
+    await this.remote.pressKeys('\uE004') // TAB
```
seanpdoyle, line 128: @sstephenson maybe its due to this line? I'm not 100% clear on how to reach into leadfoot or whatever is driving Intern to send the proper key. If this _is_ the correct character, maybe it's a timing issue?

cwgw: You can import a `keys` object from `@theintern/leadfoot` and pass `keys.TAB` to `pressKeys`.  But, I don't think the effect is any different from passing the string `'\uE004'`.

I found theintern/intern#1038, and [this comment](https://github.com/theintern/leadfoot/issues/110#issuecomment-315122224) suggests using `Element.type()` as a workaround. I'm unclear if this problem/solution is relevant for elements beyond inputs and textareas, but apparently there are at least some known issues with `pressKeys`.

seanpdoyle: @cwgw Thank you for the suggestion! I've made that change with some level of success.

@sstephenson the suite is now passing in Chrome, but the test continues to fail in Firefox.

domchristie: I think the tests will fail in Firefox for a couple of reasons:
1. The page is not long enough to be scrollable. In Chrome this doesn't appear to be an issue, so perhaps the method of measuring the "scroll position" varies. This is fixed by https://github.com/hotwired/turbo/pull/125/commits/cfbbf1413b67999e258b9507ff7c233e9fda7b3e
2. Tabbing to links in Firefox does not work unless System Preferences -> Keyboard -> Shortcuts -> All controls is checked. (I have tested this locally and it seems to work)

seanpdoyle: @domchristie extending the frame seems to have fixed my testing issues! Thank you for that suggestion!

src/tests/functional/navigation_tests.ts
```diff
@@ -116,6 +116,22 @@ export class NavigationTests extends TurboDriveTestCase {
     this.assert.equal(await this.pathname, "/src/tests/fixtures/one.html")
     this.assert.equal(await this.visitAction, "restore")
   }
+
+  async "test skip link with hash-only path"() {
+    await this.clickSelector('a[href="#main"]')
+    await this.nextBeat
```
domchristie, line 122: Do we need `this.nextBeat`? I'm thinking we shouldn't be reloading the page so I don't think this is needed.

src/tests/functional/navigation_tests.ts
seanpdoyle: It passes locally, but fails in CI. I wonder if there's a setting we can pass to `intern.json`:

https://github.com/hotwired/turbo/blob/bf3c3b0ad56249fa6a3eec3dd59f5d84ec897513/intern.json#L15-L17

src/tests/functional/navigation_tests.ts
domchristie: Hmm. Removing it doesn't cause failures as it should. Is there a way to test that a page _hasn't_ reloaded. e.g. checking that a Turbo event hasn't fired?

src/tests/functional/navigation_tests.ts
```diff
@@ -116,6 +125,33 @@ export class NavigationTests extends TurboDriveTestCase {
     this.assert.equal(await this.pathname, "/src/tests/fixtures/one.html")
     this.assert.equal(await this.visitAction, "restore")
   }
+
+  async "test skip link with hash-only path"() {
+    this.listenForVisits()
+    await this.clickSelector('a[href="#main"]')
+
+    this.assert.notOk(this.turboVisited)
```
seanpdoyle, line 133: @domchristie I've added some assertions to make sure that turbo events are not firing.

domchristie: Nice! An alternative approach, based on https://github.com/hotwired/turbo/blob/bf3c3b0ad56249fa6a3eec3dd59f5d84ec897513/src/tests/helpers/turbo_drive_test_case.ts#L45

```js
  async "test skip link with hash-only path"() {
    const bodyElementId = (await this.body).elementId
    await this.clickSelector('a[href="#main"]')

    await this.nextBeat
    this.assert.equal(await this.pathname, "/src/tests/fixtures/navigation.html")
    this.assert.equal(await this.hash, "#main")
    this.assert.equal((await this.body).elementId, bodyElementId, "does not reload page")
    this.assert.ok(await this.isScrolledToSelector("#main"))

    await this.pressTab()
    const activeElement = await this.remote.getActiveElement()
    const skippedLink = await this.querySelector("#ignored-link")
    const firstLinkWithinMain = await this.querySelector("#main a:first-of-type")
    this.assert.notOk(await activeElement.equals(skippedLink), "skips interactive elements before #main")
    this.assert.ok(await activeElement.equals(firstLinkWithinMain), "skips to first interactive element after #main")
  }
```

domchristie:
https://github.com/hotwired/turbo/pull/125 has now been updated with the latest updates (apart from the visit check). Tests pass locally with the [macOS "All controls" setting](https://github.com/hotwired/turbo/pull/57#discussion_r564129204). Let me know what you reckon.

seanpdoyle:
Thanks @domchristie. I think the latest commit is a step in the wrong direction, so I'm going to revert it so that your PR can be merged into this one.

domchristie:
@seanpdoyle 👍 _just_ discovered one bug—navigating back. `isSamePage` is incorrectly `true`, I think because it's checking against a location that's _too current_. I'll see if I can work out a fix. Apologies!

seanpdoyle:
This is passing on the "push" event and failing on the "pull_request" event. I'm not sure the difference.

src/core/drive/visit.ts
```diff
@@ -251,6 +262,15 @@ export class Visit implements FetchRequestDelegate {
     }
   }
 
+  goToSamePageAnchor() {
```
domchristie, line 265: Given that scrolling is now handled by `render`, I wonder if this should be renamed? It's still needed, but it just bypasses page rendering.

src/core/session.ts
```diff
@@ -105,9 +105,9 @@ export class Session implements HistoryDelegate, LinkClickObserverDelegate, Navi
 
   // History delegate
 
-  historyPoppedToLocationWithRestorationIdentifier(location: URL) {
+  historyPoppedToLocationWithRestorationIdentifier(location: URL, restorationIdentifier: string) {
```
domchristie, line 108: 🎉

src/tests/functional/navigation_tests.ts
```diff
@@ -116,6 +116,39 @@ export class NavigationTests extends TurboDriveTestCase {
...
+    this.assert.notOk(await this.changedBody, "does not reload page")
+    this.assert.ok(scrolledToMain)
+  }
+
+  async "test skip link with hash-only path moves focus and changes tab order"() {
+    await this.clickSelector('a[href="#main"]')
+
+    await this.pressTab()
+    const activeElement = await this.getActiveElement()
+    const equalsSkippedLink = await activeElement.equals(await this.querySelector("#ignored-link"))
```
domchristie, line 135: Would it be _so_ bad to add `tabindex=-1` on `#main` and check that it's the active element?

src/core/drive/visit.ts
seanpdoyle: I think the Snapshot re-structuring will yield many opportunities to refactor this changeset. I'm mostly interested in restoring a passing test suite so that we have a safety net for those changes.

sstephenson:
After rebasing against main, I see this failure in Firefox on macOS:

```
× firefox on mac 20.3.0 - NavigationTests - skip link with hash-only path moves focus and changes tab order (0.331s)
    AssertionError: skips to first interactive element after #main: expected false to be truthy
      at NavigationTests.test skip link with hash-only path moves focus and changes tab order @ src/tests/functional/navigation_tests.ts:162:16
      at runMicrotasks @ anonymous
      at processTicksAndRejections @ internal/process/task_queues.js:93:5
```

src/tests/functional/navigation_tests.ts
```diff
@@ -144,6 +144,37 @@ export class NavigationTests extends TurboDriveTestCase {
 
     this.assert.equal(await this.pathname, "/src/tests/fixtures/frames/hello.html")
   }
+
+  async "test skip link with hash-only path scrolls to the anchor without a visit"() {
+    await this.clickSelector('a[href="#main"]')
+    await this.nextBody
```
domchristie, line 150: I think this asserts that the link click creates a visit (and therefore [changes the body](https://github.com/hotwired/turbo/blob/8386e599769d238d8b578453d344c94e3e28da3c/src/tests/helpers/turbo_drive_test_case.ts#L56)).

https://github.com/hotwired/turbo/pull/57/commits/d6eeead682bbf6700ec3b120ec284c97bf7d0df9 tests that clicking a same-page anchor does not reload the page.

PR #58, opened 2020-12-30
Tidy location getters

domchristie:
This PR tidies `Location` getters and will make it easier to port: https://github.com/turbolinks/turbolinks/pull/285. It converts the `getX…` methods to proper getters, and delegates `origin` and `path` to the `<a>` element (as suggested by @javan: https://github.com/turbolinks/turbolinks/pull/270#issuecomment-307450179)

The anchor getter is a little more interesting. It's a workaround for the (inconsistent) native behaviour of `link.hash`. The native behaviour is as follows:

```js
link.href = '#foo'
link.hash // '#foo'

link.href = '#'
link.hash // ''

link.href = 'http://example.com'
link.hash // ''
```

This makes it impossible to distinguish whether a location has an empty hash or no hash at all.

To fix this, the `anchor` is just the text part of the hash or `undefined` if the URL does not contain a hash at all:

```js
link.href = '#foo'
location.anchor // 'foo'

link.href = '#'
location.anchor // ''

link.href = 'http://example.com'
location.anchor // undefined
```

### DIFF
--- a/src/core/location.ts
+++ b/src/core/location.ts
@@ -15,47 +15,61 @@ export class Location {
     }
   }
 
-  readonly absoluteURL: string
-  readonly requestURL: string
-  readonly anchor?: string
+  readonly link: HTMLAnchorElement = document.createElement("a")
+  private _requestURL?: string
+  private _anchor?: string
 
   constructor(url: string) {
-    const linkWithAnchor = document.createElement("a")
-    linkWithAnchor.href = url
+    this.link.href = url
+  }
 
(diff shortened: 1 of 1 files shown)


sstephenson:
Closing this in favor of #90. We want to move away from our Location class to the [DOM's URL interface](https://developer.mozilla.org/en-US/docs/Web/API/URL).

domchristie:
Any chance of getting the aforementioned hash fix in #90? /cc @seanpdoyle

PR #60, opened 2020-12-30
CI

javan:
Run the [Intern](https://theintern.io) test suite on CI with Chrome and Firefox. [Sample run](https://github.com/hotwired/turbo/runs/1626521295?check_suite_focus=true#step:7:1):

<details>
<summary>Chrome</summary>

</details>

<details><summary>Firefox</summary>

</details>


src/tests/functional/navigation_tests.ts
```diff
@@ -72,6 +72,7 @@ export class NavigationTests extends TurboDriveTestCase {
 
   async "test following a same-origin [target] link"() {
     this.clickSelector("#same-origin-targeted-link")
+    await this.nextBeat
```
seanpdoyle, line 75: Could we omit this is we moved the `await` to the `this.clickSelector` call?

intern.json
```diff
@@ -1,5 +1,4 @@
 {
-  "tunnel": "null",
```
seanpdoyle, line 2: I believe this change broke development environment test running. Was there a corresponding change to `yarn test`?

src/tests/functional/navigation_tests.ts
javan: This "fixed" [a failure in Firefox on CI](https://github.com/hotwired/turbo/runs/1626385513?check_suite_focus=true#step:7:83) that I couldn't reproduce locally. I don't think we need to add a delay to every click, and I suspect it has more to do with switching window handles https://github.com/hotwired/turbo/blob/cd09a3be3c85d9cdaae20954193cd0e4f75320e9/src/tests/functional/navigation_tests.ts#L76

intern.json
seanpdoyle: When I've reverted this line and ran it locally, the suite threw an error:

```
Tunnel started
(ノಠ益ಠ)ノ彡┻━┻
SessionNotCreatedException: [POST http://localhost:4444/wd/hub/session / {"desiredCapabilities":{"name":"intern","idle-timeout":60,"browserName":"firefox","moz:firefoxOptions":{"args":["-headless"]},"browser":"firefox"}}] Unable to create new service: GeckoDriverService
```

What needs to be ran to install the `GeckoDriverService` service?

Should we mention [Geckodriver](https://github.com/mozilla/geckodriver) in the README?

javan: Ha, this fixed development testing for me, which wasn't working because I didn't have selenium server running (I think). Can you try stopping any selenium / webdriver services you have running?

seanpdoyle: I've stopped all services and re-ran:

```
❯ yarn test
yarn run v1.22.10
Debugger listening on ws://127.0.0.1:9229/283d0b5c-6f24-4e32-ba84-39d53131fc10
✨  Done in 0.93s.
```

That's suspiciously fast 🤔.

javan: try:
```sh
$ rm -rf node_modules && yarn cache clean && yarn install && yarn build
$ yarn test
```

PR #63, opened 2020-12-31
Fix default refresh behaviour

domchristie:
[Setting `history.scrollRestoration = "manual"`](https://github.com/hotwired/turbo/blob/18b755f6210a270ad6d9eedaf1195a474e82d9cb/src/core/drive/history.ts#L31) fixes the scrolling issues mentioned in https://github.com/turbolinks/turbolinks/issues/181 and https://github.com/turbolinks/turbolinks/issues/241, however it alters the browser's refresh behaviour. When reloading a page, the scroll position is usually maintained, returning the user to their previous place (this behaviour caused the issue in https://github.com/turbolinks/turbolinks/issues/181). With `history.scrollRestoration = "manual"`, a refresh _always_ resets the scroll position to the top.

This pull request aims to fix that. It captures the scroll position just before the page is reloaded (using `onbeforeunload`), and persists it in session storage along with the current location. After a refresh, the scroll position is restored. When a view is invalidated, either by tracked asset change or a `visit-control` directive, a flag is stored in session storage. This ensures the scroll position is _not_ restored, and the user is navigated to the top as expected.

The downside of this change is that a Hard Reload, or Empty Cache + Hard Reload, won't scroll the user to the top (as is typical), rather, they'll be returned to their previous position. Hard reloads do not clear session storage, and since I don't think we can detect them, I'm not sure we can overcome this.

### DIFF
--- a/src/core/drive/history.ts
+++ b/src/core/drive/history.ts
@@ -27,8 +27,6 @@ export class History {
 
   start() {
     if (!this.started) {
-      this.previousScrollRestoration = history.scrollRestoration
-      history.scrollRestoration = "manual"
       addEventListener("popstate", this.onPopState, false)
       addEventListener("load", this.onPageLoad, false)
       this.started = true
@@ -89,6 +87,8 @@ export class History {
 
   onPageLoad = async (event: Event) => {
     await nextMicrotask()
+    this.previousScrollRestoration = history.scrollRestoration
+    history.scrollRestoration = "manual"
     this.pageLoaded = true
   }
 
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -10,12 +10,16 @@ import { PageObserver } from "../observers/page_observer"
 import { ScrollObserver } from "../observers/scroll_observer"
 import { StreamMessage } from "./streams/stream_message"
 import { StreamObserver } from "../observers/stream_observer"
-import { Action, Position, StreamSource, isAction } from "./types"
+import { SessionStorage } from "./session_storage"
+import { Action, Position, PersistedPosition, StreamSource, isAction } from "./types"
 import { dispatch } from "../util"
 import { View } from "./drive/view"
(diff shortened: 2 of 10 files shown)


domchristie:
@javan I spotted https://github.com/hotwired/turbo-site/commit/659c9c040b5b48f20983c24d47b6ceb939360992 and, inspired by that change, have updated this pull request to handle that case: https://github.com/hotwired/turbo/pull/63/commits/200c950735fd8e3c5013d4c8b974582470930e5e. A [test case for cold loads](https://github.com/hotwired/turbo/pull/63/commits/e2a21535525b4249a681d6f357201aef00d28232#diff-09ef6e1ec72f1b0adfe47be4c043247dccd43af06282b2102cdd3e4bda5241e9R120) has also been added.

sstephenson (commented):
This is a pretty heavy solution for what I've worked around in 5525c21288edc222de0a05c72538aa7f898abd0c—turns out that setting `scrollRestoration` at the right time fixes the problem without any other machinery. Thank you for the proposal, though, and please do open another PR if our test cases don't overlap 😊

domchristie:
@sstephenson yeh, I wasn't happy with the heaviness of the approach, and when I saw https://github.com/hotwired/turbo-site/commit/659c9c040b5b48f20983c24d47b6ceb939360992, I was hoping that [a timely setting/unsetting of the `scrollRestoration` mode](https://github.com/hotwired/turbo/commit/200c950735fd8e3c5013d4c8b974582470930e5e) would provide an elegant solution. Unfortunately the scroll position is still maintained when a page is invalidated, the main cause of: https://github.com/turbolinks/turbolinks/issues/181

Here's a diff with a failing test: https://github.com/hotwired/turbo/compare/main...domchristie:test_scroll_position_after_invalidation
Happy to PR it if interested?

PR #66, opened 2021-01-03
Ensure Turbo does not interfere with IFrames

seanpdoyle:
Move `[data-turbo="false"]` guard up to observers
===

Since the `FormSubmitObserver` and `LinkClickObserver` classes are
very interested in the page's HTML, extract the `elementIsNavigable`
checks from the `Session` into our utility module for re-use amongst the
observers.

This enables the `Session` to act with more confidence that any
interaction on the page is to be Turbo-enabled.

Ensure Turbo does not interfere with IFrames
===

Skip intercepting `<a>` element clicks or `<form>` element submissions
when they target an `<iframe>`. This occurs when an `<a>` declares a
[target][a-target], when a `<form>` declares a [target][form-target], or
a `<form>` element's submitting `<button>` element declares a
[formtarget][formtarget].

[formtarget]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formtarget

### DIFF
--- a/src/observers/link_click_observer.ts
+++ b/src/observers/link_click_observer.ts
@@ -1,4 +1,5 @@
 import { expandURL } from "../core/url"
+import { elementIsNavigable } from "../core/session"
 
 export interface LinkClickObserverDelegate {
   willFollowLinkToLocation(link: Element, location: URL): boolean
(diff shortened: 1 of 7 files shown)


src/observers/link_click_observer.ts
```diff
@@ -35,7 +36,7 @@ export class LinkClickObserver {
   clickBubbled = (event: MouseEvent) => {
     if (this.clickEventIsSignificant(event)) {
       const link = this.findLinkFromClickTarget(event.target)
-      if (link) {
+      if (link && !linkTargetsIframe(link) && elementIsNavigable(link)) {
```
shadowtime2000, line 39: If you want you could instead do:
```ts
const link = (this.findLinkFromClickTarget(event.target) as Element);
if (!linkTargetsIframe(link) && elementIsNavigable(link)) {
```

Issue #64, opened 2021-01-01
`data-turbo-permanent` doesn't work for frame or stream renders


stgm:
I'm creating a separate issue for this because @sstephenson mentioned this and it would be great to see this work in the future.

In my case, I have a sidebar where users can switch sidebar content using a turbo frame—but the content is not correctly reinstated when navigating back and forth using turbo drive.

_Originally posted by @sstephenson in https://github.com/hotwired/turbo/issues/21#issuecomment-750340131_

`data-turbo-permanent` has not been removed, it just doesn't (currently) work for frame or stream renders.

stgm:
Here's a simplified version of my code as an example:

```html
<div id="sidebar-content" data-turbo-permanent>
    <turbo-frame id="sidebar-frame" target="_top">
        <%= button_to '<', schedule_prev_path, form: { data: { 'turbo-frame' => 'sidebar-frame' } } %>
        <%= button_to '>', schedule_prev_path, form: { data: { 'turbo-frame' => 'sidebar-frame' } } %>
        ... page navigation links ...
    </turbo-frame>
</div>
```

stgm:
My `data-turbo-permanent` problem appears to be unrelated to `turbo-frame` after all, see #80

stgm:
@sstephenson I think this one was also resolved? I created the issue based on your comment, but I do not actually have this use case after all.

PR #71, opened 2021-01-04
Persist data-turbo-permanent element within frames

seanpdoyle:
Related to https://github.com/hotwired/turbo/issues/64

Introduce the `FrameRenderer` (as a descendant of the `Renderer`
interface, inspired by the `SnapshotRenderer` implementation) to serve
as an abstraction for the behavior the `FrameController` was previously
responsible for.

Extract `turbo-frame` content extraction from the `FrameController`
class and combine it with code copied from `SnapshotRenderer`
responsible for cloning permanent elements into the new HTML.

It's likely that there is a concept awaiting extraction, or a less
duplicative way of re-using the permanent element extraction logic, but
this is a start.

Note, this commit does not resolve the `<turbo-stream>` element
rendering shortcomings, but might blaze a trail for a `StreamRenderer`
implementation.


acetinick:
@seanpdoyle is this PR intended to improve the ability to persist content between turbo requests within a <turbo-frame> tag?

I am looking to find a way to resolve a situation where I am looking to build a live filter and have the following code.  Everything works fine, except when the turbo response comes back the text field loses focus, and looking for a way to persist the <div> section using a data-turbo-permanent attribute.

```
<turbo-frame>
  <div data-turbo-permanent>
      <%= form_for @criteria,  as: :criteria,  method: :get, url: locations_url,  data: { controller: "search" } do |f| %>
            <%= f.search_field :query, placeholder: "Search...",  data: { action: "debounced:input->search#trigger" } %>
      <% end %>
  </div>

  <table>
    <% for result in @results>
    <tr><td><%= result.name %></td></tr>
    <% end %>
  </table>

</turbo-frame>
```

seanpdoyle:
@acetinick this diff intends to make that behavior possible. It's also worth noting that any element that declares `data-turbo-permanent` _also_ has an `id` attribute.

You could modify the structure of your code example and make use of `data-turbo-target` to move the search outside of the `<turbo-frame>` to make that behavior possible before this change ships:

```html+erb
<%= form_for @criteria,  as: :criteria,  method: :get, url: locations_url,  id: "search-criteria", data: { controller: "search", turbo_frame: "search", turbo_permanent: true } do |f| %>
  <%= f.search_field :query, placeholder: "Search...",  data: { action: "debounced:input->search#trigger" } %>
<% end %>

<turbo-frame id="search">
  <table>
    <% for result in @results>
    <tr><td><%= result.name %></td></tr>
    <% end %>
  </table>

</turbo-frame>
```

src/core/renderer.ts
```diff
@@ -62,3 +66,55 @@ function copyElementAttributes(destinationElement: Element, sourceElement: Eleme
     destinationElement.setAttribute(name, value)
   }
 }
+
+function createPlaceholderForPermanentElement(permanentElement: PermanentElement) {
```
seanpdoyle, line 70: @sstephenson I've merged your latest into this branch, and followed it up by extracting some Placeholder rendering functions to the Renderer class module, then re-used them in the PageRenderer and FrameRenderer.

To help sort out what was shared and what was specific, I've also added `private` function modifiers throughout the `PageRenderer` class.

Are private and exported functions in a Class idiomatic to the Turbo codebase, or would a `renderer/util` module be more appropriate? A class of its own?

yunggindigo:
> @seanpdoyle is this PR intended to improve the ability to persist elements between turbo requests within a < turbo-frame > tag?
> 
> I am looking to find a way to resolve a situation where I am looking to build a live filter and have the following code. Everything works fine, except when the turbo response comes back the text field loses focus, and looking for a way to persist the < div > section using a data-turbo-permanent attribute.
> 
> ```
> <turbo-frame>
>   <div data-turbo-permanent>
>       <%= form_for @criteria,  as: :criteria,  method: :get, url: locations_url,  data: { controller: "search" } do |f| %>
>             <%= f.search_field :query, placeholder: "Search...",  data: { action: "debounced:input->search#trigger" } %>
>       <% end %>
>   </div>
> 
>   <table>
>     <% for result in @results>
>     <tr><td><%= result.name %></td></tr>
>     <% end %>
>   </table>
> 
> </turbo-frame>
> ```

Yes this made u be able to have a permanent element inside frames

PR #72, opened 2021-01-04
Fix form refresh behaviour when data-turbo=false.

domchristie:
Ported from https://github.com/turbolinks/turbolinks/pull/495

Turbo's initialisation currently breaks the default refresh behaviour when it's run after a POST submission and `data-turbo=false`. A POST followed by a refresh should trigger an "Are you sure…?" confirm dialog, whereas with Turbo installed, it just reloads the existing location with a GET.

When starting up, Turbo stores the initial location and restoration identifier in the session history by calling replaceState. This standardises the way to retrieve a restoration identifier on any popstate event. However, according the to the spec:

        [If the state push flag is not set], update the current entry in browsingContext's session history so that … it represents a GET request, if it currently represents a non-GET request (e.g. it was the result of a POST submission).
        https://html.spec.whatwg.org/multipage/history.html#history-1

So when a page loads after a POST, Turbo replaces the current entry with a GET to the same location, and therefore will not re-POST or display a confirm dialog upon refresh\*.

This pull request tweaks initialisation to store the initial restoration identifier and location as properties on the History instance, avoiding the need to use replaceState. When the history is popped to the initial entry, event.state will be null, and by asserting that the existing location matches the stored initial location, the restoration identifier is retrieved accordingly. Pops to non-initial entries will be handled as before.

A note on testing this programmatically: in Chrome, a POST, followed by `window.location.reload()` does not trigger an "Are you sure…?" confirm dialog, which makes this rather hard/impossible to test :/

---

\* This is observed in Chrome and Firefox, which comply with the spec. Safari behaves as we want, but does not comply with the spec.

### DIFF
--- a/src/core/drive/history.ts
+++ b/src/core/drive/history.ts
@@ -15,7 +15,9 @@ export type RestorationDataMap = { [restorationIdentifier: string]: RestorationD
 export class History {
   readonly delegate: HistoryDelegate
   location!: Location
+  initialLocation?: Location
   restorationIdentifier = uuid()
+  initialRestorationIdentifier?: string
   restorationData: RestorationDataMap = {}
   started = false
   pageLoaded = false
@@ -31,8 +33,9 @@ export class History {
       history.scrollRestoration = "manual"
       addEventListener("popstate", this.onPopState, false)
       addEventListener("load", this.onPageLoad, false)
+      this.initialLocation = Location.currentLocation
+      this.initialRestorationIdentifier = this.restorationIdentifier
       this.started = true
-      this.replace(Location.currentLocation)
     }
   }
 
@@ -41,6 +44,8 @@ export class History {
       history.scrollRestoration = this.previousScrollRestoration ?? "auto"
       removeEventListener("popstate", this.onPopState, false)
       removeEventListener("load", this.onPageLoad, false)
+      delete this.initialLocation
+      delete this.initialRestorationIdentifier
       this.started = false
     }
   }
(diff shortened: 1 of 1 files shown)


dhh:
Dom, should we try to sneak this into 7.0 final?

domchristie:
@dhh I've fixed the conflicts, but not retested yet

dhh:
Yeah, seems like there are some failing tests on this one.

dhh:
We really have to do something about these unreliable tests!

domchristie:
There were some big changes to how locations were wrapped after this PR was created, so I'm wondering if I'm missing something here.

The premise is reasonably simple:
1. Don't replace the history on start, instead store the initial restoration token as a property on the History object
2. Detect when a user has navigated Back to the initial location and use that restoration token
🤔

dhh:
I haven't looked enough into this. I only remember having a conversation with Sam at some point where he didn't think this was fixable. But maybe we just didn't consider it from the right angle.

domchristie:
FWIW, I had [a working version with Turbolinks](https://github.com/turbolinks/turbolinks/pull/495/files). If I have some time, I'll retry this approach from a blank slate

domchristie:
This should be working now. The failing test is in Chrome which demonstrates:

> A note on testing this programmatically: in Chrome, a `POST`, followed by `window.location.reload()` does not trigger an "Are you sure…?" confirm dialog, which makes this rather hard/impossible to test :/

I can revert the [test commit](https://github.com/domchristie/turbo/commit/d96d2ed8df4ad253bd6a299722e2fb7cbca83190) if preferred.

dhh:
This is great. I'm thinking we just add a skip to that test when the driver is chrome?

thomasbrus:
Seems to be  an issue still 🤔  Can reproduce it as follows:

- Comment out importing turbo
- Go to a form and submit it with empty fields (422 response)
- Paste this into the JS console: `history.replaceState({ foo: 123 }, '', location.href)`
- Then cmd+r will do a GET instead of a POST /  form resubmission. 

When Turbo is loaded, the same happens since it calls replaceState as well: https://github.com/hotwired/turbo/blob/v8.0.10/src/core/drive/history.js#L45. Setting Turbo.session.drive = false doesn't affect this since the replaceState call happens immediately when importing Turbo.

brunoprietog:
Hi @domchristie, would you mind rebasing this? Thanks!

domchristie:
Hi @brunoprietog, I'm not sure when I'll have the availability to look into this. Feel free to fork. Thanks!

PR #79, opened 2021-01-05
Scroll to top after rejected form submission

seanpdoyle:
After a form submission results in a 4xx-5xx response, scroll the window
to the top of the page, since that's what a browser does by default.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -97,6 +97,7 @@ export class Navigator {
     if (responseHTML) {
       const snapshot = PageSnapshot.fromHTMLString(responseHTML)
       await this.view.renderPage(snapshot)
+      window.scroll(0, 0)
       this.view.clearSnapshotCache()
     }
   }
--- a/src/tests/fixtures/422.html
+++ b/src/tests/fixtures/422.html
@@ -3,7 +3,7 @@
     <title>Unprocessable Entity</title>
     <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
   </head>
-  <body>
+  <body style="height: 200vh;">
     <h1>Unprocessable Entity</h1>
 
     <turbo-frame id="frame">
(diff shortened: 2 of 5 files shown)


src/core/drive/navigator.ts
```diff
@@ -89,8 +89,8 @@ export class Navigator {
     const responseHTML = await fetchResponse.responseHTML
 
     if (responseHTML) {
-      const snapshot = Snapshot.fromHTMLString(responseHTML)
-      this.view.render({ snapshot }, () => {})
+      const snapshot = Snapshot.wrap(responseHTML)
+      this.view.render({ snapshot }, () => window.scroll(0, 0))
```
seanpdoyle, line 93: I'm not sure how to manage Intern coverage for the scrolling behavior, but if you have suggestions I'll gladly add the tests.

sstephenson: Might want to take a look at the scroll restoration tests: https://github.com/hotwired/turbo/blob/5525c21288edc222de0a05c72538aa7f898abd0c/src/tests/functional/scroll_restoration_tests.ts

seanpdoyle: @sstephenson thank you! I've added some coverage.

sudara:
Looking forward to this, it's one of the [two minor UX issues](https://discuss.hotwire.dev/t/2-usability-niggles-re-turbo-drive-frame-and-forms/1969) I had updating an existing rails app to use turbo.

santib:
I'm needing this as well. Otherwise users might not see error messages (specially the more general error messages, the ones that aren't related to a specific field)

src/core/drive/navigator.ts
```diff
@@ -96,6 +96,7 @@ export class Navigator {
     if (responseHTML) {
       const snapshot = PageSnapshot.fromHTMLString(responseHTML)
       await this.view.renderPage(snapshot)
+      window.scroll(0, 0)
```
seanpdoyle, line 99: @sstephenson this used to be passed as a callback, but then the rendering interface was changed. Is this still the correct spot for it?

PR #87, opened 2021-01-07
Move `[data-turbo="false"]` guard up to observers

seanpdoyle:
Since the `FormSubmitObserver` and `LinkClickObserver` classes are
very interested in the page's HTML, extract the `elementIsNavigable`
checks from the `Session` into our utility module for re-use amongst the
observers.

This enables the `Session` to act with more confidence that any
interaction on the page is to be Turbo-enabled.

### DIFF
--- a/src/observers/link_click_observer.ts
+++ b/src/observers/link_click_observer.ts
@@ -1,3 +1,4 @@
+import { elementIsNavigable } from "../util"
 import { Location } from "../core/location"
 
 export interface LinkClickObserverDelegate {
@@ -35,7 +36,7 @@ export class LinkClickObserver {
   clickBubbled = (event: MouseEvent) => {
     if (this.clickEventIsSignificant(event)) {
       const link = this.findLinkFromClickTarget(event.target)
-      if (link) {
+      if (link && elementIsNavigable(link)) {
         const location = this.getLocationForLink(link)
         if (this.delegate.willFollowLinkToLocation(link, location)) {
           event.preventDefault()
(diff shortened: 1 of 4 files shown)


src/observers/link_click_observer.ts
```diff
@@ -35,7 +36,7 @@ export class LinkClickObserver {
   clickBubbled = (event: MouseEvent) => {
     if (this.clickEventIsSignificant(event)) {
       const link = this.findLinkFromClickTarget(event.target)
-      if (link) {
+      if (link && elementIsNavigable(link)) {
```
seanpdoyle, line 39: I'm not proficient enough in TypeScript to write this function in a way that returns a `boolean` _and_ verifies that the `link` argument is a non-null in a way that makes the `link &&` part of the conditional unnecessary. Is that possible?

shadowtime2000: Not rlly, all non null checks and stuff can't be done with TS cuz it gets removed at compile time. Or I am just not understanding the question....

seanpdoyle: The `link &&` portion of the conditional is required for the method calls within the conditional to operate with confidence that their arguments are present (unless marked as optional or `| undefined`.

I know it's possible within TypeScript to declare functions in a way that hints to the typechecker about an instances presence.

All of this might be moot, since I realize that the submitter is always optional, so the function might not be able to hint the argument's presence all of the time.

shadowtime2000: Oh now I think I understand your question, you could try `link as Element`. That should do it

seanpdoyle:
Modifying Link and Form observers is related to https://github.com/hotwired/turbo/pull/66, so I'm closing this in favor of #66 and combining the commits.

PR #90, opened 2021-01-08
Replace `Location` class with browser-provided URL

seanpdoyle:
The Turbo-internal `Location` class' responsibilities overlap almost
entirely with the browser-provided [URL][] class. In addition, Turbo's
`Location` class name conflicts with the browser-provided [Location][]
class, which is itself an almost `URL`-compliant interface.

This commit removes the Turbo-internal `Location` class, and modifies
all consuming interfaces to use `URL` instead.

The properties and functions that do not have corresponding
[URL][]-provided versions have been extracted to the `src/core/url.ts`
module.

[URL]: https://developer.mozilla.org/en-US/docs/Web/API/URL
[Location]: https://developer.mozilla.org/en-US/docs/Web/API/Location


src/core/url.ts
```diff
@@ -0,0 +1,74 @@
+export type Locatable = URL | string
+
+export function expandPath(pathOrUrl: string | URL): URL {
+  const anchor = document.createElement("a") as HTMLAnchorElement
+  anchor.href = pathOrUrl.toString()
+
+  return new URL(anchor.href)
+}
+
+export function anchor(url: URL): string {
```
seanpdoyle, line 10: @domchristie I've done my best to transform your version to one that is a function and not an instance method.

domchristie: Nice! I think I got distracted by lines 18-23, and missed 11-16 👍

seanpdoyle: Oh no! That's git debris I need to remove. Thanks for the heads up.

Intrepidd:
👍 I was about to open a ticket to ask to export it as it is needed by `navigator.history.push` which is exported, this is even better, thanks !

sstephenson (changes requested):
Cheers to modernization! 🎉

src/http/fetch_request.ts
```diff
@@ -141,3 +131,12 @@ export class FetchRequest {
     return this.abortController.signal
   }
 }
+
+function appendToURLSearchParams(url: URL, entries: Array<[string, FormDataEntryValue]>): URL {
```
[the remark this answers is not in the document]
seanpdoyle: Could we merged #108 after (and rebased off of) this work?

src/core/session.ts
```diff
@@ -201,26 +201,26 @@ export class Session implements NavigatorDelegate, PageObserverDelegate {
...
-  applicationAllowsVisitingLocation(location: Location) {
+  applicationAllowsVisitingLocation(location: URL) {
     const event = this.notifyApplicationBeforeVisitingLocation(location)
     return !event.defaultPrevented
   }
 
-  notifyApplicationAfterClickingLinkToLocation(link: Element, location: Location) {
-    return dispatch("turbo:click", { target: link, detail: { url: location.absoluteURL }, cancelable: true })
+  notifyApplicationAfterClickingLinkToLocation(link: Element, location: URL) {
+    return dispatch("turbo:click", { target: link, detail: { url: location.href }, cancelable: true })
```
seanpdoyle, line 215: @sstephenson is this a good opportunity to dispatch events with URL instances, or is a String still our best bet?

src/core/drive/history.ts
```diff
@@ -1,9 +1,8 @@
-import { Location } from "../location"
 import { Position } from "../types"
 import { nextMicrotask, uuid } from "../../util"
 
 export interface HistoryDelegate {
-  historyPoppedToLocationWithRestorationIdentifier(location: Location, restorationIdentifier: string): void
+  historyPoppedToLocationWithRestorationIdentifier(location: URL, restorationIdentifier: string): void
```
seanpdoyle, line 5: There are many methods that reference `Location` in their name that now accept `URL` instances with arguments named `location`.

A global search and replace would be a simple way to rename them, but I'm not sure which are part of the public API. Are there backward compatibility concerns with method names are positional argument names?

src/http/fetch_request.ts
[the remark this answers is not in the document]
sstephenson: Yeah, let's do that 👍

src/core/session.ts
sstephenson: I think we can just stick with strings here.

src/core/drive/history.ts
sstephenson: Thanks for noting. I'll make a pass through and clean these up.

seanpdoyle: Thank you @sstephenson ! I'm all finished on my end if you're ready to take over.

sstephenson: Decided not to rename them because it'd be too invasive (the adapter interface would change) and it's fine for the term "location" to remain part of the Turbo Drive domain language.

sstephenson (approved):

ghiculescu:
Hey, I think there is a bug in how forms are submitted since this PR was merged. Basically, all the elements of a form are being included in the URL when it's submitted.

This is what form submissions look like with the latest Turbo from npm:
image

If you clone the repo and then point your package.json at your clone on the `main` branch, they look like this:

image


If you do `git checkout 178c817dc75d9559b3b624058f28a5bbf466a841` then run against that, the bug is gone:

image

This line seems to be the culprit: https://github.com/hotwired/turbo/pull/90/files#diff-68b647dc2963716dc27c070f261d0b942ee9c00be7c4149ecb3a5acd94842d40R53

It looks like it is replacing the old behavior [here](https://github.com/hotwired/turbo/pull/90/files#diff-68b647dc2963716dc27c070f261d0b942ee9c00be7c4149ecb3a5acd94842d40L61), but that behavior only ran on `GET` requests. Now it runs on every request.

seanpdoyle:
@ghiculescu thanks for reporting this! I believe the decision was deliberate, but https://github.com/hotwired/turbo/pull/128 might handle the nuance a little more appropriately.

ghiculescu:
I don't think putting the contents of the form body in the URL are a good idea. As an example of why, different browsers implement different rules around [URL lengths](https://stackoverflow.com/a/33733386/641293) which this could now break. There's probably also security implications, eg. if your server logs URLs but not request bodies, it may suddenly start logging passwords or authenticity tokens (like my 2nd screenshot).

ghiculescu:
https://github.com/hotwired/turbo/pull/128/commits/ba323e44221eed4f1ae86c3da88f0f3872e5e72a looks like it'll fix it. Thanks @seanpdoyle

Issue #77, opened 2021-01-05
Support for globally opting-out of behaviors?


swanson:
:wave:

Wondering if there would be an appetite for adding a configuration/setting to opt-out of some of the Turbo modules.

If you only wanted the TurboDrive for links, it might be nice to support a global opt-out of the `formSubmitObserver`. If you aren't using Frames or Streams, you could opt-out of the `streamObserver`, `frameRedirector`, etc.

There are sometimes ways to opt-out of these behaviors on a per-page, per-form, or per-link basis, but I think it would smooth adoption / upgrading from Turbolinks if you could opt-in piecemeal.

I don't think it's a slam dunk, especially since the installation path is really nice and there is no config/settings (outside of the progress timeout). Happy to help implement if there is interest, but wanted to gauge support.

dhh:
I think we should wait for some real world cases that demand this before preemptively adding such configurations. I mean, I can imagine this maybe useful to some, but I'd like to see a real case before we add the additional weight.

swanson:
> I think we should wait for some real world cases that demand this before preemptively adding such configurations. I mean, I can imagine this maybe useful to some, but I'd like to see a real case before we add the additional weight.

Yeah, my only real world case is that I have a large app with lots of forms -- some are `form_with`, some are `form_for`, some are remote, some are local -- and I would like to upgrade TurboLinks to TurboDrive but...it's a big undertaking to audit every form (there are hundreds). So I have to undertake that effort before I could start using Turboframes, Streams, etc

dhh:
Yeah, that sounds fair. I'd be fine seeing a toggle for each element.

santib:
I was thinking exactly the same. In my case I'd like to:
- Have a toggle to turn on/off Turbo Drive (default: on). Why? Because I'd be able to start using Turbo Frames and Turbo Streams right away, even before making sure that none of my forms break with Turbo Drive.
- Have a toggle to turn on/off Turbo Streams via Websockets (default: on). Why? Because I'd be able to almost all of Turbo without needing to care about Redis and Websockets.

I don't know if this would be feasible with the current implementation (didn't check yet). But at least from a developer's point of view would be clear IMO.

swanson:
This has been resolved in https://github.com/hotwired/turbo/releases/tag/v7.0.0-beta.2 -- opt-out of the form behavior with `data-turbo=false` on a root element and override on a per form / page basis.

PR #95, opened 2021-01-09
Add ability to turn off link and form behaviors

swanson:
:wave:

See https://github.com/hotwired/turbo/issues/77

In an effort to ease the upgrade (or adoption) path from Turbolinks to Turbodrive, support disabling the link and/or form behavior. This allows for people to:

- Upgrade from Turbolinks to TurboDrive but opt-out of new form behavior (this may be a non-trivial effort for legacy apps)
- Use TurboFrames and TurboStreams without going "all-in" on the Turbo stack

This isn't a large changeset and omits tests, but this might be a good place to continue the discussion while looking at some actual code.

The intended usage as a consumer would be:

```js
import Turbo from "@hotwired/turbo"
Turbo.disableNavigation()
Turbo.disableFormSubmissions()
```

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -41,10 +41,10 @@ export function setProgressBarDelay(delay: number) {
   session.setProgressBarDelay(delay)
 }
 
-export function disableLinkBehavior() {
-  session.disableLinkBehavior()
+export function disableNavigation() {
+  session.disableNavigation()
 }
 
-export function disableFormBehavior() {
-  session.disableFormBehavior()
+export function disableFormSubmissions() {
+  session.disableFormSubmissions()
 }
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -40,3 +40,11 @@ export function clearCache() {
 export function setProgressBarDelay(delay: number) {
   session.setProgressBarDelay(delay)
 }
+
+export function disableLinkBehavior() {
+  session.disableLinkBehavior()
+}
+
+export function disableFormBehavior() {
+  session.disableFormBehavior()
+}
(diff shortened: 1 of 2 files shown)


santib:
@swanson Thanks for start the conversation on this 👏  This implementation adds the ability to completely turn off link and form behavior, which could be helpful. Thinking on a real app, at some point you might want to start enabling turbo drive in a per element basis (so the risk of breaking the app is lower and you can start trying it out in less riskier parts of an app).

What do you think about letting developers add `data: { turbo: true }` on links and forms to enable Turbo Drive on specific elements? Note that if we were to implement this other part, I think the current solution might need to change since it's a bit too global 🤔 Thoughts?

santib:
Implementation-wise maybe here https://github.com/hotwired/turbo/blob/main/src/core/session.ts#L249 we can have something like
```typescript
  elementIsNavigable(element?: Element) {
    const container = element?.closest("[data-turbo]")
    if (container) {
      return container.getAttribute("data-turbo") != "false"
    } else {
      return [global default value]
    }
  }
```

So that if data-turbo is specified it takes precedence, otherwise it goes with the global value (on by default)

swanson:
@santib for links this exists already: https://turbo.hotwire.dev/handbook/drive#disabling-turbo-drive-on-specific-links

I just found out the bit about having ancestors disable links and then have descendants opt-back in. So now I'm wondering if this PR should be more target for just the form behaviors.

swanson:
Ah, I missed #3 which I think will solve the same problems: toss `data-turbo=false` on your <body> tag, then override as you upgrade / migrate forms. Looks like the docs just haven't been updated yet, but https://github.com/hotwired/turbo/releases/tag/v7.0.0-beta.2 resolves this.

Issue #107, opened 2021-01-14
Drive: GET form (e.g. search form) with NO action="" adds duplicate ?q=foo&q=foo&q=foo... to URL


weaverryan:
Hi!

I'm using beta 3 with this form:

```html
<form>
        <input name="q" value="apple" type="search">
</form>
```

Notice the missing `action=""`. That's valid (it will submit to itself), but that is the key to the problem (the problem does NOT happen if there is an `action` attribute). If the original page is `/`, then on submit, the URL will become `/?q=apple`, which is correct. If you submit again, it will become `/?q=apple&q=apple`. A 3rd time will keep that pattern going - `/?q=apple&q=apple&q=apple`.

Please let me know if I can provide any more info!

Thanks!

PR #108, opened 2021-01-15
GET form submissions to the current path

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/107

The [HTML specification][] outlines the algorithm for handling form
submissions, and describes the steps involved in handling a `method=get`
request with `application/x-www-form-urlencoded` encoding.

However, it doesn't explicitly mention how to handle a `<form
method="get">` submission when the `action` attribute is omitted and the
current URL includes a query parameter that would be included within the
`<form>` element fields' values.

Determining the test cases to reproduce browser default behavior
involved was some trial and error troubleshooting in real browsers with
default (JavaScript-less) `<form>` submission handling.

The algorithm appears to be as follows:

* If there are no query parameters, append all field values as query
   parameters

* If there are existing query parameters whose keys exist as `<form>`
  fields, override them based on the form field's value

  * If there are existing query parameters whose keys exist as
    _multiple_ `<form>` fields, override the first occurrence, and then
    append subsequent values

[HTML specification]: https://html.spec.whatwg.org/#form-submission-algorithm

### DIFF
--- a/src/http/fetch_request.ts
+++ b/src/http/fetch_request.ts
@@ -137,8 +137,17 @@ export class FetchRequest {
 }
 
 function mergeFormDataEntries(url: URL, entries: [string, FormDataEntryValue][]): URL {
+  const currentSearchParams = new URLSearchParams(url.search)
+
   for (const [ name, value ] of entries) {
-    url.searchParams.append(name, value.toString())
+    if (value instanceof File) continue
+
+    if (currentSearchParams.has(name)) {
+      currentSearchParams.delete(name)
+      url.searchParams.set(name, value)
+    } else {
+      url.searchParams.append(name, value)
+    }
   }
 
   return url
--- a/src/tests/fixtures/form.html
+++ b/src/tests/fixtures/form.html
@@ -43,6 +43,22 @@
         <input type="hidden" name="greeting" value="Hello from a form">
         <input type="submit">
       </form>
+      <form action="/src/tests/fixtures/form.html?query=1" method="get" class="conflicting-values">
+        <input type="hidden" name="query" value="2">
+        <input type="submit">
+      </form>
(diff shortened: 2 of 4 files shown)


src/http/fetch_request.ts
```diff
@@ -133,8 +133,15 @@ export class FetchRequest {
...
 function mergeFormDataEntries(url: URL, entries: [string, FormDataEntryValue][]): URL {
+  const currentUrl = new URL(window.location.href)
+
   for (const [ name, value ] of entries) {
-    url.searchParams.append(name, value.toString())
+    if (currentUrl.searchParams.has(name)) {
+      currentUrl.searchParams.delete(name)
+      url.searchParams.set(name, value.toString())
+    } else {
+      url.searchParams.append(name, value.toString())
```
seanpdoyle, line 143: @sstephenson when `value` is a [File](https://developer.mozilla.org/en-US/docs/Web/API/File), will coercing to a String break uploads?

seanpdoyle: Would we be better off guarding these calls with `!(value instanceof File)`?

sstephenson: What happens when you submit an `<input type="file">` in a `<form method="get">` without Turbo? The browser just skips the field, I think?

seanpdoyle: I don't think there's anything in FetchRequest's current form that restricts this to idempotent `method="get"` requests. Am I missing something?

seanpdoyle: For what it's worth, this loop is to mutate the URL it's submitted to, and not necessarily the body we're submitting. Would omitting Files entirely avoid the problem of turning a file into a query parameter?

PR #110, opened 2021-01-15
Recursively collect delegate headers to include in fetch request

### LINKED ISSUE #86: Missing Turbo-Frame header when submitting a form
When clicking links while inside a frame, the `Turbo-Frame` header will be set.

This helps the backend to know this is a frame request and for instance to not render layout.

When submitting a form, this header is not set, I think it should be. 

Example : To respect progressive enhancement, we should be able to return a frame when the request comes from turbo, but perform a redirection when not.

tleish:
Closes #86

When turbo sends a fetch, it uses the following method in `FetchRequest` to get additional headers:

```typescript
  // FetchRequest
  get additionalHeaders() {
    if (typeof this.delegate.additionalHeadersForRequest == "function") {
      return this.delegate.additionalHeadersForRequest(this)
    } else {
      return {}
    }
  }
```

When the `this.delegate=FrameController`, the FrameController adds the `Turbo-Frame` header to the request.

```typescript
  // FrameController
  additionalHeadersForRequest(request: FetchRequest) {
    return { "Turbo-Frame": this.id }
  }
```

### TURBO-FRAME GET
For a GET form inside. a turbo-frame, `this.delegate=FrameController`.  So, form#GET requests include the `Turbo-Frame` header.

### TURBO-FRAME NON-GET (e.g. POST/PUT/DELETE)
For a NON-GET form inside. a turbo-frame, `this.delegate=FrameController`.  Which means, no `Turbo-Frame` header.

The system does detect the turbo-frame, it just wraps it in FormSubmission and assigns FrameController as the delegate.
`FormSubmission = { delegate: FrameController }`

Perhaps there are other nested delegate scenarios (e.g. ViewController). This update traverses delegates to collect all additionalHeadersForRequest.  It also ensures that headers from the closest delegate relatives override headers from the most distance delegate relatives.

### DIFF
--- a/src/http/fetch_request.ts
+++ b/src/http/fetch_request.ts
@@ -118,8 +118,11 @@ export class FetchRequest {
 
   get headers() {
     const headers = { ...this.defaultHeaders }
-    if (typeof this.delegate.prepareHeadersForRequest == "function") {
-      this.delegate.prepareHeadersForRequest(headers, this)
+    let fetchRequest:any = this;
+    while(fetchRequest = fetchRequest.delegate) {
+      if (typeof fetchRequest.prepareHeadersForRequest == "function") {
+        fetchRequest.prepareHeadersForRequest(headers, this);
+      }
     }
     return headers
   }


tleish:
Updated code to support addition of `prepareHeadersForRequest`

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/166 to extend this work and account for test coverage. @tleish I've added you as co-author. Do you mind if I close this PR in favor of #166?

Issue #115, opened 2021-01-17
turbo-frame target attribute ignored on redirect


arschrier:
Let me preface this by saying that I am completely new to turbo, so it could be that I'm doing something wrong. However, I have tried following the documentation closely, and it seems I am unable to drive navigation across turbo-frames.

In the following snippet, I would expect turbo to update `<turbo-frame id="main">` with the current time after `<a href="/path">Update time</a>` is clicked.

```html
<turbo-frame id="navigation" target="main">
    <a href="/path">Update time</a>
</turbo-frame>
<turbo-frame id="main">
    <%= Time.now %>
</turbo-frame>
```

Instead, turbo sets a `src` attribute on `<turbo-frame id="navigation" target="main" src="/path">` and tries to update the contents of that frame.

It should be noted that when setting the `data-turbo-frame="main"` attribute directly on the link, turbo updates the frame as expected.

I did some digging and thought this behavior might have something to do with the `findFrameElement` function in [frame_redirector.ts](https://github.com/hotwired/turbo/blob/3263389fd56e6fa3dd665b74ad715d6b183e1e3a/src/core/frames/frame_redirector.ts) , in which only frames referenced by the `data-turbo-frame` attribute are returned.

```ruby
private findFrameElement(element: Element) {
  const id = element.getAttribute("data-turbo-frame")
  if (id && id != "_top") {
    const frame = this.element.querySelector(`#${id}:not([disabled])`)
    if (frame instanceof FrameElement) {
      return frame
    }
  }
}
```

PR #116, opened 2021-01-17
Read frame target from turbo-frame[target]

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/115

When navigating via an `<a>` click within a `<turbo-frame>`, determine
the target of the navigation by checking for `[data-turbo-frame]`
overrides on the `<a>` element itself, and then check for frames
rreferenced by `turbo-frame[target]`, then fall back to the frame
itself.

### DIFF
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -10,6 +10,42 @@ export class FrameTests extends FunctionalTestCase {
     await this.nextBeat
     this.assert.notOk(await this.innerHTMLForSelector("#missing"))
   }
+
+  async "test following a link within a frame with a target set navigates the target frame"() {
+    await this.clickSelector("#hello a")
+    await this.nextBeat
+
+    const frameText = await this.querySelector("#frame h2")
+    this.assert.equal(await frameText.getVisibleText(), "Frame: Loaded")
+  }
+
+  async "test following a link within a descendant frame whose ancestor declares a target set navigates the descendant frame"() {
+    await this.clickSelector("#nested-root[target=frame] #nested-child a:not([data-turbo-frame])")
+    await this.nextBeat
+
+    const frame = await this.querySelector("#frame h2")
+    const nestedRoot = await this.querySelector("#nested-root h2")
+    const nestedChild = await this.querySelector("#nested-child")
+    this.assert.equal(await frame.getVisibleText(), "Frames: #frame")
+    this.assert.equal(await nestedRoot.getVisibleText(), "Frames: #nested-root")
+    this.assert.equal(await nestedChild.getVisibleText(), "Frame: Loaded")
+  }
+
+  async "test following a link that declares data-turbo-frame within a frame whose ancestor respects the override"() {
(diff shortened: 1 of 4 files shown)


src/tests/functional/frame_tests.ts
```diff
@@ -10,6 +10,38 @@ export class FrameTests extends FunctionalTestCase {
...
+
+  async "test following a link within a frame with a target set navigates the target frame"() {
+    await this.clickSelector("#hello a")
+    await this.nextBeat
+
+    const frameText = await this.querySelector("#frame h2")
+    this.assert.equal("Frame: Loaded", await frameText.getVisibleText())
+  }
+
+  async "test following a link within a frame whose ancestor declares a target set navigates the target frame"() {
```
seanpdoyle, line 22: I'm not totally clear on whether or not this is what we want.

Is each `<turbo-frame>` element to be treated as its own navigation context, or does it inherit from its ancestors?

According to the [MDN documentation for `<a target="...">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target), the default value for `[target]` when omitted is `_self` (i.e. "the current browsing context").

If we _do_ want descendant frames to inherit the `[target]` of their closest ancestor, we have some options:

```html
<!-- inherit from ancestor -->
<turbo-frame id="parent" target="_top">
  <turbo-frame id="child"> <!-- inherits target="_top" -->
    <a href="/some-other-page.html">Navigates page</a>
  </turbo-frame>
</turbo-frame>

<!-- inherit from ancestor -->
<turbo-frame id="parent" target="_top">
  <turbo-frame id="child" target="child"> <!-- overrides target to itself -->
    <a href="/some-other-page.html">Navigates #child</a>
  </turbo-frame>
</turbo-frame>
```

This also might be an opportunity to introduce `_self` or `_parent` as special case values, similar to how we treat `_top`.

```html
<!-- each frame as self-contained context -->
<turbo-frame id="parent">
  <turbo-frame id="child" target="_self"><!-- the default -->
    <a href="/some-other-page.html">Navigates #child</a>
  </turbo-frame>

  <!- ... -->
  <turbo-frame id="child" target="_parent">
    <a href="/some-other-page.html">Navigates #parent</a>
  </turbo-frame>

  <!- ... -->
  <turbo-frame id="child" target="_top">
    <a href="/some-other-page.html">Navigates the page</a>
  </turbo-frame>
</turbo-frame>
```

If we'd still like to decide separate from the bug fix for #115, we can omit this test case and replace 

```diff
- const id = element.getAttribute("data-turbo-frame")
+ const id = element.getAttribute("data-turbo-frame") || this.element.closest("turbo-frame[target]")?.getAttribute("target") || ""
```

with 

```diff
- const id = element.getAttribute("data-turbo-frame")
+ const id = element.getAttribute("data-turbo-frame") || this.element.getAttribute("target") || ""
```

seanpdoyle: Instead of defaulting to the concept of `_parent`, default to `_self`.

PR #121, opened 2021-01-18
Actions 'after' and 'before' added. Scripts enhanced for windows usage.

guitarneck:
Hello,

A simple improvment whith **after** and **before** actions. So, it give the ability to update some kind of lists.
Let says we have a `<ul...>...</ul>` list, and want to update it with `<li...>...<li>` in a desired place, _after_ or _before_ a given _target_ inside the list.

By the way, I add scripts to package.json for windows users.
Hope you find it usefull.

With unit tests process (It needed a _parent_ for this purpose). Maybe it needs additionnal checks, to insure a _parent_ exists ?

Have a nice day !


guitarneck:
The test with fixture **svg** crash. It's awaiting a `image/svg+xml` mime type, but the file has no extension and is loaded as `text/plain` in fact.
should it be named **svg.svg**, and so tests adapted ?
It fails on line 48, not in 44. Weird.

src/core/streams/stream_actions.ts
```diff
@@ -1,10 +1,18 @@
 import { StreamElement } from "../../elements/stream_element"
 
 export const StreamActions: { [action: string]: (this: StreamElement) => void } = {
+  after() {
+    this.targetElement?.parentElement?.insertBefore(this.templateContent,this.targetElement.nextSibling)
```
rainerborene, line 5: I think `nextSibling` could be `undefined` when there is only one element present. I suggest using the `insertAdjacentHTML` API for both actions. https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentHTML

guitarneck: Hi,
It can be `null` with insertBefore, this is well handle.
[https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore](https://developer.mozilla.org/en-US/docs/Web/API/Node/insertBefore)

`If this is null, then newNode is inserted at the end of parentNode's child nodes.`

insertAdjacentHTML seems not to be a good choice. Templates content are `DocumentFragment`, so insertAdjacentElement could be a better choice. But it's less compliant with all browser.

rainerborene: Cool. I didn't know about that `null` detail.

qinmingyuan:
Want this feature so much~

itsliamegan:
I realize this was just merged, but are the names of the actions set in stone? Whereas all the other turbo stream actions are verbs (replace, append, etc.) these two are adjectives and they feel out of place. Perhaps "insert" as the action name and an additional attribute like "direction" to specify before/after?

dhh:
@itsliamegan I was thinking about that too, but this should be read in context, so it's action="before" target="element", and then I think it reads fine as "before target element, THIS". Which isn't too different from action="append" target="element", which also needs a little finessing in your head to read properly as "append TO target element, THIS".

PR #124, opened 2021-01-19
Avoid redefining things when using the asset pipeline in Safari

npezza93:
While using Safari with Turbo and Stimulus(through the hotwire gem) and a Stimulus controller
that also utilizes Turbo I am seeing these errors.

When Stimulus tries to import the controller, it seems as if it is
trying to import Turbo twice causing these "already defined" errors.

While using Turbo and Stimulus through the asset pipeline
- Only define customeElements when they are not already
- Only define 'submitter' property if it isnt already defined on Event

You can recreate this error by setting up hotwire through the asset pipeline then add `import { Turbo } from "turbo"` to a stimulus controller.

### DIFF
--- a/src/elements/index.ts
+++ b/src/elements/index.ts
@@ -7,5 +7,5 @@ FrameElement.delegateConstructor = FrameController
 export * from "./frame_element"
 export * from "./stream_element"
 
-customElements.define("turbo-frame", FrameElement)
-customElements.define("turbo-stream", StreamElement)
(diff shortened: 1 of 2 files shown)


dhh:
This was because turbo was not included as a module-shim. This has been fixed in the latest version. You can fix it by including Turbo after stimulus, adding it to your importmap, and making sure its included as module-shim.

Because even with this fix, Turbo is being loaded twice, which is not what you want.

PR #125, opened 2021-01-21
Prevent reloading pages when tapping same-page anchors

domchristie:
This is a port of https://github.com/turbolinks/turbolinks/pull/285 to resolve https://github.com/turbolinks/turbolinks/issues/75, https://github.com/turbolinks/turbolinks/issues/214, and https://github.com/turbolinks/turbolinks/issues/192

## The Problem

Clicking on a link which references an anchor on the current page issues an unnecessary request. This can cause problems when linking to content which is lazily loaded (see https://agile-reef-88023.herokuapp.com/posts/1), as well as with [analytics](https://github.com/turbolinks/turbolinks/issues/75#issuecomment-255867311) and [skip links](https://github.com/hotwired/turbo/issues/42). Disabling Turbo completely on those links prevents the request, but because the view is not cached, this approach breaks the Back button.

## The Solution

This PR detects whether the visit is to an anchor on the same page, and if so, prevents the request. It takes a snapshot and performs the scroll, bypassing any rendering and `turbo:load` triggering. A similar flow handles restorations to same-page anchors. Correct focusing is achieved by temporarily adding `tabindex=-1` and calling `focus` on the target.

The focus test still seems to fail. I'm beginning to think that simulating pressing tab doesn't do anything? 

A final note: the [Turbolinks version](https://github.com/turbolinks/turbolinks/pull/285) of this PR performed [the same-page check in the session](https://github.com/turbolinks/turbolinks/pull/285/files#diff-da4d3d3a97b9ed26e47b50d9a5d2f5795434467e001f8a6deba72c361a23baa7R330-R333) (née controller). This allowed the [native adapters to check for same-page anchors](https://github.com/turbolinks/turbolinks-ios/pull/126/files#diff-db83eae543943038c85ab2f7a4c437adda02ed30411e803d0091f5a7fc32d0acR57) and handle accordingly. The session is not so accessible from the Visit, so I'd welcome some tips on how to solve this (or whether it's even necessary now?).

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -2,7 +2,7 @@ import { Adapter } from "../native/adapter"
 import { FetchMethod, FetchRequest, FetchRequestDelegate } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { History } from "./history"
-import { getAnchor } from "../url"
+import { getAnchor, getRequestURL } from "../url"
 import { PageSnapshot } from "./page_snapshot"
 import { Action } from "../types"
 import { uuid } from "../../util"
@@ -70,6 +70,7 @@ export class Visit implements FetchRequestDelegate {
   frame?: number
   historyChanged = false
   location: URL
+  isSamePage: boolean
   redirectedToLocation?: URL
   request?: FetchRequest
   response?: VisitResponse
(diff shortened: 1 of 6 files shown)


domchristie:
https://github.com/hotwired/turbo/pull/125/commits/e4bee01255f5e1b65fb6319e15a997831f06b65a is a bit of a test cheat, and I'm not proud of it!

Temporarily setting `tabindex=-1`, calling `focus()` then unsetting `tabindex`, unfocusses the element (in Chrome). However, the position in the tab sequence seems to be maintained, therefore tabbing will focus the next element as expected. If the element already has a `tabindex` (as I have added in https://github.com/hotwired/turbo/pull/125/commits/e4bee01255f5e1b65fb6319e15a997831f06b65a), it will maintain focus, and so we can test that the element is successfully focussed after navigating to it.

I hope that makes sense!?

daveverwer:
Thanks so much for this @seanpdoyle. Do you know if there's a plan to get this branch merged into main? I'd love to see this fix become part of the library.

Issue #126, opened 2021-01-22
Search Form (GET action) not working


santib:
Turbo doesn't allow a form submission to respond other than a redirect or a 4xx/5xx. But wouldn't it make sense to allow a 200 OK for search forms (GET action) ?

Currently the behavior is super weird because it sometimes works and then it stops working:
![](https://recordit.co/KUqOrRVIGy.gif)

The current workaround would be creating another endpoint that redirects there but looks hacky to me. I'm open to do a PR to allow 200 OK for Form submissions with GET action if it'd be accepted.

PR #127, opened 2021-01-22
Transform GET submissions into Visits

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/126

When submitting a `<form method="get">`, skip the Form Submission
lifecycle and instead initiate a Visit.

Testing
---

Resolve an issue that caused false positives in our test suite when
monitoring whether or not an action triggered a form submission.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -37,7 +37,12 @@ export class Navigator {
   submitForm(form: HTMLFormElement, submitter?: HTMLElement) {
     this.stop()
     this.formSubmission = new FormSubmission(this, form, submitter, true)
-    this.formSubmission.start()
+
+    if (this.formSubmission.fetchRequest.isIdempotent) {
+      this.proposeVisit(this.formSubmission.fetchRequest.url)
+    } else {
+      this.formSubmission.start()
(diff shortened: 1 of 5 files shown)


agrobbin:
@seanpdoyle would it make sense for this to also account for when there is a `data-turbo-frame` attribute on the `<form>`? I have a "search" form that, when submitted, I want to update the results Frame below it, but with this change, I am unable to do so. Or is there a different approach that would make more sense for this kind of behavior?

For reference, the reason we want to only update a single frame is we don't want the search input to lose focus while typing, so we can't just have it do a Drive visit, as that would cause the element to lose focus.

agrobbin:
Just to pre-empt the question I bet I'm going to get about using `data-turbo-permanent`, That doesn't appear to solve the focus-lost situation. I'm not sure why though!

seanpdoyle:
@agrobbin if I understand your problem correctly, you're right, this change should result in the submission driving the target frame instead of doing a page-level navigation.

agrobbin:
Ah, I must have had some old code loaded or something, you're totally right. Sorry for the confusion @seanpdoyle!

Issue #76, opened 2021-01-04
Support <form enctype>


sstephenson:
Turbo always submits forms using the `multipart/form-data` encoding. Instead, we should follow the browser's behavior: read the form submitter's `formenctype` attribute or the form's `enctype` attribute to determine the form's encoding, or fall back to `application/x-www-form-urlencoded` if none is set.

Implementation-wise, we can pass our [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instance to the [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams) constructor and use the result as the [fetch body](https://github.com/hotwired/turbo/blob/7eab20f7537338f983d58b8910a35108d81d2467/src/http/fetch_request.ts#L116).

ref. https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/enctype
https://discuss.hotwire.dev/t/post-form-using-application-x-www-form-urlencoded/1615

PR #128, opened 2021-01-22
Form Encoding

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/76

According to the HTML Specification for [Form Submission Encoding][],
there are three possible values:

* [application/x-www-form-urlencoded][] (the default value)
* [multipart/form-data][]
* [text/plain][]

[Encoding declarations on `<form method="get">` elements are
ignored][get-ignored].

This commit re-structures the `FetchRequest` constructor, and moves the
FormData to URL parameter encoding process _out_ to the `FormSubmission`
itself, since the `HTMLFormElement` knows about its [enctype][] and the
submitter its [formenctype][].

[Form Submission Encoding]: https://www.w3.org/TR/html52/sec-forms.html#selecting-a-form-submission-encoding
[application/x-www-form-urlencoded]: https://www.w3.org/TR/html52/sec-forms.html#selecting-a-form-submission-encoding
[multipart/form-data]: https://www.w3.org/TR/html52/sec-forms.html#multipart-form-data
[text/plain]: https://www.w3.org/TR/html52/sec-forms.html#plain-text-form-data
[enctype]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/enctype
[formenctype]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formenctype
[get-ignored]: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest#submitting_forms_and_uploading_files

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -24,6 +24,20 @@ export enum FormSubmissionState {
   stopped,
 }
 
+enum FormEnctype {
+  urlEncoded = "application/x-www-form-urlencoded",
+  multipart  = "multipart/form-data",
+  plain      = "text/plain"
+}
+
+function formEnctypeFromString(encoding: string): FormEnctype {
+  switch(encoding.toLowerCase()) {
+    case FormEnctype.multipart: return FormEnctype.multipart
+    case FormEnctype.plain:     return FormEnctype.plain
+    default:                    return FormEnctype.urlEncoded
+  }
+}
+
 export class FormSubmission {
   readonly delegate: FormSubmissionDelegate
   readonly formElement: HTMLFormElement
@@ -39,7 +53,7 @@ export class FormSubmission {
     this.formElement = formElement
     this.formData = buildFormData(formElement, submitter)
     this.submitter = submitter
-    this.fetchRequest = new FetchRequest(this, this.method, this.location, this.formData)
+    this.fetchRequest = new FetchRequest(this, this.method, this.location, this.body)
     this.mustRedirect = mustRedirect
   }
 
@@ -54,6 +68,18 @@ export class FormSubmission {
 
   get location(): URL {
     return expandURL(this.action)
+  }
+
+  get body() {
+    if (this.enctype == FormEnctype.urlEncoded || this.method == FetchMethod.get) {
+      return new URLSearchParams(this.formData as any)
+    } else {
+      return this.formData
+    }
+  }
+
+  get enctype(): FormEnctype {
+    return formEnctypeFromString(this.submitter?.getAttribute("formenctype") || this.formElement.enctype)
   }
 
   // The submission process
@@ -79,7 +105,7 @@ export class FormSubmission {
 
   additionalHeadersForRequest(request: FetchRequest) {
     const headers: FetchRequestHeaders = {}
-    if (this.method != FetchMethod.get) {
+    if (!request.isIdempotent) {
       const token = getCookieValue(getMetaContent("csrf-param")) || getMetaContent("csrf-token")
       if (token) {
         headers["X-CSRF-Token"] = token
--- a/src/tests/server.ts
+++ b/src/tests/server.ts
@@ -1,6 +1,7 @@
 import { Response, Router } from "express"
 import multer from "multer"
 import path from "path"
+import url from "url"
 
 const router = Router()
 const streamResponses: Set<Response> = new Set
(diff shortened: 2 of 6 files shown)


src/tests/server.ts
```diff
@@ -1,26 +1,32 @@
...
 const router = Router()
 const streamResponses: Set<Response> = new Set
 
+router.use(urlencoded({ extended: false }))
 router.use(multer().none())
 
 router.post("/redirect", (request, response) => {
-  const path = request.body.path ?? "/src/tests/fixtures/one.html"
-  response.redirect(303, path)
+  const pathname = request.body.path ?? request.query.path ?? "/src/tests/fixtures/one.html"
```
seanpdoyle, line 13: @sstephenson I'm reaching the limits of my Express middleware expertise. I was under the impression that the built-in `url encoded` middleware would access values through `request.body`, but it seems like that's most often blank and they're only accessible through `request.query`.

Additionally, any form that specifies `multipart/form-data` errors in a way that I've had trouble debugging, in spite of the `multer().none()` middleware.

src/core/drive/form_submission.ts
```diff
@@ -24,6 +24,20 @@ export enum FormSubmissionState {
...
+  multipartFormData             = "multipart/form-data",
+  textPlain                     = "text/plain"
+}
+
+export function formEnctypeFromString(method: string): FormEnctype {
+  switch (method.toLowerCase()) {
+    case FormEnctype.multipartFormData: return FormEnctype.multipartFormData
+    case FormEnctype.textPlain:         return FormEnctype.textPlain
+    default:                            return FormEnctype.applicationXWWWFormUrlEncoded
+  }
```
kaspth, line 38: Looks like `enum`s automatically have [reverse mappings](https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings) so I think this could be: 

```suggestion
  return FormEnctype[method.toLowerCase()] || FormEnctype.applicationXWWWFormUrlEncoded
```

src/core/drive/form_submission.ts
```diff
@@ -24,6 +24,20 @@ export enum FormSubmissionState {
   stopped,
 }
 
+export enum FormEnctype {
+  applicationXWWWFormUrlEncoded = "application/x-www-form-urlencoded",
+  multipartFormData             = "multipart/form-data",
+  textPlain                     = "text/plain"
```
kaspth, line 30: What do you think about shortening these a bit?

```suggestion
  urlEncoded = "application/x-www-form-urlencoded",
  multipart  = "multipart/form-data",
  plain      = "text/plain"
```

src/core/drive/form_submission.ts
```diff
@@ -24,6 +24,20 @@ export enum FormSubmissionState {
...
+}
+
+export function formEnctypeFromString(method: string): FormEnctype {
```
kaspth, line 33: Are the `export`s for `FormEnctype` and `formEnctypeFromString` leftovers? I don't see these being used outside of here?

src/core/drive/form_submission.ts
```diff
@@ -56,6 +70,24 @@ export class FormSubmission {
     return expandURL(this.action)
   }
 
+  get body() {
+    if (this.enctype == FormEnctype.applicationXWWWFormUrlEncoded) {
```
kaspth, line 74:
```suggestion
    if (this.method == FetchMethod.get || this.enctype == FormEnctype.applicationXWWWFormUrlEncoded) {
```

Why don't we also want the `get` check here? Feels a little unsafe/unclear to me to infer this purely based on the `enctype`?

src/core/drive/form_submission.ts
```diff
@@ -24,6 +24,20 @@ export enum FormSubmissionState {
...
 }
 
+enum FormEnctype {
+  urlEncoded = "application/x-www-form-urlencoded",
+  multipart  = "multipart/form-data",
+  plain      = "text/plain"
+}
+
+function formEnctypeFromString(encoding: string): FormEnctype {
+  switch(encoding.toLowerCase()) {
```
seanpdoyle, line 34: @kaspth I think reverse mapping would be great for our FetchMethod enum, but it seems to break down when there are `/` in the String value.

kaspth: Oh huh! My understanding from the docs was that it was basically like `FormEnctype[FormEnctype[urlEncoded] = "application/x-www-form-urlencoded"] = urlEncoded` and I don't get why that wouldn't work. Oh well, maybe something at the transpilation stage 😄

src/core/drive/form_submission.ts
```diff
@@ -56,6 +70,20 @@ export class FormSubmission {
...
+      return new URLSearchParams(this.formData as any)
+    } else {
+      return this.formData
+    }
+  }
+
+  get enctype(): FormEnctype {
+    const enctype = this.submitter?.getAttribute("formenctype") || this.formElement.enctype
+
+    return formEnctypeFromString(enctype) || FormEnctype.urlEncoded
```
kaspth, line 84:
```suggestion
    return formEnctypeFromString(enctype)
```

The `||` is never hit because of the default function return.

src/core/drive/form_submission.ts
kaspth: Ah! There's this bit in the documentation 😢

```
Keep in mind that string enum members do not get a reverse mapping generated at all.
```

PR #129, opened 2021-01-22
Baseuri ff

### LINKED ISSUE #123: SVG (sprite) in Frame not rendered in Firefox
**Context**

Firefox has this problem, but Safari and Chrome don't.

**Problem**

I have some SVG sprite icons that are not displayed when they are part of a dynamically loaded Turbo Frame. The embedded SVG looks like this:

```html
<svg class="bi" width="20" height="20" fill="currentColor">
    <use xlink:href="/icons/bootstrap-icons.svg#envelope"/>
</svg>
```

The sprite has already been loaded before activating the new content, because the page contains a couple of icons that are  displayed correctly.

**Solution**

The following change in Turbo will solve my problem, but I don't know what the consequences are. In effect, `createContextualFragment` is replaced by a `DOMParser` call as inspired by the Turbo Drive loading code.

```js
function fragmentFromHTML(html) {
  if (html) {
    const { documentElement } = new DOMParser().parseFromString(html, "text/html");
    return documentElement;
    // const foreignDocument = document.implementation.createHTMLDocument();
    // return foreignDocument.createRange().createContextualFragment(html);
  }
}
```

Please let me know if I can be of further assistance!

guitarneck:
Fixes #123

*document.implementation.createHTMLDocument()*, under Firefox, create a new document with baseURI pointing to `about:blank`. Its breaks any relative links in the loaded fragment..

Adding `<base>` with the attribute `href` initialised with the document.baseURI resolves it.


src/core/frames/frame_controller.ts
```diff
@@ -314,6 +314,12 @@ function readScrollLogicalPosition(value: string | null, defaultValue: ScrollLog
 function fragmentFromHTML(html?: string) {
   if (html) {
     const foreignDocument = document.implementation.createHTMLDocument()
```
stgm, line 316: @sstephenson Would you be able to comment on why it was needed to use a "foreign document" here? I'm asking because it would be possible to use `createContextualFragment()` on the current document and that also seems to work.

In short, this combination of a foreign document with `createContextualFragment()` apparently prevents Firefox to load SVG sprite icons (see the original issue). It would be interesting to understand the rationale behind using it.

sstephenson:
Closing this since I believe it was fixed in #133.

PR #133, opened 2021-01-25
Unify page and frame rendering

sstephenson:
This branch unifies the rendering system for pages and frames. Broadly speaking, rendering is the job of three collaborators: Snapshot, View, and Renderer.

* A Snapshot wraps a single DOM element and encapsulates queries on that element. Snapshots can represent an element currently on the page, an element about to be loaded onto the page, or an element from a remotely loaded document. After the changes in this branch, snapshots are now used for error pages as well as for frames.
* A View creates snapshots and performs renderings. This branch moves common behavior into the base View class and page-specific concerns into the PageView subclass, and introduces a FrameView subclass owned by FrameController.
* A Renderer is responsible for transitioning a view from its current snapshot to a new snapshot. Rendering is now an asynchronous process, and each renderer has a promise which resolves after rendering is complete. This branch expands the Renderer hierarchy to ErrorRenderer, PageRenderer, and FrameRenderer subclasses.

Possible future functionality this branch aims to simplify:

* Support for data-turbo-permanent elements in frames (#71)
* Rendering events for frames (#59)
* Custom rendering hooks for pages and frames
* Page transitions scoped to an element other than \<body>


src/core/drive/page_renderer.ts
```diff
@@ -127,42 +113,42 @@ export class SnapshotRenderer extends Renderer {
...
 
   focusFirstAutofocusableElement() {
-    const element = this.newSnapshot.findFirstAutofocusableElement()
+    const element = this.newSnapshot.firstAutofocusableElement
     if (elementIsFocusable(element)) {
       element.focus()
     }
   }
 
   getNewHeadStylesheetElements() {
```
seanpdoyle, line 130: I believe these are private functions, so maybe the payoff is limited, but is now a good opportunity to re-structure these as properties?:

```js
get newHeadStylesheetElements() {
  // ...
}
```

src/core/drive/page_snapshot.ts
```diff
@@ -0,0 +1,63 @@
...
+  static fromHTMLString(html = "") {
+    return this.fromDocument(parseHTMLDocument(html))
+  }
+
+  static fromElement(element: Element) {
+    return this.fromDocument(element.ownerDocument)
+  }
+
+  static fromDocument({ head, body }: Document) {
+    return new this(body as HTMLBodyElement, new HeadSnapshot(head))
```
seanpdoyle, line 16: I'm surprised that this `HTMLBodyElement` casting is necessary given the parameterized type in the class definition, and that `document.body` doesn't return an `HTMLBodyElement`. Is TypeScript guarding against `document.body` returning an instance of [HTMLFrameSetElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFrameSetElement)?

src/core/drive/page_renderer.ts
sstephenson: Absolutely, yeah, these should be getters.

src/core/drive/page_snapshot.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  get isVisitable() {
+    return this.getSetting("visit-control") != "reload"
+  }
+
+  // Private
+
+  getSetting(name: string): string | undefined
+  getSetting(name: string, defaultValue: string): string
+  getSetting(name: string, defaultValue?: string) {
```
seanpdoyle, line 59: What makes these overloaded type signatures necessary or beneficial instead of the most general `getSetting(name: string, defaultValue?: string): string | undefined`?

src/core/drive/page_snapshot.ts
sstephenson: Yeah, I wonder if it's for framesets? Or maybe some kind of XML doc that you might somehow have as a Document? Either way, `document.body` is an `HTMLElement` as far as the built-in typings are concerned.

seanpdoyle (approved):
This is exciting work!

src/core/drive/page_snapshot.ts
sstephenson: In this case, it's so when you call getSetting with a default value, the return type is explicitly a string (and you don't have to check for undefined).

I think there's probably a better/fancier way to write it though 😅

sstephenson: I think the reason this code looks the way it does is because it was ported from CoffeeScript to TypeScript before nullish coalescing was added to the language. It can all go away: a865647

PR #135, opened 2021-01-26
Fix navigating back to same page anchor

domchristie:
/cc @seanpdoyle

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -84,7 +84,7 @@ export class Visit implements FetchRequestDelegate {
     this.location = location
     this.isSamePage = (
       getAnchor(location) != null &&
-      getRequestURL(location) === getRequestURL(new URL(window.location.href))
+      getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation)
     )
 
     this.restorationIdentifier = restorationIdentifier || uuid()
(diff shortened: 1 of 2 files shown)


seanpdoyle:
Thanks @domchristie . Could you draft up a test to cover this behavior, or at least outline the steps you have in mind to exercise it?

domchristie:
Yep, sure! 👍

domchristie:
This has unearthed an issue with the current version on `main`—scroll positions don't seem to be restored on restoration visits. It looks like the `restorationIdentifier` is not passed through in: https://github.com/hotwired/turbo/blob/bf3c3b0ad56249fa6a3eec3dd59f5d84ec897513/src/core/session.ts#L108

seanpdoyle:
Thank you @domchristie! I'd like to list you as a co-author on the other branch. What is your git configured email address?

domchristie:
Thanks @seanpdoyle! It's christiedom at the big search company's email service :)

With regards to this PR, I'm unsure about where the logic for `isSamePage` should live. It feels like the navigator would be a sensible place. This would make it available in the native adaptors, and it could pass it to the visit on creation. The navigator does have access to the view and therefore also `View#lastRenderedLocation`. I feel like `lastRenderedLocation` is most representative of the navigation state (vs `Navigator#location`), particularly because tapping back will update the address bar before Turbo has rendered. I'd welcome any thoughts on this :)

PR #140, opened 2021-01-28
Test scroll position when page invalidated

domchristie:
Related: #63 and https://github.com/turbolinks/turbolinks/issues/181


domchristie:
Seems fixed now! 🎉

PR #152, opened 2021-02-03
Avoid request on same page

ozguruysal:
Aims to fix the issue explained at https://github.com/hotwired/turbo/issues/42#issuecomment-772725421.


dhh:
Hey @ozguruysal, I see that we didn't get this into main yet. If you're still interested pushing this forward, could you resolve the conflicts? I'll get it into main.

ozguruysal:
hi @dhh I guess I managed to resolve the conflicts. I originally sent the PR to `skip-link` branch but looks like it's already merged so I changed the target branch to `main` now. Thanks for looking into this.

ozguruysal:
Not sure what's going on here but if I fix the TS error that triggers the build error it causes a conflict. If I don't fix it,  no conflict but build fails. @dhh can you help fix this?

dhh:
Looks like there's still a merge conflict?

ozguruysal:
Correct but fixing it causes the build to fail. See [here](https://github.com/hotwired/turbo/runs/2870492258?check_suite_focus=true).

dhh:
Ah, I see. That error requires us to make sure that the argument is not null. Either through the TS bang method or a guard clause.

ozguruysal:
Yes, I fixed it [here](https://github.com/hotwired/turbo/pull/152/commits/f03079fd4864d13d4ef63a58d0280c1709ce10e1), but now it complains about the conflict. 😕

marcelolx:
@ozguruysal I think the problem is that you need to rebase this branch, the commits from the print below aren't related to your changes, right?
image

ozguruysal:
Thanks @marcelolx. @dhh I guess I've finally managed to fix the problems.

dhh:
This includes a different way to fix the history pop issue that I fixed in another branch. So let's at least remove that.

Don't know if @domchristie have any other observations on this that we should incorporate as well?

domchristie:
There's a fair bit of history on this, and I've not gone through these changes fully, but on first look I think it'd be good to reinstate the navigation tests. Also, I'm not totally sure how new native adapters interact with Turbo, but I remember having to [update the turbolinks-ios adapter](https://github.com/turbolinks/turbolinks-ios/pull/126) to be compatible with this change. I can have a closer look tomorrow.

ozguruysal:
Just let me know if you'd prefer a new PR as this is getting way too complicated. The fix I I introduced is a simple one anyways.

dhh:
@ozguruysal Maybe it is easier to split this PR into a few smaller chunks.

domchristie:
I think https://github.com/hotwired/turbo/compare/main...domchristie:skip-link-2 is clean and pretty close. Just checking if there's anything else needed to add…

dhh:
Would be great to get that fix in before we consider going with Turbo 7.0.0 final 👍

ozguruysal:
@domchristie feel free to integrate [this fix](https://github.com/hotwired/turbo/pull/152/commits/c84ad01a9ece7795b0380d72ae9a5a55105f698f) to your branch. Previously I sent the PR to your `skip-link` branch but it wasn't merged. It's quite related anyways.

domchristie:
@ozguruysal could you explain why that order matters? as opposed to: https://github.com/hotwired/turbo/pull/298/files#diff-78d8451f964182fd51330bac500ba6e71234f81aad9e8a57e682e723d0f517b3R368

ozguruysal:
I explained the issue [here](https://github.com/hotwired/turbo/issues/42#issuecomment-772725421). Overall, I guess if the target is an anchor on the same page, there is no point in making any request so it's logical to check that first and return false quickly.

domchristie:
@ozguruysal Thanks. I've added your fix here: https://github.com/hotwired/turbo/pull/298/commits/0e4193f7b6a99a576f92ff7fb927783d4ed88c32.

dhh:
Let's continue the work in https://github.com/hotwired/turbo/pull/298.

PR #156, opened 2021-02-04
turbo-frame[busy] attribute

seanpdoyle:
During the request lifecycle, `<turbo-frame>` elements will toggle the
`[busy]` boolean attribute to true when the request starts, and then
remove it when the request ends.

Additionally, form submissions were not triggering the same `[busy]` toggling behavior, so this commit adds coverage for that behavior as well.

This commit adds functional test coverage for that behavior.

### DIFF
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -25,6 +25,13 @@ export class FrameTests extends TurboDriveTestCase {
     this.assert.equal(otherEvents.length, 0, "no more events")
   }
 
+  async "test following a link driving a frame toggles the [busy] attribute"() {
+    await this.clickSelector("#hello a")
+
+    this.assert.equal(await this.nextAttributeMutationNamed("frame", "busy"), "", "sets [busy] on the #frame")
+    this.assert.equal(await this.nextAttributeMutationNamed("frame", "busy"), null, "removes [busy] from the #frame")
+  }
+
   async "test following a link to a page without a matching frame results in an empty frame"() {
     await this.clickSelector("#missing a")
     await this.nextBeat
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -186,7 +186,8 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   // Form submission delegate
 
   formSubmissionStarted(formSubmission: FormSubmission) {
-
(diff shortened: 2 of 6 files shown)


seanpdoyle:
Regardless of how we feel about aria-busy, I think the first commit to add test coverage for the attribute is worthwhile.

src/tests/functional/frame_tests.ts
```diff
@@ -5,6 +5,15 @@ export class FrameTests extends FunctionalTestCase {
     await this.goToLocation("/src/tests/fixtures/frames.html")
   }
 
+  async "test following a link driving a frame toggles the [busy] attribute"() {
```
seanpdoyle, line 8: We need to add coverage for form submissions, too.

seanpdoyle:
Since the original work uncovered some bugs we'd like to ship more immediately, I've split out https://github.com/hotwired/turbo/pull/157 to serve as the conversation venue for `aria-busy` instead of `busy`.

src/tests/helpers/functional_test_case.ts
```diff
@@ -62,6 +62,26 @@ export class FunctionalTestCase extends InternTestCase {
     return offset > -1 && offset < 1
   }
 
+  async observeAttributeChanges(elementId: string, attribute: string) {
```
seanpdoyle, line 65: @javan @sstephenson this feels unwieldy, but it was the best idea I could come up with for tracking whether or not a test toggled a boolean attribute. All other ideas were consistently losing the race between submission and subsequent page load that were unsetting the attribute.

Is there a tool in our testing harness that could better provide a seam between submission and response? A `?sleep` query parameter, maybe?

seanpdoyle: Creating a general purpose assertion would be useful in case we wanted to test that submitter buttons are marked as `[disabled]` upon submit and toggled off when the request finished.

sstephenson:
After rebasing against main, I see this failure in Firefox on macOS:

```
× firefox on mac 20.3.0 - NavigationTests - skip link with hash-only path moves focus and changes tab order (0.331s)
    AssertionError: skips to first interactive element after #main: expected false to be truthy
      at NavigationTests.test skip link with hash-only path moves focus and changes tab order @ src/tests/functional/navigation_tests.ts:162:16
      at runMicrotasks @ anonymous
      at processTicksAndRejections @ internal/process/task_queues.js:93:5
```

sstephenson:
❤️

PR #157, opened 2021-02-04
Toggle [aria-busy] instead of [busy]

seanpdoyle:
Since `<turbo-frame>` elements are custom elements, the framework has
total control over the names of the attributes.

There are existing semantics for what we've introduced as `[busy]`: the
ARIA guidelines suggest toggling [aria-busy="true"][aria-busy] when an
element is loading more content, and `aria-busy="false"` when the
content is loaded.

This provides an "interface" for loading styles through CSS attribute
selectors, and hints to assistive technologies the state of the frame.

As an alternative, we could continue to toggle the `[busy]` attribute,
and encourage consumer applications to monitor mutations to the `[busy]`
attribute (or listen to [`turbo:frame-visit` and `turbo:frame-load`
events][events]) to toggle it themselves.

[aria-busy]: https://www.w3.org/TR/wai-aria-1.1/#aria-busy
[events]: https://github.com/hotwired/turbo/pull/59

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -129,7 +129,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   requestStarted(request: FetchRequest) {
-    this.element.setAttribute("busy", "")
+    this.element.setAttribute("aria-busy", "true")
   }
 
   requestPreventedHandlingResponse(request: FetchRequest, response: FetchResponse) {
@@ -152,13 +152,14 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   requestFinished(request: FetchRequest) {
-    this.element.removeAttribute("busy")
+    this.element.removeAttribute("aria-busy")
   }
 
   // Form submission delegate
 
   formSubmissionStarted(formSubmission: FormSubmission) {
-
+    const frame = this.findFrameElement(formSubmission.formElement)
+    frame.setAttribute("aria-busy", "true")
   }
 
   formSubmissionSucceededWithResponse(formSubmission: FormSubmission, response: FetchResponse) {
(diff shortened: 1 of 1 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -152,13 +152,16 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
...
+    this.element.removeAttribute("aria-busy")
   }
 
   // Form submission delegate
 
   formSubmissionStarted(formSubmission: FormSubmission) {
-
+    const frame = this.findFrameElement(formSubmission.formElement)
+    if (frame) {
+      frame.setAttribute("aria-busy", "")
```
javan, line 163: Should the value be `"true"` here like above? https://github.com/hotwired/turbo/blob/0c2bd1446fb202eed2a234ba1ccc7250bedba950/src/core/frames/frame_controller.ts#L131-L132

src/core/frames/frame_controller.ts
```diff
@@ -152,13 +152,16 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
...
-    this.element.removeAttribute("busy")
+    this.element.removeAttribute("aria-busy")
   }
 
   // Form submission delegate
 
   formSubmissionStarted(formSubmission: FormSubmission) {
-
+    const frame = this.findFrameElement(formSubmission.formElement)
+    if (frame) {
```
javan, line 162: Is `frame` ever *not* present? Looks like it falls back to `this.element`: https://github.com/hotwired/turbo/blob/0c2bd1446fb202eed2a234ba1ccc7250bedba950/src/core/frames/frame_controller.ts#L220-L222

javan (approved):

src/core/frames/frame_controller.ts
seanpdoyle: @javan ah, I must've missed this while splitting out the separate `[aria-busy]` PR from the original: https://github.com/hotwired/turbo/pull/156/files#diff-8f44fafd854f6f716d51a2b05b5689266a9b66e0b1ad9d68570bd4e8cefd0a47R161

src/core/frames/frame_controller.ts
seanpdoyle: Yes! Thank you

Issue #147, opened 2021-02-01
Implement progressbar with <progress>


rmarronnier:
Right now Turbo can add a progressbar to the DOM.
It does this by inserting a div. Why not use the standard progress HTML element ? This way, no need for a default CSS styling. WDYT ?

Issue #61, opened 2020-12-30
Progress bar not displayed when submitting a form


coorasse:
I noticed that the progress bar is not displayed when submitting a form.
While trying to understand the reason, I noticed that `showProgressBarAfterDelay` is called only after the submit is finished.
The method is called by [`visitRequestStarted`](https://github.com/hotwired/turbo/blob/aae160bed764ca7322be1e3a9e00366787d96f7f/src/core/native/browser_adapter.ts#L28), called by `startRequest`, and by keep going back into the stack, they are called into [`formSubmissionSucceededWithResponse`](https://github.com/hotwired/turbo/blob/aae160bed764ca7322be1e3a9e00366787d96f7f/src/core/drive/navigator.ts#L72), so when the submit is already terminated.

The result is that the progress ber is not displayed.

Is this a bug, or did I miss something?

jeffochoa:
I'm having the same issue, it looks like you have to redirect AFTER handling the form-submission:

```text
Turbo Drive handles form submissions in a manner similar to link clicks. The key difference is that form submissions can issue stateful requests using the HTTP POST method, while link clicks only ever issue stateless HTTP GET requests.

After a stateful request from a form submission, Turbo Drive expects the server to return an HTTP 303 redirect response, which it will then follow and use to navigate and update the page without reloading.
```

That doesn't work for me because I want to redirect the users to a different page, and that extra request is not ideal.

An option for me would be to disable Turbo on the actual form, but right now that's not possible, although there is an open PR for that https://github.com/hotwired/turbo/pull/3

Hopefully that one will get merged soon 🤞 

@coorasse please let me know if you find a workaround for this issue.

Cheers,
Jeff.

jeffochoa:
Ok, I found a temporal solution. Using the custom turbo events, you can capture the `submit-start` to load the progress-bar manually as follows:

```script
<script>
    document.addEventListener('turbo:submit-start', function () {
        Turbo.navigator.delegate.adapter.showProgressBar();
    });
</script>
```

sudara:
I [ posted on the Hotwire forums](https://discuss.hotwire.dev/t/2-usability-niggles-re-turbo-drive-frame-and-forms/1969) asking whether it should be default for the progress bar to always show on form submission. Otherwise, repeatedly rejected forms can appear as no-ops...

wnm:
> Ok, I found a temporal solution. Using the custom turbo events, you can capture the `submit-start` to load the progress-bar manually as follows:
[...]

PR #164, opened 2021-02-07
Render progress bar during form submissions

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/61
Closes https://github.com/hotwired/turbo/issues/147

---

Extends the `Adapter` interface to support new
`formSubmissionStarted(FormSubmission)` and
`formSubmissionFinished(FormSubmission)` delegate methods to hook into a
page's form submission lifecycle. Making the Adapter aware of form
submissions provides an opportunity to show and hide the progress bar in
the same way as driving the page with Visits.

Additionally, replace the `HTMLDivElement` implementation of the
progress bar with a [progress][] element and an instance of
[HTMLProgressElement][]. When updating the progress, set the
`HTMLProgressElement.value` attribute, and let the browser-native
implementation update its width. Since the element's `value` attribute
is accessible via a property, we no longer have to synchronize via
`requestAnimationFrame()`.

To support the built-in element, there are some proprietary
pseudoelements to consider across the browsers, including:

* [-moz-progress-bar][] for the Firefox bar
* [-webkit-progress-bar][] for the Safari and Chrome bar background
* [-webkit-progress-value][] for the Safari and Chrome bar progress
* [-ms-fill][] for IE bars

[progress]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress
[HTMLProgressElement]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLProgressElement
[-moz-progress-bar]: https://developer.mozilla.org/en-US/docs/Web/CSS/::-moz-progress-bar
[-webkit-progress-bar]: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-progress-bar
[-webkit-progress-value]: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-progress-value
[-ms-fill]: https://developer.mozilla.org/en-US/docs/Archive/Web/CSS/::-ms-fill


src/core/native/adapter.ts
```diff
@@ -10,5 +11,7 @@ export interface Adapter {
   visitRequestFailedWithStatusCode(visit: Visit, statusCode: number): void
   visitRequestFinished(visit: Visit): void
   visitRendered(visit: Visit): void
+  formSubmissionStarted(formSubmission: FormSubmission): void
+  formSubmissionFinished(formSubmission: FormSubmission): void
```
seanpdoyle, line 15: @sstephenson @javan what is necessary for parity across Web/iOS/Android when changing the adapter interface?

If it requires a lot of effort, maybe it would be worthwhile to unify the `visitRequestStarted` and `visitRequestFinished` along with the new `formSubmissionStarted` and `formSubmissionFinished` methods and replace them with something more general like `fetchRequestStarted` and `fetchRequestEnded`?

seanpdoyle:
A fix for #61 (showing the progress on form submissions) deserves its own PR, separate from the fix for #147 (turning `<div class="turbo-progress-bar">` into `<progress class="turbo-progress-bar">`), we could likely tease out two separate PRS.

src/core/native/adapter.ts
seanpdoyle: We had previously proposed that we use `[data-turbo-busy]` on all elements as an indicator that a request is in-flight (https://github.com/hotwired/turbo/pull/199). If we were able to consistently set that attribute, we could manage the visibility and animation of a `<progress>` or `<div role="progress">` element through CSS or a `MutationObserver` instance.

That might be a more simplified implementation.

santib:
Any updates on this? Is there any reason to not have the progress bar on form submissions?

Issue #98, opened 2021-01-09
Lazy loading Turbo frame with same src causes infinite request loop


michiels:
Hi everyone,

I've found that Turbo will cause an infinite request loop if you lazy load a frame using `src` when the frame in the response also has the same value in `src`.

Let me explain:

The first cold response contains something like:

```
<turbo-frame src="/account">
  Grabbing auth tokens and logging you in...
</turbo-frame>
```

Through some external logic this page sets an auth token in local storage or cookie.

So the subsequent lazy loaded request to `/account` will have identified the user and will be something like:

```
<turbo-frame src="/account">
  Hi user! Welcome!
</turbo-frame>
```

It appears that Turbo lazy-loads again seeing this `src="/account"` in the response of the `fetch()`.

Because both the original frame has `src="/account"` and the new turbo frame has `src="account"` I'm seeing an infinite loop being started.

My local fix is now is to not put the `src="/account"` in the authenticated request response. This works, but it requires some extra logic in the view that I use to render the view of this frame. Eg:

```
<turbo-frame id="firmhouse-page-body" <%= 'src="/apps/firmhouse/orders"' unless @subscription.present? %>>
  <p>Loading orders...</p>

  <% if @subscription.present? %>
    <p>Hello, these are my orders.</p>

    <%= link_to "Back to Dashboard", "/apps/firmhouse/dashboard" %>
  <% end %>
</turbo-frame>
```

But perhaps it makes sense to have Turbo check if the `src` attribute is the same as the previous one or referer, and then not kick off a lazy load but just treat it as the final request.

michiels:
Perhaps to consider in addition. Why would Turbo kick off a lazy-load of the frame it has just replaced? Maybe it can just stop lazy-loading if it has just loaded a frame it has just replaced?

It should of course still lazy-load any children frames in the response. But lazy-loading the frame it has just lazy loaded seems to be something that wouldn't happen in practice.

PR #165, opened 2021-02-07
Prevent infinite looping when loading frames

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/98
Closes https://github.com/hotwired/turbo/issues/182

When a `<turbo-frame>` element declares a `[src]` attribute that
references the URL the frame is loaded from, prevent an infinitely
looping chain of requests.

To do so, extend the `FrameElement.isActive` predicate to also check for
[Node.isConnected][].

[Node.isConnected]: https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -6,7 +6,7 @@ import { parseHTMLDocument } from "../../util"
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { ViewDelegate } from "../view"
-import { expandURL, Locatable } from "../url"
+import { expandURL, urlsAreEqual, Locatable } from "../url"
 import { FormInterceptor, FormInterceptorDelegate } from "./form_interceptor"
 import { FrameView } from "./frame_view"
 import { LinkInterceptor, LinkInterceptorDelegate } from "./link_interceptor"
@@ -53,7 +53,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   sourceURLChanged() {
-    if (this.loadingStyle == FrameLoadingStyle.eager && this.element.isConnected) {
+    if (this.loadingStyle == FrameLoadingStyle.eager) {
       this.loadSourceURL()
     }
   }
@@ -232,16 +232,21 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     let element
     const id = CSS.escape(this.id)
 
-    if (element = activateElement(container.querySelector(`turbo-frame#${id}`))) {
-      return element
+    try {
+      if (element = activateElement(container.querySelector(`turbo-frame#${id}`), this.currentURL)) {
+        return element
+      }
+
+      if (element = activateElement(container.querySelector(`turbo-frame[src][recurse~=${id}]`), this.currentURL)) {
+        await element.loaded
+        return await this.extractForeignFrameElement(element)
+      }
+
+      console.error(`Response has no matching <turbo-frame id="${id}"> element`)
+    } catch (error) {
+      console.error(error)
     }
 
-    if (element = activateElement(container.querySelector(`turbo-frame[src][recurse~=${id}]`))) {
-      await element.loaded
-      return await this.extractForeignFrameElement(element)
-    }
-
-    console.error(`Response has no matching <turbo-frame id="${id}"> element`)
     return new FrameElement()
   }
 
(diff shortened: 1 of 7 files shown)


ghiculescu:
Just flagging that this will also fix https://discuss.hotwire.dev/t/trouble-with-unexpected-fetch-request-when-changing-pages

src/core/frames/frame_controller.ts
```diff
@@ -45,7 +45,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   sourceURLChanged() {
-    if (this.loadingStyle == FrameLoadingStyle.eager) {
+    if (this.loadingStyle == FrameLoadingStyle.eager && this.element.isConnected) {
```
sstephenson, line 48: The recursive frame tests fail after rebasing this against main; we'll need another way to detect frames which reference themselves. I think the guard belongs in `activateElement()` before the call to `document.importNode()`.

src/tests/functional/frame_tests.ts
```diff
@@ -15,6 +15,18 @@ export class FrameTests extends FunctionalTestCase {
     this.assert.equal(await frame.getAttribute("data-loaded-from"), currentPath)
   }
 
+  async "test a frame whose src references itself does not infinitely loop"() {
+    await this.clickSelector("#frame-self")
+
+    await this.nextEventNamed("turbo:before-fetch-response")
+
+    this.assert.equal(await this.getVisibleText("h2"), "Frames: Self")
```
sstephenson, line 23: Not sure I agree with this assertion… I think maybe we ought to avoid loading frames when we know they reference themselves, and log a specific error message explaining the problem.

sstephenson (changes requested):

sstephenson:
Here's a half-baked proposal:

```diff
diff --git a/src/core/frames/frame_controller.ts b/src/core/frames/frame_controller.ts
index 20ebbe8..36a458a 100644
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -53,7 +53,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   sourceURLChanged() {
-    if (this.loadingStyle == FrameLoadingStyle.eager && this.element.isConnected) {
+    if (this.loadingStyle == FrameLoadingStyle.eager) {
       this.loadSourceURL()
     }
   }
@@ -232,16 +232,22 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     let element
     const id = CSS.escape(this.id)
 
-    if (element = activateElement(container.querySelector(`turbo-frame#${id}`))) {
-      return element
-    }
+    try {
+      if (element = activateElement(container.querySelector(`turbo-frame#${id}`), this.currentURL)) {
+        return element
+      }
 
-    if (element = activateElement(container.querySelector(`turbo-frame[src][recurse~=${id}]`))) {
-      await element.loaded
-      return await this.extractForeignFrameElement(element)
+      if (element = activateElement(container.querySelector(`turbo-frame[src][recurse~=${id}]`), this.currentURL)) {
+        await element.loaded
+        return await this.extractForeignFrameElement(element)
+      }
+
+      console.error(`Response has no matching <turbo-frame id="${id}"> element`)
+
+    } catch (error) {
+      console.error(error)
     }
 
-    console.error(`Response has no matching <turbo-frame id="${id}"> element`)
     return new FrameElement()
   }
 
@@ -300,13 +306,24 @@ function getFrameElementById(id: string | null) {
   }
 }
 
-function activateElement(element: Node | null) {
-  if (element && element.ownerDocument !== document) {
-    element = document.importNode(element, true)
-  }
+function activateElement(element: Element | null, currentURL?: string) {
+  if (element) {
+    const src = element.getAttribute("src")
+    if (src != null && currentURL != null && urlsAreEqual(src, currentURL)) {
+      throw new Error(`Matching <turbo-frame id="${element.id}"> element has a source URL which references itself`)
+    }
 
-  if (element instanceof FrameElement) {
-    element.connectedCallback()
-    return element
+    if (element.ownerDocument !== document) {
+      element = document.importNode(element, true)
+    }
+
+    if (element instanceof FrameElement) {
+      element.connectedCallback()
+      return element
+    }
   }
 }
+
+function urlsAreEqual(left: string, right: string) {
+  return new URL(left, document.baseURI).href == new URL(right, document.baseURI).href
+}
```

seanpdoyle:
@sstephenson thank you for the feedback. I've pushed up changes to incorporate your suggestions.

PR #166, opened 2021-02-07
Prepare Frame Form Submission fetch headers

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/86
Closes https://github.com/hotwired/turbo/pull/110

When submitting a Form that is within a `<turbo-frame>` or targets a
`<turbo-frame>`, ensure that the `Turbo-Frame` header is present.

Since the constructive-style
`FetchRequestDelegate.additionalHeadersForRequest()` was replaced by the
mutative style `FetchRequestDelegate.prepareHeadersForRequest()`, this
commit introduces a readonly `FetchRequestHeaders` property created at
constructor-time so that its values can be mutated prior to the
request's submission.

Co-authored-by: tleish <tleish@users.noreply.github.com>

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -144,6 +144,8 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     if (this.formSubmission.fetchRequest.isIdempotent) {
       this.navigateFrame(element, this.formSubmission.fetchRequest.url.href)
     } else {
+      const { fetchRequest } = this.formSubmission
+      this.prepareHeadersForRequest(fetchRequest.headers, fetchRequest)
       this.formSubmission.start()
     }
   }
--- a/src/http/fetch_request.ts
+++ b/src/http/fetch_request.ts
@@ -42,6 +42,7 @@ export interface FetchRequestOptions {
 export class FetchRequest {
   readonly delegate: FetchRequestDelegate
   readonly method: FetchMethod
+  readonly headers: FetchRequestHeaders
   readonly url: URL
   readonly body?: FetchRequestBody
   readonly abortController = new AbortController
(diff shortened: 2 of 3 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -118,6 +118,8 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     if (this.formSubmission.fetchRequest.isIdempotent) {
       this.navigateFrame(element, this.formSubmission.fetchRequest.url.href)
     } else {
+      const { fetchRequest } = this.formSubmission
+      this.prepareHeadersForRequest(fetchRequest.headers, fetchRequest)
```
seanpdoyle, line 122: My main instinct was to prefer a more explicit, layered mutation over the recursive style proposed by https://github.com/hotwired/turbo/pull/110/files#diff-68b647dc2963716dc27c070f261d0b942ee9c00be7c4149ecb3a5acd94842d40, but the more I pulled on that thread, the more changes it required to stick.

Is there an alternative delegator pattern that could still leverage the late-binding values that the `get headers()` property provides without recursively walking a chain of objects that coincidentally declare `prepareHeadersForRequest()`?

tleish: the only impact we are aware of right now involves forms.  We could consider isolating the change to forms only.  Something like:

`src/core/drive/form_submission.ts`
```diff
  prepareHeadersForRequest(headers: FetchRequestHeaders, request: FetchRequest) {
    if (!request.isIdempotent) {
      const token = getCookieValue(getMetaContent("csrf-param")) || getMetaContent("csrf-token")
      if (token) {
        headers["X-CSRF-Token"] = token
      }
      headers["Accept"] = [ StreamMessage.contentType, headers["Accept"] ].join(", ")
+     if (typeof this.delegate.prepareHeadersForRequest == "function") {
+       this.delegate.prepareHeadersForRequest(headers, request);
+     }
    }
  }
```

seanpdoyle: That's a good point. What bothers me the most is that checking that `this.delegate` (a `FormSubmissionDelegate` instance) is of a different interface than the `FetchRequestDelegate`. The fact that they both declare `prepareHeadersForRequest` is a coincidence that signals to me that there's some more Typescript-driven design thinking we need to do to chain these things together in a more type-aware manner.

tleish: Agreed.

sstephenson (changes requested):

sstephenson:
Excellent, thank you!

PR #169, opened 2021-02-08
Focus [autofocus] on navigation and frame load 

seanpdoyle:
Focus [autofocus] on navigation and frame load
===

According to the `<input>` element's documentation for the
`[autofocus]` attribute:

> A Boolean attribute which, if present, indicates that the input should
> automatically have focus when the page has finished loading
>
>
> Note: An element with the `[autofocus]` attribute may gain focus
> before the `DOMContentLoaded` event is fired. No more than one element
> in the document may have the `[autofocus]` attribute. If put on more
> than one element, the **first one with the attribute receives focus**.

[autofocus]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#htmlattrdefautofocus

Focus the Snapshot that is connected to the DOM
===

During the rendering process, the `PageRenderer` replaces the current
`Snapshot` with the contents of the new `Snapshot`, while the
`FrameRenderer` extracts the contents of the new Snapshot and uses it to
replace the contents of the current `Snapshot`.

Instead of accepting a `Snapshot` as an argument,
`Renderer.focusFirstAutofocusableElement()` uses whichever `Snapshot` is
connected to the current document.

Co-authored-by: Javan Makhmali <javan@basecamp.com>

Test coverage to preserve turbo-frame attributes
===

Add test coverage for preserving turbo-frame attributes during render,
making sure that any element state is preserved while the contents are
modified.

Skip autofocus during previews
===

Pass the `isPreview` value to the `View.renderPage()` call, and then
restore the guard clause removed by [a7d02f0][].

[a7d02f0]: https://github.com/hotwired/turbo/commit/a7d02f0fde3d95d6e3af9ec72b7654d3fbeedd33

### DIFF
--- a/src/core/drive/page_renderer.ts
+++ b/src/core/drive/page_renderer.ts
@@ -16,9 +16,7 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
 
   finishRendering() {
     super.finishRendering()
-    if (this.isPreview) {
-      this.focusFirstAutofocusableElement()
-    }
+    this.focusFirstAutofocusableElement()
   }
 
   get currentHeadSnapshot() {
--- a/src/core/drive/page_renderer.ts
+++ b/src/core/drive/page_renderer.ts
@@ -16,7 +16,9 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
 
   finishRendering() {
     super.finishRendering()
-    this.focusFirstAutofocusableElement()
+    if (!this.isPreview) {
+      this.focusFirstAutofocusableElement()
+    }
   }
 
   get currentHeadSnapshot() {
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -234,7 +234,7 @@ export class Visit implements FetchRequestDelegate {
       const isPreview = this.shouldIssueRequest()
       this.render(async () => {
         this.cacheSnapshot()
-        await this.view.renderPage(snapshot)
+        await this.view.renderPage(snapshot, isPreview)
         this.adapter.visitRendered(this)
         if (!isPreview) {
           this.complete()
--- a/src/core/renderer.ts
+++ b/src/core/renderer.ts
@@ -59,10 +59,14 @@ export abstract class Renderer<E extends Element, S extends Snapshot<E> = Snapsh
   }
 
   focusFirstAutofocusableElement() {
-    const element = this.newSnapshot.firstAutofocusableElement
+    const element = this.connectedSnapshot.firstAutofocusableElement
     if (elementIsFocusable(element)) {
       element.focus()
     }
+  }
+
+  get connectedSnapshot() {
+    return this.newSnapshot.isConnected ? this.newSnapshot : this.currentSnapshot
   }
 
   get currentElement() {
--- a/src/core/snapshot.ts
+++ b/src/core/snapshot.ts
@@ -21,6 +21,10 @@ export class Snapshot<E extends Element = Element> {
     }
   }
 
+  get isConnected() {
+    return this.element.isConnected
+  }
+
   get firstAutofocusableElement() {
     return this.element.querySelector("[autofocus]")
   }
--- a/src/tests/fixtures/autofocus.html
+++ b/src/tests/fixtures/autofocus.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Autofocus</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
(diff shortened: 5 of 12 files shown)


src/core/drive/page_renderer.ts
```diff
@@ -16,9 +16,7 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
 
   finishRendering() {
     super.finishRendering()
-    if (this.isPreview) {
-      this.focusFirstAutofocusableElement()
-    }
+    this.focusFirstAutofocusableElement()
```
seanpdoyle, line 19: @javan @sstephenson what are your thoughts on this change? What's the incentive for not autofocusing elements when navigating to a page that is not a preview?

javan (approved):
Confirmed this fixes the issue I was seeing. Thank you!

src/core/drive/page_renderer.ts
javan: I think it's backwards, currently. It used to be https://github.com/hotwired/turbo/blob/269fa2436495a720ceb67215dfa8abe4d4143492/src/core/drive/page_renderer.ts#L45-L47 and got flipped around (accidentally?) in https://github.com/hotwired/turbo/commit/a7d02f0fde3d95d6e3af9ec72b7654d3fbeedd33.

seanpdoyle: @javan oooh, great digging, thank you.

Just so that I'm clear in my understanding when I revert this change, do we want:

* autofocus on navigation
* do not autofocus when visiting a reviewable page
  * i.e. autofocus element on `/a`, navigate to `/b`, navigate back to `/a` without autofocusing)
* do not autofocus when navigating through History.goBack()

seanpdoyle: I can't manage to concoct a test to cover skipping autofocus during previews, so I'll share my attempts here for posterity's sake:

```diff
diff --git a/src/core/drive/page_renderer.ts b/src/core/drive/page_renderer.ts
index 6a83d45..94d2a3e 100644
--- a/src/core/drive/page_renderer.ts
+++ b/src/core/drive/page_renderer.ts
@@ -16,7 +16,9 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
 
   finishRendering() {
     super.finishRendering()
-    this.focusFirstAutofocusableElement()
+    if (!this.isPreview) {
+      this.focusFirstAutofocusableElement()
+    }
   }
 
   get currentHeadSnapshot() {
diff --git a/src/core/drive/visit.ts b/src/core/drive/visit.ts
index 7ba354a..1187f55 100644
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -234,7 +234,7 @@ export class Visit implements FetchRequestDelegate {
       const isPreview = this.shouldIssueRequest()
       this.render(async () => {
         this.cacheSnapshot()
-        await this.view.renderPage(snapshot)
+        await this.view.renderPage(snapshot, isPreview)
         this.adapter.visitRendered(this)
         if (!isPreview) {
           this.complete()
diff --git a/src/tests/fixtures/autofocus.html b/src/tests/fixtures/autofocus.html
index dfb1354..a4f2a35 100644
--- a/src/tests/fixtures/autofocus.html
+++ b/src/tests/fixtures/autofocus.html
@@ -21,5 +21,7 @@
     <turbo-frame id="drives-frame" target="frame">
       <a id="drives-frame-target-link" href="/src/tests/fixtures/frames/form.html">#drives-frame link to frames/form.html</a>
     </turbo-frame>
+
+    <a id="navigation-link" href="/src/tests/fixtures/navigation.html">navigation.html</a>
   </body>
 </html>
diff --git a/src/tests/functional/autofocus_tests.ts b/src/tests/functional/autofocus_tests.ts
index 5db02f0..c44004a 100644
--- a/src/tests/functional/autofocus_tests.ts
+++ b/src/tests/functional/autofocus_tests.ts
@@ -19,6 +19,19 @@ export class AutofocusTests extends TurboDriveTestCase {
     this.assert.notOk(await this.hasSelector("#second-autofocus-element:focus"), "focuses the first [autofocus] element on the page")
   }
 
+  async "test does not autofocus on preview visit"() {
+    await this.clickSelector("#navigation-link")
+    await this.nextBody
+    await this.clickSelector("#autofocus-link")
+
+    this.assert.notOk(await this.hasSelector("#first-autofocus-element:focus"), "does not autofocus when previewed")
+    this.assert.notOk(await this.hasSelector("#second-autofocus-element:focus"), "does not autofocus when previewed")
+
+    await this.nextBody
+
+    this.assert.ok(await this.hasSelector("#first-autofocus-element:focus"), "autofocuses after discarding the preview")
+  }
+
   async "test navigating a frame with a descendant link autofocuses [autofocus]:first-of-type"() {
     await this.clickSelector("#frame-inner-link")
     await this.nextBeat

```

PR #177, opened 2021-02-13
Add request interceptor

kirillplatonov:
## The problem

The modern server-side embedded Shopify apps are built with JWT session tokens which are obtained on the client-side and then passed to every server request. `Turbo` works pretty well with it since it allows to navigate application without doing page reloading on every click or page submission. The biggest issue is passing JWT token to every `Turbo` request. 

## The solution

This PR introduces request interceptor support. It allows to insert async function between every fetch request. Inside this function you can do some preparation for request and add custom headers.

## Example usage

```javascript
Turbo.setRequestInterceptor(async (request) => {
  const token = await getSessionToken(window.app);
  request.addHeader("Authorization", `Bearer ${token}`);
});
```


seanpdoyle:
Could this same seam for modifying headers be achieved with a `turbo:before-fetch-request` event listener?

```js
addEventListener("turbo:before-fetch-request", ({ detail: { fetchOptions }) => fetchOptions.headers["..."] = "..."))
```

kirillplatonov:
Changing headers with `turbo:before-fetch-request` event listener will work fine for the simpler case when you need to add some static header to `Turbo` requests. Unfortunately, with Shopify embedded apps you need to retrieve the JWT token before every request because the token lifetime is short and we don't know exactly when it's expired inside the Turbo app. JWT token retrieval is async and we need to pause Turbo's fetch request invocation until it's done. So async interceptor works best for this complex case. I used [Axios interceptors](https://github.com/axios/axios#interceptors) as an inspiration for this approach.

lonroth:
@kirillplatonov Nice work! Great that we won't need to fetch JWT every x seconds with your `setRequestInterceptor` solution and rather ask for them just when we need them. 👍

Are there additional pros using `Turbo.setRequestInterceptor` VS `"turbo:before-fetch-request`? Is there any downside in pausing Turbo's fetching request invocation until it's done?

The alternative as I see it is using `turbo:before-fetch-request` for Shopify Apps and a `keepRetrievingToken` function or similar which I assume is what @seanpdoyle was referring to above? You solution for sure looks better than the code below but would love to hear more from you regarding this and why you think it's not a good option to use `keepRetrievingToken`. Is another reason that the JWT might be expired when the call is made even though it was fetched just a few seconds ago?

```
document.addEventListener("turbo:before-fetch-request", function (event) {
 var headers = event.detail.fetchOptions.headers;
 headers["Authorization"] = "Bearer " + window.sessionToken;
});
```

We can use `keepRetrievingToken` or similar to fetch new tokens every x seconds:
```
  async function retrieveToken(app) {
    var SessionToken = window["app-bridge"].actions.SessionToken

    app.dispatch(
      SessionToken.request(),
    );

    // Save a session token for future requests
    window.sessionToken = await new Promise((resolve) => {
      app.subscribe(SessionToken.ActionType.RESPOND, (data) => {
        resolve(data.sessionToken || "");
      });
    });
  }

  function keepRetrievingToken(app) {
    setInterval(() => {
      retrieveToken(app);
    }, SESSION_TOKEN_REFRESH_INTERVAL);
  }
```

lonroth:
@kirillplatonov Nice work on this PR! Great that we won't need to fetch JWT every x seconds with your `setRequestInterceptor` solution and rather ask for them just when we need them. 👍

src/tests/functional/visit_tests.ts
```diff
@@ -80,6 +90,20 @@ export class VisitTests extends TurboDriveTestCase {
...
+    this.remote.execute(() => {
+      window.Turbo.setRequestInterceptor(async (request) => {
+        request.addHeader("Authorization", "Bearer Test Token")
+      })
+    })
+  }
+
+  async resetRequestInterceptor() {
+    this.remote.execute(() => {
+      window.Turbo.setRequestInterceptor(async (request) => {})
```
ghiculescu, line 104: thoughts on this approach?

```suggestion
      window.Turbo.setRequestInterceptor(null)
```

or

```suggestion
      window.Turbo.clearRequestInterceptor()
```

kirillplatonov: Great idea. I added `window.Turbo.clearRequestInterceptor()` for resseting interceptor.

kirillplatonov:
Until it's merged I published my fork to npm:
```
yarn add turbo-edge
import * as Turbo from "turbo-edge"
```

```
yarn add turbo-rails-edge
import { Turbo } from "turbo-rails-edge"
```

dhh:
Great stuff @kirillplatonov. Could you add some documentation for this on https://turbo.hotwire.dev/reference/drive? It's in the turbo-site repo. Thanks!

src/core/request_interceptor.ts
```diff
@@ -0,0 +1,15 @@
+export class RequestInterceptor {
```
dhh, line 1: Could we not set this like we do progressBarDelay? Where it's set on the session, not as a singleton.

marcelolx: I think here in Turbo it would be completely possible

kirillplatonov: I was trying to put it in session initially. But we use interceptor in `FetchRequest` class where we don't have an access to the session. Add fetching session from delegates not worked as well since not all delegates have access to the session. That's why I end up using a singleton. Seems like the cleanest solution available.

awd:
Thanks for getting this through @kirillplatonov @dhh - useful for JWT / Shopify apps!

dhh:
@kirillplatonov Upon further reflection, I think it might be worth exploring an approach similar to https://github.com/hotwired/turbo/pull/28#issuecomment-863986177. Which basically let's the turbo:before-fetch-request pause the request. Then we could do it all in a callback instead.

kirillplatonov:
@dhh having the same pausable API for both `turbo:before-render` and `turbo:before-fetch-request` would be great. I will explore this.

PR #181, opened 2021-02-16
Observe changes to turbo-frame[disabled]

seanpdoyle:
When a Turbo Frame declares the `[disabled]` attribute, prevent any
navigation.

Once removed, if a Turbo Frame's `[loading="eager"]` and `[src]` exists,
navigate the frame.


sstephenson:
Great find and fix.

Issue #138, opened 2021-01-28
Redirect to new page on successful form submission, rerender otherwise


boardfish:
I'm trying to figure out how to do this. Right now, when I submit a form from within a `turbo_frame_tag`, I'm reliant on the response having a matching `turbo_frame_tag` to replace the content of the tag. So regardless of whether it's a successful or failed form submission, the user doesn't leave the current page right now.

On a successful submission, I'd like to redirect the user to a new page. There are a couple of angles I've considered:

- Setting the `target` attribute to `_top`, which isn't optimal because it'd replace the view in full in either case.
- Wrapping the target view in another `turbo_frame_tag`, which feels clunky and also might not account for wider layout changes.

Is there a known way to do this?

danjac:
At a guess, have a Stimulus controller listen to the turbo:submitEnd event and trigger a redirect on success by checking the Location response header.

boardfish:
I like this idea. Haven't really thought into hooking into Turbo events for this. I'll look into it and see how it goes.

boardfish:
That having been said, it might be good to build functionality into Turbo for this directly.

danjac:
Agreed: perhaps redirects should always assume "_top".

tleish:
You can use a `<turbo-stream action="replace"...>` for this.

danjac:
@tleish how would that work if we want to actually redirect to another page (including change to history etc)?

tleish:
@danjac - what backend platform are you using?

danjac:
@tleish Django.

However I think if I understood correctly the pattern would be something like:

1. Render the form inside a turbo-frame (e.g. a modal).  Turbo-frame has "_top".
2. If any validation errors, render partial HTML containing the <form> tag inside a turbo-stream with relevant target/action. Stream has ID matching the turbo-frame ID or top-level element inside the frame.
3. On success (i.e. no validation errors), just return a redirect (303). As turbo frame has "_top" it will just redirect.

boardfish:
That sounds like it could be viable. And I suppose using `_top` is also expressive of what you're trying to achieve, but there's still a dependence on that `turbo-stream` that you'd probably need to make known. I feel like I prefer the JS solution here.

tleish:
@danjac —  this is correct.  

Also, if the `turbo-frame` does not use `src=` attribute to dynamically populate it's content, then you do not even need to use `turbo-frame` for this scenario, you only need turbo-stream in the response.

[...]

PR #183, opened 2021-02-18
[RFC] Render full page if no frame came back from server

MichalSznajder:
Just ugly hack for RFC. Solves partially #138. Address bar is not updated.


brunoprietog:
I think this can be closed. It was already addressed in #863 and #867. It's documented [here](https://turbo.hotwired.dev/handbook/frames#%E2%80%9Cbreaking-out%E2%80%9D-from-a-frame).

MichalSznajder:
I forgot I even created this. Probably I used Turbo Streams in the end at that time.

PR #184, opened 2021-02-21
Don't duplicate Form Submitter [name] param

seanpdoyle:
When a `<form>` element's submitter declares a `[name]` attribute
without a `[value]` attribute, it is appended to the submission's
`FormData` instance twice. Due to a quirky difference between Safari and
Firefox + Chrome, the duplication _only_ occurs in Safari, which is a
browser that we are not running our functional test suite against.

In spite of that, this commit adds test coverage to ensure we don't
introduce a similar regression into those browsers.

Testing
---

Oddly enough, there are several occurrences of doubled-up `await await`
keywords, which this commit removes.

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -169,8 +169,8 @@ function buildFormData(formElement: HTMLFormElement, submitter?: HTMLElement): F
   const name = submitter?.getAttribute("name")
   const value = submitter?.getAttribute("value")
 
-  if (name && formData.get(name) != value) {
-    formData.append(name, value || "")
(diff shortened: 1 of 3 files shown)


seanpdoyle:
What would it take to run intern against Safari in CI? Browserstack? GitHub Actions running on macOS?

sstephenson:
> What would it take to run intern against Safari in CI? Browserstack? GitHub Actions running on macOS?

I don't recall the specifics, but there's problems with running the test suite under Safari that are not CI- or Intern-related.

sstephenson:
Thank you!

PR #191, opened 2021-02-28
Prevent Streams from {pre,app}ending duplicates

seanpdoyle:
In a circumstance where a `<form>` submission triggers a `turbo-stream`
response or a redirect back to the page with the new record on it, a
corresponding Stream update broadcast might occur in the background that
results in an `append` or `prepend` action.

In the case of the broadcast losing the race condition, update the
Stream renderer's page update to prevent prepending or appending
elements with duplicate `[id]` attributes.

Testing
---

Modify the test server to accept `action` and `target` values for
resulting Streams as request body parameters.

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -64,7 +64,7 @@ export class FormSubmission {
   }
 
   get action(): string {
-    return this.submitter?.getAttribute("formaction") || this.formElement.action
+    return this.submitter?.getAttribute("formaction") || this.formElement.getAttribute("action") || this.formElement.action
   }
 
   get location(): URL {
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -2,10 +2,14 @@ import { StreamElement } from "../../elements/stream_element"
 
 export const StreamActions: { [action: string]: (this: StreamElement) => void } = {
   append() {
+    removeMatchingElements(this.templateContent, this.targetElement)
+
     this.targetElement?.append(this.templateContent)
   },
 
   prepend() {
+    removeMatchingElements(this.templateContent, this.targetElement)
+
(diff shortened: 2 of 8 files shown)


javan (commented):
Currently, stream actions map neatly (almost 1:1) to DOM methods of the same name: https://github.com/hotwired/turbo/blob/8bce5f17cd697716600d3b34836365ebcdc04b3f/src/core/streams/stream_actions.ts#L4-L14

This change muddies that contract by adding special behavior to *some* of them, which could have unintended consequences or at the very least catch people by surprise. 

I think we should consider moving it over to [turbo-rails](https://github.com/hotwired/turbo-rails) so the remove happens in all {pre,app}end broadcasts. Basically, take [this example](https://github.com/hotwired/turbo-rails/blob/fec33d9bc767aec612b283620d2a74e78c1f90ae/app/models/concerns/turbo/broadcastable.rb#L204-L220) and bake it in:

> ```ruby
> # Render a turbo stream template asynchronously with this broadcastable model passed as the local variable using a
> # <tt>Turbo::Streams::BroadcastJob</tt>. Example:
> #
> #   # Template: entries/_entry.turbo_stream.erb
> #   <%= turbo_stream.remove entry %>
> #
> #   <%= turbo_stream.append "entries", entry if entry.active? %>
> #
> # Sends:
> #
> #   <turbo-stream action="remove" target="entry_5"></turbo-stream>
> #   <turbo-stream action="append" target="entries"><template><div id="entry_5">My Entry</div></template></turbo-stream>
> #
> # ...to the stream named "entry:5"
> def broadcast_render_later(**rendering)
>   broadcast_render_later_to self, **rendering
> end
> ```

seanpdoyle:
@javan that's a great idea! 

Outside of a Rails environment, is this enough of a problem that makes considering an alternative solution worthwhile?

PR #192, opened 2021-03-02
Activate script tags from turbo frames

### LINKED ISSUE #186: Evaluate script tags from turbo streams
It seems that <script> tags in turbostream templates are not evaluated (at least not when appending). Is this a bug, or expected behaviour?

(My use-case is that the partial I'm trying to render contains helper from chartkick. The js added by the library is not evaluated)

xtr3me:
When using a turbo frame the scripts returned from a turbostream are not evaluated and therefor not executed.
This PR adds the functionality to activate these script tags after they are inserted into the turbo-frame.

Fixes #186 for turbo-frames

Please let me know if additional fixes are needed for turbostream elements, I will add them accordingly 👍


goncalvesjoao:
I'm kinda anxious for this fix, when can we see this in production guys? 👋

dhh:
@xtr3me Can you have a look at the test failure?

xtr3me:
@dhh I will look into this, before the end of this week. I'm currently in the midst of a renovation of the house, so I need to find some spare time to address this.

xtr3me:
@dhh I have fixed the test failure. I did need to add an additional file as clicking on the link did not trigger an Turbo call anymore if it was clicked once before.

By changing the url of the link embedded in the response frame it worked again.

Issue #187, opened 2021-02-23
Add 'how to contribute' to README


mollerhoj:
I want to submit a PR, fixing a bug. I have worked on many js (and ts) projects, but it's not my favorite language. I like ruby, surprise surprise . I can't keep up with the pace in terms of js tooling. Would it be possible to add a few instructions on how to get the project up and running locally and run the specs? Might lower the barrier of entrance :-)

@sstephenson

xtr3me:
I have tried to document the steps I took when fixing a bug on this project. I hope this helps a bit, let me know if you miss anything or you want me to clarify a bit more.

brunoprietog:
Hi @mollerhoj, thanks for your interest in contributing in Turbo! Actually, there's already a link in the README that links to the contribution guide. I'll leave it [here](https://github.com/hotwired/turbo/blob/main/CONTRIBUTING.md) anyway.

PR #194, opened 2021-03-02
#187 Start the contribution guide

xtr3me:
As requested in #187 I have started on a contribution guide to lower the barrier for developers to contribute to the project.

I did not include information yet about the debugger port the is opened during the test suite as I have yet to figure out how that works myself.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,67 @@
+[![Version](https://img.shields.io/npm/v/@hotwired/turbo)](https://www.npmjs.com/package/@hotwired/turbo)
+[![License](https://img.shields.io/github/license/hotwired/turbo)](https://github.com/hotwired/turbo)
+
+# Contributing
+
+Please note we have a [code of conduct](https://github.com/hotwired/turbo/blob/main/CODE_OF_CONDUCT.md), please follow it in all your interactions with the project.
+
+## Sending a Pull Request
+The core team is monitoring for pull requests. We will review your pull request and either merge it, request changes to it, or close it with an explanation.
+
+Before submitting a pull request, please make sure the following is done:
+
+1. Fork the repository and create your branch from main.
+2. Run `yarn` in the repository root.
+3. If you’ve fixed a bug or added code that should be tested, add tests!
+4. Ensure the test suite passes (`yarn build && yarn test`).
+
+## Developing locally
+
+During the process of developing the library locally we first have to check out the repository and create a branch from main.
+
+```bash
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+
+The html files needed for the tests are stored in: `src/tests/fixtures/`
+
+### Local webserver
+Since the tests are running in headless browsers it's not easy to debug them easily without using the debugger. Sometimes it's easier to run the supplied webserver and manually click through the test fixtures.
+
+To run the webserver:
+
+```bash
+yarn start
```
seanpdoyle, line 64: This requires a build (via `yarn build`), or a separate process running `yarn watch`.

seanpdoyle: Depending on their operating system, a developer might be able to execute `open http://localhost:9000/src/tests`, and then browse from there.

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+The webserver is available on port 9000. Since the webserver is run from the root of the project the fixtures can be found using the same path as they have in the project itself, so `src/tests/fixtures/rendering.html` makes: `http://localhost:9000/src/tests/fixtures/rendering.html`
```
seanpdoyle, line 67: Since this is markdown, removing the backpacks might make the URL auto-link, so that developers reading along will be able to launch their local server from the documentation.

CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
...
+### Testing
+The library is tested by running the test suite (found in: `src/tests/*`) against headless browsers. The browsers are setup in `intern.json` check it out to see the used browser environments.
+
+The tests are using the compiled version of the library and they are themselves also compiled. To compile the tests and library:
+
+```bash
+yarn build
+```
+
+> **_Caveat:_**  Don't forget to run the build command before you run the tests everytime you changed anything, if you forget it you will eventually run into unexpected results.
```
seanpdoyle, line 45: If we recommend `yarn watch`, we might be able to omit this caveat.

CONTRIBUTING.md
xtr3me: Unfortunately the the webserver is not listing the directory contents for index less directories. When opening the page on: http://localhost:9000/src/tests they will see the following error message: `404 Not Found: /src/tests/`

I'm a bit unsure if this causes extra confusion or not, what do you think?

Issue #136, opened 2021-01-26
Ability to import Drive, Frames and Streams independently for smaller bundle sizes?


nickjj:
Hi,

In cases where you want to use Drive but not Frames / Streams it might be beneficial to be able to import Drive by itself.

I'm not sure how to calculate the size of each piece as it stands now but is this worth exploring?

Turbo as a whole is much much larger than Turbolinks 5 and using Turbolinks 5 to get Drive-like behavior isn't quite the same because Drive is so much better in regards to handling form submissions and not needing anything server side.

gregschmit:
I feel like a lot of people will want this. I submitted a PR that "kinda" fixes this. Turbo Frames/Streams are opt-in, in the sense that it doesn't mess with existing ERB/JS stuff by installing Turbo. Drive is the only part that is opt-out, meaning it's enabled and disruptive by default (e.g., it breaks lots of existing javascript). My PR would allow the user to reverse this behavior when the module is imported, and users would use `data-turbo="true"` to selectively enable it, rather than using `data-turbo="false"` to selectively disable it. It's nice particularly in projects where you have a lot of layouts, dynamically generated layouts, or customer-controlled layouts and you want to permit new Turbo functionality without requiring changes to existing stuff.

whisper-bye:
+1

nickjj:
I know it's asking a lot but I'd really like to see something happen before 1.0.

Importing Turbo into a project increases my JS payload by 188kb. That is not minified or gzipped, but it's still a lot, especially when StimulusJS is another 181kb. Before you know it, just pulling in these libraries puts you close to 400kb of JS.

bojanvidanovic:
+1

dhh:
Appreciate the instinct. But compressed, Turbo is 19kb, and Stimulus is 9kb. It's just not going to be worth the effort to optimize that further weighed against the complications to do so.

KonnorRogers:
Just to revisit, would a PR be accepted to separate the 3? You'd just need different file paths for each.

`import "@hotwired/turbo/{drive, frames,streams}"`

seems like there was a positive reaction here: https://github.com/hotwired/turbo/issues/369

dhh:
Would be happy to consider a PR, but the tolerance for added complexity in order to reach this separation is very low. If you can find a super simple way to do it, though, I see no harm in that 👍

PR #196, opened 2021-03-04
Introduce `Turbo.session.drive = false` to allow users to disable Drive by default.

### LINKED ISSUE #190: Add option to deactivate Turbo Drive
Currently, Turbo assumes that the whole framework is used from the point on it is imported. But as the whole Turbo Drive is immediately activated, this can break whole websites even though the developer might only use Turbo Frames. 

**We should either have an option to independently import the modules (#136) or for example, being able to deactivate Turbo Drive.**

Sadly, `<meta content="reload" name="turbo-visit-control">` is not enough as Turbo Drive still intercepts the native page loading and can still break things, for example, scrolling, as it scrolls to top while fetching the next page, which for example can cause an issue like wrong scroll positions with smooth-scroll CSS properties active.

gregschmit:
I have a project where customers have existing layouts and I don't want them to have to disable Turbo Drive using `data-turbo="false"` in their layouts, but I still want them to be able to use Turbo in an opt-in fashion. Right now I monkey patch it like this, which seems hacky:

```js
import { Turbo } from "@hotwired/turbo-rails";
Turbo.navigator.delegate.elementIsNavigable = function(element) {
  const container = element === null || element === void 0 ? void 0 : element.closest("[data-turbo]");
  if (container) {
    return container.getAttribute("data-turbo") == "true";
  } else {
    return false;
  }
}
```

Turbo Drive seems to be the only part of Turbo which is not opt-in, so it seems wrong to not be able to configure it to behave that way. This PR would allow developers to disable Turbo Drive by default, and then opt-in by adding `data-turbo="true"` wherever they want to use Turbo Drive (including in the `<html>` or `<body>` tags). Usage after it's in `turbo-rails`:

```js
import { Turbo } from "@hotwired/turbo-rails";
Turbo.disableDriveByDefault();
```

The name might seem a little verbose, but I thought it was good because it communicates that it's not entirely disabled, and still usable in an opt-in fashion.

Fixes #190
Related to #136

Fixes hotwired/turbo-rails#139


gregschmit:
@seanpdoyle I noticed that you're responding to some of the issues and that you're a maintainer; could you provide me with some guidance as to whether or not you think this PR provides a useful addition to the project?

gregschmit:
So `elementIsNavigable` has now been moved out of the `Session`, apparently because it's needed in a module that doesn't have access to the current session. So this PR will have to get refactored to account for that change. Would the maintainers of the project entertain the idea of having a `coreSettings` hash map where the `disableDriveByDefault` property would go?

Are the maintainers of this project just unwilling to consider this at all? I've gotten no feedback in the 2 months since I opened this, and it's a little frustrating when it seems like it's a (at lest somewhat) popular feature/enhancement, and it seems fairly straightforward to implement, but I haven't gotten any feedback about whether the maintainers of the project want this enhancement to go into the project or if they have better ideas about how I should implement this, or if you just don't really like the idea at all.

My current thought is to have options passable front `start()` (e.g., `start(disableDriveByDefault: true)`) which would be stored in something like a `coreSettings` object/hash, and then we could also expose methods if any of those settings are mutable so, for example, in `turbo-rails` where `start` is called automatically, there would be an interface to mutate those settings. This would be flexible if we ever had any other settings to add, and it would be detached from the session so `elementIsNavigable` would be able to function without being attached to the session.

gregschmit:
To be clear, I would really appreciate some feedback from a maintainer because at this point I feel like even if I refactor, this PR won't get reviewed or merged so I will have wasted my time. I feel like I've been pretty patient waiting 2 months before bringing up the fact that this PR has been ignored, so I understand that people are busy but this wait time just seems a little ridiculous.

ansonhoyt:
@gregschmit I'm watching your issue because you have an interesting idea. Did you mean to be so harsh?

Please try to stay patient as you consider how you wish to spend your time. Basecamp is going through some turmoil right now, and I'd guess that is likely to impact many projects in ways no one can predict. I wish them all the best, even as that uncertainty is impacting my choices and "wastes my time." ✌️❤️

gregschmit:
@ansonhoyt No! I definitely didn't mean to be "harsh"! I thought what I said was pretty reasonable. I didn't call anyone names or anything, I just voiced my disappointment that 2 months had passed without a comment on the PR. I definitely wasn't trying to be mean to anyone.

ghiculescu:
> So `elementIsNavigable` has now been moved out of the `Session`, apparently because it's needed in a module that doesn't have access to the current session.

(not a maintainer) I think storing a static property on `Session` is still fine. [Here](https://github.com/hotwired/turbo/blob/v7.0.0-beta.5/src/core/session.ts#L266) you'd just be referencing `Session.driveDisabledByDefault`.

gregschmit:
@ghiculescu Oh, I didn't know that would work. My understanding was that `Session` was the class and the "current" session instance would be `session`, so when the user calls `Turbo .disableDriveByDefault()` it would be an instance method call that changes the instance variable `session.disableDriveByDefault = true`, so the class variable `Session.disableDriveByDefault` would still be `false`. I'm not a JavaScript/TypeScript expert (more of a C/Python/Ruby guy), but if you can confirm that would work I can certainly make that change.

dhh:
I'm happy to see this. I appreciate the attempt to explain the default situation in the flag, but it's imo too much. I think `Turbo.drive = false` is a better default. We can document that this still allows you a per-element opt-in.

gregschmit:
Ok, I found some time to work on this again. It might be ready now. So I did move `elementIsNavigable` back into the session (renamed it to `elementDriveEnabled` to better describe it) and just imported the `session` in `frame_controller.ts` (where it was needed). AFAICT that doesn't cause any dependency issuess (circular or otherwise), so I think that is ok.

So the way to disable Drive by default is to set `Turbo.session.drive = false`. Lmk if you think a helper function on `Turbo` should be added, like `Turbo.disableDrive()`. I also added `DriveTests` to test basic Drive functionality both in the default mode and with `Turbo.session.drive = false`.

gregschmit:
So the test error is: `ScrollRestorationTests - landing on an anchor (0.01s)`. Weird, because that test actually runs successfully for me locally on both Chrome and Firefox.

gregschmit:
Ok, I merged `main` into this branch (rebasing to make merging simpler), so perhaps that test is fixed. I don't think the test failure was caused by my changes.

gregschmit:
Yeah, looks like the same error that was in `main` on `7.0.0-rc.1`: https://github.com/hotwired/turbo/runs/3128997485?check_suite_focus=true

I think this PR is good to go from a functionality standpoint, just awaiting review from maintainers.

ansonhoyt:
@gregschmit happy to see your idea merged! 🎉  I have several existing projects where it might be nice to add a little turbo, and this will make that all the easier. Thank you!

src/core/session.ts
```diff
@@ -34,6 +34,7 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
 
   readonly frameRedirector = new FrameRedirector(document.documentElement)
 
+  drive = true
   enabled = true
```
seanpdoyle, line 38: @gregschmit I'm sorry to dig up an old pull request, but could you help me understand why a `drive` property was introduced alongside [enabled](https://github.com/gregschmit/turbo/blame/67f068edefaa865e0eda78131231521d3513d609/src/core/session.ts#L38)? Could `enabled` be repurposed to serve the same purpose?

gregschmit: I might need to look at the code, but my thinking is that `enabled` on `Turbo.session` enabled/disabled everything (or the session?), whereas this feature was designed to specifically disable `Drive` behavior until enabled in the document using `data-turbo="true"`, allowing opt-in behavior. So with `Turbo.session.drive = false`, everything is still enabled (especially frames/streams), just that drive must now be opted-into rather than opted-out-of (making the behavior similar to frames/streams).

Does that make sense? I can still take a look at the code and see if i'm wrong about my assumptions of what `enabled` does.

BTW, this is the confusion I was hoping to mitigate by naming the property `disableDriveByDefault` but it was considered too verbose.

PR #199, opened 2021-03-06
Toggle `[aria-busy="true"]` during requests

seanpdoyle:
Toggle the `[aria-busy="true"]` attribute to be present on a
`<turbo-frame>` element when its navigating, a `<form>` element when
it's submitting, or on the `<html>` element during any request, then
remove the attribute after the submission or navigation completes.
 
This is a change from the `turbo-frame[busy]` attribute, but aims to
achieve the same purpose. For the sake of backwards compatibility, Turbo
will continue to toggle the `[busy]` attribute during Frame navigations.

By unifying a single, consistent attribute, consumer applications can
use a single attribute CSS selector at different depths within their
page to hide or show loading indicators.

For example, consider a loading indication spinner animation within a
submit button that animates while a submission is in progress:

```html
<style>
  .loading-spinner                      { display: none; }
  [aria-busy="true"] .loading-spinner   { display: block; }
</style>
    
<form action="/posts">
  <!-- ... -->
  <button type="submit">
    <span class="loading-spinner>...</span>
    <span>Create Post</span>
  </button>
</form>
```
    
Similarly, consider a `turbo-frame` scoped loading bar:
    
```html
<style>
  .loading-bar                          { display: none; }
  [aria-busy="true"] .loading-bar       { display: block; }
</style>
<a href="/content" data-turbo-frame="content">...</a>
<turbo-frame id="content">
  <progress id="content-loading-bar" class="loading-bar" value="0" data-turbo-permanent></progress>

  <!-- ... -->
</turbo-frame>
```
    
Finally, consider the same `.loading-bar` for page-wide navigation:
    
```html
<style>
  .loading-bar                          { display: none; }
  [aria-busy="true"] .loading-bar       { display: block; }
  </style>
    
<nav>
  <progress id="loading-bar" class="loading-bar" value="0" data-turbo-permanent></progress>
</nav>
    
<!-- ... -->
    
<a href="/posts/1">...</a>
```

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -2,7 +2,7 @@ import { FrameElement, FrameElementDelegate, FrameLoadingStyle } from "../../ele
 import { FetchMethod, FetchRequest, FetchRequestDelegate, FetchRequestHeaders } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { AppearanceObserver, AppearanceObserverDelegate } from "../../observers/appearance_observer"
-import { getAttribute, parseHTMLDocument } from "../../util"
+import { clearBusyState, getAttribute, parseHTMLDocument, markAsBusy } from "../../util"
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { ViewDelegate } from "../view"
@@ -165,7 +165,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   requestStarted(request: FetchRequest) {
-    this.element.setAttribute("busy", "")
+    [ this.element, document.documentElement ].forEach(markAsBusy)
   }
 
   requestPreventedHandlingResponse(request: FetchRequest, response: FetchResponse) {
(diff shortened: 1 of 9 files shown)


seanpdoyle:
This depends on the changes proposed by https://github.com/hotwired/turbo/pull/156, and will likely require them to ship and be released before being merged. Since it's a breaking change, we might want a release in the interim.

As a follow up to this PR, we can also implement https://github.com/hotwired/turbo/pull/164 in terms of a Mutation Observer on `html[data-turbo-busy]` instead.

seanpdoyle:
I'm still not sure toggling `[data-turbo-busy]` is better than toggling [`[aria-busy]`](https://www.w3.org/TR/wai-aria-1.1/#aria-busy) like what was proposed in https://github.com/hotwired/turbo/pull/157:

> Indicates an element is being modified and that assistive technologies MAY want to wait until the modifications are complete before exposing them to the user.

> The default value of aria-busy is false for all elements. When aria-busy is true for an element, assistive technologies MAY ignore changes to content owned by that element and then process all changes made during the busy period as a single, atomic update when aria-busy becomes false.

> If it is necessary to make multiple additions, modifications, or removals within a container element that is already either partially or fully rendered, authors MAY set aria-busy to true on the container element before the first change, and then set it to false when the last change is complete. For example, if multiple changes to a live region should be spoken as a single unit of speech, authors MAY set aria-busy to true while the changes are being made and then set it to false when the changes are complete and ready to be spoken.

> If an element with role feed is marked busy, assistive technologies MAY defer rendering changes that occur inside the feed with the exception of user-initiated changes that occur inside the article that the user is reading during the busy period.

seanpdoyle:
Applications could declare their own MutationObservers that observe changes to `[data-turbo-busy]`, and toggle `[aria-busy]` themselves, but there's an opportunity for the framework to manage that state itself.

The opportunity for styling via CSS is still there (and tools like [thoughtbot/tailwindcss-aria-attributes](https://github.com/thoughtbot/tailwindcss-aria-attributes#boolean-attributes) provide an avenue toward that.

The more I think about it, the more I think it should toggle `[aria-busy]`. I'm going to push up some changes to toggle that instead, and attempt to get CI passing.

dhh:
I'd be fine with aria-busy instead 👍

Would like to ship 7.1.0-rc1 when we've resolved this one.

dhh:
We'd need backwards compatibility, though. Already have `data-turbo-busy` for forms alone. Could consider simply setting both in both cases?

dhh:
Excellent. Do you want to take a stab at a doc PR as well for this?

Do you have anything else that's a must-have for 7.1.0?

Issue #145, opened 2021-01-31
turbo-frame with data-turbo-permanent for load once behaviour


stevehodgkiss:
Hi,

I'm trying to load a `turbo-frame[src]` once on first page load by wrapping it in a `div[data-turbo-permanent]`.

```html
<div id="user-nav-container" data-turbo-permanent>
  <turbo-frame id="user-nav" src="user_nav.html">
    <a>Sign In</a>
  </turbo-frame>
</div>
```

The turbo-frame is currently being requested on each Turbo-intercepted click. The desired behaviour is to load the frame once and then not load it on subsequent Turbo navigation (i.e. not see any `user_nav.html` fetch requests), in the same way that `data-turbo-permanent` works with regular elements.

Without the wrapping `data-turbo-permanent` div Turbo makes 2 requests to user_nav.html on each visit. This might be a separate issue because I would expect this to only one request to be made in this scenario.

With the wrapping element this increases to 3 on the second Turbo click (screenshot).

<img width="665" alt="Brave Screenshot" src="https://user-images.githubusercontent.com/62331/106375507-169c6000-63a6-11eb-85d5-56286f72e40b.png">

Example repo: https://github.com/stevehodgkiss/turbo-frame-permanent

Thanks!

stgm:
It seems you're both providing content in the frame (i.e. `<a>Sign In</a>`) and **also** providing a `src` url. Generally speaking, these are two ways to do the same thing: provide content. In most cases, you should not provide a `src` attribute, unless you want to do things like lazy-loading.

Could it be that you're looking for something like the following?

```
<turbo-frame id="user-nav">
    <a href="user_nav.html">Sign In</a>
</turbo-frame>
```

stgm:
I tried your example using the previous beta of Turbo (https://cdn.skypack.dev/@hotwired/turbo@7.0.0-beta.3) and that does seem to eliminate the problem.

stevehodgkiss:
> I tried your example using the previous beta of Turbo (https://cdn.skypack.dev/@hotwired/turbo@7.0.0-beta.3) and that does seem to eliminate the problem.

True, 7.0.0-beta.3 does eliminate the multiple XHR request issue present in the latest version.

> PS. I may have been distracted by your href-less link in the example, and it could be that you want to do lazy-loading after all.

[...]

Issue #150, opened 2021-02-03
turbo-frame with src attribute is loaded multiple times


jcoyne:
My page has 3 `turbo-frame`s each with a src:
```
<turbo-frame id="delete_work_3" src="/works/3/delete_button"></turbo-frame>
<turbo-frame id="delete_work_1" src="/works/1/delete_button"></turbo-frame>
<turbo-frame id="delete_work_3" src="/works/2/delete_button"></turbo-frame>
```

Each of these return a sensible response.  However the page immediately loads each of these 5 times (removed some from the screenshot for brevity).

<img width="389" alt="Screen Shot 2021-02-02 at 11 25 24 PM" src="https://user-images.githubusercontent.com/92044/106702467-06ef6800-65ae-11eb-9a1c-c2996dd4a62a.png">

An example response looks like this:
```xml
<turbo-stream action="replace" target="delete_work_3"><template><turbo-frame id="delete_work_3"></turbo-frame></template></turbo-stream>
```

jcoyne:
I tried this on a new rails app from scratch but I couldn't duplicate this problem, so I'm thinking I must have something misconfigured, but I can't figure out what it might be.

stgm:
Does the new rails app serve the exact same response?

It seems you're mixing streams and frames (given the example above), which I wouldn't normally expect. Is that intentional?

stgm:
The minimum response should be `<turbo-frame id="delete_work_3"></turbo-frame>` to allow Turbo Frames to replace the content of the current frame (that's what Frames is able to do). You can place any content between those tags and it will be shown instead of the previous content.

jcoyne:
@stgm I made that change and that is much simpler (Thank you! ❤️ ), however that doesn't resolve the problem of multiple requests.

stgm:
Two things that spring to mind that could cause extraneous requests:

- I'm not sure what behavior results should you manage to import Turbo twice (maybe this?), and
- setting the `src` attribute of the frame from JS will also initiate a new request

stgm:
But if you tell a bit more about the context it may be easier to spot misconceptions ;-) E.g. your code mentions a button but you are currently using the Turbo Frames feature that automatically "lazy-loads" a frame.

jcoyne:
The buttons are in the lazy loaded frame iff the current user has the authorization to do the action the button provides.

We are using this approach because we can stream updates to the page to everyone, but not everyone is authorized to do the actions.

stgm:
[...]

PR #204, opened 2021-03-11
Do not reload frame when cloning

tleish:
Fix for comment in https://github.com/hotwired/turbo/issues/150#issuecomment-797017665, and possible fix for entire #150 issue.


tleish:
This may also fix https://github.com/hotwired/turbo/issues/145

tleish:
Closing in favor of https://github.com/hotwired/turbo/compare/prevent-frames-from-loading-in-snapshots

Issue #202, opened 2021-03-10
Reload the content of a turbo-frame


coorasse:
We would like to trigger the reloading of a turbo-frame which has been lazy loaded with the `src` attribute.

How can we achieve that? If there's nothing built-in, would be something you would consider in a PR?

As a reference: https://discuss.hotwire.dev/t/how-to-auto-reload-frame/1872/5

javan:
You can re-assign a `<turbo-frame>`’s `src` to reload it:
```js
const { src } = element
element.src = null
element.src = src
```

> If there's nothing built-in, would be something you would consider in a PR?

I think wrapping that up as `element.reload()` would be a nice addition. 👍

coorasse:
To me this issue can be closed nevertheless. Your suggestion works, and I opened a PR if we want to implement the `reload()` method. Thanks.

coorasse:
The above mentioned technique doesn't work when `loading="lazy"`

marselmustafin:
Any ideas for reloading content of `loading="lazy"` frames?

jorgegorka:
@marselmustafin You can reload the content manually using `this.element.reload()` inside a stimulus controller attache to that turbo frame or you can do `document.getElementById('turbo-frame-id').reload()`

firedev:
Can I somehow mark a turbo-frame as "incomplete" so it reloads the next time it's displayed?

duffyjp:
@firedev 
> Can I somehow mark a turbo-frame as "incomplete" so it reloads the next time it's displayed?

I just implemented @jorgegorka's suggestion in the stimulus controller I use to hide/show a modal with its content loaded via turbo frame with `src:`  It works great.  Right after display I call the `.reload()` so there's zero delay but new content will indeed get loaded on top after a few ms.

A fancier alternative could be to load the content with a subscribed turbo stream and then broadcast any changes server side.  I'm loading such a small amount of data, and it's only shown temporarily so for me just reloading every time is fine.

firedev:
@duffyjp Hey here is how I did it in the end https://gist.github.com/firedev/cedf0ea3153e790498da5c72b33fcd76

scuml:
Here's a stimulus controller that resets a turbo-frame when it is removed from view.  Awesome for dialog elements. 
```
import { Controller } from "@hotwired/stimulus";

export default class extends Controller {
  /* 
  A controller that resets a lazy turbo frame when it disappears from view.
  Usage:
      <dialog>
[...]

PR #206, opened 2021-03-12
Add a "reload" method to the frame element

coorasse:
As proposed in #202, this PR adds a `.reload()` method on the frame that allows to refresh its content.

Open points:

* I am not happy with the test. It actually does not test anything interesting. I'd like the content of the frame to change upon reloading and verify that, but I have no clue on how to do that. A bit of help or an idea would be appreciated. I left it there to give an idea of what I wanted to do, but as of now, it could be removed.
* What if the frame has no src attribute? In my implementation, a call to `reload()` will simply have no effect. To me this is simple and stupid, so it should be fine.

### DIFF
--- a/src/elements/frame_element.ts
+++ b/src/elements/frame_element.ts
@@ -35,6 +35,12 @@ export class FrameElement extends HTMLElement {
     this.delegate.disconnect()
   }
 
+  reload() {
+    const { src } = this;
+    this.src = null;
+    this.src = src;
+  }
+
   attributeChangedCallback(name: string) {
     if (name == "loading") {
       this.delegate.loadingStyleChanged()
--- a/src/tests/functional/loading_tests.ts
+++ b/src/tests/functional/loading_tests.ts
@@ -84,6 +84,19 @@ export class LoadingTests extends TurboDriveTestCase {
     this.assert.equal(await contents.getVisibleText(), "Frames: #frame")
   }
 
+  async "test reloading a frame reloads the content"() {
+    await this.nextBeat
+
+    await this.clickSelector("#loading-eager summary")
+    await this.nextBeat
+
+    const frameContent = "#loading-eager turbo-frame#frame h2"
+    this.assert.ok(await this.hasSelector(frameContent))
+    // @ts-ignore
+    await this.remote.execute(() => document.querySelector("#loading-eager turbo-frame")?.reload())
+    this.assert.ok(await this.hasSelector(frameContent))
+  }
+
   async "test navigating away from a page does not reload its frames"() {
     await this.clickSelector("#one")
     await this.nextBody


src/elements/frame_element.ts
```diff
@@ -35,6 +35,12 @@ export class FrameElement extends HTMLElement {
     this.delegate.disconnect()
   }
 
+  reload() {
+    const srcAttribute = this.src;
+    this.src = null;
+    this.src = srcAttribute;
```
kaspth, line 41: We've using ES6 here, so let's do:

```suggestion
    const { src } = this
    this.src = null
    this.src = src
```

src/tests/functional/loading_tests.ts
```diff
@@ -63,6 +63,20 @@ export class LoadingTests extends TurboDriveTestCase {
...
+  
+  async "test reloading a frame reloads the content"() {
+    await this.nextBeat
+
+    await this.clickSelector("#loading-eager summary")
+    await this.nextBeat
+
+    const frameContent = "#loading-eager turbo-frame#frame h2"
+	this.assert.ok(await this.hasSelector(frameContent))
+	// @ts-ignore
```
kaspth, line 75: Why do we want this?

src/tests/functional/loading_tests.ts
```diff
@@ -63,6 +63,20 @@ export class LoadingTests extends TurboDriveTestCase {
...
+    await this.remote.execute(() => document.querySelector("#loading-eager turbo-frame")?.reload())
+	this.assert.ok(await this.hasSelector(frameContent))
```
kaspth, line 77: Looks like we're using tabs for some of these lines, can you switch to just spaces?

src/tests/functional/loading_tests.ts
```diff
@@ -63,6 +63,20 @@ export class LoadingTests extends TurboDriveTestCase {
...
+  }
+
```
kaspth, line 79: Our style before didn't add an extra line, so we should cut this one too.

kaspth:
> * I am not happy with the test. It actually does not test anything interesting. I'd like the content of the frame to change upon reloading and verify that, but I have no clue on how to do that. A bit of help or an idea would be appreciated. I left it there to give an idea of what I wanted to do, but as of now, it could be removed.

Can the `src` URL that's under test, be a test specific URL? That way it could maintain it's own counter and just increment that and display. So the test would check for a 2 and then a 3 (run it a third time so the post-reload isn't borked).

src/tests/functional/loading_tests.ts
coorasse: because otherwise typescript complains that `reload` is not an Element method. 
image
shall I cast it instead?

coorasse:
I am not sure how to achieve that. The target of `src` are always static HTML pages in the fixture folder.

acetinick:
@coorasse may i suggest an additional improvement to this method would for the reload() function to accept an optional parameter to specify the URL to set src as.

This would be useful if you don't already have one, and want to set it as a new URL,  or if for some reason the URL changes and you want it pointing somewhere else.

Issue #56, opened 2020-12-29
Replace two different turbo-frame from one response


davidjr82:
### Situation:
- There is a profile form where the user can update his own name. 
- There is, in the same page, a navigation bar where the username is shown.
- When the user submit the profile form, the response is just the turbo-frame with the profile form and a "saved" message.
- I want to return, in the same response, another turbo-frame with the navigation username updated (so it is updated in the navigation bar).
- When I tried the two turbo-frames in the same response, this won't update the navigation username.

### Question: 
- Is there anyway to do this without turbo-streams? Can 2 turbo-frames be returned in the same response that updates different parts of the page? If so, how?
- If is not possible right now, would be possible in the future to write a comma-separated string in the data-turbo-frame attribute in forms, so we can replace two different turbo-frames in the page? (in the reference, there is a way where the form can replace the navigation turbo-frame, but then the self turbo-frame can not be replaced).

Great package. Thanks!

robzolkos:
See this commit to see how to do this https://github.com/hotwired/turbo-rails/commit/5c49e57f3431956818e9e40a7e87f7bcdaf3a0e5

davidjr82:
Thanks @robzolkos 

I am not using ruby, but this looks to me like it is related to turbo-streams, returning different responses depending on the existence of a template, and not related to turbo-frames, but maybe I am wrong and not understanding the PR. What I want is to return 2 turbo-frames in the same response, not to return one or the other.

Is that what the PR does and I am misunderstanding it?

davidjr82:
By the way, here is a discussion on the hotwire discuss, of someone trying to achieve the same thing:

https://discuss.hotwire.dev/t/loading-two-frames-from-a-single-application-visit-event/1556

Intrepidd:
As far as I understand, frames are for scoped navigation. If you want to replace 2 parts of your page at once after a form submit (or from websocket) you need to use turbo streams to replace both those elements.

If the only thing you do with those blocks is bulk replace them and they don't have their own navigation context, you don't even need to use frames, turbo streams just replace from DOM id.

dhh:
If you want to replace multiple frames, you can also just target _top and replace the whole thing via drive. But otherwise there won't be a path to custom replace two frames. When you need that, you gotta go to turbo streams.

davidjr82:
[...]

PR #210, opened 2021-03-14
Render 4xx-5xx responses within frame

seanpdoyle:
Render 4xx-5xx responses within frame
===

Closes https://github.com/hotwired/turbo/issues/138

The use case
---

Consider submitting a `<form data-turbo-frame="_top">` from within a
`<turbo-frame>`:

```html
<dialog open>
  <turbo-frame id="dialog">
    <form data-turbo-target="_top" method="POST" action="/posts">
      <!-- ... -->
    </form>
  </turbo-frame>
</dialog>
```

In the case of success, the response will be a [303 redirecting to
another URL][303].

In the case of a failed submission, the response will be a [422 with an
HTML body][422]:

```ruby
class PostsController < ApplicationController
  def create
    @post = Post.create!(post_params)

    redirect_to post_url(@post)
  rescue ActiveRecord::RecordNotSaved
    render inline: <<~HTML, status: :unprocessable_entity
      <turbo-frame id="dialog">
        <form data-turbo-target="_top" method="POST" action="/posts">
          <!-- ... -->
        </form>
      </turbo-frame>
    HTML
  end
end
```

[303]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303
[422]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

The current behavior
---

Since the `<form>` element declares its target as `_top`, a successful
submission resulting in the [303][] response, the entire page redirect.

Similarly, in the case of an invalid submission, _the entire page's
HTML_ will be replaced by the response, in spite of the submission
occurring within a `<turbo-frame>` element.

The desired outcome
---

A valid submission behaves as expected and desired: redirect the `_top`
(entire page) to the URL.

In the case of a submission without Turbo, a failed submission would
result in a full re-render of the page, keeping the `<dialog>`
containing the form open. Any information or context in the "background"
would be re-rendered, but would exist.

With the introduction of Turbo Frames to the situation as a progressive
enhancement, a developer might expect the server's response to be
limited to only replacing the contents of the `<turbo-frame>` after a
failure.

A potential work-around
---

As a workaround, the server could respond with a `<turbo-stream>` to
replace the inner contents of the `<dialog>` element's `<turbo-frame>`.

Proposed change
---

Currently, the implementation treats submissions from within a
`<turbo-frame>` that target `_top` differently from elements without a
`[data-turbo-frame]` attribute, or one that refers to another element on
the page.

To account for the proposed change in behavior, _all_ frame submissions
would use the same machinery, and the different treatment would be
deferred to _after_ the submission is determined to be a success.

Implemention change
===

Importing the `navigator` reference from the package's `core` module is
a bit of a hacky workaround, but enables delegation to a Navigator in
the case of a `_top` submission redirect.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -12,6 +12,7 @@ import { FrameView } from "./frame_view"
 import { LinkInterceptor, LinkInterceptorDelegate } from "./link_interceptor"
 import { FrameRenderer } from "./frame_renderer"
 import { elementIsNavigable } from "../session"
+import { navigator } from "../../core"
 
 export class FrameController implements AppearanceObserverDelegate, FetchRequestDelegate, FormInterceptorDelegate, FormSubmissionDelegate, FrameElementDelegate, LinkInterceptorDelegate, ViewDelegate<Snapshot<FrameElement>> {
   readonly element: FrameElement
@@ -192,7 +193,16 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
 
   formSubmissionSucceededWithResponse(formSubmission: FormSubmission, response: FetchResponse) {
     const frame = this.findFrameElement(formSubmission.formElement)
-    frame.delegate.loadResponse(response)
+    const target = [ formSubmission.formElement, formSubmission.submitter, frame ]
+      .map(element => element?.getAttribute("data-turbo-frame"))
+      .find(target => target)
+
+    if (response.redirected && target == "_top") {
+      navigator.formSubmission = formSubmission
+      navigator.formSubmissionSucceededWithResponse(formSubmission, response)
+    } else {
+      frame.delegate.loadResponse(response)
+    }
   }
 
   formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
@@ -280,6 +290,9 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
       if (frameElement) {
         return !frameElement.disabled
       }
+    } else {
+      const frame = element.closest("turbo-frame:not([disabled])")
+      return this.element == frame
     }
 
     if (!elementIsNavigable(element)) {
--- a/src/core/frames/frame_redirector.ts
+++ b/src/core/frames/frame_redirector.ts
@@ -47,13 +47,30 @@ export class FrameRedirector implements LinkInterceptorDelegate, FormInterceptor
 
   private shouldRedirect(element: Element, submitter?: HTMLElement) {
     const frame = this.findFrameElement(element)
-    return frame ? frame != element.closest("turbo-frame") : false
+    if (frame) {
+      if (frame == element.closest("turbo-frame")) {
+        if (element instanceof HTMLFormElement) {
+          return frame.getAttribute("data-turbo-frame") == "_top" || element.getAttribute("data-turbo-frame") == "_top"
+        } else {
+          return frame.getAttribute("data-turbo-frame") == "_top"
+        }
+      } else {
+        return true
+      }
+    } else {
+      return false
+    }
   }
 
   private findFrameElement(element: Element) {
     const id = element.getAttribute("data-turbo-frame")
     if (id && id != "_top") {
       const frame = this.element.querySelector(`#${id}:not([disabled])`)
+      if (frame instanceof FrameElement) {
+        return frame
+      }
+    } else if (id == "_top") {
+      const frame = element.closest(`turbo-frame:not([disabled])`)
       if (frame instanceof FrameElement) {
         return frame
       }
--- a/src/tests/fixtures/form.html
+++ b/src/tests/fixtures/form.html
@@ -13,6 +13,8 @@
     </style>
   </head>
   <body>
+    <h1 id="form-title">Form</h1>
+
     <div id="standard">
       <form action="/__turbo/redirect" method="post" class="redirect">
         <input type="hidden" name="path" value="/src/tests/fixtures/form.html">
@@ -168,6 +170,14 @@
       </form>
       <div id="messages">
       </div>
+      <form action="/__turbo/redirect" method="post" class="redirect" data-turbo-frame="_top">
+        <input type="hidden" name="path" value="/src/tests/fixtures/one.html">
+        <input type="submit">
+      </form>
+      <form class="unprocessable_entity" action="/__turbo/reject" method="post" data-turbo-frame="_top">
+        <input type="hidden" name="status" value="422">
+        <input type="submit">
+      </form>
     </turbo-frame>
   </body>
 </html>
--- a/src/tests/functional/form_submission_tests.ts
+++ b/src/tests/functional/form_submission_tests.ts
@@ -202,6 +202,16 @@ export class FormSubmissionTests extends TurboDriveTestCase {
     this.assert.equal(await this.nextAttributeMutationNamed("frame", "busy"), null, "removes [busy] from the #frame")
   }
 
+  async "test frame form submission with [data-turbo-frame=_top]"() {
+    await this.clickSelector("#frame form.redirect[data-turbo-frame=_top] input[type=submit]")
+    await this.nextBody
+
+    const title = await this.querySelector("h1")
+    this.assert.equal(await this.pathname, "/src/tests/fixtures/one.html")
+    this.assert.equal(await title.getVisibleText(), "One", "follows form redirect")
+    this.assert.notOk(await this.hasSelector("#frame"), "redirects entire page")
+  }
+
   async "test frame form submission with empty created response"() {
     const htmlBefore = await this.outerHTMLForSelector("#frame")
     const button = await this.querySelector("#frame form.created input[type=submit]")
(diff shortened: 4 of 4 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -163,7 +164,16 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
 
   formSubmissionSucceededWithResponse(formSubmission: FormSubmission, response: FetchResponse) {
     const frame = this.findFrameElement(formSubmission.formElement)
-    frame.delegate.loadResponse(response)
+    const target = [ formSubmission.formElement, formSubmission.submitter, frame ]
+      .map(element => element?.getAttribute("data-turbo-frame"))
+      .find(target => target)
+
+    if (response.redirected && target == "_top") {
+      navigator.formSubmission = formSubmission
```
seanpdoyle, line 172: @sstephenson this felt like the shortest distance to achieve the outcome. The awkwardness signals to me that there is a missing concept or delegate here.

A means of delegating or directly accessing the Navigator would also enable the transformation of a frame navigation to a fully formed `Visit` to drive the page.

jon-sully:
Hey @seanpdoyle 👋🏻

Thanks for putting this together, first of all!

I'm running into some similar issues with Turbo Frames and responses/navigations so wanted to get your thoughts. The way I'm thinking about it, I'm seeing two main options:

1) The frame can target `_top` so that successful form submissions that respond with a 303 will redirect at the Drive level, and changes can be made such that 422/5xx responses (form validation failures) would be shoved back _into_ the frame as to avoid a full-page/Drive navigation (this PR)

**or** 

2) The frame can _not_ target `_top` so that any 422/5xx responses (form validation failures) will automatically be 'inside the frame' then changes can be made to such that any 303 responses to the frame (e.g. a successful form submission) could just propagate up to Drive-level to result in a full-page navigation

Side-effects wise, the former means that any time the server responds with a 422/5xx to a request coming from within a frame that's targeting `_top` the server's error(s) / issue(s) may not be rendered at all (if they're outside of the frame in the markup) while the latter means that redirection-inside-a-frame is basically dead and having a multi-step in-frame experience would be tough 😕 

I'm not sure I've got a strong opinion on either direction, but could you speak to the strategy behind choosing one or the other for the implementation of this feature?

I wonder if the real drawback here is the simple binary nature of `_top` as a target. Maybe we just need a second 'special' target like `redirect_top` that behaves normally for all responses but does propagate redirects to `_top`? Just spitballing 🙂

Tagging these because they feel relevant: https://github.com/hotwired/turbo/issues/138, [Hotwire Discourse #1562](https://discuss.hotwire.dev/t/break-out-of-a-frame-during-form-redirect/1562/)

seanpdoyle:
@jon-sully my intuition pushes me toward the first option you laid out. It's also least disruptive in terms of existing Turbo conventions and patterns.

In my experiences, there certainly is a need for some version of the second option you mentioned, with regard to an ability to "break out" of a frame and combine the scoped HTML modifications and DOM state preservation of a Turbo Frame with the URL push state changes of a Turbo Visit.

We still haven't thread that needle yet, in terms of an attribute-based "interface", but there has been some exploration (https://github.com/hotwired/turbo/pull/167).

jon-sully:
Cool! Thanks for the thoughts. Like I said, I don't think I have a strong opinion here; either way accomplishes the goal of my forms working inside a frame when they fail and when they don't... and this PR does that from what I can tell! Hopefully this gets some traction in the coming weeks 💯

tleish:
I'd like to be able to allow certain pages to respond with html that forces "top" in the response.

For example, if a users session times out and they click on a form (or link) within a frame which does not have target="_top", the response redirects the user to the login page, but to the user nothing happens on the page.  If the response returned data which told Turbo to load at the top, this would be useful in this scenario.

Growiel:
Where are we on this ? I had yet another case where this was an issue today. I work around it by using Turbo streams to update my entire page or use javascript to force a refresh, but both are, at best, temporary workarounds.

ansonhoyt:
@Growiel probably not what you intended, but asking for status suggests this issue is being neglected and asks a maintainer to give it their attention. The project is in _**beta**_, so some issues like this are expected. If you look at PR and commit activity on https://github.com/hotwired projects, you'll see quite a bit of momentum from contributors, especially Basecamp and Thoughtbot folks. This issue doesn't need extra attention.

Like you, I share some concerns around this issue. I don't have bandwidth right now to either help solve it for everyone or to dig in to work around it for myself, so I've subscribed for GitHub updates and I choose not to deploy to this production yet.

This is open source. While Basecamp made Turbo happen (with the community and standing on Turbolinks shoulders), we all have some responsibility for helping it along. I've seen several small PRs merged in from folks new to Turbo, fixing some simple errors they ran into. ✌️ ❤️

jon-sully:
_Fantastic_ answer, @ansonhoyt 👏🏻

As with all things OSS, if the issue is pressing enough to an individual, we are all capable of forking the project, fixing it in our fork, and running _that_ code wherever we need it to. Otherwise open-source is a contribution-based process we should all take part in and have good expectations around ☺️

And I'm also not assuming intent on any prior comment's tone, just speaking in a general sense!

Growiel:
I get both of your points but I can't really contribute mote in this specific case: the PR is already there and working, the only thing I can do is make sure it isn't forgotten.

Not because the project is inactive, like @ansonhoyt seems to think I meant, but because there's so much going on, it's easy for one to slip through the cracks.

Before investing the time to do what @jon-sully suggested, it's better to know if the official fix will be released or not. 

Why ? Because based on other PRs like this one: https://github.com/hotwired/turbo/issues/56 sometimes while the use case is relevant (at least 3 tickets and 2 questions on the discussions site exist), the maintainers have decided that this specific case will not be supported.

So before I invest in a custom fork, I'd rather ask if I can hope to see official support one day.

Hope it makes sense.

ansonhoyt:
@Growiel I'm just suggesting you look at the issue (in)activity in context of the project activity. I'd probably ask for status too if I see an RC release come out without this being addressed. That'd be a good signal this might be forgotten.

No worries though, I know it's easy to misread someone's comment. Sorry that I misread yours. I'm just another guy excited to get this working too. 😄

jon-sully:
Hey @seanpdoyle what lead to this one getting closed? 👀

mrhead:
@seanpdoyle What do you think about rendering 4xx responses within the frame, but replacing the whole page for 5xx responses?

5xx responses are usually designed for the whole page, and they won't fit small frames.

seanpdoyle:
@mrhead I think that's a reasonable expectation. Since Frame requests with a `5xx` status aren't handled at all, introducing that behavior feels like it could be better addressed in a separate pull request.

tleish:
> @seanpdoyle What do you think about rendering 4xx responses within the frame, but replacing the whole page for 5xx responses?
>
> 5xx responses are usually designed for the whole page, and they won't fit small frames.

I think the logic should be simpler.  If the turbo-frame response does not include the expected turbo-frame id, replace the whole page, regardless if the response is a 2xx, 3xx, 4xx or 5xx.

This solves multiple success or error cases when you want the response to turbo-frame to replace the whole page.

seanpdoyle:
> If the turbo-frame response does not include the expected turbo-frame id, replace the whole page and update the URL, regardless if the response is a 2xx, 3xx, 4xx or 5xx.

@tleish I think that's also a reasonable behavior to explore. If I'm understanding your proposal correctly, that feels outside the scope of this proposal.

This PR is scoped to handling `422` status codes served in response to a `<turbo-frame>` element that targets `_top`. In this case, there are matching `<turbo-frame>` elements in both the requesting document and the response body. Navigating the page when a matching `<turbo-frame>` is missing in the response might make this logic simpler, but that strikes me as an enhancement to make during a second pass, after this change is shipped. Does that sound reasonable?

mollerhoj:
@seanpdoyle should this not also be the case for non-form submissions? If a `turbo_frame` has a `src:` attribute, and the response is a `422` or a `500`, should that not also be rendered? Is far as I can tell, turbo has no way of dealing with this case?

glaszig:
what’s the stopper here? this missing behavior is what keeps me from using turbo and others even (wrongfully) assume this behavior, see e.g. hotwired/turbo-rails#122.

jon-sully:
I can't speak for @seanpdoyle but I believe this draft PR represents an idea that can accommodate several different functionalities, but there are a handful of other draft and/or open PRs aiming to accommodate other functionalities that may be the same. E.g. there are a lot of ideas floating around in the "make frames more flexible" space at the moment, and none has been settled on yet by the core team(s). Thus, this PR is still just an idea for consideration as it pertains to the overall Turbo API and expectations, less so a fully-featured implementation ready-for-merge.

markjerz:
Firstly, I love turbo. However, forms inside frames that should either redirect the whole page or re-render when there's an error continually trips me up.

This PR and issue #138 are a good example of how I don't seem to be the only one.

From #138 it seems that the suggested solution is to return a turbo-stream on failure as opposed to a 422 response containing the stream. 

If that is the case, could this pattern please be documented(/supported) in the handbook? I assume you have this kind of UI in Basecamp/Hey so I, at least, would really benefit from the authors of turbo indicating how they would solve this particular use case.

If that's not the case, and supporting this use-case via frames (as per this PR) is desirable can we have that instead?!

Thanks!

PR #212, opened 2021-03-16
Navigate a lazy-loaded turbo-frame

seanpdoyle:
Problem
---

Consider a `<turbo-frame>` element with `loading="lazy"`. Once it
becomes visible, the `AppearanceObserver` will delegate to the
`FrameController` to load its contents.

However, once visible, subsequent navigations of that frame will
continue to be deferred, since its `[loading]` attribute is still
deferred, in spite of the fact that the element is visible on that page.

Proposed change
---

This commit proposes to immediately navigate a `<turbo-frame>` element
with `[loading="eager"]` _or_ a `<turbo-frame>` element that has been
loaded previously.

Possible alternatives
---

As an alternative to adding another conditional to the guard clause,
could it be worthwhile to treat navigational changes driven by `<a>`
clicks or `<form>` submissions different from programatically modifying
the element's `[src]` attribute? Do Custom Elements support changing an
attribute's value without also kicking off change callbacks?

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -22,6 +22,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   formSubmission?: FormSubmission
   private resolveVisitPromise = () => {}
   private connected = false
+  private hasBeenLoaded = false
 
   constructor(element: FrameElement) {
     this.element = element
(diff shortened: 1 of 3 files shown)


sstephenson:
Intuitively I'd expect setting src on a lazy frame to take effect as soon as the element intersects the viewport, but a quick test shows that \ seems to behave the way you describe above. So, let's roll with it. Could you rebase this against main?

seanpdoyle:
@sstephenson I've rebased and updated this branch.

sstephenson:
Thank you!

Issue #101, opened 2021-01-12
Permanent elements are not always preserved


stgm:
Related to #80 

**Actual problem**

Permanent elements are not always preserved because in certain circumstances the "previous page" is cached in a state where permanent elements have been replaced by meta tags. This means that when browsing "back", the incorrectly cached page is reinstated, and the incorrectly cached meta tags aren't replaced by the current contents of their associated permanent elements.

**Circumstances**

For me this **only** happens in Safari when I navigate towards a page containing a Youtube embed. It doesn't happen when I navigate to a very similar page that doesn't have such an embed. It is also completely consistent, so it always happens.

**What does the log say**

`View.cacheSnapshot()` is responsible for cloning the DOM of the old page. In general circumstances, the line making the clone executes at the very end of the page navigation sequence, right after `Visit.loadResponse()` is completely done.

With the Youtube embed present, the cloning code from `View.cacheSnapshot()` is executed much earlier, right when the new body is assigned: `assignNewBody()` -> `replaceElementWithElement()` -> `parentElement.replaceChild()` (and before `replaceElementWithElement()` returns).

**What does the debugger say**

Unfortunately, not much. When stepping into the `parentElement.replaceChild()` call, execution immediately jumps to the cloning part of `View.cacheSnapshot()`.

Also, no Youtube JS is encountered in the debugger until after Turbo finishes the page transition.

Interestingly, if the new page itself contains a `script` tag, it will be executed right after the `parentElement.replaceChild()` call. This does not affect in any way the moment that the cloning code is run: that still depends only on the presence of the Youtube embed.

**await vs setTimeout**

The old page DOM is supposed to be cached _after rendering_ the new page because this speeds up presenting the new page. This is effected by splitting off the cloning code as a new task using `await`. Previously, this was done using `setTimeout()`.

As it happens, putting back `setTimeout()` does indeed ensure that the cloning always happens at the end of the navigation sequence.

**Conclusion**

For some reason, it seems that calling `replaceChild()` for the body, when the new body contains a Youtube embed, yields control to queued code.

Is this the Micro task queue vs the normal task queue? But why would it be triggered by the Youtube embed and not generally?

I would appreciate if anyone has any pointers on how to debug this further!

stgm:
Additional note: Firefox doesn't exhibit this behaviour.

stgm:
Minimal problematic example is here:
https://stgm.github.io/permanent-element-bug/

[...]

PR #218, opened 2021-03-26
[RFC] Cancelable rendering

### LINKED ISSUE #197: Define custom renderer
Turbo drive doesn't offer any option to define custom rendering logic AFAIK. The current implementation simply [replaces an old node with a new one](https://github.com/hotwired/turbo/blob/8bce5f17cd697716600d3b34836365ebcdc04b3f/src/core/renderer.ts#L84). This can lead to unwanted side-effects (e.g. flickering content or losing dirty input states). I want to be able to define a renderer that uses a DOM diffing library like [morphdom](https://github.com/patrick-steele-idem/morphdom). 
The easiest way would have been to use inheritance to create a custom PageView with a custom Renderer, but the necessary classes are not being exported (probably for good reasons).

Is there any clean way to implement a DOM diffing renderer?

tbo:
This PR allows to cancel the default rendering operation in any `turbo:before-render` handler. Its intended usage is to allow the application of custom rendering logic (e.g. [DOM diffing](https://github.com/turbolinks/turbolinks/issues/184)). It is compatible with full page and frame based rendering. Example usage:

```typescript
import morphdom from 'morphdom';

document.addEventListener('turbo:before-render', e => {
  e.preventDefault();
  const to: Element = (e as any).detail.newBody;
  let frame: HTMLElement | null;
  const from = to.nodeName === 'TURBO-FRAME' && (frame = document.getElementById(to.id)) ? frame : document.body;
  morphdom(from, to);
});
```
This should close #197.

I would add the necessary tests if a maintainer agrees with the general approach.

### DIFF
--- a/src/core/drive/page_snapshot.ts
+++ b/src/core/drive/page_snapshot.ts
@@ -13,7 +13,8 @@ export class PageSnapshot extends Snapshot<HTMLBodyElement> {
   }
 
   static fromDocument({ head, body }: Document) {
-    return new this(body as HTMLBodyElement, new HeadSnapshot(head))
+    // We need to clone the body because a custom renderer might mutate it.
+    return new this((body.cloneNode(true)) as HTMLBodyElement, new HeadSnapshot(head))
   }
 
   readonly headSnapshot: HeadSnapshot
--- a/src/core/view.ts
+++ b/src/core/view.ts
@@ -3,7 +3,7 @@ import { Snapshot } from "./snapshot"
 import { Position } from "./types"
 
 export interface ViewDelegate<S extends Snapshot> {
-  viewWillRenderSnapshot(snapshot: S, isPreview: boolean): void
+  viewWillRenderSnapshot(snapshot: S, isPreview: boolean): Event
   viewRenderedSnapshot(snapshot: S, isPreview: boolean): void
   viewInvalidated(): void
 }
@@ -50,12 +50,15 @@ export abstract class View<E extends Element, S extends Snapshot<E> = Snapshot<E
     }
 
     const { isPreview, shouldRender, newSnapshot: snapshot } = renderer
+
     if (shouldRender) {
       try {
         this.renderer = renderer
         this.prepareToRenderSnapshot(renderer)
-        this.delegate.viewWillRenderSnapshot(snapshot, isPreview)
-        await this.renderSnapshot(renderer)
+        const event = this.delegate.viewWillRenderSnapshot(snapshot, isPreview)
+        if (!event.defaultPrevented) {
+          await this.renderSnapshot(renderer)
+        }
         this.delegate.viewRenderedSnapshot(snapshot, isPreview)
         this.finishRenderingSnapshot(renderer)
       } finally {
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -2,7 +2,7 @@ import { FrameElement, FrameElementDelegate, FrameLoadingStyle } from "../../ele
 import { FetchMethod, FetchRequest, FetchRequestDelegate, FetchRequestHeaders } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { AppearanceObserver, AppearanceObserverDelegate } from "../../observers/appearance_observer"
-import { parseHTMLDocument } from "../../util"
+import { dispatch, parseHTMLDocument } from "../../util"
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { ViewDelegate } from "../view"
@@ -181,7 +181,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   // View delegate
 
   viewWillRenderSnapshot(snapshot: Snapshot, isPreview: boolean) {
-
+    return dispatch("turbo:before-render", { detail: { newBody: snapshot.element}, cancelable: true})
   }
 
   viewRenderedSnapshot(snapshot: Snapshot, isPreview: boolean) {
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -190,7 +190,7 @@ export class Session implements HistoryDelegate, LinkClickObserverDelegate, Navi
   }
 
   viewWillRenderSnapshot({ element }: PageSnapshot, isPreview: boolean) {
-    this.notifyApplicationBeforeRender(element)
+    return this.notifyApplicationBeforeRender(element)
   }
 
   viewRenderedSnapshot(snapshot: PageSnapshot, isPreview: boolean) {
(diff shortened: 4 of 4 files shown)


src/core/view.ts
```diff
@@ -50,12 +50,15 @@ export abstract class View<E extends Element, S extends Snapshot<E> = Snapshot<E
...
+
     if (shouldRender) {
       try {
         this.renderer = renderer
         this.prepareToRenderSnapshot(renderer)
-        this.delegate.viewWillRenderSnapshot(snapshot, isPreview)
-        await this.renderSnapshot(renderer)
+        const event = this.delegate.viewWillRenderSnapshot(snapshot, isPreview)
+        if (!event.defaultPrevented) {
+          await this.renderSnapshot(renderer)
```
seanpdoyle, line 60: Snapshot rendering involves more than setting `innerHTML` (e.g. permanent element transfer).

I wonder if _not_ preventing the default, and instead providing some form of "renderer" interface as part of the `event.detail` payload, then allow the event listeners to provide their own (via `event.detail.render = ...`).

tbo: You are right. I didn't consider that.

I was exploring another route as well. It is possible to overload an `Element`'s `replaceChild` function. Either by monkey-patching it:
```typescript
document.documentElement.replaceChild = <T extends Node>(newChild: Node, oldChild: T) {
  const clone = oldChild.cloneNode(true) as T;
  morphdom(oldChild, newChild);
  return clone;
};
```
Or by extending the elements with web components. Its usage could look like this:
```html
  <body is="morphable">
```
Turbo Frames could be extended in a similar fashion (`class MorphableTurboFrame extends TurboFrame { ...`). The good thing about this solution would be that it requires basically no change to turbo drive. Only the [cloning of the body before caching it](https://github.com/hotwired/turbo/pull/218/files#diff-1e496dfcade128e45d250f8728df33466d2fd2b35303d626be406cb4f92083a5R17) would be needed for mutating renderers. I'm still a bit on the fence what the cleanest solution would be.

tbo: Well I guess it is up to @sstephenson to decide how the API should look like. So some input about the general direction would be much appreciated.

dhh:
Do you need anything beyond what #290 offers?

tbo:
@dhh #290 doesn't quite work for [my use case](https://github.com/hotwired/turbo/issues/197). I just want to replace the render logic and leave the rest untouched. I think this case was already discussed [here](https://github.com/hotwired/turbo/pull/290#issuecomment-878496275). I might be able to monkey patch the `PageRenderer` now that it is [exposed](https://github.com/hotwired/turbo/pull/305), but I don't think that this approach is future-proof.

There is also an additional issue that hasn't been addressed in the other PRs. The snapshotting mechanism doesn't account for mutating renderers with references to the old DOM. If the DOM is being partially reused between renders (e.g. in DOM Diffing), then some mutations are applied to the snapshot as well. The default renderer simply replaces the old DOM and therefore doesn't have this issue. The [fix](https://github.com/hotwired/turbo/pull/218/files#diff-1e496dfcade128e45d250f8728df33466d2fd2b35303d626be406cb4f92083a5R17) is rather simple. I could create a separate PR for it. 

The rest of this PR is outdated and I would propose that we find an alternative solution. The discussions in #290 and #305 look promising.

dhh:
@tbo Please do investigate a proper supported solution for custom rendering 👍

internets:
> There is also an additional issue that hasn't been addressed in the other PRs. The snapshotting mechanism doesn't account for mutating renderers with references to the old DOM. If the DOM is being partially reused between renders (e.g. in DOM Diffing), then some mutations are applied to the snapshot as well. The default renderer simply replaces the old DOM and therefore doesn't have this issue. The [fix](https://github.com/hotwired/turbo/pull/218/files#diff-1e496dfcade128e45d250f8728df33466d2fd2b35303d626be406cb4f92083a5R17) is rather simple. I could create a separate PR for it.

Thanks for tracking that issue down, @tbo! I would love to see the fix for the snapshotting mechanism w/ mutating renderers in a separate PR. That (combined with the exported PageRenderer) would give us a workable solution for preventing iframe refresh.

tbo:
@internets I'll create a separate PR for the fix over the weekend.

src/core/drive/page_snapshot.ts
```diff
@@ -13,7 +13,8 @@ export class PageSnapshot extends Snapshot<HTMLBodyElement> {
   }
 
   static fromDocument({ head, body }: Document) {
-    return new this(body as HTMLBodyElement, new HeadSnapshot(head))
+    // We need to clone the body because a custom renderer might mutate it.
+    return new this((body.cloneNode(true)) as HTMLBodyElement, new HeadSnapshot(head))
```
domchristie, line 17: Unfortunately I think this breaks scrolling. I think because: https://github.com/hotwired/turbo/blob/49f82135e2a5ec8e3279f78373d69214d06a34fa/src/core/view.ts#L26 references an element that's no longer visible in the DOM

tbo: I just glanced over the implementation, but `getElementForAnchor` queries the snapshot DOM and shouldn't rely on any fixed reference within. 
I'll build a small prototype and make sure that this case is covered.

domchristie: FWIW I have been working on a proof-of-concept for custom rendering on this branch: https://github.com/hotwired/turbo/compare/main...domchristie:custom_rendering 

Demo here: https://glitch.com/edit/#!/magnificent-four-friend?path=turbo.js%3A701%3A70

If you put a debugger on line 702 of turbo.js, and inspect `element`, `element.isConnected` is false

internets: There are definitely test failures when cloning the body in `Snapshot.fromDocument`, not just with anchor scrolling but also with permanent elements, preserving video playback, and possibly others. A lot of render operations depends on the snapshot element being connected.

Turbo already clones the snapshot before putting it into the cache, but it does it after a delay. That delay is the reason why our mutated snapshot makes it into the snapshotCache. If we eliminate the delay, our mutating render works like we want it to. I'm not 100% sure of the reason behind the delay, but I believe it to be a performance optimization.

#101 describes an issue with the old solution of waiting a micro task tick (still too long for mutating renders), and the original implementation (waiting a micro task) can be seen in  https://github.com/hotwired/turbo/commit/a7d02f0#diff-9519a55c38756c9b6bee1838521a79ce71017312f257330c52f19e469083155aR32-R39 with no documentation on the reason for waiting.

Out of curiosity I did build the minimal reproduction referenced in #101 in glitch with no delay before cloning here https://glitch.com/edit/#!/half-northern-thunder?path=index.html%3A7%3A30 and I am unable to reproduce the problem outlined. So at least this wouldn't cause a regression on that :)

Here's what the method would look like:

```js
  cacheSnapshot() {
    if (this.shouldCacheSnapshot) {
      const { snapshot, lastRenderedLocation: location } = this
      this.snapshotCache.put(location, snapshot.clone())
    }
  }
```

All tests pass with that change. Thoughts?

Issue #223, opened 2021-03-29
The src attribute of turbo-frame does not reflect the actual URL when redirected


sukei:
Hello,

I'm currently building a wizard form. Each step redirect the user to the next one once validated. I follow [this issue](https://github.com/hotwired/turbo/issues/50) with great interest to improve the user experience regarding the browser history. In the meantime, I just used the stimulus controller made by [@Intrepidd](https://gist.github.com/Intrepidd/bb1ffc5944a5c1ec3a9f5582753c4b67) to reflect the navigation context of the frame to my browser history. Unfortunately, the turbo-frame didn't update the src attribute when redirected ; whether it be after following a link or after a form submission. Is there any reason that the turbo-frame src attribute to be out of step with reality?

seanpdoyle:
Thank you for opening this issue. I've opened https://github.com/hotwired/turbo/pull/224 to attempt to correct this behavior.

PR #224, opened 2021-03-29
Frames: Set `src` when a `<form>` redirects frame

seanpdoyle:
Closes [#223][]

When handling a `<form>` submission targeting a `<turbo-frame>` element
that results in a `303` redirect, update the `<turbo-frame src="...">`
to reflect the new location.

To account for the fact that `FrameController` instances are monitoring
changes to `<turbo-frame src="...">` via a Custom Elements
[attributeChangedCallback][], momentarily toggle that controller's
ability to respond to those changes to allow for modifying the `src`
attribute without triggering a redundant HTTP request.

Additionally, add coverage to ensure that navigating a `<turbo-frame>`
through an `<a>` sets the `[src]` attribute in way that is public to
the document.

[#223]: https://github.com/hotwired/turbo/issues/223
[attributeChangedCallback]: https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#using_the_lifecycle_callbacks

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -20,6 +20,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   readonly formInterceptor: FormInterceptor
   currentURL?: string
   formSubmission?: FormSubmission
+  private loadSourceURLWhenChanged = true
   private resolveVisitPromise = () => {}
   private connected = false
 
@@ -68,7 +69,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   async loadSourceURL() {
-    if (this.isActive && this.sourceURL != this.currentURL) {
+    if (this.loadSourceURLWhenChanged && this.isActive && this.sourceURL != this.currentURL) {
       const previousURL = this.currentURL
       this.currentURL = this.sourceURL
       if (this.sourceURL) {
@@ -84,9 +85,15 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     }
   }
 
-  async loadResponse(response: FetchResponse) {
+  async loadResponse(fetchResponse: FetchResponse) {
+    if (fetchResponse.redirected) {
+      this.loadSourceURLWhenChanged = false
+      this.element.src = fetchResponse.response.url
+      this.loadSourceURLWhenChanged = true
+    }
+
     try {
-      const html = await response.responseHTML
+      const html = await fetchResponse.responseHTML
       if (html) {
         const { body } = parseHTMLDocument(html)
         const snapshot = new Snapshot(await this.extractForeignFrameElement(body))
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -20,9 +20,9 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   readonly formInterceptor: FormInterceptor
   currentURL?: string
   formSubmission?: FormSubmission
-  private loadSourceURLWhenChanged = true
   private resolveVisitPromise = () => {}
   private connected = false
+  private settingSourceURL = false
 
   constructor(element: FrameElement) {
     this.element = element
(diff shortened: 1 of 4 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -72,9 +73,15 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     }
   }
 
-  async loadResponse(response: FetchResponse) {
+  async loadResponse(fetchResponse: FetchResponse) {
+    if (fetchResponse.redirected) {
+      this.listenForChangestoSrc = false
```
seanpdoyle, line 78: I'm dissatisfied with toggling the `listenForChangesToSrc` private property, but it illustrates the point and helps pass the suite.

When responding to successful form submissions, the `Navigator` proposes a visit.

https://github.com/hotwired/turbo/blob/8bce5f17cd697716600d3b34836365ebcdc04b3f/src/core/drive/navigator.ts#L86-L88

Unfortunately, Frames don't have the same access to a Navigator instance at the moment, so our options are limited.

src/core/frames/frame_controller.ts
```diff
@@ -72,9 +73,15 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     }
   }
 
-  async loadResponse(response: FetchResponse) {
+  async loadResponse(fetchResponse: FetchResponse) {
+    if (fetchResponse.redirected) {
+      this.loadSourceURLWhenChanged = false
```
seanpdoyle, line 78: I'm dissatisfied with toggling the `loadSourceURLWhenChanged` private property, but it illustrates the point and helps pass the suite.

When responding to successful form submissions, the `Navigator` proposes a visit.

https://github.com/hotwired/turbo/blob/8bce5f17cd697716600d3b34836365ebcdc04b3f/src/core/drive/navigator.ts#L86-L88

Unfortunately, Frames don't have the same access to a Navigator instance at the moment, so our options are limited.

src/core/frames/frame_controller.ts
```diff
@@ -20,6 +20,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   readonly formInterceptor: FormInterceptor
   loadingURL?: string
   formSubmission?: FormSubmission
+  private loadSourceURLWhenChanged = true
```
sstephenson, line 23: Let's name the property after the state it represents rather than the behavior it triggers.

src/core/frames/frame_controller.ts
sstephenson: I don't think there's any problem per se with toggling a private property, but that logic should live elsewhere. It's not operating at the same level of abstraction as the rest of this method.

sstephenson (changes requested):
Suggest pulling the logic out into a `sourceURL` setter:

```diff
diff --git a/src/core/frames/frame_controller.ts b/src/core/frames/frame_controller.ts
index 996da12..e070198 100644
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -20,9 +20,9 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   readonly formInterceptor: FormInterceptor
   currentURL?: string
   formSubmission?: FormSubmission
-  private loadSourceURLWhenChanged = true
   private resolveVisitPromise = () => {}
   private connected = false
+  private settingSourceURL = false
 
   constructor(element: FrameElement) {
     this.element = element
@@ -69,7 +69,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   async loadSourceURL() {
-    if (this.loadSourceURLWhenChanged && this.isActive && this.sourceURL != this.currentURL) {
+    if (!this.settingSourceURL && this.isActive && this.sourceURL != this.currentURL) {
       const previousURL = this.currentURL
       this.currentURL = this.sourceURL
       if (this.sourceURL) {
@@ -87,9 +87,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
 
   async loadResponse(fetchResponse: FetchResponse) {
     if (fetchResponse.redirected) {
-      this.loadSourceURLWhenChanged = false
-      this.element.src = fetchResponse.response.url
-      this.loadSourceURLWhenChanged = true
+      this.sourceURL = fetchResponse.response.url
     }
 
     try {
@@ -285,6 +283,12 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     }
   }
 
+  set sourceURL(sourceURL: string | undefined) {
+    this.settingSourceURL = true
+    this.element.src = sourceURL ?? null
+    this.settingSourceURL = false
+  }
+
   get loadingStyle() {
     return this.element.loading
   }
```

seanpdoyle:
@sstephenson thank you for the review. I've incorporated your feedback and have pushed up changes to reflect your suggestions.

sstephenson:
Wonderful!

PR #222, opened 2021-03-28
Preserve playback state in permanent HTMLMediaElements

seanpdoyle:
Closes [#221][]

Add test coverage as part of the `rendering_tests.ts`, in an effort to
reproduce [#221][].

[#221]: https://github.com/hotwired/turbo/issues/221

PR #226, opened 2021-04-01
Persist playback state of permanent HTMLMediaElements

### LINKED ISSUE #221: Is Turbo "data-turbo-permanent" designed to work differently to Turbolinks "data-turbolinks-permanent"?
Hello.

Currently I am using Turbolinks for some projects. But after release Turbo and this phrase in Turbolinks README:

> Please note that Turbolinks is no longer under active development. It has been superseded by a new framework called Turbo, which is part of the Hotwire umbrella.

I decided to move projects to Turbo. In the process of migration I find out, that Turbo "data-turbo-permanent" work differently, than Turbolinks "data-turbolinks-permanent" with audio/video tag.

I did example, which contain the same codebase, except one use Turbo and other use Turbolinks (changed data attributes and events listeners only)

Here example with turbolinks: 

https://turbo-permanent.catware.org/turbolinks/ (click play audio and after this travel by "HOME" and "ABOUT" links - audio works without any issues, audio controls still working)

The same code, but with Turbo:

https://turbo-permanent.catware.org/turbo/ (click play audio and after this travel by "HOME" and "ABOUT" links - audio will stop immediately after first page visit)

Here source code for

Turbolinks 
JS: https://github.com/le0pard/turbo-permanent-issue/blob/main/webpack/turbolinks_controllers/audio_controller.js 
HTML: https://github.com/le0pard/turbo-permanent-issue/blob/main/source/layouts/turbolinks.html.erb#L57-L63

Turbo 
JS: https://github.com/le0pard/turbo-permanent-issue/blob/main/webpack/turbo_controllers/audio_controller.js
HTML: https://github.com/le0pard/turbo-permanent-issue/blob/main/source/layouts/turbo.html.erb#L56-L62

Code the same, except for Turbo in HTML used "data-turbo-permanent" instead "data-turbolinks-permanent" and in JS used events "turbo:before-cache" and "turbo:load" instead "turbolinks:before-cache" and "turbolinks:load"

I read about some solutions https://github.com/turbolinks/turbolinks/issues/157#issuecomment-423275861 , but "pause/play" solution is noticeable for the end user.

In the end my question: is this designed to work differently? Is this mean, that Turbolinks still better for use cases, where needed "permanent" element with audio/video tag or it is just bug in Turbo, which do not have Turbolinks?

Versions:
Turbolinks: 5.2.0
Turbo: 7.0.0-beta.4

Thanks

sstephenson:
* Fixes #221; permanent media elements (`<audio>`, `<video>`) now correctly preserve their state across page and frame changes.
* Extracts a new internal class for managing the transition of permanent elements from one snapshot to another.
* Replaces the internal `replaceElementWithElement()` helper with `Element#replaceWith()`. Older browsers will need a polyfill.
* Supersedes #222.


src/tests/fixtures/rendering.html
```diff
@@ -5,6 +5,22 @@
     <title>Turbo</title>
     <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
     <script src="/src/tests/fixtures/test.js"></script>
+    <script type="module">
+      const permanentAudioElementLoaded = new Promise(resolve => {
+        addEventListener("DOMContentLoaded", () => {
+          const element = document.getElementById("permanent-audio")
+          element.addEventListener("canplaythrough", () => resolve(element), { once: true })
```
seanpdoyle, line 12: Could this be declared as `addEventListener("canplaythrough", () => ... { once: true, capture: true })` so that we're not declaring nested handlers?

Issue #227, opened 2021-04-02
Unable to opt out of Turbo for forms within a frame


patricklindsay:
According to [the docs](https://turbo.hotwire.dev/handbook/drive#disabling-turbo-drive-on-specific-links-or-forms) it's possible to disable Turbo for specific forms. However this doesn't seem possible within a Turbo frame, regardless of whether I add `data-turbo="false"` directly within the form tag or within a parent tag Turbo seems to ignore it and still carries out a fetch. Is this by design or an issue? I'm using `7.0.0-beta.4`

I can do;
```
<form action="https://third-party-domain.example/form-url'" data-turbo-frame="_top">
...
</form
```
Which will submit the form without a fetch however for my use case the form action is external and the required UX is for the form to be submitted and for the response to open in a new tab. Previous to use Turbo this behaviour was achieved by adding `target='blank'` on the form. I tried `data-turbo-frame="_blank"` but this didn't work. 

Note that as the form is external, when it is submitted I receive the following CORs error;

```
Access to fetch at 'https://third-party-domain.example/form-url' from origin 'http://localhost:3000' has been blocked by CORS policy: Request header field turbo-frame is not allowed by Access-Control-Allow-Headers in preflight response.
```

cjaoude:
Great 👍  allows Livewire forms to work properly.

PR #228, opened 2021-04-02
Skip form[data-turbo="false"] submissions in frame

seanpdoyle:
Closes [#227][]

Adds a set of conditional guards to the
`FrameController.shouldInterceptFormSubmission()` method, similar to the
[Session.willSubmitForm()][] method, implementing a [identical guard
logic][]

[#227]: https://github.com/hotwired/turbo/issues/227
[Session.willSubmitForm()]: https://github.com/hotwired/turbo/blob/8bce5f17cd697716600d3b34836365ebcdc04b3f/src/core/session.ts#L157-L159
[identical guard logic]: https://github.com/hotwired/turbo/blob/8bce5f17cd697716600d3b34836365ebcdc04b3f/src/core/session.ts#L252-L259

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -11,6 +11,7 @@ import { FormInterceptor, FormInterceptorDelegate } from "./form_interceptor"
 import { FrameView } from "./frame_view"
 import { LinkInterceptor, LinkInterceptorDelegate } from "./link_interceptor"
 import { FrameRenderer } from "./frame_renderer"
+import { elementIsNavigable } from "../session"
 
 export class FrameController implements AppearanceObserverDelegate, FetchRequestDelegate, FormInterceptorDelegate, FormSubmissionDelegate, FrameElementDelegate, LinkInterceptorDelegate, ViewDelegate<Snapshot<FrameElement>> {
   readonly element: FrameElement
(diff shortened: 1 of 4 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -295,3 +295,12 @@ function activateElement(element: Node | null) {
     return element
   }
 }
+
+function elementIsNavigable(element: Element) {
```
seanpdoyle, line 299: For the sake of passing the suite, this function is directly copied from the implementation in `Session`.

It's unclear to me the difference between `FormSubmitObserver` (and its `Session` delegate) and `FormInterceptor` (and its `FrameController` delegate). Is the concept of a page-level delegate dissimilar enough from a frame-level delegate that the two concepts should be distinctly represented with their own interfaces?

Would it be worthwhile to re-use the `FormSubmitObserver` interface in the case of a `Session` and `FrameController`, then leave it up to the delegate implementations to have different behaviors?

There feels like there is a concept we're missing here, which is leading to the duplicated guard logic.

sstephenson: > For the sake of passing the suite, this function is directly copied from the implementation in Session.

Let's demote `Session#elementIsNavigable()` to a function, export it from `session`, and import it here.

> Is the concept of a page-level delegate dissimilar enough from a frame-level delegate that the two concepts should be distinctly represented with their own interfaces?

I'm pretty sure the duplication is an artifact of the original \<turbolinks-frame> implementation in HEY. Please do explore unification, if you're so inclined (in a separate branch 😄)!

sstephenson (changes requested):

src/core/frames/frame_controller.ts
seanpdoyle: @sstephenson I've opened https://github.com/hotwired/turbo/pull/232 to remove `FormInterceptor` and use `FormSubmitObserver` universally.

PR #236, opened 2021-04-10
Revert "Revert "Add test coverage for toggling data-turbo-preview""

seanpdoyle:
This reverts commit 59fbd8ff4681fc8a541a7282e99c1e5f3d3c3344.

---

While the original implementation change for this commit was covered
elsewhere, there is still value in the tests to add coverage for the
behavior.


seanpdoyle:
I've re-opened this as a draft to continue to debug the CI failures.

PR #238, opened 2021-04-12
Remove stale elements before cache

ayrton:
Some content is inherently stale the moment you've shown it to users. Eg in Ruby on Rails it's common that you use flash messages to communicate a message from your controller to your users.

When a user navigates away and returns to the original page, turbo will briefly show cached content:

https://user-images.githubusercontent.com/440926/107358271-78aa4300-6ad3-11eb-8345-3da01cfe7e3d.mov

This PR adds a hook by adding `data-turbo-cache="false"` to individual dom elements to remove them from Turbo's cache.

--

_Originally posted about the issue in https://github.com/hotwired/turbo-rails/issues/117_

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -1,5 +1,6 @@
 import { Adapter } from "./native/adapter"
 import { BrowserAdapter } from "./native/browser_adapter"
+import { CacheObserver } from "../observers/cache_observer"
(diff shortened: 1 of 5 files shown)


dhh:
I dig this. Can you add a test?

ayrton:
> I dig this. Can you add a test?

Added tests:

```
✓ firefox on mac 20.3.0 - CacheObserverTests - removes stale elements (0.621s)
```

Let me know if something's still missing.

PR #239, opened 2021-04-12
Replace children with duplicate id instead of append/prepend

### LINKED ISSUE #132: Duplicate content, Turbo stream added an already existing item
I am currently working on an app that heavily leans on Hotwire (with Rails), I love working with it! Our apps feels so much more responsive because of Hotwire. 🚀 

There is one thing that I'm struggling with right now. Our app uses a lot of background jobs and sometimes the jobs are delayed for a couple of seconds.

This is our problem: Our app has a form where users create `Vehicles`. When the vehicle is created, I redirect the user back to the Vehicle#index. What sometimes happens, is that after the page is rendered by the server, the Turbo Stream job is picked up. Which sends the created vehicle to the user, even if the page already has that content. If that happens the user sees two identical vehicles (with the same `dom_id`'s), one rendered by the page render, and one added from the Turbo Stream.

It would help if Turbo would check if an `id` already existed, before blinding appending it to the target. Is that something you would be open to? If so, I would love to contribute to this project.

jeromecornet:
This PR introduces 2 new actions `update_children_or_append` and `update_children_or_prepend` which are very similar to `append` and `prepend` except that:
- the contents of the duplicate children are replaced in place (wherever they are in the target) with the contents of the matching children of the template
- the rest of the template (non duplicate children) are appended/prepended to the target

Assuming the following contrived example:
Target: 
```html
<ul id="vehicles">
  <li id="vehicle_1">Old car</li>
  <li id="vehicle_2">New car</li>
</ul>
```
Stream element:
```html
<turbo-stream action="append" target="vehicles">
  <template>
      <li id="vehicle_1">Old car with updates</li>
      <li id="vehicle_3">Newer car</li>
  </template>
</turbo-stream>
```

With the current code, the result is:
```html
<ul id="vehicles">
  <li id="vehicle_1">Old car</li>
  <li id="vehicle_2">New car</li>
  <li id="vehicle_1">Old car with updates</li>
  <li id="vehicle_3">Newer car</li>
</ul>
```

This PR changes it so the result is
```html
<ul id="vehicles">
  <li id="vehicle_1">Old car with updates</li>
  <li id="vehicle_2">New car</li>
  <li id="vehicle_3">Newer car</li>
</ul>
```

For context his was originally introduced as a fix for https://github.com/hotwired/turbo/issues/132#issuecomment-813040051 but after https://github.com/hotwired/turbo/issues/132#issuecomment-817927043 I rebased this PR over #240 and introduced new commands

### DIFF
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -11,6 +11,17 @@ export const StreamActions: { [action: string]: (this: StreamElement) => void } 
     this.targetElement?.prepend(this.templateContent)
   },
 
+  update_children_or_append() {
+    this.replaceDuplicateChildren()
+    this.targetElement?.append(this.templateContent)
+  },
+
+  update_children_or_prepend() {
+    this.replaceDuplicateChildren()
+    this.targetElement?.prepend(this.templateContent)
+  },
+
+
   remove() {
     this.targetElement?.remove()
   },
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -2,10 +2,12 @@ import { StreamElement } from "../../elements/stream_element"
 
 export const StreamActions: { [action: string]: (this: StreamElement) => void } = {
   append() {
+    this.removeDuplicateTargetChildren()
     this.targetElement?.append(this.templateContent)
   },
 
   prepend() {
+    this.removeDuplicateTargetChildren()
     this.targetElement?.prepend(this.templateContent)
   },
 
--- a/src/elements/stream_element.ts
+++ b/src/elements/stream_element.ts
@@ -34,6 +34,13 @@ export class StreamElement extends HTMLElement {
       targetChild.remove();
     })
   }
+ 
+  replaceDuplicateChildren() {
+    this.duplicateChildren.forEach(({targetChild, templateChild}) => {
+      templateChild.remove();
+      targetChild!.replaceWith(templateChild);
+    })
+  }
 
   get duplicateChildren() {
     return [...this.templateContent?.children].map(templateChild => {
--- a/src/elements/stream_element.ts
+++ b/src/elements/stream_element.ts
@@ -28,7 +28,20 @@ export class StreamElement extends HTMLElement {
   disconnect() {
     try { this.remove() } catch {}
   }
+ 
+  removeDuplicateTargetChildren() {
+    this.duplicateChildren.forEach(({targetChild}) => {
+      targetChild.remove();
+    })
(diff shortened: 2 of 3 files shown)


tleish:
I think solution allows developers the flexibility to cover 1 of 2 scenarios.

## 1. Replace or Append/Prepend Multiple Items
If a developer always wants existing items to stay in the same position and not jump to the end.
```html
<turbo-stream action="append" target="list">
  <template>
    <li id="item_2">Item 2 (UPDATED)</li>
    <li id="item_3">Item 3 (UPDATED)</li>
  </template>
</turbo-stream>
```

## 2. Remove and Append/Prepend Multiple Items
If a developer always wants the latest item to show up at the beginning or end.
```html
<turbo-stream action="remove" target="item_2"></turbo-stream>
<turbo-stream action="remove" target="item_3"></turbo-stream>
<turbo-stream action="append" target="list">
  <template>
    <li id="item_2">Item 2 (UPDATED)</li>
    <li id="item_3">Item 3 (UPDATED)</li>
  </template>
</turbo-stream>
```

## More Native Solution?

Even though I proposed it earlier, I wonder if we might want to consider a simpler version of solution 2.  Looking back to one of your original solutions with a new turbo-stream attribute, I wonder if we might want to consider adding a `duplicates` attribute:

```html
<turbo-stream action="append" target="list" duplicates="remove">
  <template>
    <li id="item_2">Item 2 (UPDATED)</li>
    <li id="item_3">Item 3 (UPDATED)</li>
  </template>
</turbo-stream>
```
With the existing code, we'd only need to add one function and then call `replaceDuplicateChildren` or `removeDuplicateChildren` depending on the parameter.
```javascript
  removeDuplicateChildren() {
    this.duplicateChildren.forEach(({targetChild}) => {
      targetChild.remove();
    })
  }
```

Options for `duplicates` attribute might be:
- replace (default)
- remove
- ignore (maybe some want duplicates)

Some issues to consider:
- Is it clear that duplicates is immediate children elements with an id attribute and not all elements or all ancestors?  Since turbo-stream actions deal only with ids, it's intuitive to me and I like the simple attribute name `duplicates`, but others might prefer a name of `duplicate-ids`?
- I think the default should be replace, but do others think the same way?

jeromecornet:
@tleish I updated the description and code after https://github.com/hotwired/turbo/issues/132#issuecomment-817969961

dhh:
I think we're covered with a good default solution in #240. IMO, that'll tackle the vast majority of cases. For something more intricate, it's already possible to have full control by rerendering the entire container.

PR #240, opened 2021-04-12
Remove target children with duplicate ID before append/prepend

### LINKED ISSUE #132: Duplicate content, Turbo stream added an already existing item

jeromecornet:
Fixes #132

@dhh this implements Option C of https://github.com/hotwired/turbo/issues/132#issuecomment-813040051

When appending/prepending, children of the template may contain the same ID as children of the existing target.
_(this is common when using turbo-rails or other frameworks where the append/prepend action comes both as a result of the response to the original action and as a broadcasted action from a background job)_.

When this is the case, the generated result contains duplicate children ID which is undesirable.

This PR changes the behaviour of the append/prepend actions so that the existing children of the target which match the children of the template are removed before appending the template.

Assuming the following contrived example:
Target: 
```html
<ul id="vehicles">
  <li id="vehicle_1">Old car</li>
  <li id="vehicle_2">New car</li>
</ul>
```
Stream element:
```html
<turbo-stream action="append" target="vehicles">
  <template>
      <li id="vehicle_1">Old car with updates</li>
      <li id="vehicle_3">Newer car</li>
  </template>
</turbo-stream>
```

With the current code, the result is:
```html
<ul id="vehicles">
  <li id="vehicle_1">Old car</li>
  <li id="vehicle_2">New car</li>
  <li id="vehicle_1">Old car with updates</li>
  <li id="vehicle_3">Newer car</li>
</ul>
```

This PR changes it so the result is
```html
<ul id="vehicles">
  <li id="vehicle_2">New car</li>
  <li id="vehicle_1">Old car with updates</li>
  <li id="vehicle_3">Newer car</li>
</ul>
```

### DIFF
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -2,10 +2,12 @@ import { StreamElement } from "../../elements/stream_element"
 
 export const StreamActions: { [action: string]: (this: StreamElement) => void } = {
   append() {
+    this.removeDuplicateTargetChildren()
     this.targetElement?.append(this.templateContent)
   },
 
   prepend() {
+    this.removeDuplicateTargetChildren()
     this.targetElement?.prepend(this.templateContent)
   },
 
--- a/src/elements/stream_element.ts
+++ b/src/elements/stream_element.ts
@@ -28,7 +28,20 @@ export class StreamElement extends HTMLElement {
   disconnect() {
     try { this.remove() } catch {}
   }
+ 
+  removeDuplicateTargetChildren() {
+    this.duplicateChildren.forEach(({targetChild}) => {
+      targetChild.remove();
+    })
+  }
 
+  get duplicateChildren() {
+    return [...this.templateContent?.children].map(templateChild => {
+      let targetChild = [...this.targetElement!.children].filter(c => c.id === templateChild.id)[0]
+      return { targetChild , templateChild }
+    }).filter(({targetChild}) => targetChild);
+  }
+  
   get performAction() {
     if (this.action) {
       const actionFunction = StreamActions[this.action]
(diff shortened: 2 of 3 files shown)


acushlakoncept:
I am currently having this issue, how do I apply the solution in my code as the PR is yet to be merged?

jeromecornet:
> I am currently having this issue, how do I apply the solution in my code as the PR is yet to be merged?

@acushlakoncept if you are using turbo-rails, I have created a temporary fork with a patched turbo.js until this gets merged.
`
gem "turbo-rails", github: "jeromecornet/turbo-rails", branch: "dedup" 
`

acushlakoncept:
Thank you @jeromecornet, is `gem "turbo-rails" ...` different from `gem 'hotwire-rails'` that I had?
Sorry I am new with turbo stuff

jeromecornet:
@acushlakoncept The gem `hotwire-rails` depends on the other gem `turbo-rails`, so if you only have `hotwire-rails` in your Gemfile, add another line as mentioned above to force the version of `turbo-rails`, until both `turbo` and `turbo-rails` get updated with that fix.

PR #241, opened 2021-04-13
Clear a target's content from the DOM

ayrton:
Introduce a little helper to clear the content of a target in the DOM. This is useful eg when you are closing a modal

__Before:__

```js
<turbo-stream action="replace" target="message_1">
  <template>
    <div id="message_1">
    </div>
  </template>
</turbo-stream>
```

__After:__

```js
<turbo-stream action="clear" target="message_1">
</turbo-stream>
```

--

_Originally opened a PR in https://github.com/hotwired/turbo-rails/pull/131_

### DIFF
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -14,7 +14,9 @@ export const StreamActions: { [action: string]: (this: StreamElement) => void } 
   },
 
   clear() {
-    this.targetElement?.replaceWith("")
+    if (this.targetElement) {
+      this.targetElement.innerHTML = ""
+    }
   },
 
   replace() {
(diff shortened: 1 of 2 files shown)


sstephenson:
You can already do this with the `update` stream action and an empty body:

```html
<turbo-stream action="update" target="message_1">
  <template></template>
</turbo-stream>
```

ayrton:
The idea is that the following becomes a bit more succinct:

__Before:__

```rb
# create.turbo_stream.erb
<%= turbo_stream.update :modal, "" %>
```

Or a bit more natural to read:

```rb
# create.turbo_stream.erb
<%= turbo_stream.update :modal %>
```

__After:__

```rb
# create.turbo_stream.erb
<%= turbo_stream.clear :modal %>
```

But I understand it's a fine line between what to add as API and what not

PR #251, opened 2021-04-20
Fixes #250 (frameElement.complete returns always false)

### LINKED ISSUE #250: frameElement.complete returns always false on beta5
I had a function which was checking if a frame finished loading that relies on the `complete` attribute.
After upgrading to `beta5` a call to `frameElement.complete` always returns `false` although the frame finished loading.

coorasse:
#250 

I guess this should be tested better, shall I add a regression test?


dhh:
Yes, please do add a test 👍

PR #255, opened 2021-04-22
Support Custom `<turbo-stream action="...">` value

seanpdoyle:
By dispatching the `turbo:before-stream-render` action with an
`event.detail.actions` value set to the default `StreamActions`
instance, clients can merge in their own actions, or override existing
ones by setting mutating `event.detail.actions` in their own event
listeners.

### DIFF
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -1,6 +1,10 @@
 import { StreamElement } from "../../elements/stream_element"
 
-export const StreamActions: { [action: string]: (this: StreamElement) => void } = {
+export interface StreamOperations {
+  [action: string]: (this: StreamElement) => void
+}
+
+export const StreamActions: StreamOperations = {
   append() {
     this.targetElement?.append(this.templateContent)
   },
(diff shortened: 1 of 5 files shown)


rainerborene:
This looks promising. It will be possible to create custom action to replace the default [`assignNewBody`](https://github.com/hotwired/turbo/blob/ffe81b23f5b924ff1bd3d4440569ec0b1f07f86b/src/core/drive/page_renderer.ts#L90) behaviour?

seanpdoyle:
@rainerborene it's possible! This changeset is scoped to Turbo Streams, but if this ends up being a viable proposal, it might be viable to add the same kind of hooks to `turbo:before-render`.

To state the subtext directly, [PageRenderer.assignNewBody()](https://github.com/hotwired/turbo/blob/ffe81b23f5b924ff1bd3d4440569ec0b1f07f86b/src/core/drive/page_renderer.ts#L90) is similar enough to a `<turbo-stream action="update">` operation that that's potential to re-use the concepts and code.

leonardoprg:
@seanpdoyle why did you close this PR? 🤔

PR #263, opened 2021-05-01
Allow a frame to navigate to previous url

### LINKED ISSUE #245: Turbo frame navigation after following a form redirection is break when the URL has been visited before.
Hello, please note that this behavior has changed since the beta-5. Instead of trying to explain the issue, I'm going to describe the behavior I'm facing.

- I have a link that target a frame with `data-turbo-frame="my-id"`.
- When I click to this link, a request is issued. Then, the `<turbo-frame id="my-id">` is successfully updated and its `src` attribute too.
- I have some forms within the frame (a multi-step wizard in fact), when I submit the forms, I'm redirected to the next step, and since the beta-5, the `src` attribute of the frame is updated too. Everything is good so far, I'm very grateful for this one.
- But now, When I click to the original link, no request is issued and the frame remains on the wizard step URL.

My guess: The frame keep somewhere the state of the current active URL and didn't trigger a query when the same one is requested again. But, this state is not updated when following form redirection. So, when a link that match the current state is clicked again, the frame didn't move.

### LINKED ISSUE #249: can only interact with a turbo frame once?
It's probably something silly, but I couldn't figure it out yet...

The problem: **After interacting with a specific turboframe, I cannot interact with it again**.

First off: **the server response does contain the correct turbo-frame-id** :)

Further details:

I have a simple form to inline edit the title of an item. It works on all items on the list perfectly. I click edit, update, click save, and the change is reflected. However if I try to do this another time on the same item, it just doesn't do anything. No errors. If I edit another untouched item, it works fine.

When I inspect the element after updating, it still has the turbo-frame-id provided in the responses, but somehow also added `src` which wasn't returned by the server... I'm not sure if this is a potential clue?

See attached video that hope demonstrates this?

CleanShot 2021-04-17 at 09 07 32

using: turbo-rails gem 0.5.9, turbo-rails 7.0.0-beta.5, rails 6.0 with webpacker 5.2.1

### LINKED ISSUE #265: After a form submission, links in a frame back to the form don't click
Here's a simple server to replicate the issue. It seems to exist in beta 5, but not beta 4.

```ruby
    require 'sinatra'

    get '/test' do
      <<~HTML
<html>
<head>
<script type="module">
  import hotwiredTurbo from 'https://cdn.skypack.dev/@hotwired/turbo';
</script>
</head>
<body>
  <turbo-frame id="message_1">
    <a href="test2">Edit message</a>
  </turbo-frame>
</body>

      HTML
    end

    get '/test2' do
      <<~HTML
<html>
<head>
<script type="module">
  import hotwiredTurbo from 'https://cdn.skypack.dev/@hotwired/turbo';
</script>
</head>
<body>
  <turbo-frame id="message_1">
    <form action="test" method="post"><button type="submit">back to list via button</button></form>
    <a href="test">Back to list</a>
  </turbo-frame>
</body>


      HTML
    end

    post '/test' do
      redirect '/test'
    end
```

- Start the server (`ruby server.rb`) then navigate to http://localhost:4567/test
- Click "edit message"
- Click "back to list"
- Click "edit message" -> this works fine. You can keep clicking the links back and forth forever.
- Click "back to list via button" (ie. submit a form)
- Click "edit message" -> nothing happens, no errors are logged

t27duck:
This is my attempt at a fix based on discussion and the previous PR as well as adding a test for it.

The gist: A frame will only update/make a request if `currentURL` and `sourceURL` are different. `currentURL` was not get updated properly so it was still pointing to the previous value. If `sourceURL` changes back to the original URL of the frame, the values would match and result in the frame not navigating.

Example:
- Frame gets src set to "a". `sourceURL` and `currentURL` are set to "a"
- The frame's src is changed to "b". While `sourceURL` is updated, `currentURL` is still "a". Frame navigates to "b".
- The frame's src is changed back to "a". Because `currentURL` was never changed to "b", navigation doesn't happen because `sourceURL` == `currentURL`. 

Fixes #245 
Fixes #249 (I presume)
Fixes #265 (Reported)

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -19,7 +19,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   readonly appearanceObserver: AppearanceObserver
   readonly linkInterceptor: LinkInterceptor
   readonly formInterceptor: FormInterceptor
-  currentURL?: string
+  currentURL?: string | null
   formSubmission?: FormSubmission
   private resolveVisitPromise = () => {}
   private connected = false
@@ -312,6 +312,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   set sourceURL(sourceURL: string | undefined) {
     this.settingSourceURL = true
     this.element.src = sourceURL ?? null
+    this.currentURL = this.element.src
     this.settingSourceURL = false
   }
 
@@ -337,7 +338,7 @@ function getFrameElementById(id: string | null) {
   }
 }
 
-function activateElement(element: Element | null, currentURL?: string) {
+function activateElement(element: Element | null, currentURL?: string | null) {
   if (element) {
     const src = element.getAttribute("src")
     if (src != null && currentURL != null && urlsAreEqual(src, currentURL)) {
--- a/src/tests/fixtures/frames.html
+++ b/src/tests/fixtures/frames.html
@@ -40,5 +40,8 @@
     </turbo-frame>
 
     <a id="frame-self" href="/src/tests/fixtures/frames/self.html" data-turbo-frame="frame">Visit self.html</a>
+
+    <a id="navigate-form-redirect" href="/src/tests/fixtures/frames/form-redirect.html" data-turbo-frame="form-redirect">Visit form-redirect.html</a>
+    <turbo-frame id="form-redirect"></turbo-frame>
   </body>
 </html>
--- a/src/tests/fixtures/frames/form-redirect.html
+++ b/src/tests/fixtures/frames/form-redirect.html
@@ -1,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Frame</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+  </head>
+  <body>
+    <h1>Page One Form</h1>
+    <turbo-frame id="form-redirect">
+      <h2 id="form-redirect-header">Form Redirect</h2>
+
+      <form action="/__turbo/redirect" method="post" enctype="multipart/form-data">
+        <input type="hidden" name="path" value="/src/tests/fixtures/frames/form-redirected.html">
+        <input type="submit" id="submit-form">
+      </form>
+    </turbo-frame>
+  </body>
+</html>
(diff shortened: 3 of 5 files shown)


tleish:
I might think you would want this use the following change instead to ensure that `this.element.src` and `this.currentURL` match exactly.

```diff
  set sourceURL(sourceURL: string | undefined) {
    this.settingSourceURL = true
    this.element.src = sourceURL ?? null
+   this.currentURL = this.element.src
    this.settingSourceURL = false
  }
```

t27duck:
> I might think you would want this use the following change instead to ensure that `this.element.src` and `this.currentURL` match exactly.
> 
> ```diff
>   set sourceURL(sourceURL: string | undefined) {
>     this.settingSourceURL = true
>     this.element.src = sourceURL ?? null
> +   this.currentURL = this.element.src
>     this.settingSourceURL = false
>   }
> ```

Good idea. Updated. Thanks for pointing that out.

ghiculescu:
Thanks for this, it looks like it fixes https://github.com/hotwired/turbo/issues/265 too.

acetinick:
For those needing this fix until its merged, I have published it as separate npm package here if anyone needs it.

https://www.npmjs.com/package/@peopleforce/turbo

bobmaerten:
I successfully tested on our stack with @acetinick 's module, the revisit problem seems to be gone.
Well played @t27duck!
Hope to have a release soon enough 🙏🏻

sukei:
> I patched this PR into my local code and I can also confirm that it fixed the bug ❤️

Hope it get merged soon!

t27duck:
> Hope it get merged soon!

To be perfectly honest, at this point probably DHH is the only one left with merge and release powers for this repo. So the odds of this (or any future releases of turbo) are pretty slim at this point.

sukei:
Thank you for the status report. I was thinking of waiting, but finally I'll fork it out while waiting for the official support.

acetinick:
@dhh would you be able to review and merge this PR? It's quite a critical bug that is affecting lots of people.  Thanks

Bramjetten:
@t27duck Thanks for fixing this 🙏

@dhh bumped the `turbo-rails` gem to 0.5.10 two days ago, causing more people to run into this bug now. Is @dhh the only one that can review this PR?

dhh:
Thanks for working on this @t27duck. I'll get this merged and cut a new release shortly.

dhh:
Fix included in beta 7 that was just released 👍

PR #277, opened 2021-06-04
Turn links with methods into form submissions

dhh:
Rails UJS offered a path to turn links into form submissions via the data-method attribute. This PR recreates that capability under data-turbo-method (but also with data-method for compatibility with existing apps built to Rails UJS). 

The link is turned into a form that's added next to the link in the dom, such that all the frame scopes still apply, and corresponding behavior is triggered.

For accessibility reasons, we should continue to promote buttons rather than method links, but this makes Hotwire much easier to fit into existing applications that were built to Rails UJS standards.

### DIFF
--- a/src/core/frames/link_interceptor.ts
+++ b/src/core/frames/link_interceptor.ts
@@ -1,3 +1,5 @@
+import { dispatch } from "../../util"
+
 export interface LinkInterceptorDelegate {
   shouldInterceptLinkClick(element: Element, url: string): boolean
   linkClickIntercepted(element: Element, url: string): void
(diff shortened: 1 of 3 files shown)


marcelolx (approved):

Intrepidd:
This seems to only work within a turbo-frame, is it expected behaviour ? I am not sure as it is advertised to be a replacement for data-method links

dhh:
@Intrepidd Ah yes, need to expand it to all cases. If you want to take a swing at a PR for that 👍

Intrepidd:
@dhh I took a swing at it in #299 , I need a little bit of guidance though before wrapping it up

PR #280, opened 2021-06-04
Set window.Turbo on import

excid3:
This makes sure Turbo is available for Request.js and Turbo Native out
of the box.

Related to rails/request.js#6


excid3:
Added a test to make sure the ESM module also sets `window.Turbo`

This one should be ready to go @dhh @marcelolx

michelson:
are you sure that this a good idea @excid3 ? 
what happens if a user does not want to expose this to window ? wouldn't be better to delegate this to the user ?

marcelolx (approved):
Tested here and is working as expected.

marcelolx:
> are you sure that this a good idea @excid3 ?
> what happens if a user does not want to expose this to window ? wouldn't be better to delegate this choice to the user ?

@michelson Have you a scenario where it would be a bad idea Turbo exposing itself to the window? I'm asking because I really don't know if there would be a problem or not

dhh:
I think this is indeed the better default, but I'd also like to see an option where someone could import Turbo without triggering window + autostart. That would be a separate build target, though. That's more of a nice-to-have.

Having window.Turbo set is needed for all the native integrations anyway. And it makes it easier to drop Turbo into a site and start using it right away.

michelson:
> @michelson Have you a scenario where it would be a bad idea Turbo exposing itself to the window? I'm asking because I really don't know if there would be a problem or not
the only scenario I can imagine right now is for example some sort of widget app that could be imported on sites (like chat widgets like intercom) I would expect that app will self contained and not mess with the rest of the app.

vfonic:
@michelson exactly. That's exactly the use-case for _not_ having `window.Turbo` automatically assigned on import. My app is providing a code snippet that any website can add to their page and load some widgets. This all worked great, until I hit a website that also has Rails/Turbo and assigns `window.Turbo`, just like my app does.

There's a version disparity between my app's Turbo and main website Turbo and other issues that I'm looking to avoid, but I don't know how. Any ideas? Was a separate build ever made?
Wouldn't it be better to do: `import { Turbo } from "./core"` and pass that non-global variable to Request.js constructor and wherever else it's needed?

BTW I also had to remove `Stimulus` from being globally available. Luckily that was easy as that code was generated and part of my Rails app code. Too bad that's not the case with `Turbo`:

Screenshot 2025-03-18 at 15 17 51

https://github.com/rails/request.js/pull/6#issuecomment-854798055
Screenshot 2025-03-18 at 15 19 02

Would be great if this was more "omakase". Rails/Turbo/Hotwire provide a way and if you don't like it, you can "choose your own adventure". But in this case "choose your own adventure" would be: build and compile your own version of Turbo without that one line of code: `window.Turbo = ...`

Also, Turbo is not very specific, unique word for a global variable name `window.Turbo`. There's a TypeScript/JavaScript package manager called turbo. It's got 4M+ downloads on NPM: https://www.npmjs.com/package/turbo
If, for whatever reason, this or some other "turbo" package would assign `window.Turbo`...well, we'd get reminded why we moved away from assigning to a global namespace. I strongly believe assigning to global namespace should be left to user to manage.

PR #285, opened 2021-06-16
Don't overwrite children without ID when appending/prepending

### LINKED ISSUE #284: Remove children with duplicate ID's
Pull request #240 seems to be also removing children without an explicit ID set. Is that the intended behavior?

[This line](https://github.com/hotwired/turbo/pull/240/files#diff-abe0bd02e4af84735411ee9b586a062b9191db56bf54a8df92d6cbb04539bda6R40) is calling `id` on the element, which returns empty string if it does not have one.

jeromecornet:
Fixes #284
cc: @dhh @pbstriker38

This was indeed a mistake in #240 so I updated the append/prepend tests to use <span> elements rather than plain text to reproduce the problem and added the fix for it.


jeromecornet:
The CI failure looks like an unrelated (flaky?) test

dhh:
Yeah, that test is super flaky. Would be nice to have someone investigate that as well.

Thanks for the quick fix! I'll get another beta release out in not too long.

src/elements/stream_element.ts
```diff
@@ -36,7 +36,7 @@ export class StreamElement extends HTMLElement {
   }
 
   get duplicateChildren() {
-    return [...this.templateContent?.children].map(templateChild => {
+    return [...this.templateContent?.children].filter(templateChild => !!templateChild.id).map(templateChild => {
       let targetChild = [...this.targetElement!.children].filter(c => c.id === templateChild.id)[0]
       return { targetChild , templateChild }
```
tleish, line 41: @jeromecornet 

I'd like to suggest a slight adjustment.  Since the method is "find duplicate children in the TARGET with the same ID as the TEMPLATE", I think the code should filter the TEMPLATE children with id's.

```diff
get duplicateChildren() {<br class="Apple-interchange-newline">
  return [...this.templateContent?.children].map(templateChild => {
-   let targetChild = [...this.targetElement!.children].filter(c => c.id === templateChild.id)[0]
+   let targetChild = [...this.targetElement!.children].filter(c => c.hasAttribute('id') && c.id === templateChild.id)[0]
    return { targetChild , templateChild }
```
It's also a bit more efficient (less loops).

src/tests/unit/stream_element_tests.ts
```diff
@@ -8,14 +8,22 @@ export class StreamElementTests extends DOMTestCase {
   }
 
   async "test action=append"() {
-    const element = createStreamElement("append", "hello", createTemplateElement(" Streams"))
+    const element = createStreamElement("append", "hello", createTemplateElement("<span> Streams</span>"))
+    const element2 = createStreamElement("append", "hello", createTemplateElement("<span> and more</span>"))
```
tleish, line 12: Would we want to make this more clear what we are testing?

```
"<span> Element without ID</span>"
"<span> another element without ID</span>"
```

tleish (changes requested):

src/elements/stream_element.ts
jeromecornet: > I think the code should filter the TARGET children with id's.
> It's also a bit more efficient (less loops)
Either I don't understand your comment, or I simply don't agree (or both 😄)

the bit I added: `.filter(templateChild => !!templateChild.id)` runs before the `map`, so it ensures that the inner loop is only executed for the ids that exist in the template. So there are less loops in my version than in the ones you suggest, no ?

In my code, the only way `c.id === templateChild.id` in the inner `filter` is that `c.id` is not empty (since the filter ensures that `templateChild.id` is not empty), so the added test ` c.hasAttribute('id')` is redundant.

Am I missing something ?

src/tests/unit/stream_element_tests.ts
jeromecornet: Maybe, although since the goal of this change is to catch regressions, I'm not sure it's _that_ important in the long run. (i.e. if the code is broken at some point in the future this test will fail and it will be obvious why).
I would have updated the code, but the PR is now merged, so if you feel strongly about it feel free to open another PR for it.

src/elements/stream_element.ts
tleish: Clarification: "less loops" assuming the `this.targetElement` has more children than `this. templateContent`.  Otherwise, if the inverse I can see why we might filter by target.

One additional efficiency I noticed the code generates `[...this.targetElement!.children]` array for every templateChild in the loop.  It could be created once before the looping over `this.templateContent?.children`.

tleish: @jeromecornet 

I recognize the redundancy of both changes.  I was suggesting filtering the targetChildren without ids instead of filtering out the templateChildren without ids.  I was not suggesting we do both.  They both accomplish the same result, it was feedback. based on code comprehension/readability and not functionality.

FYI, I started my code review prior to being merged.

src/tests/unit/stream_element_tests.ts
tleish: Again, opinionated feedback based on future code comprehension and readability.  If a year from now a developer is trying to understand "Why are we testing adding span elements without ID's, seems redundant.  We could simplify this test and only check if one item is added".  Adding the additional information provides clarity for future developers to understand that we are testing elements without ids.

src/elements/stream_element.ts
dhh: Was just working on this code in the branch to add css selectors for multiple replaces, and came up with this that I found much easier to follow/read:

```js
  removeDuplicateTargetChildren() {
    this.duplicateChildren.forEach(c => c.remove())
  }
  
  get duplicateChildren() {
    const existingChildren = this.targetElements.flatMap(e => [...e.children]).filter(c => !!c.id)
    const newChildrenIds   = [...this.templateContent?.children].filter(c => !!c.id).map(c => c.id)
  
    return existingChildren.filter(c => newChildrenIds.includes(c.id))
  }
```

Have not considered performance ramifications, though.

dhh: The targetElements is from the PR that allows CSS selectors for targets.

tleish: I like the suggested changes.

PR #289, opened 2021-06-20
Include url in `turbo:before-fetch-request` event

ctrochalakis:
Hello all and thank you for Turbo.

Although the `turbo:before-fetch-request` event includes the request's `fetchOptions` (method, headers, etc), it doesn't include the URL itself. With this patch applied you can access it with `event.detail.url`.

If it gets accepted I can send a PR to add a reference to the `events` documentation in https://github.com/hotwired/turbo-site/.

### DIFF
--- a/src/http/fetch_request.ts
+++ b/src/http/fetch_request.ts
@@ -132,7 +132,7 @@ export class FetchRequest {
 
   private async allowRequestToBeIntercepted(fetchOptions: RequestInit) {
     const requestInterception = new Promise(resolve => this.resolveRequestPromise = resolve)
-    const event = dispatch("turbo:before-fetch-request", { cancelable: true, detail: { fetchOptions, resume: this.resolveRequestPromise } })
+    const event = dispatch("turbo:before-fetch-request", { cancelable: true, detail: { fetchOptions, url: this.url.href, resume: this.resolveRequestPromise } })
     if (event.defaultPrevented) await requestInterception
   }
 }
--- a/src/tests/functional/visit_tests.ts
+++ b/src/tests/functional/visit_tests.ts
@@ -5,7 +5,7 @@ declare const Turbo: any
 
 export class VisitTests extends TurboDriveTestCase {
   async setup() {
-    this.goToLocation("/src/tests/fixtures/visit.html")
+    await this.goToLocation("/src/tests/fixtures/visit.html")
   }
 
   async "test programmatically visiting a same-origin location"() {
--- a/src/tests/functional/visit_tests.ts
+++ b/src/tests/functional/visit_tests.ts
@@ -71,6 +71,14 @@ export class VisitTests extends TurboDriveTestCase {
     this.assert(await this.changedBody)
   }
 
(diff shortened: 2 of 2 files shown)


src/tests/functional/visit_tests.ts
```diff
@@ -71,6 +71,14 @@ export class VisitTests extends TurboDriveTestCase {
     this.resetRequestInterceptor()
   }
 
+  async "test turbo:before-fetch-request event.detail"() {
```
ctrochalakis, line 74: I am getting a timeout error when I run the whole suite, but everything seems to work if I run this specific test in isolation (by adding `intern.configure({ grep: /turbo:before/ })` in `runner.js`).

It's strange since it's basically identical to the request interceptor test.


```
× firefox on linux 5.10.0-0.bpo.5-amd64 - VisitTests - turbo:before-fetch-request event.detail (2.5s)
    TimeoutError: Timeout reached on firefox on linux 5.10.0-0.bpo.5-amd64 - VisitTests - turbo:before-fetch-request event.detail#
      at Timeout._onTimeout @ src/lib/Test.ts:226:24
      at ontimeout @ timers.js:436:11
      at tryOnTimeout @ timers.js:300:5
      at listOnTimeout @ timers.js:263:5
      at Timer.processTimers @ timers.js:223:10


```

dhh: We are seeing the same on #299. Only for Firefox. Odd. Please do investigate!

kirillplatonov:
@ctrochalakis could you please resolve merge conflicts? The `FetchRequest` class has been recently updated.

ctrochalakis:
I looked into the test failure yesterday, it was a race between the page reload and the events tracking setup. I've opened #310 with a fix.

I rebased the PR on top of the main branch (the #310 fix is also included here).

src/http/fetch_request.ts
```diff
@@ -132,7 +132,8 @@ export class FetchRequest {
 
   private async allowRequestToBeIntercepted(fetchOptions: RequestInit) {
     const requestInterception = new Promise(resolve => this.resolveRequestPromise = resolve)
-    const event = dispatch("turbo:before-fetch-request", { cancelable: true, detail: { fetchOptions, resume: this.resolveRequestPromise } })
+    const fetchUrl = this.url.href
+    const event = dispatch("turbo:before-fetch-request", { cancelable: true, detail: { fetchOptions, url: fetchUrl, resume: this.resolveRequestPromise } })
```
kirillplatonov, line 136: Maybe just pass the value inline? The `fetchUrl` variable seems redundant here.

```suggestion
    const event = dispatch("turbo:before-fetch-request", { cancelable: true, detail: { fetchOptions, url: this.url.href, resume: this.resolveRequestPromise } })
```

kirillplatonov:
@ctrochalakis nice catch with visits tests race condition! 👍

src/http/fetch_request.ts
ctrochalakis: Sure, fixed :)

PR #295, opened 2021-06-29
History needs to visit with a restoration identifier to restore scroll position

dhh:
We don't need to go through proposeVisit, because the visit already passed that check to get into the history.

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -109,9 +109,9 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
 
   // History delegate
 
-  historyPoppedToLocationWithRestorationIdentifier(location: URL) {
+  historyPoppedToLocationWithRestorationIdentifier(location: URL, restorationIdentifier: string) {
     if (this.enabled) {
-      this.navigator.proposeVisit(location, { action: "restore", historyChanged: true })
+      this.navigator.startVisit(location, restorationIdentifier, { action: "restore", historyChanged: true })
     } else {
       this.adapter.pageInvalidated()
     }
--- a/src/tests/functional/visit_tests.ts
+++ b/src/tests/functional/visit_tests.ts
@@ -76,6 +76,7 @@ export class VisitTests extends TurboDriveTestCase {
     await this.drainEventLog()
     await this.nextBeat
 
+    this.cancelNextVisit()
     await this.goBack()
     this.assert(await this.changedBody)
   }


src/core/session.ts
```diff
@@ -109,9 +109,9 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
 
   // History delegate
 
-  historyPoppedToLocationWithRestorationIdentifier(location: URL) {
+  historyPoppedToLocationWithRestorationIdentifier(location: URL, restorationIdentifier: string) {
     if (this.enabled) {
-      this.navigator.proposeVisit(location, { action: "restore", historyChanged: true })
```
domchristie, line 114: I wonder if calling `proposeVisit` might be needed for the native adapters?
I think this issue was [previously solved](https://github.com/hotwired/turbo/pull/57/commits/e92d9f8d6c96191dd961266e2c779380021eb224#diff-4fdcafb26950e01d47ac3215a38de2bbb3d4788ca980214293f334cfec78a7c0R110) by passing the `restorationIdentifier` along in the options to `proposeVisit`.

dhh: I checked that with the Android adapter, and it did not seem to be the case: https://github.com/hotwired/turbo-android/blob/556cfae61408811fe488bd877416b27e53cb35ca/turbo/src/main/assets/js/turbo_bridge.js. But will test before rolling further. We also have a test that's explicitly asserting that we're not going through proposeVisit (since that implies the right to cancel): https://github.com/hotwired/turbo/blob/main/src/tests/functional/visit_tests.ts#L74-L81

dhh: I'm not sure it'd work just to pass the restorationIdentifier as part of the option set either. Where this problem was triggered was in BrowserAdapter#visitPropoposedToLocation, which just always generates a new uuid for the restorationIdentifier: https://github.com/hotwired/turbo/blob/main/src/core/native/browser_adapter.ts#L18. And passing the restorationIdentifier all the way from history popped meant passing it through like 3-4 times.

dhh: Ahh, I see that this other PR added restorationIdentifier to the VisitOptions. Yeah, that would also be a way to go.

dhh: Verified that historyPoppedToLocationWithRestorationIdentifier is not used by the native adapters. So fine to proceed with this simple fix to the problem as it appears in browsers.

Issue #176, opened 2021-02-12
ActionCable connection not reestablished after authentication with Turbo


jonsgreen:
So I recognize that there are other issues similar or related to this one like https://github.com/hotwired/turbo/issues/45 and if this has been covered elsewhere I apologize. 

I am noticing that if I have a channel identified by current_user then if I submit a Turbo enabled authentication form (e.g. with Devise) then I don't have access to the current user through the ActionCable channel. A full refresh of the page establishes connection immediately. This behavior is also fixed by adding a turbo-false data element to the form.  A final option that sort of works is to do something like this (note that this is with an ApplicationCable::Connection identified by both session id and current_user): 
```ruby
class Users::SessionsController < Devise::SessionsController
  def create
     super do
        ActionCable.server.remote_connections.where(session_id: session.id, current_user: nil).disconnect
     end
end
```
The problem with this last solution is that it takes several seconds for ActionCable to detect the stale connection which can lead to confusing UX for the user. The initial connection on page load is quite snappy in comparison. 

My conclusion is that when Turbo handles the redirect the connection to ActionCable does not get refreshed. 

Maybe just disabling turbo in the form is good enough but I am curious if there is a way and whether it would make sense for Turbo to trigger ActionCable to reconnect in such cases.

jonsgreen:
In case someone else bumps into this here is one of probably many solutions to deal with this problem if you are using [CableReady](https://cableready.stimulusreflex.com)

first you need multiple identifiers for your connection to deal with unauthenticated users: 

```ruby
module ApplicationCable
  class Connection < ActionCable::Connection::Base
    identified_by :current_user
    identified_by :session_id

    def connect
      self.current_user = env["warden"].user(:user)
      self.session_id = request.session.id
      reject_unauthorized_connection unless self.current_user || self.session_id
    end
  end
end
```

then you will want a SessionChannel for when users are not yet authenticated: 

```ruby
class SessionChannel < ApplicationCable::Channel
  def subscribed
    stream_for session_id
  end
end
```

```javascript
import CableReady from 'cable_ready'
import consumer from './consumer'

consumer.subscriptions.create('SessionChannel', {
  received (data) {
    if (data.cableReady) CableReady.perform(data.operations)
  },

  connected () {
    document.addEventListener('reconnect', this.reconnect)
  },

  reconnect() {
    consumer.disconnect()
    consumer.connect()
  }

})
```

Then finally in your session_controller: 

```ruby
[...]

PR #297, opened 2021-07-01
Ignore noscript elements inside head

en30:
This pull request prevents Turbo from evaluating `<noscript>` elements inside the `<head>` element.

As a descendant of the `<head>` element, a `<noscript>` element can include `<style>` and `<link>` elements. This is useful for example, to show modals for script-enabled users and always to show modal contents for script-disabled users.
ref: [&lt;noscript&gt; - HTML: HyperText Markup Language | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/noscript)

Currently, Turbo evaluates stylesheets inside `<head>` elements even though those are inside `<noscript>` elements. This causes unexpected style for script-enabled users after a page transition.

When Turbo requests, JavaScript is definitely enabled. So `<noscript>` elements can be safely ignored.

### DIFF
--- a/src/core/drive/head_snapshot.ts
+++ b/src/core/drive/head_snapshot.ts
@@ -7,7 +7,9 @@ type ElementDetails = { type?: ElementType, tracked: boolean, elements: Element[
 type ElementType = "script" | "stylesheet"
 
 export class HeadSnapshot extends Snapshot<HTMLHeadElement> {
-  readonly detailsByOuterHTML = this.children.reduce((result, element) => {
+  readonly detailsByOuterHTML = this.children
+    .filter((element) => !elementIsNoscript(element))
+    .reduce((result, element) => {
     const { outerHTML } = element
     const details: ElementDetails
       = outerHTML in result
@@ -91,6 +93,11 @@ function elementIsTracked(element: Element) {
 function elementIsScript(element: Element) {
   const tagName = element.tagName.toLowerCase()
   return tagName == "script"
+}
+
+function elementIsNoscript(element: Element) {
+  const tagName = element.tagName.toLowerCase()
+  return tagName == "noscript"
 }
 
 function elementIsStylesheet(element: Element) {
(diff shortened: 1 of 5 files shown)


tleish:
FYI, this was discussed back with the turbolinks classic project (see: [Problem with CSS in <noscript> · Issue #176 · turbolinks/turbolinks-classic · GitHub ](https://github.com/turbolinks/turbolinks-classic/issues/176)).  Code was then added to address this.

However because of unintended side effects, it was reverted (see: [Turbolinks removes noscript tags implicitly · Issue #523 · turbolinks/turbolinks-classic · GitHub](https://github.com/turbolinks/turbolinks-classic/issues/523)).  

The recommended approach at the time was to have devs implement their own solutions for their specific needs, adding something like the following might help.

```
document.addEventListener('turbo:before-render', function (event) {
  event.detail.newBody.querySelectorAll('noscript').forEach((e) => e.remove());
});

en30:
I read [Turbolinks removes noscript tags implicitly · Issue #523 · turbolinks/turbolinks-classic · GitHub](https://github.com/turbolinks/turbolinks-classic/issues/523) just now.
The issue is caused by removing `<noscript>` inside the `<body>` element, which can contain meaningful content to show. But this pull request only removes `<noscript>` inside the `<head>` element. So I think this pull request does not lead to the unanticipated side effects.

I submitted this because I could not find an API to manipulate the `<head>` element before Turbo processes it.

en30:
With dynamic backend, there is a workaround, not rendering `<noscript>` elements based on whether a request is from Turbo or not. But Turbo is useful for static sites too! So I think it is better to be handled by Turbo.

en30:
The test suite passes on my laptop but failed on CI.
Am I missing something 🤔?

dhh:
@en30 Some of the tests are a little flaky.

en30:
@dhh I see. It happens. Thank you.

PR #298, opened 2021-07-01
Prevent reloading pages when tapping same-page anchors

domchristie:
This is a port of https://github.com/hotwired/turbo/pull/57 and https://github.com/turbolinks/turbolinks/pull/285 to resolve https://github.com/turbolinks/turbolinks/issues/75, https://github.com/turbolinks/turbolinks/issues/214, https://github.com/turbolinks/turbolinks/issues/192, and https://github.com/hotwired/turbo/issues/42

## The Problem

Clicking on a link which references an anchor on the current page issues an unnecessary request. This can cause problems when linking to content which is lazily loaded (see https://agile-reef-88023.herokuapp.com/posts/1), as well as with [analytics](https://github.com/turbolinks/turbolinks/issues/75#issuecomment-255867311) and [skip links](https://github.com/hotwired/turbo/issues/42). Disabling Turbo completely on those links prevents the request, but because the view is not cached, this approach breaks the Back button.

## The Solution

This PR detects whether the visit is to an anchor on the same page, and if so, prevents the request. It takes a snapshot and performs the scroll, bypassing any rendering and `turbo:load` triggering. A similar flow handles restorations to same-page anchors. Correct focusing is achieved by temporarily adding `tabindex=-1` and calling `focus` on the target.

## Notes

1. I'm getting a reliably failing test (`chrome on linux - FrameTests - evaluates frame script elements on each render (0.297s) AssertionError: expected 1 to equal 2`), which doesn't _seem_ to be related to these changes, and it's passing locally on my machine
2. the [Turbolinks version](https://github.com/turbolinks/turbolinks/pull/285) of this PR performed [the same-page check in the session](https://github.com/turbolinks/turbolinks/pull/285/files#diff-da4d3d3a97b9ed26e47b50d9a5d2f5795434467e001f8a6deba72c361a23baa7R330-R333) (née controller). This allowed the [native adapters to check for same-page anchors](https://github.com/turbolinks/turbolinks-ios/pull/126/files#diff-db83eae543943038c85ab2f7a4c437adda02ed30411e803d0091f5a7fc32d0acR57) and handle accordingly. The session is not so accessible from the Visit, so I'd welcome some tips on how to solve this (or whether it's even necessary now?).

## Todo
- [x] Manual tests to double-check
- [x] Investigate Firefox bug when navigating back after first anchor click
- [x] Trigger hash change event when scrolling to same-page anchor

Fixes https://github.com/hotwired/turbo/issues/42
Closes https://github.com/hotwired/turbo/pull/152

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -9,6 +9,7 @@ import { PageSnapshot } from "./page_snapshot"
 export type NavigatorDelegate = VisitDelegate & {
   allowsVisitingLocation(location: URL): boolean
   visitProposedToLocation(location: URL, options: Partial<VisitOptions>): void
+  notifyApplicationAfterVisitingSamePageLocation(oldURL: URL, newURL: URL): void
 }
 
 export class Navigator {
@@ -122,6 +123,10 @@ export class Navigator {
   locationWithActionIsSamePage(location: URL, action: Action): boolean {
     return getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation) &&
       (getAnchor(location) != null || action == "restore")
+  }
+
+  visitScrolledToSamePageLocation(oldURL: URL, newURL: URL) {
+    this.delegate.visitScrolledToSamePageLocation(oldURL, newURL)
   }
 
   // Visits
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -2,7 +2,7 @@ import { Action, isAction } from "../types"
 import { FetchMethod } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { FormSubmission } from "./form_submission"
-import { expandURL, Locatable } from "../url"
+import { expandURL, getAnchor, getRequestURL, Locatable } from "../url"
 import { Visit, VisitDelegate, VisitOptions } from "./visit"
 import { PageSnapshot } from "./page_snapshot"
 
@@ -117,6 +117,11 @@ export class Navigator {
 
   visitCompleted(visit: Visit) {
     this.delegate.visitCompleted(visit)
+  }
+
+  locationWithActionIsSamePage(location: URL, action: Action): boolean {
+    return getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation) &&
+      (getAnchor(location) != null || action == "restore")
   }
 
   // Visits
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -310,8 +310,9 @@ export class Visit implements FetchRequestDelegate {
   }
 
   scrollToAnchor() {
-    if (getAnchor(this.location)) {
-      this.view.scrollToAnchor(getAnchor(this.location))
+    const anchor = getAnchor(this.location)
+    if (anchor != null) {
+      this.view.scrollToAnchor(anchor)
       return true
     }
   }
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -2,7 +2,7 @@ import { Adapter } from "../native/adapter"
 import { FetchMethod, FetchRequest, FetchRequestDelegate } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { History } from "./history"
-import { getAnchor, getRequestURL } from "../url"
+import { getAnchor } from "../url"
 import { PageSnapshot } from "./page_snapshot"
 import { Action } from "../types"
 import { uuid } from "../../util"
@@ -15,6 +15,7 @@ export interface VisitDelegate {
 
   visitStarted(visit: Visit): void
   visitCompleted(visit: Visit): void
+  locationWithActionIsSamePage(location: URL, action: Action): boolean
 }
 
 export enum TimingMetric {
@@ -90,11 +91,7 @@ export class Visit implements FetchRequestDelegate {
     this.referrer = referrer
     this.snapshotHTML = snapshotHTML
     this.response = response
-
-    this.isSamePage = (
-      getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation) &&
-      (getAnchor(location) != null || this.action == "restore")
-    )
+    this.isSamePage = this.delegate.locationWithActionIsSamePage(this.location, this.action)
   }
 
   get adapter() {
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -84,7 +84,7 @@ export class Visit implements FetchRequestDelegate {
     this.location = location
     this.isSamePage = (
       getAnchor(location) != null &&
-      getRequestURL(location) === getRequestURL(new URL(window.location.href))
+      getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation)
     )
 
     this.restorationIdentifier = restorationIdentifier || uuid()
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -16,6 +16,7 @@ export interface VisitDelegate {
   visitStarted(visit: Visit): void
   visitCompleted(visit: Visit): void
   locationWithActionIsSamePage(location: URL, action: Action): boolean
+  visitScrolledToSamePageLocation(oldURL: URL, newURL: URL): void
 }
 
 export enum TimingMetric {
@@ -313,6 +314,10 @@ export class Visit implements FetchRequestDelegate {
       } else {
         this.scrollToAnchor() || this.scrollToTop()
       }
+      if (this.isSamePage) {
+        this.delegate.visitScrolledToSamePageLocation(this.view.lastRenderedLocation, this.location)
+      }
+
       this.scrolled = true
     }
   }
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -2,7 +2,7 @@ import { Adapter } from "../native/adapter"
 import { FetchMethod, FetchRequest, FetchRequestDelegate } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { History } from "./history"
-import { getAnchor } from "../url"
+import { getAnchor, getRequestURL } from "../url"
 import { PageSnapshot } from "./page_snapshot"
 import { Action } from "../types"
 import { uuid } from "../../util"
@@ -70,6 +70,7 @@ export class Visit implements FetchRequestDelegate {
   frame?: number
   historyChanged = false
   location: URL
+  isSamePage: boolean
   redirectedToLocation?: URL
   request?: FetchRequest
   response?: VisitResponse
@@ -81,6 +82,11 @@ export class Visit implements FetchRequestDelegate {
   constructor(delegate: VisitDelegate, location: URL, restorationIdentifier: string | undefined, options: Partial<VisitOptions> = {}) {
     this.delegate = delegate
     this.location = location
+    this.isSamePage = (
+      getAnchor(location) != null &&
+      getRequestURL(location) === getRequestURL(new URL(window.location.href))
+    )
+
     this.restorationIdentifier = restorationIdentifier || uuid()
 
     const { action, historyChanged, referrer, snapshotHTML, response } = { ...defaultOptions, ...options }
@@ -234,10 +240,14 @@ export class Visit implements FetchRequestDelegate {
       const isPreview = this.shouldIssueRequest()
       this.render(async () => {
         this.cacheSnapshot()
-        await this.view.renderPage(snapshot, isPreview)
-        this.adapter.visitRendered(this)
-        if (!isPreview) {
-          this.complete()
+        if (this.isSamePage) {
+          this.adapter.visitRendered(this)
+        } else {
+          await this.view.renderPage(snapshot, isPreview)
+          this.adapter.visitRendered(this)
+          if (!isPreview) {
+            this.complete()
+          }
         }
       })
     }
@@ -248,6 +258,15 @@ export class Visit implements FetchRequestDelegate {
       this.location = this.redirectedToLocation
       this.history.replace(this.redirectedToLocation, this.restorationIdentifier)
       this.followedRedirect = true
+    }
+  }
+
+  goToSamePageAnchor() {
+    if (this.isSamePage) {
+      this.render(async () => {
+        this.cacheSnapshot()
+        this.adapter.visitRendered(this)
+      })
     }
   }
 
@@ -346,9 +365,13 @@ export class Visit implements FetchRequestDelegate {
   }
 
   shouldIssueRequest() {
-    return this.action == "restore"
-      ? !this.hasCachedSnapshot()
-      : true
+    if (this.action == "restore") {
+      return !this.hasCachedSnapshot()
+    } else if (this.isSamePage) {
+      return false
+    } else {
+      return true
+    }
   }
 
   cacheSnapshot() {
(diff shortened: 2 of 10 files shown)


dhh:
That test is flaky. I reran the suite and it passed.

Reviewing whether we need something for the native adapters.

src/core/drive/visit.ts
```diff
@@ -251,6 +261,15 @@ export class Visit implements FetchRequestDelegate {
     }
   }
 
+  goToSamePageAnchor() {
```
domchristie, line 264: I wonder if there's a better name for this? It originates from [the original code](https://github.com/turbolinks/turbolinks/pull/285/files#diff-d34bf5280590ab473d1ad488821d664a0746c3459a2d1f3c2d7ea6037be2b042R169-R177) that actually did the scrolling, but now this is handled by `render`.

dhh:
I think we're probably OK re: native apps. The iOS + Android adapters just call the visit code directly: https://github.com/hotwired/turbo-ios/blob/main/Source/WebView/turbo.js and https://github.com/hotwired/turbo-android/blob/main/turbo/src/main/assets/js/turbo_bridge.js.

src/core/drive/visit.ts
dhh: I think that's still an accurate naming as even if the actual jump is done by render, it's triggered here.

domchristie:
Do we not need some kind of equivalent to: https://github.com/turbolinks/turbolinks-ios/pull/126/files#diff-db83eae543943038c85ab2f7a4c437adda02ed30411e803d0091f5a7fc32d0acR57 ? Feel free to ignore if I'm way out-of-date on this!

dhh:
Let's double check with @jayohms who runs Mobile at Basecamp. He's out for a bit, but we'll make sure we're reading this right.

domchristie:
👍 to clarify my thoughts: I don't think that clicking on a same-page anchor in a native app should trigger a visit, Turbo should just perform the scrolling.

Navigating back from a same-page visit might be interesting. For iOS, this is probably less of an issue: normal visits get pushed onto a stack, and the back button controls the navigation between these screens. I don't think there'd be any expectation that tapping **< Back** would handle the screen's scroll position. For Android, with its hardware back button, the expectations might be different(?)

terracatta:
I just wanted to say that I tried this branch on my Rails app that was originally using the following code with Turbolinks to fix hash nav behavior.

```js
// Workround for odd hash navigation behavior in turbolinks that messes
// with the pushstate history
document.addEventListener('turbolinks:click', function (event) {
  var anchorElement = event.target
  var isSamePageAnchor = (
    anchorElement.hash &&
    anchorElement.origin === window.location.origin &&
    anchorElement.pathname === window.location.pathname
  )

  if (isSamePageAnchor) {
    Turbolinks.controller.pushHistoryWithLocationAndRestorationIdentifier(
      event.data.url,
      Turbolinks.uuid()
    )
    window.dispatchEvent(new HashChangeEvent("hashchange"))
    event.preventDefault()
  }
})
```

I first ported the Rails app over to Turbo and then applied this PR. It worked great in the sense that no pages/requests were fired for hash links, but I had some stimulus code (`data-action="hashchange@window->tabbed-nav#navigate"`) that also relied on the dispatching of the `hashchange` event to `window`, which this PR does not attempt to synthetically fire.

I forked this PR and just added a quick line of code to fire the event here https://github.com/kolide/turbo/commit/fbed8a95ab2abc125b17150f41e952145e520047#diff-78d8451f964182fd51330bac500ba6e71234f81aad9e8a57e682e723d0f517b3R269. 

I'm likely going to refactor my tabbed nav in my Rails app to not have to rely on this event being fired, but I felt it was worth dropping a comment here in case you all felt firing the `hashchange` event was a good idea.

jayohms:
> 👍 to clarify my thoughts: I don't think that clicking on a same-page anchor in a native app should trigger a visit, Turbo should just perform the scrolling.

@domchristie Agreed! This would be a nice change in the mobile apps.

> Navigating back from a same-page visit might be interesting. For iOS, this is probably less of an issue: normal visits get pushed onto a stack, and the back button controls the navigation between these screens. I don't think there'd be any expectation that tapping < Back would handle the screen's scroll position. For Android, with its hardware back button, the expectations might be different(?)

I don't see this as problematic for iOS or Android. On Android, in most situations now, the system back button should work just like you'd expect the top toolbar back arrow to behave. Both mobile apps can behave the same way here: the same-page scroll is transparent to the apps and will not push onto the navigation stack.

> the Turbolinks version of this PR performed the same-page check in the session (née controller). This allowed the native adapters to check for same-page anchors and handle accordingly. The session is not so accessible from the Visit, so I'd welcome some tips on how to solve this (or whether it's even necessary now?).

I tested this out in the mobile demo app here: https://github.com/hotwired/turbo-native-demo/compare/same-page-anchor

Some additional work will need to be done to prevent the mobile apps from following anchor visit navigation. I haven't spent enough time yet to see what a proper solution will look like, but it doesn't look like it'll be as straightforward as the proposed Turbolinks mobile adapter solutions, since it's wrapped up in the `Visit` class, like you mentioned.

If you have any ideas in mind, let me know! And I'll take I'll take closer look at it this week, too.

jayohms:
The mobile adapters do have access to the `currentVisit` here: https://github.com/hotwired/turbo-android/blob/main/turbo/src/main/assets/js/turbo_bridge.js#L105

Here's a quick spike at the minimal set of changes (with clean up necessary) to support the mobile behavior I think we'd want:
```diff
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -328,8 +328,8 @@ export class Visit implements FetchRequestDelegate {
     }
   }
 
-  scrollToAnchor() {
-    const anchor = getAnchor(this.location)
+  scrollToAnchor(location: URL = this.location) {
+    const anchor = getAnchor(location)
     if (anchor != null) {
       this.view.scrollToAnchor(anchor)
       return true
@@ -340,6 +340,11 @@ export class Visit implements FetchRequestDelegate {
     this.view.scrollToPosition({ x: 0, y: 0 })
   }
 
+  locationIsSamePageAnchor(location: URL) {
+    return getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation) &&
+      getAnchor(location) != null
+  }
+
```

We can then combine that with this change in the mobile adapter: https://github.com/hotwired/turbo-android/pull/175

One thing to note here is that the mobile apps will not see the new `#anchor` url when the scroll is peformed. The original location for the screen will be preserved. I don't think it should cause a negative side effect, though. And a positive is that a native screen location is immutable for each destination on the back stack, so we couldn't update the url without a new navigation visit, anyway.

@domchristie Thoughts?

domchristie:
Thanks @jayohms. It's looking like it's doable then :D

A possible slight complication here is that [determining a same-page visit also requires the `action`](https://github.com/hotwired/turbo/pull/298/files#diff-78d8451f964182fd51330bac500ba6e71234f81aad9e8a57e682e723d0f517b3R94-R97)

Having access to the `currentVisit` is handy, although I'm not 100% sure about asking the it if a proposed location is for same page. It _kind of_ works if you consider a same-page click to be a child of the main visit; and it does have the benefit of access to the scrolling functions.

Another approach might be to go via `Turbo.navigator`. In the native bridge this might look like:

```js
visitProposedToLocation(location, options) {
  if (window.Turbo && Turbo.navigator.locationWithActionIsSamePage(location, options.action)) {
    Turbo.navigator.view.scrollToAnchorFromLocation(location)
  } else {
    TurboSession.visitProposedToLocation(location.toString(), JSON.stringify(options))
  }
}
```

A bigger changeset (and the above breaks the law of demeter :/), but might work in terms of code organisation. 

```diff
diff --git a/src/core/drive/navigator.ts b/src/core/drive/navigator.ts
index 998fb51..7b5d71d 100644
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -2,7 +2,7 @@ import { Action, isAction } from "../types"
 import { FetchMethod } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { FormSubmission } from "./form_submission"
-import { expandURL, Locatable } from "../url"
+import { expandURL, getAnchor, getRequestURL, Locatable } from "../url"
 import { Visit, VisitDelegate, VisitOptions } from "./visit"
 import { PageSnapshot } from "./page_snapshot"
 
@@ -119,6 +119,11 @@ export class Navigator {
     this.delegate.visitCompleted(visit)
   }
 
+  locationWithActionIsSamePage(location: URL, action: Action): boolean {
+    return getRequestURL(location) === getRequestURL(this.location) &&
+      (getAnchor(location) != null || action == "restore")
+  }
+
   // Visits
 
   get location() {
diff --git a/src/core/drive/visit.ts b/src/core/drive/visit.ts
index 5d8808e..2feee5b 100644
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -2,7 +2,7 @@ import { Adapter } from "../native/adapter"
 import { FetchMethod, FetchRequest, FetchRequestDelegate } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { History } from "./history"
-import { getAnchor, getRequestURL } from "../url"
+import { getAnchor } from "../url"
 import { PageSnapshot } from "./page_snapshot"
 import { Action } from "../types"
 import { uuid } from "../../util"
@@ -15,6 +15,7 @@ export interface VisitDelegate {
 
   visitStarted(visit: Visit): void
   visitCompleted(visit: Visit): void
+  locationWithActionIsSamePage(location: URL, action: Action): boolean
 }
 
 export enum TimingMetric {
@@ -90,11 +91,7 @@ export class Visit implements FetchRequestDelegate {
     this.referrer = referrer
     this.snapshotHTML = snapshotHTML
     this.response = response
-
-    this.isSamePage = (
-      getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation) &&
-      (getAnchor(location) != null || this.action == "restore")
-    )
+    this.isSamePage = this.delegate.locationWithActionIsSamePage(this.location, this.action)
   }
 
   get adapter() {
diff --git a/src/core/session.ts b/src/core/session.ts
index 8374962..7c5fb38 100644
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -156,6 +156,10 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
     this.notifyApplicationAfterPageLoad(visit.getTimingMetrics())
   }
 
+  locationWithActionIsSamePage(location: URL, action: Action): boolean {
+    return this.navigator.locationWithActionIsSamePage(location, action)
+  }
+
   // Form submit observer delegate
 
   willSubmitForm(form: HTMLFormElement, submitter?: HTMLElement): boolean {
diff --git a/src/core/view.ts b/src/core/view.ts
index e8ee0ae..eb01e55 100644
--- a/src/core/view.ts
+++ b/src/core/view.ts
@@ -1,6 +1,7 @@
 import { Renderer } from "./renderer"
 import { Snapshot } from "./snapshot"
 import { Position } from "./types"
+import { getAnchor } from "./url"
 
 export interface ViewDelegate<S extends Snapshot> {
   viewWillRenderSnapshot(snapshot: S, isPreview: boolean): void
@@ -21,7 +22,7 @@ export abstract class View<E extends Element, S extends Snapshot<E> = Snapshot<E
 
   // Scrolling
 
-  scrollToAnchor(anchor: string) {
+  scrollToAnchor(anchor: string | undefined) {
     const element = this.snapshot.getElementForAnchor(anchor)
     if (element) {
       this.scrollToElement(element)
@@ -31,6 +32,10 @@ export abstract class View<E extends Element, S extends Snapshot<E> = Snapshot<E
     }
   }
 
+  scrollToAnchorFromLocation(location: URL) {
+    this.scrollToAnchor(getAnchor(location))
+  }
+
   scrollToElement(element: Element) {
     element.scrollIntoView()
   }
```

jayohms:
@domchristie I prefer the approach in your solution 👍. I realized that the `currentVisit` won't be available in the native adapter on the initial cold boot, so that would prevent same-page scrolling from working properly in the mobile apps in that situation.

Want to commit your solution and I can try it out and we see if there are any opportunities to improve the PR?

domchristie:
> Want to commit your solution and I can try it out and we see if there are any opportunities to improve the PR?

@jayohms 👍 Done

domchristie:
@terracatta I've added the hash change event as a todo

jayohms:
@domchristie The new implementation works great with https://github.com/hotwired/turbo-android/pull/175 (and we can make the same change to the iOS adapter). Note that we won't be able to release new versions of the native adapters until this is merged and a new Turbo version is available, since the new navigator API is required.

src/core/drive/navigator.ts
```diff
@@ -119,6 +119,11 @@ export class Navigator {
     this.delegate.visitCompleted(visit)
   }
 
+  locationWithActionIsSamePage(location: URL, action: Action): boolean {
+    return getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation) &&
+      (getAnchor(location) != null || action == "restore")
```
jayohms, line 124: Can you elaborate on what situation the `|| action == "restore"` condition is handling? I don't think this should ever affect the native apps, since restores are initiated by the apps themselves, but want to make sure.

domchristie: @jayohms sure, this refers to the changes in: https://github.com/hotwired/turbo/pull/298/commits/0800ec84f9e79a02af6c4d5592274dec7179bf31 It prevents unnecessary requests when navigating back _from_ a same-page anchor to a location with no anchor, when caching is disabled or has been cleared. e.g.

1. Visit `example.com` which has Turbo caching disabled
2. Click link to `example.com#main`
3. Tap Back

Without this change, Turbo would re-request `example.com` because the anchor is `undefined`. If the request URLs are equal, and the action is `restore`, we can be sure that the visit is a same-page one.

jayohms: Perfect, makes sense 👍

jayohms:
@domchristie I'm good with this from the native app perspective. Thanks for doing this!

I'll make sure we ship native adapter updates once this is available.

dhh:
So we're good to merge now, @domchristie?

domchristie:
> So we're good to merge now, @domchristie?

@dhh just a few things to check off, including triggering the hashchange event, then I think we'll be good.

domchristie:
@dhh @terracatta 7387badf2ddead412a974d09640c94ee42cdefe2 triggers a hashchange event on same-page visits.

A couple of point on this:
- it does make use of the `Visit` process so won't be dispatched on native clients, which bypasses it (cc @jayohms)
- it's fired twice when navigating Back/Forward: one that we trigger (with the correct old/new URLs) and one that the browser to automatically triggers (with the old/new URLs both set to the new location 🤷‍♂️)

domchristie:
@dhh OK, I've completed some manual tests using a build from this branch, and this demo: https://stripe-root-spatula.glitch.me/

There's still an issue in Firefox when navigating back to the initial location, e.g.:

1. visit https://stripe-root-spatula.glitch.me/ on Firefox
2. tap "Skip to content"
3. tap Back (the scroll position should be that the top, but it does not change)

I have checked that `scrollToTop` called, but it seems to ignore it 🤷‍♂️

dhh:
Excellent work, Dom 👏

acetinick:
hi @domchristie ,  I just tried upgrading to latest release and looks like its broken some existing behaviour I have across my app.   

We were using Turbo.visit() to force redirect of same page to anchor after modal window will close to refresh just added record, and now with this PR doesn't trigger the visit.

Is there a way to override this behaviour or maybe introduce Turbo.forceVisit() ?

marispro:
same to me. Would be good to know how to disable this.

domchristie:
@acetinick @marispro this should be fixed by https://github.com/hotwired/turbo/pull/324: any visit with `action: replace` will load from the server. Hope that solves your cases.

domchristie:
>There's still an issue in Firefox when navigating back to the initial location … this is a bug with Firefox when using `history.scrollRestoration = 'manual'` ([minimal reproduction](https://glitch.com/edit/#!/tide-closed-salary?path=script.js%3A14%3A2)). It seems to be fixed in Firefox Nightly 🎉

Confirmed to be fixed in Firefox v92 🎉

PR #299, opened 2021-07-02
Make links with method work everywhere, not only inside turbo frames

Intrepidd:
This is the continuation of #277 that introduced `data-turbo-method`. Unfortunately it only worked within turbo-frames.
This PR aims to make it work globally.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -122,11 +122,11 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   // Link interceptor delegate
 
   shouldInterceptLinkClick(element: Element, url: string) {
-    const linkMethod = element.getAttribute("data-turbo-method") || element.getAttribute("data-method")
-    if (linkMethod)
+    if (element.hasAttribute("data-turbo-method")) {
       return false
-
-    return this.shouldInterceptNavigation(element)
+    } else {
+      return this.shouldInterceptNavigation(element)
+    }
   }
 
   linkClickIntercepted(element: Element, url: string) {
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -122,6 +122,10 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   // Link interceptor delegate
 
   shouldInterceptLinkClick(element: Element, url: string) {
+    const linkMethod = element.getAttribute("data-turbo-method") || element.getAttribute("data-method")
+    if (linkMethod)
+      return false
+
     return this.shouldInterceptNavigation(element)
   }
 
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -133,8 +133,24 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
 
   followedLinkToLocation(link: Element, location: URL) {
     const action = this.getActionForLink(link)
-    this.visit(location.href, { action })
+    this.convertLinkWithMethodClickToFormSubmission(link) || this.visit(location.href, { action })
   }
+
+  convertLinkWithMethodClickToFormSubmission(link: Element) {
+    const linkMethod = link.getAttribute("data-turbo-method") || link.getAttribute("data-method")
+
+    if (linkMethod) {
+      const form = document.createElement("form")
+      form.method = linkMethod
+      form.action = link.getAttribute("href") || "undefined"
+
+      link.parentNode?.insertBefore(form, link)
+      return dispatch("submit", { target: form })
+    } else {
+      return false
+    }
+  }
+
 
   // Navigator delegate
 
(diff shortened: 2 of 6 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -122,6 +122,10 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   // Link interceptor delegate
 
   shouldInterceptLinkClick(element: Element, url: string) {
+    const linkMethod = element.getAttribute("data-turbo-method") || element.getAttribute("data-method")
```
Intrepidd, line 125: This should probably be extracted as it is also used [here](https://github.com/hotwired/turbo/pull/299/files#diff-4fdcafb26950e01d47ac3215a38de2bbb3d4788ca980214293f334cfec78a7c0R140)

Where would be the best place to do so ?

src/tests/functional/form_submission_tests.ts
```diff
@@ -331,13 +331,22 @@ export class FormSubmissionTests extends TurboDriveTestCase {
     this.assert.ok(fetchOptions.headers["Turbo-Frame"], "submits with the Turbo-Frame header")
   }
 
-  async "test link method form submission"() {
-    await this.clickSelector("#link-method-redirect")
```
Intrepidd, line 335: For the record, this test passed by mistake. As the link was not within a frame, it was not turned into a form, and the redirect endpoint supports get requests.

src/tests/functional/form_submission_tests.ts
```diff
@@ -331,13 +331,22 @@ export class FormSubmissionTests extends TurboDriveTestCase {
...
+  }
+
+  async "test link method form submission outside frame"() {
+    await this.clickSelector("#link-method-outside-frame")
+
+    await this.nextBeat
+
+    const message = await this.querySelector("#frame div.message")
+    this.assert.equal(await message.getVisibleText(), "Link!")
   }
```
Intrepidd, line 350: Those tests timeout on my machine, not really sure why, as it works perfectly fine from the browser.

src/core/frames/frame_controller.ts
dhh: I've actually wanted to drop compatibility with data-method, as it actually makes coexisting with UJS more confusing. So we should just proceed with data-turbo-method as the only attribute, and then we don't need an extraction. Then we can even inline it:

```javascript
if (element.getAttribute("data-turbo-method")) {
  return false
} else {
  return this.shouldInterceptNavigation(element)
}
```

Then we also prevent multiple exit wounds.

src/core/session.ts
```diff
@@ -133,9 +133,25 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
 
   followedLinkToLocation(link: Element, location: URL) {
     const action = this.getActionForLink(link)
-    this.visit(location.href, { action })
+    this.convertLinkWithMethodClickToFormSubmission(link) || this.visit(location.href, { action })
```
dhh, line 136: Yeah, this is a nicer place to do this 👍

Intrepidd:
Thanks ! I also removed the mention of `data-method` in session.ts

Any insight on the timeout-ing tests. ?

dhh:
Test failures are appearing only in Firefox, not Chrome. That might be a clue to try fixing this? Thanks for working on it!

Screen Shot 2021-07-06 at 16 35 02

Intrepidd:
Nice catch, so on firefox I have the following when I click on the link : 
CleanShot 2021-07-07 at 11 19 09@2x

My Accept header is bad : 
image

Looks like the form is submitted instantly rather than being intercepted again by turbo ? Does it ring a bell ? I'll keep digging

Intrepidd:
I fixed it, the event was not cancelable so it triggered anyway, adding `cancelable: true` did the trick.

dhh:
Great work on this 👏

src/core/session.ts
```diff
@@ -133,9 +133,25 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
...
   }
 
+  convertLinkWithMethodClickToFormSubmission(link: Element) {
+    const linkMethod = link.getAttribute("data-turbo-method")
+
+    if (linkMethod) {
+      const form = document.createElement("form")
+      form.method = linkMethod
```
tleish, line 144: So, this only supports data-turbo-method of get and post, not delete?

Intrepidd: It does work with any method that works in a form AFAIK, are you encountering any issue?

tleish: My mistake.  When initially testing, it didn't work correctly.    Since It creates the form differently in Rails UJS (post with hidden delete input), I assumed it didn't work.  Digging into it more, it works as expected.

tleish: Actually, this is a potential issue with the Rails stack.  Prior UJS library converted links to POST and added hidden _method.  This made handling redirect_to easy.  Now, if using DELETE or POST, Turbo XHR redirects do not work unless adding a status: 3xx on the response.  (see #320).

Intrepidd: Interesting, I think rails `form_for` and other form helpers actually use a `POST` form and a `_method` attribute. 

But I think `_method` is a rails construct right ? So it would not work with another backend, which is not what we want :(

PR #306, opened 2021-07-12
Pausable requests

kirillplatonov:
A new revision of https://github.com/hotwired/turbo/pull/177 using pausable API instead of `RequestInterceptor` singleton.

Usage example:
```javascript
addEventListener('turbo:before-fetch-request', async function(event) {
  // pause request
  event.preventDefault()
  
  // fetch and set token
  const token = await getSessionToken(window.app)
  event.detail.fetchOptions.headers['Authorization'] = `Bearer ${token}`

  // continue request
  event.detail.resume()
})
```


dhh:
Wonderful. Now just need some documentation for both request and render pausing!

awd:
Thanks @kirillplatonov this will work great!

gnat:
With high regards, popping in from the [htmx](https://htmx.org) project- wanted to say, brilliant feature! It's an area htmx needs to polish up its UX.

I'm glad we have a lot of very smart people working on the HTML over the wire ecosystem!

PR #314, opened 2021-07-16
Fix pausable rendering for previews

### LINKED ISSUE #313: Pauseable rendering breaks after rendering preview
If the fetch response completes before the paused preview render completes, the visit will attempt to render twice whilst a render is in progress, raising:
https://github.com/hotwired/turbo/blob/49f82135e2a5ec8e3279f78373d69214d06a34fa/src/core/view.ts#L50

A quick fix is to disable previews, but ideally, in the case of animating page transitions, the preview would animate in first, and then be seemlessly replaced by the server response.

kirillplatonov:
Resolves #313

### DIFF
--- a/src/core/drive/page_view.ts
+++ b/src/core/drive/page_view.ts
@@ -22,7 +22,7 @@ export class PageView extends View<Element, PageSnapshot, PageViewRenderer, Page
 
   renderError(snapshot: PageSnapshot) {
     const renderer = new ErrorRenderer(this.snapshot, snapshot, false)
-    this.render(renderer)
+    return this.render(renderer)
   }
 
   clearSnapshotCache() {
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -199,6 +199,7 @@ export class Visit implements FetchRequestDelegate {
       const { statusCode, responseHTML } = this.response
       this.render(async () => {
         this.cacheSnapshot()
+        if (this.view.renderPromise) await this.view.renderPromise
         if (isSuccessful(statusCode) && responseHTML != null) {
           await this.view.renderPage(PageSnapshot.fromHTMLString(responseHTML))
           this.adapter.visitRendered(this)
@@ -241,6 +242,7 @@ export class Visit implements FetchRequestDelegate {
         if (this.isSamePage) {
           this.adapter.visitRendered(this)
         } else {
+          if (this.view.renderPromise) await this.view.renderPromise
           await this.view.renderPage(snapshot, isPreview)
           this.adapter.visitRendered(this)
           if (!isPreview) {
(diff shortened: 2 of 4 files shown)


kirillplatonov:
@domchristie it should solve the issue for preview. If fetch will finish before preview rendering is complete - it will wait for it and then just replace the content without a second pause.

dhh:
Think this needs a bump after I merged Dom's fix for same-page anchor links.

kirillplatonov:
@dhh this test keeps failing in Chrome even on main. Investigating.
> chrome on linux - FrameTests - evaluates frame script elements on each render (0.2s)

domchristie:
Thanks @kirillplatonov! That fixes the error, but unfortunately it still results in some unexpected behaviour. When rendering the response after a preview, the `before-render` event is not pausable. For example:

```js
addEventListener('turbo:before-render', (event) => {
  event.preventDefault()
  console.time('RENDER')
  setTimeout(() => event.detail.resume(), 2000)
})

addEventListener('turbo:render', (event) => {
  console.timeEnd('RENDER')
})
```

The timer should always be ~2000ms. It is for the preview, and for visits with no preview, but it's immediate for the render after a preview. Hope that makes sense!? Here's a demo: https://glitch.com/edit/#!/scratched-held-mistake?path=public%2Fscript.js%3A1%3A39

kirillplatonov:
@domchristie It was done intentionally to replace the preview without interception:
https://github.com/hotwired/turbo/blob/main/src/core/view.ts#L79

We can get rid of it. But in that case, both preview and real response will be animated. How do you think it should work?

domchristie:
To provide a bit more context: if the response completes whilst the preview is animating in, the new page will appear to animate twice. The animations will start on the preview, then once the response HTML comes in, the animations restart on the new HTML—it looks a bit glitchy. 

I naively thought that replacing the body would just carry on the animation as it left off; but now I think about it, whenever an element is rendered, any applicable animations will appear to restart.

Now I'm thinking it might be preferable to wait for the preview animations to complete before resuming with the response render. It's possible to determine whether a render is a preview or not, so we can switch off animations for the response. For a visit with a preview, the flow might look something like:

1. animate out
2. wait for animate out then animate in the preview
3. wait for the preview to animate in then immediately render the response

I _think_ that might work?

kirillplatonov:
It should already work this way. The only concern I have is that we probably should skip animation inside the event handler, not in the core. So the example might looks like that:
```javascript
document.addEventListener('turbo:before-render', async (event) => {
  event.preventDefault()
  // Skip animation for response if preview is rendered
  if (document.documentElement.hasAttribute("data-turbo-preview")) {
    await animateOut()
  }
  event.detail.resume()
})
```
And then we won't need to check if preview is rendered here: https://github.com/hotwired/turbo/blob/main/src/core/view.ts#L79

What do you think?

domchristie:
> It should already work this way. The only concern I have is that we probably should skip animation inside the event handler, not in the core.

Unfortunately, I think the `!previewIsRendered` is preventing it from working this way. Your example is exactly what I had in mind. (And always allowing developers to pause a render is a more consistent API 👍)

kirillplatonov:
Sounds good. Prepared a fix: https://github.com/hotwired/turbo/pull/316

PR #316, opened 2021-07-17
Fix rendering interception after preview

kirillplatonov:
We can skip interception after preview inside the event handler. So no need to do it in core.
See the discussion here: https://github.com/hotwired/turbo/pull/314#issuecomment-881958039


kirillplatonov:
cc @dhh

domchristie:
This works really nicely. Demo: https://experienced-internal-napkin.glitch.me 👍

PR #319, opened 2021-07-18
Emit `turbo:submit-failed` when form is rejected

terracatta:
## Description
This PR adds a new event to the Turbo Drive form submission lifecycle called `turbo:submit-failed`.This event is dispatched when a form is rejected by the server or the server throws an error. 

## Why a new event?
In a situation where a form submission is successful, Turbo Drive will follow the redirect and kick off a `turbo:visit` event. A lot of JS libs explicitly look out for this event to correct their state with the DOM is redrawn. One example is [Intercom](https://github.com/intercom/intercom-rails/issues/336). In this situation, everything works as expected.

But when a form is rejected by the server, new HTML is rendered in the DOM, but no `turbo:visit` event is ever fired. This means external scripts that may need to redraw HTML elements in a situation where they may already have state attached to `window` cannot do so.

I think the event could also be useful in other JS contexts (ex: like sending a report to an error monitoring service, or scrolling the window to a specific position), but I wanted to be honest about my need for it.

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -136,6 +136,8 @@ export class FormSubmission {
 
   requestSucceededWithResponse(request: FetchRequest, response: FetchResponse) {
     if (response.clientError || response.serverError) {
+      this.result = { success: false, fetchResponse: response }
+      dispatch("turbo:submit-failed", { target: this.formElement, detail: { formSubmission: this, ...this.result } })
       this.delegate.formSubmissionFailedWithResponse(this, response)
     } else if (this.requestMustRedirect(request) && responseSucceededWithoutRedirect(response)) {
       const error = new Error("Form responses must redirect to another location")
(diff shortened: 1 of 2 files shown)


src/core/drive/form_submission.ts
```diff
@@ -136,6 +136,8 @@ export class FormSubmission {
 
   requestSucceededWithResponse(request: FetchRequest, response: FetchResponse) {
     if (response.clientError || response.serverError) {
+      this.result = { success: false, fetchResponse: response }
+      dispatch("turbo:submit-failed", { target: this.formElement, detail: { formSubmission: this, ...this.result } })
       this.delegate.formSubmissionFailedWithResponse(this, response)
```
terracatta, line 141: I am actually not confident this code is ever called. If the fetchResponse is not a 2xx it will not be `ok` and therefore only the `requestFailedWithResponse` method will be called.

I added it here for safety, but this branch of code might not ever be called.

en30:
After a form submission, `turbo:submit-end` is dispatched even when the request is failed, and it has `FormSubmission` as its event detail.
https://turbo.hotwired.dev/reference/events
Is `turbo:submit-end` not enough for your use case?

terracatta:
> Is turbo:submit-end not enough for your use case?

It is, I actually didn't realize this because I think sometimes `submit-end` is not firing reliably in Safari. I'm going to close this as I don't think this is needed. Thanks @en30

PR #327, opened 2021-07-28
Fire 'turbo:after-fetch-render' event after turbo frame renders the view

kapantzak:
Hello all!

This PR introduces the `turbo:after-fetch-render` event which
gets fired right after a turbo-frame element renders its view.

You can access the current turbo-frame id with `event.detail.frame`
and the fetch response object with `event.detail.fetchResponse`.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -2,7 +2,7 @@ import { FrameElement, FrameElementDelegate, FrameLoadingStyle } from "../../ele
 import { FetchMethod, FetchRequest, FetchRequestDelegate, FetchRequestHeaders } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { AppearanceObserver, AppearanceObserverDelegate } from "../../observers/appearance_observer"
-import { parseHTMLDocument } from "../../util"
+import { parseHTMLDocument, dispatch } from "../../util"
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { ViewDelegate } from "../view"
@@ -107,6 +107,8 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
         const renderer = new FrameRenderer(this.view.snapshot, snapshot, false)
         if (this.view.renderPromise) await this.view.renderPromise
         await this.view.render(renderer)
+        const frame = this.element.id;
+        dispatch("turbo:after-frame-render", { cancelable: true, detail: { fetchResponse, frame } })
       }
     } catch (error) {
       console.error(error)
--- a/src/tests/fixtures/frames.html
+++ b/src/tests/fixtures/frames.html
@@ -51,5 +51,9 @@
 
     <a id="navigate-form-redirect" href="/src/tests/fixtures/frames/form-redirect.html" data-turbo-frame="form-redirect">Visit form-redirect.html</a>
     <turbo-frame id="form-redirect"></turbo-frame>
+
+    <turbo-frame id="part">
+      <a id="frame-part" href="/src/tests/fixtures/frames/part.html">Load #part</a>
+    </turbo-frame>
   </body>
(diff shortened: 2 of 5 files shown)


t27duck:
Any reason why `event.detail.frame` couldn't be `this.element` which should be the actual turbo frame element instead of only its id?

kapantzak:
> Any reason why `event.detail.frame` couldn't be `this.element` which should be the actual turbo frame element instead of only its id?

Not really, just 2 concerns:
- Is it better to carry the whole frame object inside the event's details instead of just its id?
- `this.assert.equal(frame.id, 'part')` fails (`id` is undefined)

If you think that we should send the object itself, please let me know and I'll change it!

t27duck:
> 
>     * `this.assert.equal(frame.id, 'part')` fails (`id` is undefined)
> 

I'm curious then how this is working in the code:

```javascript
  const frame = this.element.id;
  dispatch("turbo:after-frame-render", { cancelable: true, detail: { fetchResponse, frame } })
```

If `frame = this.element.id` is getting the correct id, then why would `frame = this.element` being sent as part of the event detail result in `frame.id` being undefined?

If the id were to be the only thing passed, then perhaps naming it `frameId` instead of `frame` would make it more descriptive. To me, "frame" implies some form of object rather than its id.

kapantzak:
> I'm curious then how this is working in the code:

I guess that the testing suite represents the elements in a different way, causing the `frame` element to have an undefined id.

> If the id were to be the only thing passed, then perhaps naming it frameId instead of frame would make it more descriptive. To me, "frame" implies some form of object rather than its id.

Agree with the naming, if we pass only the id, `frameId` would be a better name 👍

dhh:
I think the right event name should probably be `turbo:frame-render` to match `turbo:render`, no? Since the style is that these are automatically "after".

dhh:
Also, we should probably stick with the style where we declare this in Session following the form of `notifyApplicationAfterFrameRender`.

dhh:
We should pull in turbo:frame-load as well: https://github.com/hotwired/turbo/pull/59/

dhh:
#59 is using this.element to put in the event, rather than the ID. I think that's a better form to follow here as well.

kapantzak:
Great! So, I'll try to implement notifyApplicationAfterFrameRender inside session and pass the element itself rather than the id, asap.

I will also pull in turbo:frame-load as well.

kapantzak:
I've pushed this [commit](https://github.com/hotwired/turbo/pull/327/commits/95928789cd630fdf3353c98e38a522841f9b154c) which introduces the `notifyApplicationAfterFrameRender` inside session.

There, I've passed the frame element as the target of the event and removed it from the `detail` object, so, now we can get the frame element from the `event.target` property. I think it makes more sense this way, WDYT?

dhh:
Looks good 👍

zeto:
@kapantzak (thanks for the contribution!) context:
```ruby
const html = await fetchResponse.responseHTML
if (html) {
  const { body } = parseHTMLDocument(html)
  const snapshot = new Snapshot(await this.extractForeignFrameElement(body))
  const renderer = new FrameRenderer(this.view.snapshot, snapshot, false, false)
  if (this.view.renderPromise) await this.view.renderPromise
  await this.view.render(renderer)
  session.frameRendered(fetchResponse, this.element)
  session.frameLoaded(this.element)
  this.fetchResponseLoaded(fetchResponse)
}
```

Assuming 🤔 the main intention to provide `fetchResponse` in the Event detail is to have access to the HTML Document, the most likely (?) use case would be something like:

```ruby
document.addEventListener("turbo:frame-render", () => {
  event.detail.fetchResponse.responseHTML.then((html) => {
    const content = (new DOMParser()).parseFromString(html, "text/html")
  })
})
```

In order to avoid reparsing what was just parsed, the detail of the Event could have the already parsed `body` instead of the `fetchResponse` (or in addition to), seeing as at the time the Promise has already been fulfilled. Similar to the `turbo:before-render` event. Would this make sense ?

PR #328, opened 2021-07-30
Propose a `replace` visit on redirection

jaysson:
* Updated the browser adapter to remove the call to `followRedirect` as it now happens in the visit itself
* Updated the tests to reflect the new behaviour

This should also provide a solution to these issues on native adapters:
- https://github.com/hotwired/turbo-ios/issues/30
- https://github.com/hotwired/turbo-android/issues/128

In addition, it allows path properties to be applied correctly to redirection destinations in native adapters.


src/tests/functional/form_submission_tests.ts
```diff
@@ -24,7 +24,7 @@ export class FormSubmissionTests extends TurboDriveTestCase {
 
     this.assert.notOk(await this.formSubmitted)
     this.assert.equal(await this.pathname, "/src/tests/fixtures/one.html")
-    this.assert.equal(await this.visitAction, "advance")
+    this.assert.equal(await this.visitAction, "replace")
```
jayohms, line 27: Is `replace` the expected action for GET form submissions? 

It looks like we'll want a new test to validate normal visit redirects result in a new visit proposal with a `replace` action.

jayohms (commented):
Thanks for doing this @jaysson 👍, been meaning to get this fixed for awhile!

src/tests/functional/form_submission_tests.ts
jaysson: The sequence:
- Visit the original URL with advance action
- If there is a redirection, propose replace visit to the destination with the existing response. That way it doesn't fire one more fetch call, instead reuses the existing response.

User will see the same behaviour as before. Just that programmatically the sequence looks like `advance` + `replace` instead of just `advance`.

As long as there is a redirect, `replace` action is expected at the end. In this test, form is submitted to `/__turbo/redirect` (ref: https://github.com/jaysson/turbo/blob/replace-on-redirect/src/tests/fixtures/form.html#L41), and that endpoint returns a 303 to `/src/tests/fixtures/one.html` resulting in `replace` action.

jaysson:
@jayohms Added two tests. One verifies that the replace visit is happening, and another verifies that it's not replacing the current page, but pushes a new one in the navigation stack.

jayohms:
Thanks @jaysson, I'll take a closer look at this before the end of the week and test in the native apps.

jayohms (approved):
@jaysson This works great on the native side, thanks!

Can you please resolve the merge conflict and let's make sure that the tests pass this time around.

dhh:
Fixed the merge and tests passed 👍

domchristie:
@jayohms @jaysson might there be an alternative way to achieve this without proposing another Visit?

I work on an animation library that animates on `turbo:visit`, assuming it is a user-initiated action. With this change, clicking a link that redirects will trigger the animation twice, and there's no easy way to determine which is the user initiated visit and which is the one triggered by `followRedirect`

Related: https://github.com/hotwired/turbo/issues/877

jayohms:
@domchristie The main issue I see is that we map url paths to view controllers/fragments in the path configuration files. The path properties that each path maps to in the path config files determine how/where to navigate. The turbo native libraries propose visits and load the view controller/fragment *before* the visit actually starts.

So:
- A visit proposal is initiated
- The native libraries look up the path config properties to determine what view controller/fragment to load
- The view controller/fragment is presented
- The proposed visit is started
- The redirect is detected and a new "replace" visit is proposed
- The native libraries look up the path config properties to determine what view controller/fragment to load
- The view controller/fragment is presented
- The redirect "replace" visit is started

Since the original url can have different path properties than the redirect url, the double-visit is the best way to work around these issues. Maybe passing along metadata with the `turbo:visit` event could be another path?

PR #329, opened 2021-07-31
Prevent unexpected side-effects of same-page visit events.

### LINKED ISSUE #321: Expose isSamePage to turbo:visit events
#298 was a very welcome change, as now I can drop `data: { turbo: !current_page?(thing_path) }` from links that are anchors on the current page. However, `turbo:visit` events now fire for visits to anchors on the same page. I'm sure this is potentially useful in some ways, but for now, it's meant that I've had to go from hooking into `turbo:visit` to using `turbo:before-visit` as follows:

```js
import isEqual from "lodash.isequal"

document.addEventListener("turbo:before-visit", function (e) {
  const attributes = ({ origin, pathname }) => ({ origin, pathname })
  if (
    !isEqual(attributes(window.location), attributes(new URL(e.detail.url)))
  ) {
    // do things here
  }
})
```

What I'm doing there is pulling the `origin` (`http://example.com`) and the pathname (`/blog`, without an anchor or any query params) from the window location before the request is processed, and comparing it with those the target URL. If they're identical, then it's safe to say we shouldn't do what would otherwise be triggered under `turbo:visit`.

It'd be nicer to be able to, for example, do this:

```js
document.addEventListener("turbo:visit", function (e) {
  if(e.samePage) { return; }
  // do things here
})
```

domchristie:
Do not trigger unexpected events on same-page visits. Fixes #321

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -154,8 +154,8 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
 
   // Navigator delegate
 
-  allowsVisitingLocation(location: URL) {
-    return this.applicationAllowsVisitingLocation(location)
+  allowsVisitingLocationWithAction(location: URL, action?: Action) {
+    return this.locationWithActionIsSamePage(location, action) || this.applicationAllowsVisitingLocation(location)
   }
 
   visitProposedToLocation(location: URL, options: Partial<VisitOptions>) {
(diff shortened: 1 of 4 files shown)


src/core/session.ts
```diff
@@ -154,8 +154,8 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
 
   // Navigator delegate
 
-  allowsVisitingLocation(location: URL) {
-    return this.applicationAllowsVisitingLocation(location)
+  allowsVisitingLocationWithAction(location: URL, action?: Action) {
+    return this.locationWithActionIsSamePage(location, action) || this.applicationAllowsVisitingLocation(location)
```
domchristie, line 158: We allow the visit, it's just a silent one. Running the `locationWithActionIsSamePage` check first, prevents the `turbo:before-visit` event from being triggered in `applicationAllowsVisitingLocation`

PR #334, opened 2021-08-05
Add support for a reload action in a turbo stream.

defsdoor:
<turbo-stream action="reload" target="mainFrame"/>

I've implemented this in a turbo:before-stream-render event handler and it seems to be a logical and possibly desirable function so I added it to turbo directly.

My use case for this was to push via a cable socket and leave the fetching to the web browser, with the content loading then, therefore, being subject to all the normal server side request validation and scoping.

Writing  tests for this is a little out of my depth at the moment.

### DIFF
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -27,6 +27,14 @@ export const StreamActions: { [action: string]: (this: StreamElement) => void } 
     this.targetElements.forEach(e => e.replaceWith(this.templateContent))
   },
 
+  reload() {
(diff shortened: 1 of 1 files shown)


dhh:
I think this is too specific. I'd rather see us move forward with the idea where a stream can trigger an event, and then you can process that event with dedicated logic in a stimulus controller or elsewhere.

defsdoor:
I like the idea of being able to send a turbo stream that triggers a custom event and payload to a frame and let a stimulus controller attach and respond to it.

Is that the sort of thing you are thinking of ?

PR #339, opened 2021-08-15
Add an option to Refill a Turbo Frame

dfmoreto:
Currently, when we click on a link and a Turbo Frame and the requested content is loaded the `src` attribute of frame is filled and we can't refill this Turbo Frame again with the same URL request.

For example, the code below, where we have a link targeting a TurboFrame.
```
<a href="/my-sample-url" data-turbo-frame="my-turbo-frame">My Link</a>

<turbo-frame id="my-turbo-frame">
</turbo-frame>
```
If I click on link, Turbo Frame loads the content as expected. If I click on the link for a second time, the content of TurboFrame will remain the same as expected to avoid reloading all over again.

But there are some cases we need to reload the content again for some update that could be happen on server side.

I'm proposing to add a new attribute called `refill` to enable a Turbo Frame to be reloaded with the same URL when a request is done.

An example with this new thing would be this code:
```
<a href="/my-sample-url" data-turbo-frame="my-turbo-frame">My Link</a>

<turbo-frame id="my-turbo-frame" refill="true">
</turbo-frame>
```
As TurboFrame has an attribute `refill` set as `true`, whenever we click on the link, the content on TurboFrame will be reloaded.

**:red_circle: :red_circle: :red_circle: DISCLAIMER: I need a help to finish the specs for this feature. I've built the test structure, but I couldn't find a way to clean the content of a TurboFrame on specs. If someone is able to help me on that, it'd be great! =D**

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -77,7 +77,9 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   async loadSourceURL() {
-    if (!this.settingSourceURL && this.enabled && this.isActive && this.sourceURL != this.currentURL) {
+    let canLoadSource = !this.settingSourceURL && this.enabled && this.isActive && 
+                        (this.refillable || this.sourceURL != this.currentURL)
+    if (canLoadSource) {
       const previousURL = this.currentURL
       this.currentURL = this.sourceURL
       if (this.sourceURL) {
@@ -312,6 +314,10 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     if (this.element.src) {
       return this.element.src
     }
+  }
+
+  get refillable() {
+    return this.element.refill
   }
 
   set sourceURL(sourceURL: string | undefined) {
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -78,7 +78,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
 
   async loadSourceURL() {
     let canLoadSource = !this.settingSourceURL && this.enabled && this.isActive && 
-                        (this.refillable || this.sourceURL != this.currentURL)
+                        (this.reloadable || this.sourceURL != this.currentURL)
     if (canLoadSource) {
       const previousURL = this.currentURL
       this.currentURL = this.sourceURL
@@ -316,8 +316,8 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
     }
   }
 
-  get refillable() {
-    return this.element.refill
+  get reloadable() {
+    return this.element.reloadable
   }
 
   set sourceURL(sourceURL: string | undefined) {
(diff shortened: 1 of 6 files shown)


kirillplatonov:
@dfmoreto maybe we could use `data-turbo-reload` attributes instead of refill?

dfmoreto:
@kirillplatonov I think `data-turbo-reload` is too verbose as it's altrady a TurboFrame attribute. What do you think about using only `reload` (or `reloadable`) in TurboFrame?

kirillplatonov:
@dfmoreto Oh, right. Then maybe `data-turbo-frame-reload`?
I do think that having `data-turbo` namespace is good. By using `reload` or `reloadable` it's not clear that it's related to turbo.

dfmoreto:
> Then maybe data-turbo-frame-reload
I agree with you at this point, `frame-reload` is a good one. 
About prefixing with `data-turbo`, TurboFrame tag has another attributes that aren't prefixed by `data-turbo`. Prefixing this won't disconnect from the pattern of others?

adzienis:
Hey, I think this is a great idea, but is there a way for this functionality to be more generic? There was this recent pull request https://github.com/hotwired/turbo/pull/334 that was unfortunately closed due to narrow scope, which would have helped with your original motivation of updating based on side effects from the server. I believe functionality like this could just be rewritten with a small stimulus controller (honestly probably just any sort of basic event listener, and just call reload on the turbo frame since it will mutate the src attribute accordingly), if custom events were emitted from a stream directly as suggested.

I just feel like having a more general way of directly refreshing a turbo frame from server events could be useful (like imperatively emitting a reload event). For instance, I've forked this repo to support such a reload event just so I can directly invalidate the targeted turbo frames through websockets, and refetch, since I think such an action has at least been useful to me.

dfmoreto:
@adzienis yes, it could be resolved with this PR, but my proposal here is more simple: is to only allow a TurboFrame to receive a content again even if it comes from same source as previous request.
You're right, both this and that PR could be resolved with a Stimulus controller. But although it makes sense to me to have a Stimulus controller to handle events, I can't see as good way to create one with the single purpose of clean `src` attribute on TurboFrame to allow it to receive the content from same URL again (in my opinion it should be TurboFrame responsibility to enable itself to receive a content again)
So, I think it's better o keep this reload action with this single and strict purpose.

dfmoreto:
@kirillplatonov I've just renamed from `refill` to `data-turbo-frame-reload`. I've tested in some applications and it didn't seem as verbose on the code as I though and it (maybe) can open doors to have more data-turbo attributes on a TurboFrame in the future.
Can you help me with the tests? I couldn't complete the test for this feature. On the file `src/tests/functional/frame_tests.ts` line 160, I couldn't find a way to erase the TurboFrame content to check if the content is fulfilled again after a new action with same URL.

kirillplatonov:
@dfmoreto actually, it seems like we should force reload by default, without additional config. Like normal web links. So it's a bug in the current frame behavior. I opened a PR on top of your's that fixes that: https://github.com/hotwired/turbo/pull/349

Thank you for bringing this in! 🙌

Issue #340, opened 2021-08-17
Links with method not work if used inside another form


davidgil:
I detect an issue with the data-turbo-method attribute on links that are included inside another form, this is a follow up of:
https://github.com/hotwired/turbo/pull/299

In session.ts: 
`
convertLinkWithMethodClickToFormSubmission(link: Element) {
    const linkMethod = link.getAttribute("data-turbo-method")

    if (linkMethod) {
      const form = document.createElement("form")
      form.method = linkMethod
      form.action = link.getAttribute("href") || "undefined"

      link.parentNode?.insertBefore(form, link)
      return dispatch("submit", { cancelable: true, target: form })
    } else {
      return false
    }
  }
`
the "virtual" form created is injected in the parent node of the link, the problem is that if the link is inside another form this functionality dont work. The solution could be to inject the form at the end of body?

`
convertLinkWithMethodClickToFormSubmission(link: Element) {
    const linkMethod = link.getAttribute("data-turbo-method")

    if (linkMethod) {
      const form = document.createElement("form")
      form.method = linkMethod
      form.action = link.getAttribute("href") || "undefined"

      document.body.appendChild(form)

      return dispatch("submit", { cancelable: true, target: form })
    } else {
      return false
    }
  }
`

dhh:
I like that idea. Do make a PR for it 👍

davidgil:
My first PR so... 
https://github.com/hotwired/turbo/pull/341

Intrepidd:
@dhh Can we reopen this issue ? The commit was reverted so we are still facing this issue. I would be glad to give it a swing, any pointers on the way you would like to see it done ?

PR #341, opened 2021-08-17
Links with method not work if used inside another form

davidgil:
https://github.com/hotwired/turbo/issues/340

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -145,13 +145,12 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
       form.method = linkMethod
       form.action = link.getAttribute("href") || "undefined"
 
-      link.parentNode?.insertBefore(form, link)
+      document.body.appendChild(form)
       return dispatch("submit", { cancelable: true, target: form })
     } else {
       return false
     }
   }
-
 
   // Navigator delegate
 
--- a/src/tests/fixtures/form.html
+++ b/src/tests/fixtures/form.html
@@ -171,5 +171,8 @@
       </div>
     </turbo-frame>
     <a href="/__turbo/messages?content=Link!&type=stream" data-turbo-method="post" id="link-method-outside-frame">Stream link outside frame</a>
+    <form>
+      <a href="/__turbo/messages?content=Link!&type=stream" data-turbo-method="post" id="link-method-inside-form">Stream link inside form</a>
+    </form>
   </body>
 </html>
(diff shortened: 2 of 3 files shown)


src/core/session.ts
```diff
@@ -145,14 +145,13 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
       form.method = linkMethod
       form.action = link.getAttribute("href") || "undefined"
 
-      link.parentNode?.insertBefore(form, link)
+      document.body.appendChild(form)
```
tleish, line 148: Does this cause multiple forms to stack up at the bottom of the body if pressing submit, back, submit, back? 
 Should the code cleanup the form before turbo caches the page?

davidgil: @tleish yes i thought you are right, the problem was also there with previus solution. I have no idea how clean the page before caching, do you have a clue? is there some other parts with same problem so i can take a look and try to improve my PR?

dhh:
Let’s deal with accumulation in a separate PR.

src/core/session.ts
tleish: The following would self cleanup:

```diff
      const form = document.createElement("form")
      form.method = linkMethod
      form.action = link.getAttribute("href") || "undefined"
+     form.addEventListener("submit", () => form.remove())

      document.body.appendChild(form)
      return dispatch("submit", { cancelable: true, target: form })
```

willcosgrove: I'm seeing a behavior where the form is actually removed from the DOM by this event listener before Turbo is able to handle the form submission. Turbo checks to make sure that the form is mounted to the DOM before it handles the submission.

https://github.com/hotwired/turbo/blob/aa9724dd23309a56b3e7c6ee62c673fa63799038/src/core/frames/form_interceptor.ts#L23-L25

Could this instead be?

```diff
      const form = document.createElement("form")
      form.method = linkMethod
      form.action = link.getAttribute("href") || "undefined"
-     form.addEventListener("submit", () => form.remove())
+     document.documentElement.addEventListener("turbo:before-cache", () => form.remove())

      document.body.appendChild(form)
      return dispatch("submit", { cancelable: true, target: form })
```

Or even?

```diff
      const form = document.createElement("form")
      form.method = linkMethod
      form.action = link.getAttribute("href") || "undefined"
-     form.addEventListener("submit", () => form.remove())
+     form.addEventListener("submit", () => setTimeout(() => form.remove()))

      document.body.appendChild(form)
      return dispatch("submit", { cancelable: true, target: form })
```

PR #343, opened 2021-08-19
Add Github Codespace configuration

edwinv:
With the introduction of Github Codespaces, it is easier than ever to contribute to an open source project. In my pursuit of fixing a bug in Turbo, I wanted to use a Codespace. In order to run the test suite, some additional tools are needed in the devcontainer:

- Java for starting a webserver
- Firefox for browser testing
- Chromium for browser testing

My goal is to successfully run `yarn test` after booting a Codespace. I'm running into some issues I don't know how to solve, so this is a WIP PR with a call for some help:

1. Firefox is green for all but one spec about the scroll position. 

  ```
  × firefox on linux 5.4.0-1055-azure - NavigationTests - navigating back to anchored URL (0.783s)
      AssertionError: scrolled to #main: expected false to be truthy
        at NavigationTests.test navigating back to anchored URL @ src/tests/functional/navigation_tests.ts:182:16
        at runMicrotasks @ anonymous
        at processTicksAndRejections @ node:internal/process/task_queues:96:5
  ```

2. Chromium doesn't start due to this error, but the weird thing it that `chromedriver` has the exact same version.

```
This version of ChromeDriver only supports Chrome version 91
Current browser version is 90.0.4430.212 with binary path /usr/bin/chromium
```

Anyone suggestions how to solve these last two errors to enable Codespaces for this project?

If you want to test these changes or want to experiment yourself, you can [open this PR with Codespace too](https://docs.github.com/en/codespaces/developing-in-codespaces/using-codespaces-for-pull-requests#opening-a-pull-request-in-codespaces).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,21 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.191.1/containers/typescript-node/.devcontainer/base.Dockerfile
+
+# [Choice] Node.js version: 16, 14, 12
+ARG VARIANT="16-buster"
+FROM mcr.microsoft.com/vscode/devcontainers/typescript-node:0-${VARIANT}
+
+# [Optional] Uncomment this section to install additional OS packages.
+ENV CHROMIUM_VERSION 90.0.4430.212-1~deb10u1
+RUN apt-get update \
+    && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends \
+        chromium-common=${CHROMIUM_VERSION} chromium-driver=${CHROMIUM_VERSION} chromium-sandbox=${CHROMIUM_VERSION} \
+        openjdk-11-jdk \
+        firefox-esr
+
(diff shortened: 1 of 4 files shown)


edwinv:
Last two commits fix the open issues:

- By peeking at [the way chromedriver is installed in the `ubuntu-latest`](https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/google-chrome.sh) images for Github Actions, I found a way to install the correct version of chromedriver together with chrome itself.
- The issue with Firefox seems to be related to timing: the assertion is probably done before the page is fully loaded. By waiting for the next body, everything works fine.

To test this PR, click the green "Code" button on the top right of this PR and start a Codespace. After starting, execute `yarn test` in the terminal and see that all tests pass.

kirillplatonov:
That looks great. Just played with it. Can we add `yarn install` to post create commands?

edwinv:
> That looks great. Just played with it. Can we add `yarn install` to post create commands?

I think `yarn build` already installed the dependencies, because everything worked fine. But adding it explicitly is better, so everyone can see what is happing. 👍

Issue #332, opened 2021-08-03
Turbo Drive: "Referer" header set to target URL, not current URL


awh-tokyo:
When using Turbo Drive, the Referer header seems to be set to the target URL, not the current URL:

![Screen Shot 2021-08-03 at 12 33 25](https://user-images.githubusercontent.com/689566/127953799-547aa015-451b-4ce7-af88-49c2c1e02788.png)

This was following a link from /visits/ so I would have expected that to be the Referrer.

Is this by design?  Is there some other way to know the referring page?  I need to redirect back to different URLs once editing is done depending on the referring URL.

awh-tokyo:
EDIT:  I'll note that by adding data-turbo="false" to those links, it works as expected and I receive the appropriate Referer header.  I've done that for now, which isn't my favourite solution, but lets me get back to work.

zzeligg:
I confirm this occurs with v7.0.0-rc.1. It breaks `:back` url functionality in Rails link / url helpers.

It was introduced with the Pausable requests changes: #306 

When commenting out line 82 in `fetch_request.ts` :

https://github.com/hotwired/turbo/blob/main/src/http/fetch_request.ts#L80

which simply disables `await this.allowRequestToBeIntercepted(fetchOptions)` in `FetchRequest#perform()`, then the Referer header sent is the correct one.

edwinv:
In #344 I've added a test to ensure the header is set correctly. The test currently fails and I can confirm what @zzeligg found out about the changes in #306.

PR #344, opened 2021-08-20
Add test for correct referrer header

edwinv:
The current version of Turbo contains a bug in which the `referer` header is not correctly set for requests (see #332 and https://github.com/hotwired/turbo-rails/issues/211) This potentially breaks a lot of functionality in Rails apps. I noticed there is no test assuring the header is set correctly, so I added one.

Like stated in #332, the bug is introduced with the changes in #306 by @kirillplatonov and @dhh. When you change `fetch_requests.ts:82` back to to old version, the test is green.

```
// await this.allowRequestToBeIntercepted(fetchOptions)
dispatch("turbo:before-fetch-request", { detail: { fetchOptions } })
```

I'm not familiar enough with the internals of Turbo to understand why the header changes, so any help to actually fix the issue is welcome!


kirillplatonov:
@edwinv thank you for adding the test. I will take a look at this issue with Pausable API.

kirillplatonov:
https://github.com/hotwired/turbo/pull/345

edwinv:
Perfect! Closing this PR in favor of #345, because it contains both the tests and the fix.

PR #346, opened 2021-08-21
fixup: late lifecycle attachment

guybedford:
I haven't tested this change, but in theory this should be the right sort of approach to handle late attachment of page state for module scripts that are delayed, in particular as happens when using es-module-shims.

Once tested and assuming there aren't any bugs, this should resolve https://github.com/hotwired/turbo-rails/issues/216.

### DIFF
--- a/src/observers/page_observer.ts
+++ b/src/observers/page_observer.ts
@@ -26,6 +26,7 @@ export class PageObserver {
         this.stage = PageStage.loading
       }
       document.addEventListener("readystatechange", this.interpretReadyState, false)
+      this.interpretReadyState(document.readyState);
       addEventListener("pagehide", this.pageWillUnload, false)
       this.started = true
     }


dhh:
This unfortunately does not work. If we're not in the correct ready state, the rest of the library won't work.

Seems like we have to fix this in the shim. That it basically has to send a right event for libraries that depend on readystatechange. Any other library that does this will fail in the same way otherwise.

dhh:
Seems like we could fix this universally if the shim did a `document.dispatchEvent(new Event("readystatechange"))` once it's done. If I manually set that up, it works.

guybedford:
> If we're not in the correct ready state, the rest of the library won't work.

Looking at the code it seems to me like the `this.pageIsInteractive()` call in `pageIsComplete` can handle the late attachment the issue is just not getting the event. What else doesn't work about the immediate double transition here?

Will follow up re the dispatch on the es-module-shims issue, but there are a few things to consider around this.

dhh:
Solved by es-module-shims 0.12.7

PR #349, opened 2021-08-22
Force frame reload on every link click

kirillplatonov:
Based on https://github.com/hotwired/turbo/pull/339 this PR fixes frame's behavior and forces reload on every link click (like normal web links).


dfmoreto:
@kirillplatonov if TurboFrame should be reloadable every time, wouldn't be better to remove the verification `this.sourceURL != this.currentURL`?
But I have another point: as TurboFrame verifies if current URL is different from source URL, I thought the default behavior was not processing same request to avoid wasting a request to receive same content again. The `reloadable` attribute from another PR would be like the dev "taking the risk" of reloading it with same request URL.

kirillplatonov:
@dfmoreto removing `this.sourceURL != this.currentURL` will break the existing behavior on controller reconnect: https://github.com/hotwired/turbo/blob/main/src/tests/functional/loading_tests.ts#L109

We have other ways to load frame URLs aside from links (forms, lazy loading, frame source change). I didn't change the behavior for any of them, only for links. So it won't waste a request unless the user clicked on the link again. It should be safe.

en30:
I agree that a turbo frame should reload on every click based on the analogy with browsers' behavior.
But, when a Turbo user
1. click an anchor to load a turbo frame,
2. disconnect the turbo frame,
3. reconnect the turbo frame,

does this PR change behavior at 3 from **not reloading** to **reloading**, doesn't it?

I mean in a test case in `src/tests/functional/frame_tests.ts`
```ts
async "test ..."() {
    await this.clickSelector("#hello a")
    await this.nextEventNamed("turbo:before-fetch-request")

    await this.remote.execute(() => {
      window.savedElement = document.querySelector("#frame")
      window.savedElement?.remove()
    })
    await this.nextBeat

    await this.remote.execute(() => {
      if (window.savedElement) {
        document.body.appendChild(window.savedElement)
      }
    })
    await this.nextBeat

    const eventLogs = await this.eventLogChannel.read()
    const requestLogs = eventLogs.filter(([name]) => name == "turbo:before-fetch-request")
    this.assert.equal(requestLogs.length, 0)
}
```

kirillplatonov:
@en30 you're right, it's a bug. Thanks for noticing that and adding a test case!
Opened a PR to fix that: https://github.com/hotwired/turbo/pull/350

dfmoreto:
@kirillplatonov, I got it. Thanks for explanation.
I'm happy on participating a little bit on this resolution :D

PR #353, opened 2021-08-25
Handle server errors for form submissions

### LINKED ISSUE #208: Turbo should probably render post 500 status response HTML even if trackedElementsAreIdentical is false.
In ASP.NET Core we have developer exception page that is automatically generated when there is a server error (500). The developer exception page doesn't know anything about Turbo. The developer exception page is correctly rendered with Turbo on a GET request with a status code of 500 but not on a post request, Turbo just reload the page. Should Turbo just render any post 500 status response HTML even if trackedElementAreIdentical is false?

FireShot Capture 269 - Internal Server Error - https___localhost_5001_server_error

### LINKED ISSUE #234: 5xx Responses trigger full reload for form submissions.
Hello.

I am using:

- Rails  6.1.3.1
- turbo-rails 0.59
- turbo v7.0.0beta4

I noticed that if i visit some Turbo enabled links that may tigger a 5xx response, we see the classic page informing us about the server error (i.e)

Screenshot 2021-04-08 at 3 46 34 PM

When i submit a form to create, update or delete a resource, that may trigger a 5xx response, we don't get the same response. The page fully reloads and the submission failure error is written on the browser console.  That seems, to be related to the fact that the response headers do not include the same assets,  turbo detects that change and triggers a full reload on the page. But that is not happening for the first use case (link navigation).

In fact if we remove data-turbo-track=reload,  then even form submission will result in the error page.

After searching around the codebase i found that in normal turbo visit navigation, error responses are handled by an ErrorRenderer:

https://github.com/hotwired/turbo/blob/db96ea903b50e1ae3ceef89a7285ef7a52657dad/src/core/drive/visit.ts#L203

https://github.com/hotwired/turbo/blob/db96ea903b50e1ae3ceef89a7285ef7a52657dad/src/core/drive/page_view.ts#L23

After form submission, it is handled by the  PageRenderer that will check for asset changes and reload if there are any changes
https://github.com/hotwired/turbo/blob/db96ea903b50e1ae3ceef89a7285ef7a52657dad/src/core/drive/navigator.ts#L94

Is there any possibility we could have the same behaviour for form submissions? Or is it something that i should handle somehow at application level?( i took the idea from the test here: https://github.com/hotwired/turbo/blob/4670f2b57c5d0246dfc0f6d10ff7d9a52a63fdca/src/tests/fixtures/500.html)

 #208 is definitely related.

Thanks a lot for your time

kirillplatonov:
Closes #208, closes #234.

Call `View.renderError` when receiving server error (5XX) after form submission.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -100,7 +100,11 @@ export class Navigator {
 
     if (responseHTML) {
       const snapshot = PageSnapshot.fromHTMLString(responseHTML)
-      await this.view.renderPage(snapshot)
+      if (fetchResponse.serverError) {
+        await this.view.renderError(snapshot)
+      } else {
+        await this.view.renderPage(snapshot)
+      }
       this.view.scrollToTop()
       this.view.clearSnapshotCache()
     }
--- a/src/tests/fixtures/form.html
+++ b/src/tests/fixtures/form.html
@@ -5,7 +5,7 @@
     <title>Form</title>
     <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
     <script src="/src/tests/fixtures/test.js"></script>
-    <style>
+    <style id="form-fixture-styles">
       dialog {
         display: block;
         position: static;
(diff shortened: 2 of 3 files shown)


buncis:
can we do this for 403 too?

PR #356, opened 2021-08-27
Seperate progress bars for form submits and visits

### LINKED ISSUE #355: Progress bar does not disappear after form submission on safari
Hi, I am using:

- mac OS Big Sur 11.5.2
- Safari 14.1.2
- Rails 6.1.4.1
- Webpacker 6.0.0.rc5
- Turbo 6.0.0.rc2

After form submission, the progress bar does not disappear. I noticed the issue only on Safari. Here is a video with the issue:

https://user-images.githubusercontent.com/425283/131020729-9b092463-4e5c-4de2-9062-2c6d48c5f875.mp4

i also created a sample app in order to check and replicate the issue locally:

[Sample app](https://github.com/thanosbellos/turbo-webpacker-safari-progress-bar-bug)

terracatta:
This PR fixes #355. 

This issue occurs due to a race condition and possibly a bug in Safari when a form is submitted via Turbo Drive that redirects. In this case, _both_ the form submission _and_ the visit caused by the redirect call `showFormProgressBarAfterDelay()`.  This should be fine (and is for Firefox and Chrome). The original code stored the timeout for the progress bar under the same property so there is only one to clear.

But for some reason that I cannot explain, if the timing of these calls is _just_ right and `window.clearTimeout(x)` is called at the same time `x = window.setTimeout(...)`, Safari gets confused. While `window.clearTimeout` appears to clear the timeout, there is still some remnant of the one that was just set in the background. This remnant `window.setTimeout` still fires after both the redirect visit and the form submission are both complete and it causes this lingering progress bar to remain on the screen.

I don't really understand what's going on here, but what I _do_ know is after I apply this PR, I am unable to reproduce the issue on any of the three browsers. I think the reason why is we aren't sharing the same underlying timeout property, and nothing is attempting to set the property at the precise time something else is trying to clear it.

### DIFF
--- a/src/core/native/browser_adapter.ts
+++ b/src/core/native/browser_adapter.ts
@@ -9,7 +9,8 @@ export class BrowserAdapter implements Adapter {
   readonly session: Session
   readonly progressBar = new ProgressBar
 
-  progressBarTimeout?: number
+  visitProgressBarTimeout?: number
+  formProgressBarTimeout?: number
 
   constructor(session: Session) {
     this.session = session
@@ -29,7 +30,7 @@ export class BrowserAdapter implements Adapter {
   visitRequestStarted(visit: Visit) {
     this.progressBar.setValue(0)
     if (visit.hasCachedSnapshot() || visit.action != "restore") {
-      this.showProgressBarAfterDelay()
+      this.showVisitProgressBarAfterDelay()
     } else {
       this.showProgressBar()
     }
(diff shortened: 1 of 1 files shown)


georgeclaghorn:
This change seems to have broken the progress bar for form submissions from inside frames. It continues forever:

https://user-images.githubusercontent.com/94129/132101831-bf8859b8-3612-40f1-9fcf-4ec0d5fecb70.mov

thanosbellos:
@georgeclaghorn  What version of turbo are you using?

I had the same issue on turbo-v7.0.0.rc2 but after upgrading to turbo-v7.0.0.rc3(which includes this pr) i haven't noticed the issue yet.

georgeclaghorn:
Ah, sorry, I thought this was in rc2. Upgraded.

Issue #336, opened 2021-08-09
No interface for listening to Turbo errors


tomwaddington:
Any errors within Turbo are directly output with `console.errors`, meaning 3rd party exception tracking libraries do not (by default) capture possible failures within the page.

As we migrate to Turbo, I'd like to keep track of any user-facing errors recognised by Turbo,

I'd be happy to try and improve this, possibly by emitting a `CustomEvent` that could be listened to by the host page?

dhh:
Happy to see a PR for this 👍

PR #357, opened 2021-08-27
Error reporting event

tomwaddington:
Rather than just emitting any errors to the console, we can emit them as `turbo:error` `CustomEvent`s, allowing the application developer to monitor and raise these to any exception trackers.

I think this would make migration to Turbo easier, see #336.

I'd appreciate a quick 👀  on this approach, and then I'd be happy to update the PR with some tests.


src/util.ts
```diff
@@ -49,3 +49,8 @@ export function uuid() {
     }
   }).join("")
 }
+
+export function reportError(error) {
+  console.warn(error)
```
gobijan, line 54: Shouldn't be this console.error to stay true to the method name and the old function?

Issue #19, opened 2020-12-22
Issues reading response on turbo:submit-end


cj:
I am trying to read the error response from `turbo:submit-end`, but it is failing with:

```
base.js:96 Uncaught (in promise) TypeError: Failed to execute 'text' on 'Response': body stream already read
```

The test code I have is as follows:

```javascript
  async turboSubmit(event) {
    const {
      success,
      fetchResponse: { response },
    } = event.detail

    if (success) {
      return
    }

    console.log(await response.text())
  }
```

dhh:
Fixed by #364

PR #364, opened 2021-08-30
Fix for get response html/text in locked body

ferdyrurka:
Hi,

I detected a small bug during migrating from turbolinks to turbo. :)

Zrzut ekranu z 2021-08-30 20-19-56

It appears when he wants to get `responseHTML` or `responeText` on event: `turbo: before-fetch-response`

### DIFF
--- a/src/http/fetch_response.ts
+++ b/src/http/fetch_response.ts
@@ -44,12 +44,12 @@ export class FetchResponse {
   }
 
   get responseText(): Promise<string> {
-    return this.response.text()
+    return this.response.clone().text()
   }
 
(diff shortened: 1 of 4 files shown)


dhh:
Nice. Could you add a test as well?

ferdyrurka:
sure, tomorrow I will add tests :)

marphi:
@ferdyrurka Does this pr resolve issue #19?

ferdyrurka:
@marphi yes :)


I checked this use case it with the following code:

```
document.addEventListener('turbo:submit-end',  async (event) => {
    console.log(await event.detail.fetchResponse.responseHTML)
});
```

PR #368, opened 2021-09-01
Fix restoring a visit without a cached snapshot available

jayohms:
There may not always be a cached snapshot available when restoring a visit. We need to issue a new request in that situation.

This fixes a regression from #298. Bug report here: https://github.com/hotwired/turbo-android/issues/184

cc @domchristie

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -372,9 +372,7 @@ export class Visit implements FetchRequestDelegate {
   }
 
   shouldIssueRequest() {
-    const hasCachedSnapshot = this.hasCachedSnapshot()
-
-    if (this.action == "restore" && !hasCachedSnapshot) {
+    if (this.action == "restore" && !this.hasCachedSnapshot()) {
       return true
     } else if (this.isSamePage) {
       return false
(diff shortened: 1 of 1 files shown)


domchristie:
I think this "breaks" the intended behaviour when navigating Back from a same-page anchor with turbo caching disabled 🥴

E.g.
1. Visit page with `<meta name="turbo-cache-control" content="no-cache">`
2. Tap "Skip" to jump the the main content
3. Tap Back

There won't be a cached snapshot, so it'll make a request, even though it should just jump back :/

jayohms:
Hmmm, that is a dilemma 🤔. Let me think about this a bit more. The existing behavior breaks the native apps, so we definitely need to address that aspect.

domchristie:
I'm not familiar with Android logs, so I'm not sure I understand the problem. Could you describe the flow that is causing the error?

jayohms:
@domchristie I've outlined the issue here: https://github.com/hotwired/turbo-android/pull/185

I was actually able to resolve the issue in the Android library and the iOS library isn't affected by this bug, so there shouldn't need to be any additional changes necessary here. Let me know if that doesn't make sense or you have any questions. Thanks!

PR #370, opened 2021-09-02
Fix frame reloading from links/forms nested inside a turbo-frame

### LINKED ISSUE #365: Frame not updating a second time unless I manually clear src attribute
Hi folks. Cudos to all the people building this thing. I think it bodes well for the future of software development.

I am having a similar issue to some others I have researched: https://github.com/hotwired/turbo/pull/263, https://github.com/hotwired/turbo/pull/360, https://github.com/hotwired/turbo/pull/339, https://github.com/hotwired/turbo/pull/349.

I have a link above an empty turbo-frame as so:

````
<a class="icon-link" data-turbo-frame="new_chapter_at_1_form" data-controller="text-tooltip" data-text-tooltip-text-value="New chapter" href="/versions/16/chapters/new?q%5Border%5D=1">
      <i class="fas fa-plus"></i>
</a>
<turbo-frame id="new_chapter_at_1_form">
  
</turbo-frame>
````

The response to the link is as so:
````
<turbo-frame id="new_chapter_at_1_form">
  <div id="new_chapter_at_1_form_div">
    //form here
    ...
    //Cancel button that removes enclosing div with id="new_chapter_at_1_form_div" from dom using a stimulus controller
  </div>
</turbo-frame>
````

The issue is that if I click the cancel button to remove the div containing the form, the link can't load the form again because the src gets set on the turbo-frame. The request succeeds and the browser returns the html, but it doesn't update the frame. If I clear the src manually, then it works again. 

I tried to use a stimulus controller to clear the src attribute in the `connect()` lifecycle callback, but it never actually calls it because the turbo-frame is there from the beginning and never actually get reloaded (just its contents, as originally intended). I have seen some other stimulus controllers that apparently do what I need, but they look overly complex for such a simple task (not a javascript guy, hence the appeal of hotwire/turbo).

I am hoping there is an easy solution to this, but I am using `7.0.0-rc.3` which was supposed to have a fix to the similar issues I listed above, so I am at a loss and think there might be another facet to this bug that hasn't been fixed yet.

### PR DESCRIPTION
**Pull request by tleish**
Closes #365

### DIFF
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -190,6 +190,14 @@ export class FrameTests extends TurboDriveTestCase {
     await this.nextEventNamed("turbo:before-fetch-request")
   }
 
+  async "test an inner/outer link reloads frame on every click"() {
+    await this.clickSelector("#inner-outer-frame-link")
+    await this.nextEventNamed("turbo:before-fetch-request")
+
+    await this.clickSelector("#inner-outer-frame-link")
+    await this.nextEventNamed("turbo:before-fetch-request")
+  }
+
   async "test reconnecting after following a link does not reload the frame"() {
     await this.clickSelector("#hello a")
     await this.nextEventNamed("turbo:before-fetch-request")
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -214,7 +214,6 @@ export class FrameTests extends TurboDriveTestCase {
     await this.nextEventNamed("turbo:before-fetch-request")
   }
 
-
   async "test reconnecting after following a link does not reload the frame"() {
     await this.clickSelector("#hello a")
     await this.nextEventNamed("turbo:before-fetch-request")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by WriterZephos** [comment]
I am not a JavaScript expert but looking at https://github.com/hotwired/turbo/pull/360 it seems like it was supposed to already test for this. Is there an existing test producing a false positive?

**Comment by tleish** [comment]
FYI: The failed chrome test passed on firefox in the same build.  Also in my local environment the failed test passes.

**Comment by tleish** [comment]
> I am not a JavaScript expert but looking at #360 it seems like it was supposed to already test for this. Is there an existing test producing a false positive?

#360 fixed the case where the link existed outside a frame, but Turbo uses different code when the link exists inside a frame.

**Comment by kirillplatonov** [comment]
@tleish could you please rebase from `main`? This should help with failing tests.

File: src/tests/functional/frame_tests.ts [hunk 1]
```diff
@@ -177,6 +177,31 @@ export class FrameTests extends TurboDriveTestCase {
...
+
+  async "test an inner/outer form reloads frame on every submit"() {
+    await this.clickSelector("#inner-outer-frame-submit")
+    await this.nextEventNamed("turbo:before-fetch-request")
+
+    await this.clickSelector("#inner-outer-frame-submit")
+    await this.nextEventNamed("turbo:before-fetch-request")
+  }
+
+
```
kirillplatonov, line 204:
```suggestion

```

tleish: rebased and removed extra line

kirillplatonov (approved):

PR #379, opened 2021-09-06
Add data-confirm behavior to turbo

virolea:
## Background

Following [this merged PR from Rails](https://github.com/rails/rails/pull/43112#issuecomment-908214854) and the deprecation of `rails-ujs`, it has been proposed that `confirm` and `disable` behaviors are added to Turbo. 

As a first contribution to this repo, I started with `confirm` as it appeared to be simpler. If we agree on an API for `disable` (should it work submit buttons? inputs? I proposed a change on `rails ujs` [here](https://github.com/rails/rails/pull/43031) to know if we wanted to extend it to remote "elements" as well) I'll be happy to work on it . 

## Implementation 

One can add a `data-turbo-confirm` attribute to any turbo-tracked form to have a dialog asking for confirmation before submitting the request : 

```html
<form action="/action" method="post" data-turbo-confirm="Are you sure?">
  <!-- ... -->
</form>
```

Capture d’écran 2021-09-06 à 18 50 14


t27duck:
Any reason this couldn't be applied to links as well?

acetinick:
I like the idea of this.  I think if everyone agrees, we should also extend this to work for <a> tags that have data-method attributes attached to them too.

dhh:
Need to make a clean break AND allow for coexistence of UJS and Turbo. So prefer to keep this on data-turbo-confirm.

Don’t think this makes sense on regular links, but on data-turbo-method links yes.

acetinick:
@

> Need to make a clean break AND allow for coexistence of UJS and Turbo. So prefer to keep this on data-turbo-confirm.
> 
> Don’t think this makes sense on regular links, but on data-turbo-method links yes.

Sorry meant data-turbo-method in my comment. Updated

virolea:
Following your advices, I have extended the behavior to `data-turbo-method` links. Since this links create "shadow" forms when clicked, we only had to forward the `data-turbo-confirm` to the form when appropriate and everything falls back into place.

dhh:
See a failing test:

```
× chrome on linux - FormSubmissionTests - link method form submission inside frame with confirmation cancelled (0.386s)
116
    NoSuchElement: [POST http://localhost:4444/wd/hub/session/a665b96dc88bc2a976a45e18366ae51f/element / {"using":"css selector","value":"#frame div.message"}] no such element: Unable to locate element: {"method":"css selector","selector":"#frame div.message"}
117
```

virolea:
@dhh thanks, the last push should have fixed it

songjiz:
What do you think about making the confirm method overridable just like [this PR](https://github.com/rails/rails/pull/32404/files) ?

virolea:
Hey @songjiz ! 

This idea crossed my mind as well :) I even started to work on this, at first but I wanted to validate the `confirm` behavior first, before improving it. I'll propose a solution for this today !

virolea:
Solution pushed for overriding the native confirm modal. 

A new configuration method is available on the `Turbo` object for setting up the new method : 

```typescript 
const newConfirmMethod = (message, formElement) => { confirm("overriden method") }

Turbo.setConfirmMethod(newConfirmMethod)
```

kylekeesling:
I just noticed that links that only use `data-turbo-confirm`, but do not also use `data-turbo-method`,  are not currently honored with this implementation. Is there a specific reason that it wouldn't be desirable to support confirmation on all links? 

If the goal is to keep parity with the exiting rails-ujs `data-confirm` attribute, it currently does support confirmation of normal links.

jasonfb:
*new Rails app created after Oct 2021* (affects only developers with the yanked versions installed on their machines)

Make sure to uninstall the yanked 7.1.0 and 7.1.1 from your machine with `gem uninstall turbo-rails`

and also fix any Rails apps created since October 2021 by fixing the Gemfile. 


https://stackoverflow.com/questions/70671324/new-rails-7-turbo-app-doesnt-show-the-data-turbo-confirm-alert-messages-dont-f

janko:
I don't see that the ability to change `Turbo.setConfirmMethod` is documented on the official website. It would be great to have that, for others to more easily discover this feature 😉

PR #382, opened 2021-09-07
Replace FormInterceptor with FormSubmitObserver

seanpdoyle:
Unify `<form>` element submission event listening by replacing all
`FormInterceptor` call sites and `FormInterceptorDelegate` implementers,
replacing them with the `FormSubmitObserver` pattern.


dhh:
Worth rebasing too for 7.2.0 👍

PR #383, opened 2021-09-07
Support navigating a Frame to its current `src`

seanpdoyle:
Rename `hasBeenLoaded` to `willEagerLoad`
===

First, rename `hasBeenLoaded` to `willEagerLoad`, and flip the default
from `false` to `true`.

Next, set the flag to `false` if the `FrameController` is initialized
with a `loading="lazy"` frame. Finally, set the flag to `true` once the
element appears in the viewport for the first time.

Support navigating a Frame to its current `src`
===

Re-opens [#248][]
Closes [#245][]

Update the `FrameController.currentURL` from within `set sourceURL()` to
ensure that `currentURL` does not still refer to the previous load's
`sourceURL`.

[#245]: https://github.com/hotwired/turbo/issues/245
[#248]: https://github.com/hotwired/turbo/pull/248

Replace `reloadable` attribute with `currentURL` property
===

Follow-up to [#263][]

As an alternative to leaking internal `FrameController` state to the DOM
as the `[reloadable]` attribute, this commit replaces those assignments
and guard clause checks with references to `currentURL` instead.

When responding to `<a>` clicks and `<form>` submissions, reset the
`<turbo-frame>` element delegates `currentURL = null`.

Adding a "public" (that is, package-internal) accessor to the
`FrameElementDelegate` feels less committal than an end-user facing
`[reloadable]` attribute, and achieves the same outcome.

[#263]: https://github.com/hotwired/turbo/pull/263

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -19,7 +19,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   readonly appearanceObserver: AppearanceObserver
   readonly linkInterceptor: LinkInterceptor
   readonly formInterceptor: FormInterceptor
-  currentURL?: string | null
+  currentURL: string | null = null
   formSubmission?: FormSubmission
   private currentFetchRequest: FetchRequest | null = null
   private resolveVisitPromise = () => {}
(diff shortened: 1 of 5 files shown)


seanpdoyle:
Closing in favor of https://github.com/hotwired/turbo/pull/487.

PR #384, opened 2021-09-07
Add disable-with Behavior to Turbo

jacobdaddario:
Following some discussion in the GoRails Discord a while back, it looks like there's some popular sentiment that the old Rails UJS behaviors for `confirm` and `disable-with` should be rolled into Turbo. Seeing that #379 implements the `confirm` behavior, I wanted to get the ball rolling on this.

As of right now, the behavior works in the most baseline fashion possible. On a form submission, the submit tag is disabled. Unfortunately, I'm unfamiliar with Intern, so I had some trouble getting the test to work correctly. For some reason I couldn't seem to query for the button fast enough before the page changed.

Additionally, as I see it right now, there's 3 major items that still need to be addressed:

1. Getting the buttons to be un-disabled when the form is a permanent element on the page.
2. Getting the buttons to be un-disabled when the form is targeting a different Turbo frame.
3. Making sure that the buttons are not disabled when navigating using the back button after a full redirect.

I have an idea of how to handle the first two using the `FormSubmissionDelegate` interface, but I'm not sure about the 3rd. I think it involves the way history is tracked.

So if anyone can offer some insight on getting tests working and on how we'd make this work with how Turbo manages history, it'd be a huge help.

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -27,15 +27,15 @@ export enum FormSubmissionState {
 
 enum FormEnctype {
   urlEncoded = "application/x-www-form-urlencoded",
-  multipart  = "multipart/form-data",
-  plain      = "text/plain"
+  multipart = "multipart/form-data",
+  plain = "text/plain"
 }
 
 function formEnctypeFromString(encoding: string): FormEnctype {
-  switch(encoding.toLowerCase()) {
+  switch (encoding.toLowerCase()) {
     case FormEnctype.multipart: return FormEnctype.multipart
-    case FormEnctype.plain:     return FormEnctype.plain
-    default:                    return FormEnctype.urlEncoded
+    case FormEnctype.plain: return FormEnctype.plain
+    default: return FormEnctype.urlEncoded
   }
 }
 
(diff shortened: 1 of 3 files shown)


dhh:
Thanks for working on this! I do like to see this come to Turbo. Does feel like a natural place.

Can you cleanup the PR so it doesn't have style changes? Then it's easier to review.

jacobdaddario:
Of course! This is still very much WIP, just wanted to bookmark this to see if anyone else had ideas. I’ll likely have time to clean this up this weekend.

Do you see styles changing on the element or are you referring to code style?

dhh:
No sweat. I just meant the style changes that are showing up in the diff unrelated to your changes:

Screen Shot 2021-09-08 at 15 43 18

seanpdoyle:
Given the need for this behavior, and Given that #384 relies on a new attribute (`data-turbo-disable-with`) and won't be a drop in replacement of `data-disable-with`, could it be worth considering an alternative like https://github.com/hotwired/turbo/pull/386 instead of re-implementing the Rails UJS behavior?

dhh:
Yeah, I really like the simplicity of just always disabling the submitter until the request is finished. Should just be standard. We even moved away from using extra submission wording in HEY as well. Just wasn't required. I dig that, @seanpdoyle 👍

dhh:
Let's proceed with #386. Thanks for kicking this off @jacobdaddario 🙏

PR #386, opened 2021-09-10
Toggle [disabled] on form submitter

seanpdoyle:
During a form submission, toggle the [disabled][] attribute on prior to
`turbo:submit-start`, and remove it prior to firing `turbo:submit-end`.

If callers need to control the submitter's [disabled][] attribute more
finely, they can declare listeners to do so.

Combined with CSS rules, consumer applications can hide and show
submission text similar to RailsUJS's support for `data-disable-with`:

```css
button                  .show-when-disabled { display: none; }
button[disabled]        .show-when-disabled { display: initial; }

button                  .show-when-enabled { display: initial; }
button[disabled]        .show-when-enabled { display: none; }
```

```html
<button>
  <span class="show-when-enabled">Submit</span>
  <span class="show-when-disabled">Submitting...</span>
</button>
```

Styling descendant submitters based on their ancestors is only possible
with `<button>` elements, since `<input type="submit">` do not have text
content or descendants of their own.

[disabled]: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -145,6 +145,7 @@ export class FormSubmission {
 
   requestStarted(request: FetchRequest) {
     this.state = FormSubmissionState.waiting
+    this.submitter?.setAttribute("disabled", "")
     dispatch("turbo:submit-start", { target: this.formElement, detail: { formSubmission: this } })
     this.delegate.formSubmissionStarted(this)
   }
@@ -178,6 +179,7 @@ export class FormSubmission {
 
   requestFinished(request: FetchRequest) {
     this.state = FormSubmissionState.stopped
+    this.submitter?.removeAttribute("disabled")
     dispatch("turbo:submit-end", { target: this.formElement, detail: { formSubmission: this, ...this.result }})
     this.delegate.formSubmissionFinished(this)
   }
--- a/src/tests/functional/form_submission_tests.ts
+++ b/src/tests/functional/form_submission_tests.ts
@@ -86,6 +86,13 @@ export class FormSubmissionTests extends TurboDriveTestCase {
     await this.nextEventNamed("turbo:load")
   }
 
+  async "test standard POST form submission toggles submitter [disabled] attribute"() {
+    await this.clickSelector("#standard-post-form-submit")
+
+    this.assert.equal(await this.nextAttributeMutationNamed("standard-post-form-submit", "disabled"), "", "sets [disabled] on the submitter")
+    this.assert.equal(await this.nextAttributeMutationNamed("standard-post-form-submit", "disabled"), null, "removes [disabled] from the submitter")
+  }
+
   async "test standard GET form submission"() {
     await this.clickSelector("#standard form.greeting input[type=submit]")
     await this.nextBody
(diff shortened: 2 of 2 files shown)


jacobdaddario:
This is so slick. I don’t know why I thought I couldn’t use the request callbacks to achieve this. The use of CSS selectors to handle the disable behaviors is super cool too.

That said, I wonder if it’s worth trying to implement the behavior for `submit` given that [submit inputs seems to be considered semantic for the Rails form builder.](https://guides.rubyonrails.org/form_helpers.html#dealing-with-basic-forms) `submit` is useful for I18n as well thanks to Rails’ built-in localization of helpers.

Also there’s the question of if we want to continue the hack of disable working on links. I figure no since Hotwire focuses on progressive enhancement, but I wanted to throw it out there.

pinzonjulian:
The `SubmitEven.submitter` discussion is very interesting

That makes me think that it may be out of Turbo's reach/responsibility to do something too specific about it. I personally think that disabling a button in a `busy` form is super important for all apps and Hotwire should probably provide a solution but I starting to gear more towards leaving that responsibility to Stimulus. I just added this functionality to an app and I liked having that control; from toggling styles the way I like to being responsible for accessibility. In my case, styles are tied to `aria-disabled` so this just works:

```js

  static targets = ["submitter"]

  disableSubmitter(event){
    // Although this should be the correct implementation,
    // there is a bug in safari at the moment that doesn't allow
    // the submitter to be located this way.
    // In the future it's likely that Turbo will have this functionality baked
    // in so we won't have to do it manually.
    // https://github.com/hotwired/turbo-rails/pull/257
    // event.detail.formSubmission.submitter.setAttribute("disabled", "")
    this.submitterTarget.setAttribute("aria-disabled", "true")
  }

  enableSubmitter(event){
    this.submitterTarget.setAttribute("aria-disabled", "false")
  }
}

```

```erb
<%= form_with model: @user,  data: { controller: "form", action: 'turbo:submit-start->form#disableSubmitter turbo:submit-end->form#enableSubmitter') } do |form| %>
 <%= form.button data: { form_target: "submitter" } %>
<%= end %>
```

What I personally don't like about this approach is that it's very verbose. I could probably cut a couple of characters by changing `disableSubmitter` to  `toggleSubmitter` and having it set the listener for `submit-end`?. 

Also, maybe I'm doing something wrong but I'd like for the actions to be declared on the button rather than the form. Is that possible? 

I think having an answer for questions like these in the context of Stimulus might be worth exploring to understand if this functionality is better of over there. I still love the idea of this being part of a Standard Library.

pinzonjulian:
I also found something interesting while working with Activestorage direct uploads yesterday. It also sets the `disabled` attribute upon form submission while it waits for the blob ids to return back from the server and upon success it doesn't remove the disabled attribute. 

Maybe it's something we could think of holistically(?)

dhh:
Let's start with `disabled`, so we stay congruent with ActiveStorage as well. Then we can explore options for `aria-disabled` separately.

tleish:
The advantage of using `disabled` is it automatically prevents users who like to double-click buttons from submitting the form multiple times.  When using `aria-disabled`, you must also add additional javascript to check the state of `aria-disabled` to prevent double clicking?  If considering `aria-disabled` in the future, consider how to make it easy to also prevent double clicking.

seanpdoyle:
@tleish that's a good point.

Turbo already declares [FormSubmitObserver](https://github.com/hotwired/turbo/blob/b419c0de0b82025b1c7ba644efcb12c57c23d05d/src/observers/form_submit_observer.ts#L34-L46) and [FormInterceptor](https://github.com/hotwired/turbo/blob/main/src/core/frames/form_interceptor.ts#L24-L34) classes that listen for `submit` events.

If `[aria-disabled="true"]` were toggled, those listeners could be extended to ignore or cancel submissions where `event.submitter.getAttribute("aria-disabled") == "true"`.

PR #393, opened 2021-09-15
Add test for Turbo reloading page when a new tracked element is introduced.

nashby:
Is it expected behaviour that when a new tracked script element is introduced (notice that already tracked `<script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>` is not changed) turbo does page reload or it a bug?


dhh:
That's a bug.

Issue #257, opened 2021-04-23
Ability to override frame-target from server response


acetinick:
Currently if the response from turbo requests returns a redirect response whilst inside a frame, there is no way server side to force a redirect to target="_top", or conditionally control what turbo to do with the response.

There are use cases especailly for modal windows where upon succesful save we would want to perform a redirect back to the index page,  where as 422 would do normal turbo frame replacement within current frame.

would be nice to support something like this.  Which will allow us to control targets from serverside frame targets by setting frame target via the http response headers.  

 if @record.save
  redirect_to settings_locations_url, target: "_top"
else
  render partial: 'form', status: :unprocessable_entity
end

What everyones thoughts are on this? or is there something I am missing to make this easier.

sukei:
I got some use cases where controlling that behavior from the server would be useful. Session expiration is one of them.

WriterZephos:
I like this idea and think it is inline with the programming model turbo is built for, which is letting the server do the work.

saltysealion:
How is this problem currently being solved? I'm at this exact situation and not sure what I can do to work around it.

dhh:
I'd be happy to see something like this.

seanpdoyle:
Unfortunately, sending a response with `Turbo-Frame: _top` is incompatible with the browser built-in `fetch` API.

A fetch `Response` resulting in a redirect _deliberately prevents_ access to the intermediate  redirect response with a status in the `300...399` range.

According to the [2.2.6. Responses](https://fetch.spec.whatwg.org/#responses) section of the specification:

> Except for the last <a href="https://url.spec.whatwg.org/#concept-url">URL</a>, if any, a <a href="https://fetch.spec.whatwg.org#concept-response">response</a>’s <a href="https://fetch.spec.whatwg.org#concept-response-url-list">URL list</a> is not exposed to script as that would violate <a href="https://fetch.spec.whatwg.org#atomic-http-redirect-handling">atomic HTTP redirect handling</a>.

The [Atomic HTTP redirect handling](https://fetch.spec.whatwg.org/#atomic-http-redirect-handling) section of the specification states:

[...]

PR #397, opened 2021-09-17
Override Frame response target from server

seanpdoyle:
https://github.com/hotwired/turbo/issues/257

---

While it's still unclear how servers will persist a `Turbo-Frame:`
header override across destructive actions and the `GET` requests that
result from their `303 See Other` responses, the client side concern is
more straightforward:

  Whenever a response with a `Turbo-Frame: _top` header is handled by a
  `FrameController`, propose a full-page Visit.

This commit adds test coverage for both `<a>` element initiated `GET`
requests, as well as `<form>` element initiated `POST` requests.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -172,8 +172,16 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   async requestSucceededWithResponse(request: FetchRequest, response: FetchResponse) {
-    await this.loadResponse(response)
-    this.resolveVisitPromise()
+    const { location, redirected, statusCode } = response
+
+    if (redirected && response.header("Turbo-Frame") == "_top") {
+      const responseHTML = await response.responseHTML
+
+      session.visit(location, { response: { statusCode, redirected, responseHTML } })
+    } else {
+      await this.loadResponse(response)
+      this.resolveVisitPromise()
+    }
   }
 
   requestFailedWithResponse(request: FetchRequest, response: FetchResponse) {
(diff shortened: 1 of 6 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -175,8 +175,12 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   async requestSucceededWithResponse(request: FetchRequest, response: FetchResponse) {
-    await this.loadResponse(response)
-    this.resolveVisitPromise()
+    if (response.redirected && response.header("Turbo-Frame") == "_top") {
+      session.visit(response.location)
```
mrhead, line 179: Please correct me if I'm wrong, but from my understanding, fetch automatically follows all redirects and gives us the latest GET response. Once Turbo detects the `Turbo-Frame` header in the latest response, it will visit the same URL again. This leads to duplicate requests, and it breaks Rails' flash messages.

So, instead of revisiting the latest URL, we should replace the whole page with the provided content.

seanpdoyle: That's a good point. In this case, we'd want to load the contents _and still_ to translate the `<turbo-frame>`-level navigation into a page-wide `Visit`. I think we want to call `session.visit(response.location, { response: { statusCode: response.statusCode, redirected: response.redirected, responseHTML: ... })` so that the `Visit` is in the session-level history, but does not result in an additional HTTP request.

davidjr82:
@seanpdoyle shouldn't it be better to do it always there is no matching frame in the response, instead of doing it only when a header is set in the response? Just following the comment in this issue:

https://github.com/hotwired/turbo/issues/257#issuecomment-1188397132

anyway, this is a good workaround because allow us to have a solution for a common case, thanks!

seanpdoyle:
I'm closing this for now in favor of exploring other solutions like https://github.com/hotwired/turbo/pull/445 and https://github.com/hotwired/turbo-rails/pull/367.

Issue #361, opened 2021-08-30
Update URL search params after a form submits


kapantzak:
Consider that you have some filters inside a form that fetches the content of a turbo-frame element on submit. Wouldn't that be nice if it automatically updated the current URL search params with the respective field names and values in order to match the current state of the ui?

Of course, this behavior should have to be optional and can be enabled by adding a specific data-attribute (e.g. data-turbo-url-base) to the respective form element.

Also, we might want this to happen only for forms that have the GET method.

**Example:**
```
<form action="/path/to/action"
  method="get"
  data-turbo-frame="frame_1"
  data-turbo-url-base="_current">

  <label for="name">Name</label>
  <input id="name" type="text" name="name">

  <label for="name">Age</label>
  <input id="age" type="number" name="age">

  <input type="submit" value="Submit">
</form>
```

Submitting this form after filling the fields with the following values
```
name: 'John'
age: 39
```
would update the URL search params like so:
```
// Previous
https://www.example.com/listing

// Next
https://www.example.com/listing?name=John&age=39
```

**data-turbo-url-base possible values**:
- `_form`: The form's action (default)
- `_current`: The current location
- `/custom/url/path/`: Provide any custom path

If you think that this would be a useful feature I can start implementing asap.

PR #398, opened 2021-09-17
Push history state from frame navigations

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/50
Closes https://github.com/hotwired/turbo/issues/361
Closes https://github.com/hotwired/turbo/pull/167

---

Extend of built-in support for `<a>` elements with [data-turbo-action][]
(with `"replace"` or `"advance"`) to also encompass `<turbo-frame>`
navigations.

Account for the combination of of `[data-turbo-frame]` and
`[data-turbo-action]` to navigate the target `<turbo-frame>` _and_
navigate the page's history push state, supporting:

* `turbo-frame[data-turbo-action="..."]`
* `turbo-frame a[data-turbo-action="..."]`
* `a[data-turbo-frame="..."][data-turbo-action="..."]`
* `form[data-turbo-frame="..."][data-turbo-action="..."]`
* `form[data-turbo-frame="..."] button[data-turbo-action="..."]`
* `form button[data-turbo-frame="..."][data-turbo-action="..."]`

Whenever a Turbo Frame response is loaded that was initiated from one of
those submitters, forms, anchors, or turbo-frames annotated with a
`[data-turbo-action]`, the subsequent firing `turbo:frame-render` event
will create a `Visit` instance that will skip rendering, won't result in
a network request, and will instead only update the snapshot cache and
history.

[data-turbo-action]: https://turbo.hotwired.dev/handbook/drive#application-visits

Add `linkClickIntercepted` to FrameElementDelegate
===

Expand the `FrameElementDelegate` interface to include a
`linkClickIntercepted` to match its existing
`formSubmissionIntercepted`, then replace a manual `setAttribute` and
`src` assignment with a delegation to the `FrameElementDelegate`
instance.

Extract `getAttribute` utility function
===

For cases where we need to find an attribute value from a collection of
elements, use `getAttribute` instead of a long chain of `||` and `?`
operators.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -3,6 +3,7 @@ import { FetchMethod } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { FormSubmission } from "./form_submission"
 import { expandURL, getAnchor, getRequestURL, Locatable, locationIsVisitable } from "../url"
+import { getAttribute } from "../../util"
 import { Visit, VisitDelegate, VisitOptions } from "./visit"
 import { PageSnapshot } from "./page_snapshot"
 
@@ -158,7 +159,7 @@ export class Navigator {
 
   getActionForFormSubmission(formSubmission: FormSubmission): Action {
     const { formElement, submitter } = formSubmission
-    const action = submitter?.getAttribute("data-turbo-action") || formElement.getAttribute("data-turbo-action")
+    const action = getAttribute("data-turbo-action", submitter, formElement)
     return isAction(action) ? action : "advance"
   }
 }
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -39,6 +39,7 @@ export enum VisitState {
 export type VisitOptions = {
   action: Action,
   historyChanged: boolean,
+  willRender: boolean
   referrer?: URL,
   snapshotHTML?: string,
   response?: VisitResponse
@@ -46,7 +47,8 @@ export type VisitOptions = {
 
 const defaultOptions: VisitOptions = {
   action: "advance",
-  historyChanged: false
+  historyChanged: false,
+  willRender: true
 }
 
 export type VisitResponse = {
@@ -69,6 +71,7 @@ export class Visit implements FetchRequestDelegate {
   readonly referrer?: URL
   readonly timingMetrics: TimingMetrics = {}
 
+  willRender: boolean
   followedRedirect = false
   frame?: number
   historyChanged = false
@@ -87,7 +90,8 @@ export class Visit implements FetchRequestDelegate {
     this.location = location
     this.restorationIdentifier = restorationIdentifier || uuid()
 
-    const { action, historyChanged, referrer, snapshotHTML, response } = { ...defaultOptions, ...options }
+    const { action, historyChanged, referrer, snapshotHTML, response, willRender } = { ...defaultOptions, ...options }
+    this.willRender = willRender
     this.action = action
     this.historyChanged = historyChanged
     this.referrer = referrer
@@ -201,7 +205,7 @@ export class Visit implements FetchRequestDelegate {
   }
 
   loadResponse() {
-    if (this.response) {
+    if (this.response && this.willRender) {
       const { statusCode, responseHTML } = this.response
       this.render(async () => {
         this.cacheSnapshot()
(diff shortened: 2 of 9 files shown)


mhw:
I'd like to use this, but in my case I'm navigating the frame by setting the `src` attribute from a Stimulus controller. Do you think this should/could be included?

seanpdoyle:
> I'd like to use this, but in my case I'm navigating the frame by setting the src attribute from a Stimulus controller. Do you think this should/could be included?

@mhw `<a>` clicks and `<form method="get">` submissions both result in modifying the frame's `src` attribute, which ends up using the same code paths as something like `FrameElement.src =` or `frame.setAttribute("src", ...)`. Does that cover your use case?

mhw:
Ah yes, yes it does. I can put the URLs in hidden links with the right attributes and `click()` those from JavaScript. I'm having an issue with it not working in some cases, but I think that's going to be due to something in my code. It looks to be working well though.

tleish:
FYI, one developer suggested the turbo should have the following convention to navigate frames using js (I thought was a good idea).  

```javascript
Turbo.visit(location, { frame: 'frameId' })
```

Something like this might solve this issue, and provide a new way to navigate frames:
see: https://discuss.hotwired.dev/t/navigate-with-js/3215/5

mhw:
Hmm; this is working well for me in the case where the link is within a `turbo-frame` element, but I found it wasn't working when the link is outside of any `turbo-frame`. Part of my HTML looks roughly like this:

```
<details>
<summary>
<span>Label</span>
<a data-turbo-frame="frame" data-turbo-action="replace" href="/frame/frame.html" hidden="">scalar Url</a>
</summary>
<turbo-frame id="frame"></turbo-frame>
</details>
```

When the details element is opened a Stimulus controller activates the link in the `summary` element with `click()` and the frame contents load, but the URL isn't updated.

I've dug into the Turbo implementation and found that in this case the link clicks are being captured by the `LinkInterceptor` attached to the `html` element, rather than one attached to a `turbo-frame` element. As a result they are handed off to `FrameRedirector` instead of `FrameRenderer`. But I'm not sure if there should be further events triggered following this that lead to the changes in this PR.

I then looked at the functional tests you've added to see when I'd missed, and just got baffled further. I can see that the `test navigating frame with a[data-turbo-action=advance] pushes URL state` test passes when run by intern, but if I run `yarn run start`, browse to `http://localhost:9000/src/tests/fixtures/frames.html`, and then click the `Navigate #frame with data-turbo-action="advance"` link, then the `turbo-frame` navigates but the URL does not update.

I expect the issue is at my end somewhere, but any suggestions would be gratefully received.

mhw:
That's great - with those changes it's working nicely for me now. Good spot on the duplicated `id` in the test fixture: I thought I'd looked everywhere, but missed that.

I tried out your suggestion of using the `loading="lazy"` attribute, but the code path that uses when the frame becomes visible (`elementAppearedInViewport` -> `loadSourceURL` -> `visit`) doesn't hit the changes in this PR and so it doesn't update the URL.

That said, I'm not sure `loading="lazy"` _should_ be in scope of this PR. It feels different somehow from updating the URL in response to a form submission or link click.

seanpdoyle:
> That's great - with those changes it's working nicely for me now. Good spot on the duplicated `id` in the test fixture: I thought I'd looked everywhere, but missed that.

I'm happy to hear it suits the needs of your use case!

> I tried out your suggestion of using the `loading="lazy"` attribute, but the code path that uses when the frame becomes visible (`elementAppearedInViewport` -> `loadSourceURL` -> `visit`) doesn't hit the changes in this PR and so it doesn't update the URL.

Did you also add the `[data-turbo-action="replace"]` attribute to the lazily-loaded frame? I was hoping that all frame loading code paths traversed the portion that checks for `data-turbo-action`.

mhw:
> Did you also add the `data-turbo-action="replace"` attribute to the lazily-loaded frame? I was hoping that all frame loading code paths traversed the portion that checks for `data-turbo-action`.

I did: here's the `turbo-frame` element I've tried:

```
<turbo-frame loading="lazy" data-turbo-action="replace" id="frame_57" src="/frame/57">
...
</turbo-frame>
```

dhh:
Good to proceed with this @seanpdoyle, if you want to square the merge conflict.

javierav:
@seanpdoyle how do you cover this case described in #167:

> 
> You can also manually override the URL to whatever value you may need:
> 
> ```
> <turbo-frame id="history">
>  <a href="/messages" data-turbo-history-url="/special-messages">Show messages</a>
> </turbo-frame>
> ```

In my case, I want to refresh the content of the frame with a different URL than the one that should be promoted to the navigation bar.

Issue #396, opened 2021-09-16
nested frames with a form will not redirect


jcoyne:
Given I have a template that looks like this:
```erb
<%= turbo_frame_tag 'modal_thing' do %>
  <%= turbo_frame_tag 'create_tag', target: '_top' do %>
    <%= form_with url: '/bags/1', method: 'patch' do |form| %>
      <input name="title">
      <button type="submit">go</button>
    <% end %>
  <% end %>
<% end %>
```

And my server responds with a 302 redirect.

Then the page does not redirect on form submission and I see this error in the javascript console:
```
Response has no matching <turbo-frame id="modal_thing"> element turbo.es2017-esm.js:5181
    _callee25$ turbo.es2017-esm.js:5181
    Babel 8
    _callee22$ turbo.es2017-esm.js:4906
    Babel 10
    formSubmissionSucceededWithResponse turbo.es2017-esm.js:5065
    requestSucceededWithResponse turbo.es2017-esm.js:1048
    _callee2$ turbo.es2017-esm.js:616
    Babel 8
    _callee$ turbo.es2017-esm.js:559
    Babel 12
    _callee4$ turbo.es2017-esm.js:972
    Babel 8
    formSubmissionIntercepted turbo.es2017-esm.js:4992
    submitBubbled turbo.es2017-esm.js:1214
    (Async: EventListener.handleEvent)
    start turbo.es2017-esm.js:1226
    connect turbo.es2017-esm.js:4785
    connectedCallback turbo.es2017-esm.js:143
    (Async: LifecycleConnectedCallback)
    js turbo.es2017-esm.js:5621
    Webpack 9

```

If I remove the outer containing turbo-frame, then it works as expected.

boardfish:
I've been having exactly the same problem. I've tried applying `target="_top"` to the form, the inner frame, and the outer frame.

If `target: '_top'` is applied to the inner frame and/or the form, the error refers to the outer frame as in the OP.

If `target: '_top'` is applied to both frames and the form, the error reads `Response has no matching <turbo-frame id="\39 modal_thing"> element`.

boardfish:
In my case, the nesting of the frames is necessary because the route with the form is loaded using the `src` attribute, and the form itself is in a Turbo frame on its own page so that it can be rerendered in isolation.

boardfish:
This Stimulus controller I wrote *should* work as a stopgap in the meantime:

```js
/* form_controller.js */
import { Controller } from "stimulus"
[...]

PR #399, opened 2021-09-18
Fix form in nested turbo-frame with target=_top

tleish:
#396 

The class LinkInterceptor checks if the target is a valid Element and if the closest turbo-frame matches the current turbo-frame. 

```javascript
    class LinkInterceptor {
        constructor(delegate, element) {
          //...
          if (this.respondsToEventTarget(event.target)) 

          respondsToEventTarget(target) {
              const element = target instanceof Element
                  ? target
                  : target instanceof Node
                      ? target.parentElement
                      : null;
              return element && element.closest("turbo-frame, html") == this.element;
          }
```

The FormInterceptor checks if the target is a HTMLFormElement, but does not check if the closest turbo-frame matches the current turbo-frame.

```javascript
    class FormInterceptor {
        constructor(delegate, element) {
            //...
                if (event.target instanceof HTMLFormElement) {
```

As a result, a nested turbo-frame will bubble up to the parent turbo-frame when the turbo-frame is not disabled or the target == "_top".  The parent turbo-frame should not be handling form events nested in a child turbo-frame.

### DIFF
--- a/src/core/frames/form_interceptor.ts
+++ b/src/core/frames/form_interceptor.ts
@@ -21,8 +21,8 @@ export class FormInterceptor {
   }
 
   submitBubbled = <EventListener>((event: SubmitEvent) => {
-    if (event.target instanceof HTMLFormElement) {
-      const form = event.target
+    const form = event.target
+    if (form instanceof HTMLFormElement && form.closest("turbo-frame, html") == this.element) {
       const submitter = event.submitter || undefined
       if (this.delegate.shouldInterceptFormSubmission(form, submitter)) {
         event.preventDefault()
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -75,7 +75,18 @@ export class FrameTests extends TurboDriveTestCase {
     this.assert.notOk(await this.hasSelector("#nested-child"))
   }
 
-  async "test following a form within a nested frame with target top"() {
+  async "test following a form within a nested frame with form target top"() {
+    await this.clickSelector("#nested-child-navigate-form-top-submit")
+    await this.nextBeat
+
+    const frameText = await this.querySelector("body > h1")
+    this.assert.equal(await frameText.getVisibleText(), "One")
+    this.assert.notOk(await this.hasSelector("#frame"))
+    this.assert.notOk(await this.hasSelector("#nested-root"))
+    this.assert.notOk(await this.hasSelector("#nested-child"))
+  }
+
+  async "test following a form within a nested frame with child frame target top"() {
     await this.clickSelector("#nested-child-navigate-top-submit")
     await this.nextBeat
 
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -75,6 +75,17 @@ export class FrameTests extends TurboDriveTestCase {
     this.assert.notOk(await this.hasSelector("#nested-child"))
   }
 
+  async "test following a form within a nested frame with target top"() {
+    await this.clickSelector("#nested-child-navigate-top-submit")
+    await this.nextBeat
+
+    const frameText = await this.querySelector("body > h1")
+    this.assert.equal(await frameText.getVisibleText(), "One")
+    this.assert.notOk(await this.hasSelector("#frame"))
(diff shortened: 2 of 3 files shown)


tleish:
@seanpdoyle - check my logic on this.

seanpdoyle:
Would https://github.com/hotwired/turbo/pull/382 cover this bug?

src/core/frames/form_interceptor.ts
```diff
@@ -21,8 +21,8 @@ export class FormInterceptor {
   }
 
   submitBubbled = <EventListener>((event: SubmitEvent) => {
-    if (event.target instanceof HTMLFormElement) {
-      const form = event.target
+    const form = event.target
+    if (form instanceof HTMLFormElement && form.closest("turbo-frame, html") == this.element) {
```
seanpdoyle, line 25: What is the significance of the second `html` selector?

Also, if the frame that the `<form>` is nested within matches `turbo-frame[disabled]`, wouldn't we want events to bubble past it to the outer frame?

seanpdoyle: According to the commit message, the `element.closest("turbo-frame, html") == this.element` is lifted directly from the `LinkInterceptor`. The `[disabled]` checks happen within the delegates' `shouldInterceptLinkClick` and `shouldInterceptFormSubmission`, which on further consideration feels more aligned with the delegator pattern than leaking that concept up into the observer layer.

tleish: > What is the significance of the second html selector?

I had the same question about the logic in [link_interceptor.ts#L58](https://github.com/hotwired/turbo/blob/c54b443d26b525644b60ef3a04b47dba0c72ef6f/src/core/frames/link_interceptor.ts#L58), but figured I'd follow the same pattern to be safe.

> Also, if the frame that the <form> is nested within matches turbo-frame[disabled], wouldn't we want events to bubble past it to the outer frame?

I also wondered this, and do not know the answer.  Under what circumstances are a frame disabled?

I saw you were the last one that worked with this code, which is why I pinged you for review.

> The [disabled] checks happen within the delegates' shouldInterceptLinkClick and shouldInterceptFormSubmission, which on further consideration feels more aligned with the delegator pattern than leaking that concept up into the observer layer.

I originally placed the code in `shouldInterceptFormSubmission`, but then went this direction after seeing how LinkInterceptor handled it.  

```diff
private shouldInterceptNavigation(element: Element, submitter?: Element) {
  const id = submitter?.getAttribute("data-turbo-frame") || element.getAttribute("data-turbo-frame") || this.element.getAttribute("target")

- if (!this.enabled || id == "_top") {
+ if (!this.enabled || id == "_top" || this.element != element.closest("turbo-frame") || id == "_top") {
  }
    return false
  }
``` 

It seemed odd to me to handle it one way for links and another for forms when the overall concept is very similar.  Almost seems like there's some shared code that should exist here.  

FYI, I've seen this happen a couple of times with turbo, where updates are made to link logic, but not forms or the other way around.

src/tests/functional/frame_tests.ts
```diff
@@ -75,6 +75,17 @@ export class FrameTests extends TurboDriveTestCase {
     this.assert.notOk(await this.hasSelector("#nested-child"))
   }
 
+  async "test following a form within a nested frame with target top"() {
```
boardfish, line 78: A test for a `form` with `data-turbo-frame="_top"` might also be helpful here: https://turbo.hotwired.dev/reference/frames#frame-with-overwritten-navigation-targets

src/core/frames/form_interceptor.ts
seanpdoyle: I've opened https://github.com/hotwired/turbo/pull/402 and https://github.com/hotwired/turbo/pull/382 to attempt to reduce the surface area of the delegate interfaces, and to attempt to use consistent method naming across similar scenarios.

I don't think they'd address this particular issue entirely, but it's a start.

boardfish: I've got a piece of functionality ready to go that relies on the issue being fixed, so this PR getting merged would be helpful to me. Are those two PRs a prerequisite for this one to go through? I should also ask whether there's anything I can do to help progress this further.

PR #402, opened 2021-09-21
Expand the FrameElementDelegate interface

seanpdoyle:
Expand the `FrameElementDelegate` to extend from both
`FormInterceptorDelegate` and `LinkInterceptorDelegate`.

After that change, replace conditionals in the `FrameRedirector` with
calls to the `FrameElement` instance's delegate when one is found.

This reduces the surface area for duplicate logic and improves
guarantees of consistency across semantically similar situations.


seanpdoyle:
Closing this in favor of https://github.com/hotwired/turbo/pull/430.

PR #403, opened 2021-09-21
Target must be in the document.body to be the event emitter

dhh:
It's possible for this not to be true if the target element has been removed from the dom between when a form submission was made and the response is processed. This can happen as a race condition when the same turbo stream removal is triggered by both websocket and server response.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -4,9 +4,9 @@ export function dispatch(eventName: string, { target, cancelable, detail }: Part
   const event = new CustomEvent(eventName, { cancelable, bubbles: true, detail })
 
   if (target && document.body.contains(target as Element)) {
-    void target.dispatchEvent(event);
+    target.dispatchEvent(event);
   } else {
-    void document.documentElement.dispatchEvent(event);
+    document.documentElement.dispatchEvent(event);
   }
 
   return event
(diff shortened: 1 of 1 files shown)


src/util.ts
```diff
@@ -2,7 +2,13 @@ export type DispatchOptions = { target: EventTarget, cancelable: boolean, detail
 
 export function dispatch(eventName: string, { target, cancelable, detail }: Partial<DispatchOptions> = {}) {
   const event = new CustomEvent(eventName, { cancelable, bubbles: true, detail })
-  void (target || document.documentElement).dispatchEvent(event)
+
+  if (target && document.body.contains(target as Element)) {
```
seanpdoyle, line 6: Would [Node.isConnected](https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected) be an interchangeable check here?

src/util.ts
```diff
@@ -2,7 +2,13 @@ export type DispatchOptions = { target: EventTarget, cancelable: boolean, detail
...
+    void target.dispatchEvent(event);
```
seanpdoyle, line 7: What is the significance of the `void` here? Could it be omitted?

dhh: Good question. Left that as it was, but it did strike me as odd.

src/util.ts
dhh: Good call. Works on shadow doms as well.

src/util.ts
dhh: Couldn't find a purpose. Stripped it 👍

PR #405, opened 2021-09-24
Update SubmitEvent Polyfill to Workaround Safari 15 bug

### LINKED ISSUE #390: TurboDrive fires on forms despite being disabled in Safari 15 beta
Over the last week I've noticed a big uptick in JS exceptions in my app. All of the exceptions were for browsers running the Safari 15 beta. After investigating, I learned that on a form submission, Turbo Drive would fire even if the submit button had the attribute `data-turbo=false`

After further investigation, I discovered the `submitter` property of the `SubmitEvent` was always null. It seems Safari 15 beta now implements the SubmitEvent, but due to a bug in their implementation, they do not populate the `submitter` property.

Since the `submitter` property is `null` TurboDrive cannot inspect its attributes to see if it should delegate control of the form submission back to the browser. 

[This bug was recently patched](https://bugs.webkit.org/show_bug.cgi?id=229660), but it is unclear when it will hit Safari 15 beta users, it's possible it won't for months.

I'm going to continue looking for the most elegant workaround, but I wanted to raise the issue here in case there was an appetite to extend the existing polyfill to officially fix this unfortunate situation.

terracatta:
Fixes #390 

This PR updates the Polyfill to work around the Safari 15 bug where the submitter property is sometimes null when a form is submitted with element of type button vs an element of type submit. More info https://bugs.webkit.org/show_bug.cgi?id=229660

### DIFF
--- a/src/polyfills/submit-event.ts
+++ b/src/polyfills/submit-event.ts
@@ -17,12 +17,23 @@ function clickCaptured(event: Event) {
 }
 
 (function() {
-  if ("SubmitEvent" in window) return
   if ("submitter" in Event.prototype) return
+
+  let prototype;
+  // Certain versions of Safari 15 have a bug where they won't
+  // populate the submitter. This hurts TurboDrive's enable/disable detection.
+  // See https://bugs.webkit.org/show_bug.cgi?id=229660
+  if ("SubmitEvent" in window && /Apple Computer/.test(navigator.vendor)) {
+    prototype = window.SubmitEvent.prototype;
+  } else if ("SubmitEvent" in window) {
+    return; // polyfill not needed
+  } else {
+    prototype = window.Event.prototype;
+  }
 
   addEventListener("click", clickCaptured, true)
 
-  Object.defineProperty(Event.prototype, "submitter", {
+  Object.defineProperty(prototype, "submitter", {
     get(): HTMLElement | undefined {
       if (this.type == "submit" && this.target instanceof HTMLFormElement) {
         return submittersByForm.get(this.target)
--- a/src/globals.d.ts
+++ b/src/globals.d.ts
@@ -9,4 +9,5 @@ interface Node {
 
 interface Window {
   Turbo: typeof import("./core/index")
+  SubmitEvent: typeof Event
 }


terracatta:
@dhh ok I manually tested the broken behavior using Safari 15 with `yarn start` and I can confirm this fixes the issue with the least amount of required changes.

terracatta:
cc @ParamagicDev since you use a similar polyfill for your mrujs project.

ParamagicDev:
Thank you @terracatta , making a note on Mrujs so I can hotfix this.

dhh:
Do we need any additional verification that this didn’t break other browsers handling it? Will this automatically become irrelevant when safari is fixed or does it require a change then?

terracatta:
> Do we need any additional verification that this didn’t break other browsers handling it?

I don't _think_ so. First, the way it's written I only run a new behavior if it's Safari (via testing the navigator.vendor property) that says it supports SubmitEvent. Second, I've been running my ES6 polyfill on a high-traffic app that generates exceptions for console errors, and I am not seeing any regression in browsers that are already working fine.

> Will this automatically become irrelevant when safari is fixed or does it require a change then?

So I download the latest Safari Technical Preview `Release 132 (Safari 15.4, WebKit 16613.1.1.5)`  which according to the release notes, has the underlying bug fixed (which I verified using the test case attached to the original Safari bug report). On this version, the polyfill still runs (even though it doesn't need to) but it doesn't do anything bad like throw errors/warnings or regress the behavior we want. 

I cannot think of a clean way to test if we are on the fixed version of Safari without actually creating a fake form that produces the bug, submitting it via JS, preventing default, and checking the value of submitter. This might be possible, but I'm not sure if the juice is worth the squeeze.

dhh:
Great work. Let’s put a note to revisit this in 3 months or whatever and hope that the version in the wild no longer is a meaningful percentage so we can clean this up 👍

src/polyfills/submit-event.ts
```diff
@@ -17,12 +17,23 @@ function clickCaptured(event: Event) {
...
 
 (function() {
-  if ("SubmitEvent" in window) return
   if ("submitter" in Event.prototype) return
 
+  let prototype;
+  // Certain versions of Safari 15 have a bug where they won't
+  // populate the submitter. This hurts TurboDrive's enable/disable detection.
+  // See https://bugs.webkit.org/show_bug.cgi?id=229660
+  if ("SubmitEvent" in window && /Apple Computer/.test(navigator.vendor)) {
```
ParamagicDev, line 26: Silly question, since Safar iOS is probably similar to desktop Safari, is there a chance this will need to be fixed when iOS 15 ships?

ParamagicDev: Appears navigator.vendor can only be one of:
`"Google Inc."`, `"Apple Computer Inc."`, or `""`

So I think we should be okay (you also tested it as well, thank you!)

PR #406, opened 2021-09-24
Allow Drive to submit a form when spot-enabled but submitter is null 

### LINKED ISSUE #392: Turbo not handling form submission when globally disabled but locally enabled and called with requestSubmit
I have Turbo drive globally disabled in my app.

```js
Turbo.session.drive = false
```

I have a form that is invisible to the user, no submit button, but with `data-turbo=true` set on it. It is used to handle a QR code scan to send the results of the scan back to the server. So upon a QR scan, an input in the form gets set to the value of the scan, and then `form.requestSubmit()` is called. But Turbo does not do anything to intercept this submission. I am trying to do a turbo stream response from the server, but the turbo stream mime type is not getting added to the accept header.

I think I've tracked down the reason for this. When my `form.requestSubmit()` gets called, this should run:

https://github.com/hotwired/turbo/blob/11840c1ba45cc82e59830b23353db140fcd768c5/src/core/session.ts#L190-L192

`form` will be set, but `submitter` will not, because I'm not passing in a submitter in my `form.requestSubmit()` call (primarily because I don't have a submitter, no submit button on this form currently).

`this.elementDriveEnabled(form)` should return true because I have  `data-turbo=true` set on the form. But `this.elementDriveEnabled(submitter)` when submitter is undefined, what will it return?

https://github.com/hotwired/turbo/blob/11840c1ba45cc82e59830b23353db140fcd768c5/src/core/session.ts#L305-L324

First it's closest `[data-turbo]` is found, which will be undefined, because it itself is undefined, and then the answer depends entirely on what `Turbo.session.drive` is set to, which in my case is `false`.

Now, what I've discovered through the process of writing this issue is that `form.requestSubmit()` [should default `submitter` to the form if one is not provided](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit#:~:text=If%20you%20omit%20the%C2%A0submitter%C2%A0parameter%2C%20the%20form%20element%20itself%20is%20used%20as%20the%20submitter.), but that is not the behavior I'm seeing in the latest Chrome, or in the latest Safari. Both of them pass `null` as the `submitter` which gets turned into `undefined` here:

https://github.com/hotwired/turbo/blob/11840c1ba45cc82e59830b23353db140fcd768c5/src/observers/form_submit_observer.ts#L33-L36

If the browsers were behaving like the documentation says that they should, I would not have run into this issue. But, in my opinion, I think it would make sense that if there is no submitter, that the answer of `willSubmitForm` should be based solely on the form.

```ts
willSubmitForm(form: HTMLFormElement, submitter?: HTMLElement): boolean { 
  return this.elementDriveEnabled(form) && (!submitter || this.elementDriveEnabled(submitter))
}

// or more clearly

willSubmitForm(form: HTMLFormElement, submitter?: HTMLElement): boolean {
  if (submitter) {
    return this.elementDriveEnabled(form) && this.elementDriveEnabled(submitter)
  } else {
    return this.elementDriveEnabled(form)
  }
}
```

---

_edit:_

I should just also include the workaround in case anyone else is running into this. I was unable to just pass the form element as the submitter, as Chrome (maybe other browsers would have as well) was raising an error. So I added a hidden submit button and passing that as the submitter in the `requestSubmit()` call.

```js
form.requestSubmit(hiddenSubmitButton)
```

terracatta:
Fixes #392 

When you disable Turbo Drive wholesale with `Turbo.session.drive = false` and spot enable it on a form without a submit button with `data-turbo=true` Turbo Drive still wouldn't fire because it couldn't verify that the submitter (which was null) had a container where turbo was enabled.

This pulls in the fix @willcosgrove recommends as it's wise to never assume the DOM placement and logic for a submitter that is null.

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -189,7 +189,7 @@ export class Session implements FormSubmitObserverDelegate, HistoryDelegate, Lin
   // Form submit observer delegate
 
   willSubmitForm(form: HTMLFormElement, submitter?: HTMLElement): boolean {
-    return this.elementDriveEnabled(form) && this.elementDriveEnabled(submitter)
+    return this.elementDriveEnabled(form) && (!submitter || this.elementDriveEnabled(submitter))
   }
 
   formSubmitted(form: HTMLFormElement, submitter?: HTMLElement) {
--- a/src/tests/fixtures/drive_disabled.html
+++ b/src/tests/fixtures/drive_disabled.html
@@ -5,6 +5,19 @@
     <title>Drive (Disabled by Default)</title>
     <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
     <script src="/src/tests/fixtures/test.js"></script>
+    <script type="module">
+      addEventListener("click", event => {
+        if (event.target.id == "requestSubmit") {
+          event.preventDefault()
+          const form = event.target.closest('form')
+          if(typeof form.requestSubmit === "function"){
+            form.requestSubmit()
+          } else {
+            form.submit()
+          }
+        }
+      })
+    </script>
     <script>
       Turbo.session.drive = false
     </script>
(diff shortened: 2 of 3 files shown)


dhh:
Would be good with a test for this.

terracatta:
> Would be good with a test for this.

I agree. Working on it.

terracatta:
@dhh Ok I have a working test that simulates @willcosgrove's initial situation. Let me know if you need anything else.

dhh:
Boom!

Issue #401, opened 2021-09-20
Turbo Drive treats external URLs redirected from same-origin URLs as same-origin


TastyPi:
According to the Turbo Drive docs https://turbo.hotwired.dev/handbook/drive#setting-a-root-location:

> By default, Turbo Drive only loads URLs with the same origin—i.e. the same protocol, domain name, and port—as the current document. A visit to any other URL falls back to a full page load.

However, I have encountered two scenarios where this is not the case:

1. Form POST request to same-origin returns a 303 redirect response to another domain
2. An anchor with a link to a same-origin URL returns a 303 redirect response to another domain

In both cases, Turbo Drive attempted to load the external URL, which in my particular case causes a CORS error due to cross-origin requests not being allowed by the external page. I've fixed both cases by adding `data-turbo=false` to the `<form>` and `<a>` respectively, but based on the wording of the documentation I would have expected them to both fallback to a full page load.

Is it possible for Turbo Drive to handle redirects in the expected way?

ghiculescu:
Turbo Drive tells its underlying `fetch` request to follow redirects: https://github.com/hotwired/turbo/blob/c9c1c11610f12442a6342200e396c12a30ed957d/src/http/fetch_request.ts#L120

I think the problem is that Turbo can't know ahead of time if the page will redirect to a third party URL. So it has no way of knowing it if should send a `fetch` or not.

https://developer.mozilla.org/en-US/docs/Web/API/fetch suggests there's an alternative option (`manual`) which I suppose could be supported instead, it'd be a fair bit more code to support redirects rather than letting the browser do it for us.

![image](https://user-images.githubusercontent.com/509837/134253866-6dc08933-1241-4cfc-a45b-561c01d514d9.png)

TastyPi:
Makes sense, I figured it would be something to do with automatically follwing redirects. If someone (maybe me, not sure if I have the time) did contribute a pull request which changed the behaviour, would it be likely to be accepted or would it be rejected due to adding too much complexity for a corner case like this? I'm fine with my solution of disabling Turbo if it wouldn't, but it might be worth at least updating the docs to mention this corner case.

ghiculescu:
I think a docs PR is a good start. I’m not a maintainer so can’t answer the other question. My gut feel is this isn’t a very common case, as you say.

mxlje:
I have the same problem. In my case I want to redirect users to an external payment provider during the checkout, and the URL has to be generated on the server.

[...]

PR #410, opened 2021-09-28
fix: allow programmatic visits to go cross-origin

ParamagicDev:
Currently, when calling `Turbo.visit("https://example.com", { action: "advance" })` , the site will not navigate to the `https://example.com` and in addition a number of errors pop up here:

Screen Shot 2021-09-27 at 6 04 09 PM

Most of these errors have to do with trying to fetch cross-origin as well as adding history for a cross-origin url. This PR takes previous art from Turbolinks here:

https://github.com/turbolinks/turbolinks/blob/80cd172ed5b3a35eda85c58e428de9f1f1d06fc8/src/controller.ts#L70-L80

and uses previous art from Turbo's own link delegator here:

https://github.com/hotwired/turbo/blob/5982a679f19f9c745c51e85885e99069cf3d0c4a/src/core/session.ts#L334-L336

Unfortunately, I ran into multiple issues trying to get tests for this due to my chromedriver only supporting up to v92 and chrome itself being v94. I tried manually installing and upgrading both and was unsuccessful. Regardless, I think drawing on prior art and having done some manual testing as well, all appears to be well.

### Video Before:

https://user-images.githubusercontent.com/26425882/135016268-da859358-d0ce-4b75-aa54-e74b3aa3345f.mov

### Video After:

https://user-images.githubusercontent.com/26425882/135016032-0d5be5d0-be6f-455e-80a6-a0b70e9142ae.mov

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -2,7 +2,7 @@ import { Action, isAction } from "../types"
 import { FetchMethod } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { FormSubmission } from "./form_submission"
-import { expandURL, getAnchor, getRequestURL, Locatable, isPrefixedBy, isHTML } from "../url"
+import { expandURL, getAnchor, getRequestURL, Locatable, locationIsVisitable } from "../url"
 import { Visit, VisitDelegate, VisitOptions } from "./visit"
 import { PageSnapshot } from "./page_snapshot"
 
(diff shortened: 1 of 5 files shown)


Matt-Yorkley:
You're a star! :star2: 

Just for clarity; this should close issue https://github.com/hotwired/turbo/issues/401 right? Looks like there's a few folks that would be excited to see it fixed (myself included)

ParamagicDev:
@Matt-Yorkley I believe that issue has to do with when you submit a form and then redirect to an external URL. This will not fix that. This is purely for programmatic visits.

Matt-Yorkley:
> @Matt-Yorkley I believe that issue has to do with when you submit a form and then redirect to an external URL. This will not fix that. This is purely for programmatic visits. IE: `Turbo.visit`

Ah, okay! I assumed if a Stream responded with a redirect to an external URL it would hit the same code path when it came to navigating..? I'll take your word for it though! :ok_hand::smile:

Issue #358, opened 2021-08-27
data-turbo-method new functionality breaking out of frame


acetinick:
Related I believe to this PR #341,  what I am seeing now is that when using data-turbo-method within a turbo frame instead of doing the normal turbo request, will break out of the current frame.  As the form is created dynamically end of the body now it always does a full page refresh, rather than using the frame in context of the parent frame.

I think perhaps solution here is to also find the immediate parent turbo-frame element (if exists) and use the ID of that frame to inject as a data-turbo-frame attribute into the dynamic form created.  That way I pretty sure will do turbo request as expected within frame.

dhh:
Doh. Yes of course. This is a regression. Cc @davidgil.

acetinick:
@dhh perhaps PR #341 should be reverted in another release of hotwire until it gets resolved?

PR #411, opened 2021-09-30
Load data-turbo-method links within the correct turbo frame

spohlenz:
This is a continuation of @davidgil's work in #341, and addresses the problem raised by @acetinick in #358.

In addition it also addresses some other concerns that had been raised:

* cleans up the created forms after they are submitted
* allows `data-turbo-frame` targets on links with `data-turbo-method` set


Intrepidd:
Anything else needed to move this forward ? I'm actually using a forked version of turbo with these changes in production and it's running fantastic.

There are a few conflicts following recent development on `main`, mind rebasnig ?

spohlenz:
I've rebased and verified that the tests are still passing.

PR #412, opened 2021-09-30
Replace LinkInterceptor with LinkClickObserver

seanpdoyle:
Follow-up to https://github.com/hotwired/turbo/pull/382

---

In the same style as [hotwired/turbo#382][], replace instances of
`LinkInterceptor` with `LinkClickObserver`, making the necessary
interface changes in classes that used to extend
`LinkInterceptorDelegate`.

Conditional logic that was once covered in the `LinkInterceptor` is
moved into the predicate methods of the delegates (for example, the
`FrameController.willFollowLinkToLocation` and
`FrameRedirector.willFollowLinkToLocation`).

Since the recently introduced [FormLinkInterceptor][] was named after
the `LinkInterceptor`, this commit also renames that class (and its call
sites) to match the `LinkClickObserver`-suffix, along with the
`LinkInterceptorDelegate` method structure.

[hotwired/turbo#382]: https://github.com/hotwired/turbo/pull/382
[FormLinkInterceptor]: https://github.com/hotwired/turbo/commit/f8a94c5f761c4f080d68a87459d3916eb4fb6993


dhh:
Same re: rebase and 7.2.0.

Issue #413, opened 2021-10-02
Feature Request: declarative connection to a Turbo Stream using the `turbo-stream` custom tag


delitescere:
Currently, some JavaScript is needed to connect a Turbo Stream to an EventSource (SSE) or WebSocket (by the way, the docs aren't very clear about how to do this).

It would be superb if it could be done without JavaScript, rather using an incantation of the `turbo-stream` custom element.

Current imperative mechanism:
```javascript
(window['EventSource'] && window['Turbo']) ?
  Turbo.connectStreamSource(new EventSource('/my-turbo-stream')) :
  console.warn('Turbo Streams over SSE not available');
```

Suggested declarative mechanism:
```html
<turbo-stream src="/my-turbo-stream">
```

If the URL scheme was `ws:` then Turbo would create a new `WebSocket`, otherwise it would be an `EventSource`.

The addition of a `src` attribute to the custom `turbo-stream` tag need not (but may) be conflated with the use of it in directives to altering the DOM as content arrives on the stream or as the result of a form submission (i.e. the existing capabilities of the custom tag).

seanpdoyle:
Could you share a use case that you have in mind to help demonstrate this feature's utility?

There seems to be some confusion between concepts. 

A `<turbo-stream>` represents a single DOM operation, and disappears once it's connected to the page.

The `@hotwired/turbo-rails` package's [`<turbo-cable-stream-source>`](https://github.com/hotwired/turbo-rails/blob/v0.8.1/app/javascript/turbo/cable_stream_source_element.js) is a persistently connected element. Its `[channel]` and `[signed-stream-name]` attributes are forwarded to the Rails application's underlying Action Cable connection. Action Cable does not yet support connecting over `EventSource` instances, or [`ActionController::Live`](https://edgeapi.rubyonrails.org/classes/ActionController/Live.html) integration. My hunch is that the `channel` and `stream-name` attributes are serving the purpose of a `ws://` or `wss://` web socket URL at a different layer of abstraction.

Since the `<turbo-cable-stream-source>` element is declared in `@hotwired/turbo-rails`, that repository might be a better place to open an Issue. If your use case is outside of Rails, I'm interpreting the scope of the Issue as much larger, including the introduction of a backend agnostic version of the `<turbo-cable-stream-source>`.

Am I misinterpreting anything here?

delitescere:
[...]

PR #415, opened 2021-10-03
Turbo stream source

seanpdoyle:
Improve test coverage for streams over SSE
===

Add a `<form>` element to the `src/tests/fixtures/stream.html` file so
that tests can exercise receiving `<turbo-stream>` elements
asynchronously over an `EventSource` instance that polls for Server-sent
Events.

Extend the `src/tests/server.ts` to account for the `Accept:
text/event-stream` requests that are made by browsers via the `new
EventSource(...)` instance.

Introduce `<turbo-stream-source>`
===

Closes https://github.com/hotwired/turbo/issues/413

The `<turbo-stream-source>` element accepts a `[src]` attribute, and
uses that to connect Turbo to poll for streams published on the server
side.

When the element is connected to the document, the stream source is
connected. When the element is disconnected, the stream is disconnected.

When declared with an `ws://` or `wss://` URL, the underlying Stream
Source will be a `WebSocket` instance. Otherwise, the connection is
through an `EventSource`.

Since the document's `<head>` is persistent across navigations, the
`<turbo-stream-source>` is meant to be mounted within the `<body>`
element.

Typical full page navigations driven by Turbo will result in the
`<body>` being discarded and replaced with the resulting document. It's
the server's responsibility to ensure that the element is present on
each page that requires streaming.


delitescere:
Great!

I am not sure about the head / body assertion being made, nor the assumption about full page updates by Turbo (I certainly would never do that).

dhh:
This is great, @seanpdoyle, we just need to pair it up with the existing tag that was put in turbo-rails. Or rather, make turbo-rails use this more generic approach directly.

dhh:
Ah, yes, fair enough. This is mostly then for use outside of Rails. Which is fine.

dhh:
There's a few lint errors, if you want to sort.

PR #416, opened 2021-10-03
Expose FormSubmission and FetchRequest

yuki24:
I'm working to add Hotwire support for [Shoelace](https://shoelace.style/), a Web Component-based design system. One challenge wth integrating custom elements is that the normal `<form>` tag fires a `submit` event with all the form controls automatically captured while form inputs in a shadow DOM would be excluded from it.

Thankfully, both Hotwire and Shoelace are built in a very standardized way, and I was able to get them working well with these classes exposed in Hotwire. It would be great if these two constants could be exposed for reusability.

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -4,12 +4,13 @@ import { Locatable } from "./url"
 import { StreamMessage } from "./streams/stream_message"
 import { StreamSource } from "./types"
 import { VisitOptions } from "./drive/visit"
+import { FormSubmission } from "./drive/form_submission"
 import { PageRenderer } from "./drive/page_renderer"
 import { PageSnapshot } from "./drive/page_snapshot"
 
 const session = new Session
 const { navigator } = session
-export { navigator, session, PageRenderer, PageSnapshot }
+export { navigator, session, PageRenderer, PageSnapshot, FormSubmission }
 
 /**
  * Starts the main session.
--- a/src/index.ts
+++ b/src/index.ts
@@ -8,3 +8,4 @@ window.Turbo = Turbo
 Turbo.start()
 
 export * from "./core"
+export { FetchRequest } from "./http/fetch_request"


seanpdoyle:
At the moment, applications and third-party libraries integrate with Turbo Form Submissions with the browser standardized `submit` event like you'd mentioned. I'm opposed to increasing the surface area of public API, since that increases the tool's maintenance burden, and locks the team and project into maintaining a consistent interface, constructors, etc.

Similarly, the current implementation of `FormSubmission` relies on a `FormSubmissionDelegate` TypeScript interface, which is the entry point to another series of dependent interfaces we'd need to commit to and maintain.

When fired, [`turbo:submit-start` and `turbo:submit-end` events](https://turbo.hotwired.dev/reference/events) expose instances of `FormSubmission` objects as part of the CustomEvent's `event.detail` property. Those instances are already constructed, with a `FormSubmissionDelegate`, are useful for data access, but are otherwise inert.

 My initial reaction is that it'd be less disruptive for the Shoelace form code to synthesize and fire its own [SubmitEvent](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent) instances, and let them bubble up to be handled by Turbo.

@yuki24 if the `FormSubmission` class were available, how would Shoelace invoke it? Could you share example code, pseudo code, or a use case?

yuki24:
@seanpdoyle Thanks for your quick reply. Using the `SubmitEvent` certainly seems like a reasonable approach. One blocker I see on top of my head is how [Turbo constructs the `FormData`](https://github.com/hotwired/turbo/blob/acb003d9879311b7f44fe12499ad333605ad3cdd/src/core/drive/form_submission.ts#L173) for the submission. It is totally reasonable to do this if you use a usual form, but most Web Component libraries have [their own form components](https://shoelace.style/components/form) and thus the `<form>` tag is not used. They do provide [a way to get a `formdata` object](https://shoelace.style/components/form?id=methods) though, so it would be great if we could somehow override [the `buildFormData` function](https://github.com/hotwired/turbo/blob/acb003d9879311b7f44fe12499ad333605ad3cdd/src/core/drive/form_submission.ts#L172).

In case this is helpful. here is how my current implementation works:

1. Capture Shoelace's `sl-submit` event:

    ```ts
    // Call this function in the entry-point file:
    start() {
      if (!this.started) {
        addEventListener("sl-submit", this.submitBubbled.bind(this), false)
        this.started = true
      }
    }
    
    submitBubbled<EventListener>(event: CustomEvent) {
      if (!event.defaultPrevented) {
        const form = event.target instanceof SlForm ? event.target : undefined
    
        if (form) {
          const method = form.getAttribute("method")
    
          if (method != "dialog" && this.delegate.willSubmitForm(form)) {
            this.delegate.formSubmitted(form)
          }
        }
      }
    }
    ```
2. In the `formSubmitted` function, instantiate an `SlFormSubmittion` object so the navigator can submit the request:

    ```ts
    formSubmitted(form: SlForm) {
      const { navigator } = this.session
      const submitter = document.activeElement as HTMLElement
      const formSubmission = new SlFormSubmission(navigator, form, this.fakeFormElement(form), submitter, true)

      navigator.stop()
      navigator.formSubmission = formSubmission

      if (navigator.formSubmission.isIdempotent) {
        navigator.proposeVisit(formSubmission.fetchRequest.url, {
          action: navigator.getActionForFormSubmission(formSubmission),
        })
      } else {
        navigator.formSubmission.start()
      }
    }
    ```

And here is what the `SlFormSubmission` looks like:

```ts
export class SlFormSubmission extends FormSubmission {
  readonly formData: FormData
  readonly fetchRequest: FetchRequest
  readonly slForm: SlForm

  constructor(delegate: FormSubmissionDelegate, formElement: SlForm, fakeFormElement: HTMLFormElement, submitter?: HTMLElement, mustRedirect = false) {
    super(delegate, fakeFormElement, submitter, mustRedirect)

    this.formData = formElement.getFormData()
    this.fetchRequest = new FetchRequest(this, this.method, this.location, this.body, this.formElement)
    this.slForm = formElement
  }
}
```

seanpdoyle:
@yuki24 thank you for sharing snippets of the dependent code. I'm not very familiar with using Custom elements to implement alternatives to built-in elements like `<form>`. Could you also share an example of end-user facing code?

pinzonjulian:
I'm not an expert in Web Components but I do feel like they will become a common practice very soon. 

I recommend you bring @jaredcwhite and @ParamagicDev in to the conversation. They've implemented web components and shoelace specifically and I'm sure they can shed some light into this :)

ParamagicDev:
@yuki24 I've had a couple thoughts about this. The code you linked is fairly specific to Shoelace. I think perhaps an easier alternatives would be to have end users capture the submit event specific to their web component, then re-export the submit even with both an `event.detail.submitter` and `event.detail.formData`

Then in the FormSubmission class, it can then have an optional parameter for `formData` and then it sets `this.formData` like this:

`this.formData = formData ?? buildFormData(form)`

https://github.com/hotwired/turbo/blob/acb003d9879311b7f44fe12499ad333605ad3cdd/src/core/drive/form_submission.ts#L56

I think this allows for it to be extensible without having to have workarounds specific to every WC.

So in theory, the end user should do something like this:

```js
document.addEventListener("sl-submit", (event) => {
  const { formData, submitter } = event.detail
  const event = new CustomEvent("submit", {bubbles: true, detail: { formData, submitter }})
  event.currentTarget.dispatchEvent(event)
})
```

And then Turbo _should_ be able to just handle it. Obviously totally untested and kinda pseudo-code, but thats where my head is at.

seanpdoyle:
Would pairing a synthetic `submit` dispatched from Shoelace with a `turbo:submit-start` suffice?

The `turbo:submit-start` event has a `event.detail.formSubmission.formData` property, which exposes the request's [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData) instance directly. If Shoelace were able to intervene at that point, it could re-write the behind the scenes `FormData` instance with whatever data is available in the `sl-submit`.

yuki24:
@seanpdoyle There are two problems with relying on the `turbo:submit-start` event:

1. Web Component-based design systems may not rely on the `<form>` tag, which means this turbo event would not get fired due to this check: https://github.com/hotwired/turbo/blob/acb003d9879311b7f44fe12499ad333605ad3cdd/src/observers/form_submit_observer.ts#L35-L38
2. Creating a `FetchRequest` object implicitly relies on the `body()` getter, which relies on the formData getter: https://github.com/hotwired/turbo/blob/acb003d9879311b7f44fe12499ad333605ad3cdd/src/core/drive/form_submission.ts#L79 which in turn would require replacing the `this.featchRequest` attribute.

We could "fake" a form element and set it to `event.target`, but that seems hacky and it probably wouldn't work with the `<turbo-frame>` that [checks if the form tag is the direct closest child](https://github.com/hotwired/turbo/blob/acb003d9879311b7f44fe12499ad333605ad3cdd/src/core/frames/form_interceptor.ts#L25).

Here is an example code of Shoelace. As you can see, there is no `<form>` tag we could use as a `event.target`:

```html
<sl-form>
  <!-- Shoelace automatically captures all fomr controls: -->
  <sl-input name="name" type="text" label="Name"></sl-input>

  <!-- Clicking this button triggers a `sl-submit`event: -->
  <sl-button submit>
    Submit
  </sl-button>
</sl-form>
```

yuki24:
I'm closing this PR as I ended up translating a `sl-submit` event to a synthetic `submit` as suggested above. I also ended up capturing a `formdata` event to append input data. This feels like a more reliable way than the `turbo:submit-start` event since it's the browser that fires it (not looking at IE11 for now).

A more standardized approach is definitely preferable, but the Form Participation API is nowhere near production use, so it's probably too early to bring up this topic.

Thanks everyone for all your insights and feedback!

Issue #417, opened 2021-10-04
Setting `TurboFrameElement.src` twice in quick succession can show wrong source


TastyPi:
If `src` is set twice in quick succession and the second source loads before the first source, then the `turbo-frame` will display the first source instead of discarding that request.

My use case is a kind of reactive preview, where I update the src of a `turbo-frame` on each input event of a form. Typing two characters in quick succession can be fast enough to trigger this issue. I'm currently working around it by using `await TurboFrameElement.loaded`.

```js
export default class extends Controller {
  static targets = ["frame", ...]

  async load() {
    // Wait for the frame to finish loading whatever it is currently loading, but ensure we only wait once.
    if (this.loading) {
      return
    }
    this.loading = true
    await this.frameTarget.loaded
    this.loading = false
    this.frameTarget.src = ...
  }
}
```

seanpdoyle:
Generally, I think [debouncing the event](https://thoughtbot.com/blog/hotwire-typeahead-searching#searching-while-typing) that triggers the `src` navigation could be the most effective step you could take in this scenario to mitigate this issue. I'd even recommend doing that even after a fix for this issue ships.

With that being said, thank you for sharing this context. I believe a fix for this could involve changing the `FrameController` to cancel any pending promises (via a [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) instance) from previous requests, which is triggered after changes to the `src` attribute:

https://github.com/hotwired/turbo/blob/acb003d9879311b7f44fe12499ad333605ad3cdd/src/core/frames/frame_controller.ts#L86

TastyPi:
I agree debouncing would be a good idea for this use case, the bug could present itself if one of the requests took longer than the debounce time. Unlikely, but possible.

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/418. I'm not sure how to add a test to the suite to cover this scenario. Could you try it out to see if it addresses your scenario?

PR #418, opened 2021-10-04
Frames: Abort prior requests during navigation

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/417

---

If there is an unresolved `FetchRequest` in-flight when navigating a
Turbo Frame, [abort][] its underlying [AbortController][] instance.

[abort]: https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort
[AbortController]: https://developer.mozilla.org/en-US/docs/Web/API/AbortController

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -21,6 +21,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   readonly formInterceptor: FormInterceptor
   currentURL?: string | null
   formSubmission?: FormSubmission
+  private currentFetchRequest: FetchRequest | null = null
   private resolveVisitPromise = () => {}
   private connected = false
   private hasBeenLoaded = false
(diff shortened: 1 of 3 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -233,11 +234,15 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   // Private
 
   private async visit(url: Locatable) {
-    const request = new FetchRequest(this, FetchMethod.get, expandURL(url), undefined, this.element)
+    const request = new FetchRequest(this, FetchMethod.get, expandURL(url), new URLSearchParams, this.element)
+
+    this.currentFetchRequest?.abortController.abort()
```
packagethief, line 239: Could we use `FetchRequest`'s `cancel()` here instead of manually aborting?

https://github.com/hotwired/turbo/blob/3e4becaaf293e5ca6e962f8151c494185241dd17/src/http/fetch_request.ts#L80-L82

PR #421, opened 2021-10-12
Fire `turbo:click` event when submitting `<form method="get">`

seanpdoyle:
Prior to this change, navigating a page via an `<a>` element and
navigating a page via a `<form method="get">` element both resulted in
`turbo:visit` events.

However, clicking an `<a>` fired a `turbo:click` event, whereas
submitting a `<form method="get">` by clicking its `<input
type="submit">` or `<button>` element _did not_ fire the click for the
submitter.

This commit adds support for firing `turbo:click` events for a
`<form method="get">` element's submitter.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -40,15 +40,10 @@ export class Navigator {
     this.currentVisit.start()
   }
 
-  submitForm(form: HTMLFormElement, submitter?: HTMLElement) {
+  submitForm(formSubmission: FormSubmission) {
     this.stop()
-    this.formSubmission = new FormSubmission(this, form, submitter, true)
-
-    if (this.formSubmission.isIdempotent) {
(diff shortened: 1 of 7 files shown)


dhh:
I wouldn't expect a form submission, even if its a GET method, to trigger turbo:click, when other form submissions do not. What gave rise to wanting this?

dhh:
It is an odd duck, but making get-forms fire turbo:click doesn't quite make sense to me. I get that they under the good are actually more like links than forms, but that's not how a user would think of it. If anything, it should fire more of the same like post-forms. But I also think that maybe it's just fine as it is? It is a bit of a hybrid, so imo OK that it has a different event stack.

seanpdoyle:
The more I think about it, the more I think what's proposed in #424 is more viable.

I'm closing this in favor of https://github.com/hotwired/turbo/pull/424.

PR #423, opened 2021-10-13
Persisting the snapshots

jaysson:
Persisting the snapshots opens up the following use cases:
- Offline browsing
- Faster page loads across browser or mobile app sessions

However, I am not sure about the best way to design this and how well can this perform.

- Should we use indexeddb with websql and localStorage fallbacks? `localForage` library is the standard for this behaviour, but that adds a dependency
- Is it better to expose some sort of `persist` and `hydrate` callbacks so that the actual persistence doesn't have to be part of the library?


dhh:
This adds a good amount of complexity, and to truly give offline powers, you need much more (like graceful error states etc). I think for now we're better off with the simpler model of letting the browser do prefetching.

Issue #122, opened 2021-01-19
Forms with data-turbo-action?


jwagener:
Hello,

are form submissions meant to abide by the data-turbo-action attribute of a form/submit button similar to the way links behave?
Something akin to: https://github.com/jwagener/turbo/commit/f55dbb3ec6df8d2a108d208ce0c03eb5c714da63

My use case is that a devise login page that should be replaced by a logged in screen upon successful login in my native app.

Is that something worthwhile to create a PR for when cleaned up and tested?
Or have I gotten a wrong understanding of something?

Thanks!
Johannes

tleish:
That's an interesting thought.  Would [data-turbo="false"](https://turbo.hotwire.dev/handbook/drive#disabling-turbo-drive-on-specific-links-or-forms) be another way to solve this?

PR #424, opened 2021-10-14
`GET` Forms: fire `submit-start` and `submit-end`

seanpdoyle:
Closes https://github.com/hotwired/turbo/pull/421
Closes https://github.com/hotwired/turbo/issues/122

---

Fire the same sequence of events for  `<form method="get">` submissions
as for `<form method="post">` submissions.

The [FormSubmission.prepareHeadersForRequest][] already accounts for
`[method="get"]` forms by omitting the `Accept:` header with the custom
`turbo-stream` MIME type.

Visit actions
---

Prior to this change, a `<form method="get">` would _always_ result in
two Visits: the first with `{ action: "advance" }`, then a second with
`{ action: "replace" }`.

Since the response to a `<form method="get">` has the potential to be a
[200 OK][] or a [201 Created][], this commit also modifies the `Visit`
class to skip the `followRedirect()` call when the request is idempotent
and the response is not a redirect.

Test changes
---

The `eventLogs` mechanisms we have in place declared in
`src/tests/fixutres/test.js` cannot properly serialize `Element`
instances, so adding `turbo:submit-start` and `turbo:submit-end`
listeners to serialize events for our test suite isn't possible in the
current configuration. To that end, this commit adds assertions for
`<form>` submit event sequences for all events except those two. In
their place, the suite adds listeners to set `[data-form-submit-start]`
and `[data-form-submit-end]` to the `<html>` element when they fire.

[FormSubmission.prepareHeadersForRequest]: https://github.com/hotwired/turbo/blob/58d2261274533a80a2c5efda7da211b3f20efcbb/src/core/drive/form_submission.ts#L136-L144
[200 OK]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200
[201 Created]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -44,11 +44,7 @@ export class Navigator {
     this.stop()
     this.formSubmission = new FormSubmission(this, form, submitter, true)
 
-    if (this.formSubmission.isIdempotent) {
-      this.proposeVisit(this.formSubmission.fetchRequest.url, { action: this.getActionForFormSubmission(this.formSubmission) })
-    } else {
-      this.formSubmission.start()
-    }
+    this.formSubmission.start()
   }
 
   stop() {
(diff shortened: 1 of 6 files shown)


dhh:
```
Since the response to a <form method="get"> has the potential to be a
200 OK or a 201 Created, this commit also modifies the Visit
class to skip the followRedirect() call when the request is idempotent
and the response is not a redirect
```

I don't know if this is a good idea. A GET action should never respond in 200 OK or 201 Created, since it should never be a writing action. Having it as such wrecks havoc with DB balancing and proxies.

Issue #325, opened 2021-07-26
Absence of turbo:load event for 4xx html responses


fwilkens:
Hi,

For those of us on the turbolinks -> turbo upgrade path, it was great to see https://github.com/hotwired/turbo/pull/39 in response to https://github.com/hotwired/turbo-rails/issues/12 -- thanks for that. One thing I've noticed, however, is that when a 422 html response is rendered by Turbo in cases like this, `turbo:load` is not fired. I just checked in one of our apps with turbolinks, and it looks like Turbolinks _would_ fire `turbolinks:load` in such a case.

The outcome with Turbo is that the 422 html response is rendered, but javascript functionality that hinges on `turbo:load` is broken. My expectation was that `turbo:load` would fire -- is that a reasonable expectation or is there something I'm missing here? One obvious solution is to refactor these cases to use a frame or stream I guess, but my hope after seeing https://github.com/hotwired/turbo/pull/39 was that I could avoid frame/stream refactors until after the initial swap of turbo in for turbolinks and validating that Drive worked more/less ootb with our apps (following the notes in https://github.com/hotwired/turbo-rails/blob/main/UPGRADING.md).

Here's a repo demonstrating the lack of the `turbo:load` event in such a case (forked and slightly modified demo chat app) https://github.com/fwilkens/hotwire-rails-demo-chat/tree/demo_422_no_event. To see the issue, submit the New room form without a name, and note that there's no console log for the event listener on `turbo:load`.

willnet:
@fwilkens I know it's inconvenient, but this is an intentional specification.

https://github.com/hotwired/turbo/issues/85#issuecomment-756191950

fwilkens:
Thanks @willnet -- not sure how missed that issue. I guess I'll close this issue, since 85 was closed with an explanation.

Issue #85, opened 2021-01-07
turbo:load won't be triggered when a FormSubmission request fails


nyrf:
When a FormSubmission request fails and return 400-500 code, `turbo:load` won't be triggered. eg:

`application.js`

```js
document.addEventListener('turbo:load', function () {
  console.log('turbo load')
  document.getElementById('demo').replaceWith('demo replace')
})

```

`new.html.erb`

```
<div id="demo">demo</div>
<%= form_with(model: @post) do |f| %>
  <%= f.label :title %>
  <%= f.text_field :title %>
  <%= f.submit %>
<% end %>
```

`posts_controller.rb`

```ruby
def create
    @post = Post.new(post_params)

    respond_to do |format|
      if @post.save
        format.html { redirect_to @post, notice: "Post was successfully created." }
        format.json { render :show, status: :created, location: @post }
      else
        format.html { render :new, status: :unprocessable_entity }
        format.json { render json: @post.errors, status: :unprocessable_entity }
      end
    end
  end
```

```ruby
class Post < ApplicationRecord
  validates :title, presence: true
end
```

After submit form, it should show `demo replace`, but it showed `demo` because of `turbo:load` hasn't be triggered.

![error2](https://user-images.githubusercontent.com/299532/103898975-a9900480-5130-11eb-9327-93d1129afd20.gif)

More info https://github.com/SimoTod/alpine-turbo-drive-adapter/issues/29#issuecomment-756062646

sstephenson:
This is by design. The `turbo:load` event fires in response to a successful visit, and form submissions only trigger visits when the server responds with a redirect.

nyrf:
So, the PR https://github.com/hotwired/turbo/pull/39  is useless?  We often use `turbo:load` event to run some codes, if this is by design, after submit form fail and return 422, the code inside `turbo:load` won't run. So, how to run the codes inside `turbo:load`?

sstephenson:
You shouldn't listen for the `turbo:load` event unless you specifically want to know that a visit—a navigation that changes the URL—has finished. 

The Hotwire way is to put the code you want to run in the `connect()` method of a Stimulus controller, and then add a matching `data-controller` attribute to an element so that the code runs whenever the element appears on the page.

nyrf:
> The Hotwire way is to put the code you want to run in the `connect()` method of a Stimulus controller, and then add a matching `data-controller` attribute to an element so that the code runs whenever the element appears on the page.

I see, thanks.

[...]

PR #425, opened 2021-10-18
Fire `turbo:frame-load` event during form submission

seanpdoyle:
Prior to this change, whenever a `<form>` submission drove a
`<turbo-frame>` element, `turbo:frame-render` events would fire but
`turbo:frame-load` events would not. The implementation to handle
navigations initiated by `[src]`-based attribute changes diverges from
the code paths that handle form submissions and `4xx` response
rendering.

Regardless of any plans to re-structure those code paths, it's important to ensure that
`turbo:frame-load` events fire regardless of how the `<turbo-frame>` was
navigated.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -88,7 +88,6 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
           this.appearanceObserver.stop()
           await this.element.loaded
           this.hasBeenLoaded = true
-          session.frameLoaded(this.element)
         } catch (error) {
           this.currentURL = previousURL
           throw error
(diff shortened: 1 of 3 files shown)


daniel-rikowski:
As far as I understand this is the feature people were asking for in #85, #325 and #520, but just for Turbo Frames and not the whole page.

If I may ask, what was the reasoning for this change? I.e. why is a `turbo:frame-load` event triggered for `4xx` form responses in a frame, but no `turbo:load` for `4xx` form responses on the page?


Given this statement:

> it's important to ensure that **`turbo:frame-load`**  events fire regardless of how the **`<turbo-frame>`** was
navigated.

I fully agree. And I would also argue:

> it's important to ensure that **`turbo:load`** events fire regardless of how the **`<body>`** was
navigated.


I might be missing something here, sorry if this is a dumb question...


Thank you for your great work, BTW!

Issue #326, opened 2021-07-27
Turbo frame doesn't render response HTML if status is 4xx or 5xx


janko:
I'm trying to implement custom error pages in a way that will work with turbo frames. I'm using the strategy where I have `config.exceptions_app = self.routes` in my `config/application.rb`, and I have an `ErrorsController` that's rendering the error pages (404 & 500 for now).

This works with Turbo Drive, and how I want it to behave with Turbo Frames is that an error message would be rendered inside the turbo frame instead of the content. However, it seems that Turbo Frames doesn't render the response HTML if the status if 4xx or 5xx. Keep in mind that this is regular asynchronus frame loading with a GET request, it's not form submission.

So, what I'm basically asking for is this functionality – https://github.com/hotwired/turbo/pull/39, but to have it work for GET request as well.

tleish:
A PR that might help closed not too long ago without explanation by the author https://github.com/hotwired/turbo/pull/210.

PR #430, opened 2021-10-27
Extract `FrameVisit` to drive `FrameController`

seanpdoyle:
Might close https://github.com/hotwired/turbo/pull/793

The problem
---

Programmatically driving a `<turbo-frame>` element when its `[src]`
attribute changes is a suitable end-user experience in consumer
applications. It's a fitting black-box interface for the outside world:
change the value of the attribute and let Turbo handle the rest.

However, internally, it's a lossy abstraction.

For example, when the `FrameRedirector` class listens for page-wide
`click` and `submit` events, it determines if their targets are meant to
drive a `<turbo-frame>` element by:

1. finding an element that matches a clicked `<a>` element's `[data-turbo-frame]` attribute
2. finding an element that matches a submitted `<form>` element's `[data-turbo-frame]` attribute
3. finding an element that matches a submitted `<form>` element's
   _submitter's_ `[data-turbo-frame]` attribute
4. finding the closest `<turbo-frame>` ancestor to the `<a>` or `<form>`

Once it finds the matching frame element, it disposes of all that
additional context and navigates the `<turbo-frame>` by updating its
`[src]` attribute. This makes it impossible to control various aspects
of the frame navigation (like its "rendering" explored in
[hotwired/turbo#146][]) outside of its destination URL.

Similarly, since a `<form>` and submitter pairing have an impact on
which `<turbo-frame>` is navigated, the `FrameController` implementation
passes around a `HTMLFormElement` and `HTMLSubmitter?` data clump and
constantly re-fetches a matching `<turbo-frame>` instance.

Outside of frames, page-wide navigation is driven by a `Visit` instance
that manages the HTTP life cycle and delegates along the way to a
`Visit` delegate. It also pairs calls to visit with an option object to
capture additional context.

The proposal
---

This commit introduces the `FrameVisit` class. It serves as an
encapsulation of the `FetchRequest` and `FormSubmission` lifecycle
events involved in navigating a frame.

It's implementation draws inspiration from the `Visit` class's delegate
and option structures. Since the `FrameVisit` knows how to unify
both `FetchRequest` and `FormSubmission` hooks, the resulting callbacks
fired from within the `FrameController` are flat and consistent.

Extra benefits
---

The biggest benefit is the introduction of a DRY abstraction to
manage the behind the scenes HTTP calls necessary to drive a
`<turbo-frame>`.

With the introduction of the `FrameVisit` concept, we can also declare a
`visit()` and `submit()` method for `FrameElement` delegate
implementations in the place of other implementation-specific methods
like `loadResponse()` and `formSubmissionIntercepted()`.

In addition, these changes have the potential to close
[hotwired/turbo#326][], since we can consistently invoke
`loadResponse()` across `<a>`-click-initiated and
`<form>`-submission-initiated visits. To ensure that's the case, this
commit adds test coverage for navigating a `<turbo-frame>` by making a
`GET` request to an endpoint that responds with a `500` status.

[hotwired/turbo#146]: https://github.com/hotwired/turbo/pull/146
[hotwired/turbo#326]: https://github.com/hotwired/turbo/issues/326

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -1,18 +1,12 @@
 import { FrameElement, FrameLoadingStyle } from "../../elements/frame_element"
-import { FetchMethod, FetchRequest } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { AppearanceObserver } from "../../observers/appearance_observer"
 import {
-  clearBusyState,
   dispatch,
   getAttribute,
-  parseHTMLDocument,
-  markAsBusy,
   uuid,
-  getHistoryMethodForAction,
-  getVisitAction
+  getHistoryMethodForAction
 } from "../../util"
-import { FormSubmission } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { getAction, expandURL, urlsAreEqual, locationIsVisitable } from "../url"
 import { FormSubmitObserver } from "../../observers/form_submit_observer"
@@ -21,18 +15,15 @@ import { LinkInterceptor } from "./link_interceptor"
 import { FormLinkClickObserver } from "../../observers/form_link_click_observer"
 import { FrameRenderer } from "./frame_renderer"
 import { session } from "../index"
-import { StreamMessage } from "../streams/stream_message"
 import { PageSnapshot } from "../drive/page_snapshot"
 import { TurboFrameMissingError } from "../errors"
+import { FrameVisit } from "./frame_visit"
 
 export class FrameController {
-  fetchResponseLoaded = (_fetchResponse) => Promise.resolve()
-  #currentFetchRequest = null
-  #resolveVisitPromise = () => {}
   #connected = false
   #hasBeenLoaded = false
   #ignoredAttributes = new Set()
-  action = null
+  #frameVisit = null
 
   constructor(element) {
     this.element = element
@@ -68,6 +59,11 @@ export class FrameController {
       this.linkInterceptor.stop()
       this.formSubmitObserver.stop()
     }
+  }
+
+  visit(options) {
+    const frameVisit = new FrameVisit(this, this.element, options)
+    return frameVisit.start()
   }
 
   disabledChanged() {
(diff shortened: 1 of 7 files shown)


seanpdoyle:
@manuelpuyol @kevinmcconnell I'd really appreciate a review of this changeset, if you're able to provide some.

manuelpuyol:
I'm seeing some odd behaviors with requests when navigating with frames in this branch:

1 - Navigating away from a frame that has a `src` set creates an instantly canceled fetch request

https://user-images.githubusercontent.com/11280312/206230499-b5b550e7-2a2c-446d-a167-8f8f2647a9e7.mov

2 - After canceling a frame visit (by clicking a link before the fetch completes) when going back in history, the frame will refetch.

https://user-images.githubusercontent.com/11280312/206231369-f9ac41bc-6bbe-4189-bcd4-153f1da2168a.mov

Note that when I navigate without canceling a previous frame nav, the back button **won't** cause a new fetch, but after I cancel a nav it does.

3 - The behavior above causes an error similar to #604 where we can reach a bad state by quickly navigating tabs and history

https://user-images.githubusercontent.com/11280312/206232242-6ee04207-9017-41b6-ac11-1c70fc083e9d.mov

seanpdoyle:
@manuelpuyol I've pushed up some changes to move when the `[complete]` attribute is set. Could you pull down the latest and try to reproduce the behavior you mentioned above?

manuelpuyol:
@seanpdoyle I pushed some tests in https://github.com/seanpdoyle/turbo/pull/1, we could check those!

seanpdoyle:
thank you @manuelpuyol !

I've pushed up those tests as part of 7f3ff394049de6329ce8aff7ed0ca86d16b0c563. They seem to overlap a little with the tests I've introduced. Once CI is passing, I'll try to pick out the best parts from each and coalesce the two.

Thank you for helping troubleshoot!

manuelpuyol (approved):

seanpdoyle:
@kevinmcconnell @afcapel with 7.2.5 released, could this changeset serve as a start toward 7.3.x development?

kevinmcconnell:
@seanpdoyle we'll definitely take a look at this shortly! Sorry for the delay in getting to it so far, but I've finally carved out some time to catch up on the backlog so will be getting to it soon.

src/core/frames/frame_controller.ts
```diff
@@ -145,39 +134,30 @@ export class FrameController
...
-          await this.loadFrameResponse(fetchResponse, document)
-        } else {
-          await this.handleUnvisitableFrameResponse(fetchResponse)
-        }
+    const html = await fetchResponse.responseHTML
+    if (html) {
+      const pageSnapshot = PageSnapshot.fromHTMLString(html)
+
+      if (pageSnapshot.isVisitable) {
+        await this.loadFrameResponse(fetchResponse, pageSnapshot, frameVisit)
```
seanpdoyle, line 151: @kevinmcconnell @afcapel I've rebased this commit to incorporate the changes made in https://github.com/hotwired/turbo/pull/867, namely the expectation that [there will be a `PageSnapshot` instance at-the-ready](https://github.com/hotwired/turbo/commit/1e78f3b1ef4e5263c4d6fd2003ae8298e27b3b3c#diff-8f44fafd854f6f716d51a2b05b5689266a9b66e0b1ad9d68570bd4e8cefd0a47).

seanpdoyle:
@kevinmcconnell @afcapel are either of you able to review these changes?

Issue #378, opened 2021-09-06
Turbo frames conditional rendering


james-reading:
I want to stick a form in a modal, using a link that will request the `new` action to render the form HTML which then gets injected into the modal and opened.

This is a pretty common pattern that I've implemented many times before with Rails UJS.

I am now trying to do the same thing with Turbo.

```
<%# index.html.erb %>

<%= link_to 'New Post', new_post_path, data: { 'turbo-frame': 'modal-body' } %>

<div class="modal" data-controller="modal">
  <%= turbo_frame_tag "modal-body", 'data-action': 'turbo:frame-render->modal#show' %>
</div>
```

```
<%# new.html.erb %>

<h1>New Post</h1>

<%= turbo_frame_tag "modal-body" do %>
  <%= form_with(model: @post) do |form| %>
    <div class="field">
      <%= form.label :title %>
      <%= form.text_field :title %>
    </div>

    <div class="actions">
      <%= form.submit %>
    </div>
  <% end %>
<% end %>

<%= link_to 'Show', @post %> |
<%= link_to 'Back', posts_path %>
```

So I have a pre-rendered empty modal with `turbo_frame_tag` in the modal body. When I click the link, the controller renders the `new` template which has the form wrapped in a matching `turbo_frame_tag`. The form gets injected into the modal and I have a stimulus controller which opens the modal on `turbo:frame-render`.

This works pretty well, but there seems to be a pretty big limitation - The form HTML has to be identical for both the modal, and for when I visit the new URL directly. For example, if I need a close button when in a modal, but a cancel link when visiting the page directly, this doesn't seem possible to handle server-side.

With Rails UJS, My `new` controller action could respond to `format.js` for the modal and `format.html` when visiting directly, making it possible to render different HTML. As `new` action is a GET, this rules out `format.turbo_stream` which is for non-GET requests only.

acetinick:
Hi @james-reading , this might be a duplicate of this one https://github.com/hotwired/turbo/issues/257 ?

james-reading:
> Hi @james-reading , this might be a duplicate of this one #257 ?

Hmm I think it might be slightly different. I am not talking about conditionally setting the response turbo frame. I mean conditionally rendering different HTML based on the request turbo frame.

What I'd like to do is something like

```
class PostsController < ApplicationController
  def new
    @post = Post.new

    case request.headers['Turbo-Frame']
    when 'modal-body'
      render partial: 'posts/modal_body'
    when nil
      render :new
    end
[...]

PR #431, opened 2021-10-31
Support custom rendering in `turbo:before{-frame,}-render` events

seanpdoyle:
Renderer: Extract `renderElement()`
===

For each renderer (`PageRenderer`, `ErrorRenderer`, `FrameRenderer`)
extract their rendering logic into a class-`static-` property named
`renderElement()`. That function accepts two arguments: the first is the
current version of the element being rendered, the second is the new
version of the element being rendered.

The `PageRenderer` and `ErrorRenderer` types are genericized to accept
`HTMLBodyElement` instances. The `FrameRenderer` generics accept a
`FrameElement` instance. This means that the `renderElement()` function
shares the same class-level generics.

Extract `ViewRenderOptions`
===

Introduce the `ViewRenderOptions<E extends Element>` to store the
`resume()` function dispatched with the `turbo:before-render` event.

Future commits will add additional properties to this option interface.

Support custom rendering from `turbo:before-render`
===

The problem
---

The rendering process during a page-wide navigation is opaque, and
cannot be extended. Proposals have been made to use third-party
rendering tools like [morphdom][], or other animation libraries.

Outside of the HTML manipulation, Turbo is also responsible for loading
script, transposing permanent elements, etc.

How might these tools integrate with Turbo in a way that's compliant
with permanent elements.

The solution
---

When publishing a `turbo:before-render` event, dispatch it with a
`render()` function property in addition to the `resume()`.

This way, consumer applications can override rendering:

```js
import morphdom from "morphdom"

    morphdom(currentElement, newElement)
  }

  // or, more tersely
  detail.render = morphdom
})
```

Potentially Closes [hotwired/turbo#197][]
Potentially Closes [hotwired/turbo#378][]
Potentially Closes [hotwired/turbo#218][]

[morphdom]: https://github.com/patrick-steele-idem/morphdom
[hotwired/turbo#218]: https://github.com/hotwired/turbo/pull/218
[hotwired/turbo#378]: https://github.com/hotwired/turbo/issues/378
[hotwired/turbo#197]: https://github.com/hotwired/turbo/issues/197

Introduce `turbo:before-frame-render` event
===

The problem
---

Similar to `turbo:before-render` events and the `<body>` element,
rendering `<turbo-frame>` events is opaque and isn't extensible.

The solution
---

Publish a `turbo:before-frame-render` event, dispatch it with a
`render()` function property in addition to the `resume()`.

This way, consumer applications can override rendering in the same style
as `turbo:before-render` events.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -7,7 +7,7 @@ import {
 import { FetchMethod, FetchRequest, FetchRequestDelegate, FetchRequestHeaders } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { AppearanceObserver, AppearanceObserverDelegate } from "../../observers/appearance_observer"
-import { clearBusyState, getAttribute, parseHTMLDocument, markAsBusy } from "../../util"
+import { clearBusyState, dispatch, getAttribute, parseHTMLDocument, markAsBusy } from "../../util"
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { ViewDelegate, ViewRenderOptions } from "../view"
@@ -19,6 +19,7 @@ import { FormLinkInterceptor, FormLinkInterceptorDelegate } from "../../observer
 import { FrameRenderer } from "./frame_renderer"
 import { session } from "../index"
 import { isAction } from "../types"
+import { TurboBeforeFrameRenderEvent } from "../session"
 
 export class FrameController
   implements
@@ -259,8 +260,22 @@ export class FrameController
 
   // View delegate
 
-  allowsImmediateRender(_snapshot: Snapshot, _options: ViewRenderOptions<FrameElement>) {
-    return true
+  allowsImmediateRender({ element: newFrame }: Snapshot<FrameElement>, options: ViewRenderOptions<FrameElement>) {
+    const event = dispatch<TurboBeforeFrameRenderEvent>("turbo:before-frame-render", {
+      target: this.element,
+      detail: { newFrame, ...options },
+      cancelable: true,
+    })
+    const {
+      defaultPrevented,
+      detail: { render },
+    } = event
+
+    if (this.view.renderer && render) {
+      this.view.renderer.renderElement = render
+    }
+
+    return !defaultPrevented
   }
 
   viewRenderedSnapshot(_snapshot: Snapshot, _isPreview: boolean) {}
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -10,7 +10,7 @@ import { AppearanceObserver, AppearanceObserverDelegate } from "../../observers/
 import { clearBusyState, getAttribute, parseHTMLDocument, markAsBusy } from "../../util"
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
-import { ViewDelegate } from "../view"
+import { ViewDelegate, ViewRenderOptions } from "../view"
 import { getAction, expandURL, urlsAreEqual, locationIsVisitable } from "../url"
 import { FormInterceptor, FormInterceptorDelegate } from "./form_interceptor"
 import { FrameView } from "./frame_view"
@@ -259,7 +259,7 @@ export class FrameController
 
   // View delegate
 
-  allowsImmediateRender(_snapshot: Snapshot, _resume: (value: any) => void) {
+  allowsImmediateRender(_snapshot: Snapshot, _options: ViewRenderOptions) {
     return true
   }
 
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -29,7 +29,7 @@ export class FrameController
     FrameElementDelegate,
     FormLinkInterceptorDelegate,
     LinkInterceptorDelegate,
-    ViewDelegate<Snapshot<FrameElement>>
+    ViewDelegate<FrameElement, Snapshot<FrameElement>>
 {
   readonly element: FrameElement
   readonly view: FrameView
@@ -44,7 +44,7 @@ export class FrameController
   private connected = false
   private hasBeenLoaded = false
   private ignoredAttributes: Set<FrameElementObservedAttribute> = new Set()
-  private previousContents?: DocumentFragment
+  private previousFrameElement?: FrameElement
 
   constructor(element: FrameElement) {
     this.element = element
@@ -131,7 +131,14 @@ export class FrameController
       if (html) {
         const { body } = parseHTMLDocument(html)
         const snapshot = new Snapshot(await this.extractForeignFrameElement(body))
-        const renderer = new FrameRenderer(this, this.view.snapshot, snapshot, false, false)
+        const renderer = new FrameRenderer(
+          this,
+          this.view.snapshot,
+          snapshot,
+          FrameRenderer.renderElement,
+          false,
+          false
+        )
         if (this.view.renderPromise) await this.view.renderPromise
         await this.view.render(renderer)
         this.complete = true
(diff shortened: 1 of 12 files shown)


seanpdoyle:
This is a companion PR to https://github.com/hotwired/turbo/pull/146. https://github.com/hotwired/turbo/pull/146 is meant to expose the out-of-the-box rendering in a plug-and-play way, this PR intends to allow for total overrides.

src/core/frames/frame_controller.ts
```diff
@@ -217,8 +217,15 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
 
   // View delegate
 
-  allowsImmediateRender(snapshot: Snapshot, resume: (value: any) => void) {
-    return true
+  allowsImmediateRender({ element: newFrame }: Snapshot, options: ViewRenderOptions<FrameElement>) {
+    const event = dispatch("turbo:before-frame-render", { detail: { newFrame, ...options }, cancelable: true })
```
seanpdoyle, line 221: We need to dispatch this event with the FrameElement as its target.

aaronjensen:
@seanpdoyle this and #146 are very interesting. Are you still pursuing getting these merged or do you have alternative ways of accomplishing the same things?

seb-jean:
Nice PR

seanpdoyle:
If this were to ship, it might be worthwhile to reverse the changes made in https://github.com/hotwired/turbo/pull/568 and https://github.com/hotwired/turbo/pull/305 so that the `PageRenderer` and `FrameRenderer` classes remain private implementation details.

rainerborene:
I think this is pretty useful. Right now we're monkey patching the `PageRenderer` class.
```js
import { PageRenderer } from "@hotwired/turbo"

Object.assign(PageRenderer.prototype, {
  assignNewBody() {
    const container = document.querySelector("#app")
    const newContainer = this.newElement.querySelector("#app")

    if (container && newContainer) {
      container.replaceWith(newContainer)
    } else {
      document.body.replaceWith(this.newElement)
    }
  }
})
```

dhh:
Be happy to see this refreshed for 7.2.0.

dhh:
This could use a solid doc PR with examples of how and why to overwrite this.

scottnicolson:
> I think this is pretty useful. Right now we're monkey patching the `PageRenderer` class.
> 
> ```js
> import { PageRenderer } from "@hotwired/turbo"
> 
> Object.assign(PageRenderer.prototype, {
>   assignNewBody() {
>     const container = document.querySelector("#app")
>     const newContainer = this.newElement.querySelector("#app")
> 
>     if (container && newContainer) {
>       container.replaceWith(newContainer)
>     } else {
>       document.body.replaceWith(this.newElement)
>     }
>   }
> })
> ```

@rainerborene I tried this approach while we wait for version 7.2.0.  It only worked for forward links for me.  i.e. if I used the browser back button (both chrome and safari) then the page would not update.  Did you not experience this issue with your patch?

manuelpuyol:
@seanpdoyle I'm experiencing some issues with back/forward buttons with custom renders. I created a very simple example here https://github.com/hotwired/turbo/compare/main...manuelpuyol:turbo:showcase-custom-render-bug?expand=1

There, instead of replacing the `<body>`, I'm replacing an element marked by `data-turbo-body`, but that change makes back/forward navigation inconsistent, sometimes loading the correct page:


https://user-images.githubusercontent.com/11280312/198410331-61539cfd-5d04-4bfd-a100-bcbe758f47e7.mov

brunoprietog:
Hi @seanpdoyle 

I also have a similar problem.

In my case, I am replacing the div#app element instead of the body, because I have additional elements I want to keep.

I can reproduce it by entering one page, navigating to another, navigating back to the previous page, and trying to navigate to the current page again.

What I found is that the turbo:before-render event is executed twice in some cases. When that happens, the first time the event runs, I get the new element with all elements except div#app.

If I don't replace the detail.render method, I can see that when it happens that the event is emited twice, the first time the new element is an empty body, and the second time it is the new element.

I don't know what's going on, but I temporarily fixed it like this:

```
const findAppElement = (element) => element.querySelector("#app")

document.addEventListener("turbo:before-render", ({ detail }) => {
    const currentAppElement = findAppElement(currentElement)
    const newAppElement = findAppElement(newElement)
    if (newAppElement) currentAppElement.replaceWith(newAppElement)
    else currentElement.prepend(currentAppElement)
  }
})
```

By doing this, the second time the event is emited I really get the new element and can replace it correctly.

It would be great if you could find a way to fix this so that Turbo handles this case correctly and other people don't have this problem, because it can't consistently rely on the new element. Actually, Turbo should not emit this event twice in any case.

Thanks!

scottnicolson:
@manuelpuyol I resolved my issue adding a in a delay as detailed [here](https://github.com/hotwired/turbo/pull/305#issuecomment-1183505367)

Please note that it is Janky, as I point out [here](https://github.com/hotwired/turbo/pull/305#issuecomment-1285334349).  You should be aware that using `data-turbo-permanent` with this approach doesn't work.  I got around that by using a stimulus controller to clear the Turbo cache in certain circumstances where I was using `data-turbo-permanent`.  I would say just use with caution and keep our fingers crossed that https://github.com/hotwired/turbo/pull/711 gets some attention soon.

manuelpuyol:
thanks for the tip @scottnicolson !

> I got around that by using a stimulus controller to clear the Turbo cache in certain circumstances where I was using data-turbo-permanent

Could you elaborate a bit more here? We do have some `data-turbo-permanent` elements

dylanfisher:
This is a great addition to Turbo, but I don't think it's documented in the [handbook](https://turbo.hotwired.dev/reference/events) yet. Would be awesome if there was so documentation for it.

scottnicolson:
> Could you elaborate a bit more here? We do have some `data-turbo-permanent` elements

@manuelpuyol I had an index page of items and each item had a bookmark icon.  The bookmark icon also showed up on the show page for each item.  The bookmark icon was marked as `data-turbo-permanent` because if you toggled the bookmark on the show page and used the browser back button to go back to the index, I wanted the bookmark state to be correct. I was using a turbo-stream to update the bookmark state as I didn't want it to affect the browser history.   After updating the `assignNewBody` method the bookmarks would not show on the show page when navigating from the index page.  I decided to forego the hassle and instead just used a stimulus controller to clear the Turbo cache when the bookmark was toggled.  This ensured all stale pages would be reloaded (even on browser back).

Also worth noting that https://github.com/hotwired/turbo/issues/623 makes it so `data-turbo-permanent` delivered in a turbo stream response are now ignored, so my bookmarking solution got worse with the introduction of that PR.

trsteel88:
Out of curiosity, is there a reason Turbo doesn't use morphdom by default. It seems like this would be better for merging the dom?

It looks like morphdom is better when you're focused on a input field which is changing the state of the current frame. When the frame reloads with just Turbo, you lose focus. If you use morphdom, it retains its current focused state.

marcoroth:
Morphdom is good in a lot of use-cases, but it also has its drawbacks and weird side effects. Some of the behavior feels super random and unpredictable if you give it improper markup for example. Which is why I think it shouldn't be enabled by default, because it would probably break more. 

But it should be straightforward enough to configure Turbo to do all the updates using morphdom, if you think its worth it for all the updates in your app.

There's also https://github.com/marcoroth/turbo-morph if you are looking for a pre-configured Turbo Stream action.

seb-jean:
With this feature, how to avoid creating a Stimulus controller. Indeed, I followed a tutorial to create a pagination that scrolls infinitely (https://github.com/thoughtbot/hotwire-example-template/tree/hotwire-example-pagination#replacing-frames-with-their-content).


I created this Stimulus controller below but I think we can do better using this feature.

```js
// app/javascript/controllers/element_controller.js

  replaceWithChildren({ target }) {
    this.element.replaceWith(...target.children)
  }
}
```

Originally it was using `[rendering="replace"]` from PR https://github.com/hotwired/turbo/pull/146 but it was closed.

PR #439, opened 2021-11-09
Add requestSubmit() polyfill by default

dhh:
So we can rely on requestSubmit everywhere and know it'll work on Safari.


dhh:
Should add documentation too that this is the way to do it.

dwightwatson:
Great to see this by default. It looks like this is the polyfill by javan, which [seems to have an issue on iOS Safari](javan/form-request-submit-polyfill/issues/3) so not entirely bulletproof yet. Thought it's worth noting here in case others run into similar issues.

PR #442, opened 2021-11-12
Contain `[aria-busy]` toggling within Frames

seanpdoyle:
Setting `[aria-busy="true"]` on the `<html>` element during a
`<turbo-frame>` navigation might be too aggressive given the frame's
value proposition of a compartmentalized request-response cycle. If
applications style the page based on `html[aria-busy="true"]`, having
that toggled during (eager or lazy) Frame loading could lead to a
disorienting experience.

If that isn't the case for applications, they can synchronize the
attribute themselves with a `MutationObserver`:

```js
const observer = new MutationObserver(([ { target } ]) => {
  document.documentElement.setAttribute("aria-busy", target.busy)
}, { subtree: true, attributeFilter: ["busy"] })

observer.observe(document.body)
```

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -165,7 +165,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   requestStarted(request: FetchRequest) {
-    [ this.element, document.documentElement ].forEach(markAsBusy)
+    markAsBusy(this.element)
   }
 
   requestPreventedHandlingResponse(request: FetchRequest, response: FetchResponse) {
@@ -188,13 +188,13 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   requestFinished(request: FetchRequest) {
-    [ this.element, document.documentElement ].forEach(clearBusyState)
+    clearBusyState(this.element)
   }
 
   // Form submission delegate
 
   formSubmissionStarted({ formElement }: FormSubmission) {
-    [ formElement, this.findFrameElement(formElement), document.documentElement ].forEach(markAsBusy)
+    markAsBusy(formElement, this.findFrameElement(formElement))
   }
 
   formSubmissionSucceededWithResponse(formSubmission: FormSubmission, response: FetchResponse) {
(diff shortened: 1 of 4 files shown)


dhh:
I think if we can't use aria-busy in a uniform way, we should use data-turbo-busy, and then leave it as an exercise for the app to choose when aria-busy might make sense for their flow.

seanpdoyle:
I think Turbo's usage of `[aria-busy]` on the `<html>` element for page-wide navigations and the `<turbo-frame>` (and submitted `<form>`) is appropriate.

The issue this change is addressing would be the same with `[data-turbo-busy]`.

Imagine an application that renders a page-wide spinner instead of a progress bar. If a `<turbo-frame>` navigation toggled `html[data-turbo-busy]` (or `html[aria-busy="true"]`, the attribute itself is unimportant), and CSS was hooked off `[data-turbo-busy]`, the spinner would be rendered whenever a lazy- or eagerly- loaded frame was navigated.

dhh:
Ahhh, I misunderstood the original issue. Yes, that this is about turbo-frame being self-contained. Not whether aria-busy or not makes sense. I thought you meant it wouldn't work for the turbo-frame. Yup, all good with this then.

seanpdoyle:
Switching to `[data-turbo-busy]` leaves the door open for the same style of mutation observer, where the client applications could own the logic for setting `[aria-busy]`:

```js
const observer = new MutationObserver((mutations) => {
  for (const { target } of mutations) {
    if (target.localName == "html") {
      // somehow determine if the change was initiated by a `<turbo-frame>` element or a page visit
    } else if (target.localName == "form" || target.localName == "turbo-frame") {
      target.setAttribute("aria-busy", target.hasAttribute("data-turbo-busy"))
    }
  }
  document.documentElement.setAttribute("aria-busy", target.busy)
}, { subtree: true, attributeFilter: ["data-turbo-busy"] })

observer.observe(document.body)
```

That's possible, but the issue feels the same for `html`-scoped attribute changes.

dhh:
You could actually make a fair case that aria-busy should not be set on html, but that data-turbo-busy should. Since you might have a global style spinner. But... I think it's fine to see how this is going to be used first.

PR #443, opened 2021-11-12
Pass CI on `main` branch

seanpdoyle:
Merging [9c74f77][] introduced a TypeScript build error, which is
failing our CI build. This commit replaces the old `formSubmitted`
helper with the newer `formSubmitEnded`.

[9c74f77]: https://github.com/hotwired/turbo/commit/9c74f77d69b6caa789b891e3fa2b92ae93b33409#diff-225e8f674fa1e4fd108c22a917fd5673e51909d4a5d8958e4874ae9e402dca2f

Attempt to resolve flaky tests
---

Tests involving successive `this.goBack()` and `this.goForward()` calls
are flaky due to timing issues:

```
× firefox on linux 5.11.0-1020-azure - FrameTests - navigating back then forward after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames next contents (0.629s)
367
    StaleElementReference: [GET http://localhost:4444/wd/hub/session/01c3334a-e303-4d29-b392-1e4f1d4d7f5c/element/a97830bd-56a4-4b7a-86a0-9543cf697716/text] The element reference of <h2> is stale; either the element is no longer attached to the DOM, it is not in the current frame context, or the document has been refreshed
368
    For documentation on this error, please visit: https://www.seleniumhq.org/exceptions/stale_element_reference.html
369
370
    System info: host: 'fv-az214-396', ip: '10.1.0.47', os.name: 'Linux', os.arch: 'amd64', os.version: '5.11.0-1020-azure', java.version: '11.0.11'
```

In order to resolve that, this commit is more explicit with timing:
navigating forward and back will wait on `turbo:load` events to fire,
which won't occur until the page renders the cached version _and_
re-fetches the contents. Hopefully, this addresses the stale element
failures, since the referenced element won't be replaced while being
read from.

This commit introduces the `this.getVisibleTextForSelector()` helper to
combine the access-and-read action to a single Promise chain, instead of
sequential access and read commands.

### DIFF
--- a/src/tests/functional/form_submission_tests.ts
+++ b/src/tests/functional/form_submission_tests.ts
@@ -584,14 +584,14 @@ export class FormSubmissionTests extends TurboDriveTestCase {
     await this.clickSelector('#dialog-formmethod-turbo-frame [formmethod="dialog"]')
     await this.nextBeat
 
-    this.assert.notOk(await this.formSubmitted)
+    this.assert.notOk(await this.formSubmitEnded)
   }
 
   async "test form submission targetting frame skipped within method=dialog"() {
     await this.clickSelector('#dialog-method-turbo-frame button')
     await this.nextBeat
 
-    this.assert.notOk(await this.formSubmitted)
+    this.assert.notOk(await this.formSubmitEnded)
   }
 
   async "test form submission targetting frame skipped with submitter formmethod=dialog"() {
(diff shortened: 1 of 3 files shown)


seanpdoyle:
@dhh this passes CI and resolves a test suite regression.

dhh:
Solved via https://github.com/hotwired/turbo/pull/444

PR #444, opened 2021-11-13
Resolve Frame-to-Page Visit event ordering

seanpdoyle:
The problem
---

By attempting to avoid unnecessary renders and events by introducing the
`willRender:` Visit option, the initial `<turbo-frame
data-turbo-action="...">` implementation was skipping several crucial
lifecycle hooks. For example, the resulting Visit would fire a
`turbo:render`, but would not fire a `turbo:load`. This left the
`<html>` element in an inconsistent state without cleaning up any
`[data-turbo-preview]` or `[aria-busy]` attribute modifications.

The solution
---

Forego the `willRender:` option, and instead propose a visit with a
pre-populated `statusCode`, `redirected`, and `responseHTML` value so
that the `Session` (including all of its hooks) can transparently handle it
the same as other `Visit` instances.

The result is much simpler than the original implementation: a promoted
Visit doesn't receive any specialized treatment, so it stands to
benefits from all the existing plumbing.

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -41,7 +41,6 @@ export type VisitOptions = {
   action: Action,
   delegate: Partial<VisitDelegate>
   historyChanged: boolean,
-  willRender: boolean
   referrer?: URL,
   snapshotHTML?: string,
   response?: VisitResponse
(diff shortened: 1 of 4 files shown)


seanpdoyle:
If this were merged, we could skip and close https://github.com/hotwired/turbo/pull/443 since https://github.com/hotwired/turbo/pull/444/commits/23963f120b5700fae5afa84750658e3a29e04ccd resolves the build error and https://github.com/hotwired/turbo/pull/444/commits/c5479d48ed10f38ff6d7ca03aa9de500c596696b fixes the flaky test more directly.

Issue #432, opened 2021-11-01
Error is logged in console when response has no matching `<turbo-frame>`


alexandreruban:
When loading a frame and the response does not have a matching `<turbo-frame>` with the same `id`, an error is logged in the console in `FrameController#extractForeignFrameElement`. It was discussed in this issue #31.

However, I think this causes an issue in most Rails apps. 

#### Steps to reproduce

Let's imagine a simple CRUD controller on a `quote` resource. The destroy action in the controller would look like this:

```rb
# app/controllers/quotes_controller.rb

def destroy
  @quote.destroy # Destroy the quote
  redirect_to quotes_path # Redirect to index page
end
```

In the `QuotesController#index` page, I render the collection of quotes:

```erb
<%# app/views/quotes/index.html.erb %>

<%= render @quotes %>
```

The markup for a quote is in the `app/views/quotes/_quote.html.erb` partial:

```erb
<%# app/views/quotes/_quote.html.erb %>

<%= turbo_frame_tag quote do %>
  <div class="quote">
    <span>Irrelevant markup</span>
    <%= button_to "Destroy", quote, method: :delete %> 
  </div>
<% end %>
```

#### Actual behavior

When clicking on the "Destroy" button for a quote, the quote is removed, but I get an error in console: `Response has no matching <turbo-frame id="quote_1"> element`. The empty frame remains in the DOM.

#### Expected behavior

The `<turbo-frame id="quote_1">` is removed from the DOM completely. No error is logged in the console.

#### What do you think?

If you agree with the expected behavior, I would love to make a PR. 

If what is expected when destroying a resource is to always distinguish the `html` and `turbo_stream` formats, then I would love to document it somewhere:

```rb 
# app/controllers/quotes_controller.rb

def destroy
  @quote.destroy
  respond_to do |format|
    format.html { redirect_to quotes_path }
    format.turbo_stream
  end
end
```

kylekeesling:
It looks like you may be missing a required view. In order to remove the quote from the page you need a corresponding `destroy.turbo_stream.erb` view or you’d need to explicitly write the code to remove the quote from the page in your controller response. Would look something like:

in a view - destroy.turbo_stream.erb
`
turbo_stream.remove @quote
`

Or in your controller:
`
respond_to do |format|
  format.html { redirect_to quotes_path } 
  format.turbo_stream { turbo_stream.remove @quote } 
`

tleish:
@alexandreruban - You might be interested in these discussions:

see: 
[...]

Issue #94, opened 2021-01-08
Some thoughts about turbo-frame-tags when the redirected render doesn't contain any turbo-frame-tags that match the original


kobaltz:
Firstly and perhaps most importantly, thank you so much for Turbo. I really enjoy it!

Let's take a very simple example of a new Rails application with `hotwire-rails` created. We create a scaffold for products and within the form partial of the products, we wrap the form elements in a `turbo_frame_tag 'product'`. The expected behavior now is if there are any validation errors, it renders the page with the validation errors. Turbo accomplishes this already. 

However, if there are no validation errors and the record persists successfully, then I would expect it to render the display of the show page. However, because there is no turbo-frame-tag on the show page, it doesn't do anything and we still have the form displayed.

I think that this is very unexpected behavior even though things make sense with turbo. I know that this functionality has caused me a few headaches initially and also the community with trying to find proper workarounds.

I'm not the best Javascript person out there, but dabbled around in it a bit, so I would leave any PR to the experts. However, I did pull in  the turbo library from turbo-rails into my application to tinker with things and came up with a minor-ish fix.

Within the `FrameController` class, I modified the `loadResponse` function. I added the `const html` which before was simply inside of the `fragmentFromHTML`, but the important bit is the `else` statement. It looks like the `element` is the rendered result from our Rails application and the turbo-frame-tag has been plucked out as a fragment of the original rendered response. So, if no `element` is found, then that means that the rendered page did not have any turbo-frame-tags which means that our UI will look like it did not perform any action. 

My thought here is to replace the body of the current page, with the entire contents of the rendered response in the event that no turbo-frame-tag was found within the response.

```
  async loadResponse(response) {
    const html = await response.responseHTML
    const fragment = fragmentFromHTML(html);
    const element = await this.extractForeignFrameElement(fragment);
    if (element) {
      await nextAnimationFrame();
      this.loadFrameElement(element);
      this.scrollFrameIntoView(element);
      await nextAnimationFrame();
      this.focusFirstAutofocusableElement();
    } else {
      document.getElementsByTagName("html")[0].innerHTML = html
    }
  }
```

sstephenson:
Thanks for opening this issue! What we intend to do when there's no matching frame in the response is to empty out the frame's contents and log a message to the console. Hoping to get that in for the next beta.

kobaltz:
[...]

Issue #31, opened 2020-12-23
Better developer experience when lazy loading turbo-frame doesnt find a matching frame to replace


swanson:
I got tripped up for a bit when trying to add a turbo-frame with lazy loading. It seemed like it wasn't working correctly, but the issue was a mistake on my part: I didn't properly add the frame to the response I was loading. There are instructions for doing so in the guide, but I can see this tripping up people.

My workflow went like this:
- Build a page
- Go to add a turbo-frame to lazy load the page somewhere else in the app
- Expected: to lazy load the page with the application layout (nested nav bar, etc), then I would know it was working and add the frame to slice out the content I didn't want duplicated
- Actual: the lazy loading frame did nothing and seemed to not work

Until I looked at the Rails logs to see the request was being made, I had no indication of what was wrong.

Would it be possible to console error or warn the developer if Turbo wasn't able to find a suitable replacement frame?

Intrepidd:
Not to mention it looks quite easy to push a production breaking bug by unknowingly changing a link to something that won't return a frame. Without extensive integration testing this could go unnoticed.

I suggest to throw an error when an expected frame is not found so it can at least be reported through error monitoring.

JulianFeinauer:
The easiest solution would be to push an error log in the js console, right? As soon as we get a 404 / 500 response on the AJAX request, or?

swanson:
Related https://github.com/hotwired/turbo/issues/94

sstephenson:
(Sorry for not chiming in here as well! ✌️)

swanson:
All good ✌️

sstephenson:
As of 3263389 Turbo empties the frame and logs an error to the console when a matching frame is missing.

Thank you for the suggestion!

brunodrugowick:
I had a working sample that stopped after this `v7.0.0-beta.3` release. I'm using the latest over CDN so I noticed.

If I change back to any prior `v7.0.0-beta` version my code works. On the `v7.0.0-beta.3` I get:

```
Response has no matching <turbo-frame id="beverages_list"> element turbo.js:740:13
    extractForeignFrameElement https://cdn.skypack.dev/-/@hotwired/turbo@v7.0.0-beta.3-tvFsO8IW4m2YjmlcLZZq/dist=es2020/@hotwired/turbo.js:740
[...]

PR #445, opened 2021-11-14
Introduce `turbo:frame-missing` event

seanpdoyle:
Closes [hotwired/turbo#432][]
Follow-up to [hotwired/turbo#94][]
Follow-up to [hotwired/turbo#31][]

When a response from _within_ a frame is missing a matching frame, fire
the `turbo:frame-missing` event.

There is an existing [contract][] that dictates a request from within a
frame stays within a frame.

However, if an application is interested in reacting to a response
without a frame, dispatch a `turbo:frame-missing` event. The event's
`target` is the `FrameElement`, and the `detail` contains the
`fetchResponse:` key. Unless it's canceled (by calling
`event.preventDefault()`), Turbo Drive will visit the frame's URL as a
full-page navigation.

The event listener is also a good opportunity to change the
`<turbo-frame>` element itself to prevent future missing responses.

For example, if the reason the frame is missing is access (an expired
session, for example), the call to `visit()` can be made with `{ action:
"replace" }` to remove the current page from Turbo's page history.

[contract]: https://github.com/hotwired/turbo/issues/94#issuecomment-756968205
[hotwired/turbo#432]: https://github.com/hotwired/turbo/issues/432
[hotwired/turbo#94]: https://github.com/hotwired/turbo/issues/94
[hotwired/turbo#31]: https://github.com/hotwired/turbo/issues/31


tleish:
In the above example, does `Turbo.visit(url, { response })` update the page/url with the existing response or generate another server request?

seanpdoyle:
@tleish if there's a `response` key in the `Partial<VisitOptions>` passed to the `visit` call, the HTTP request is "simulated" by the Visit, and does not result in an additional call:

https://github.com/hotwired/turbo/blob/v7.1.0-rc.1/src/core/drive/visit.ts#L171-L173

As part of this draft, I'm investigating ways to more easily transform the `FetchResponse` into `Turbo.visit` arguments so that client applications don't need to deconstruct them themselves.

tleish:
> there's a `response` key in the `Partial<VisitOptions>` passed to the `visit` call, the HTTP request is "simulated" by the Visit, and does not result in an additional call:

A good solution.  It would be even better if a developer didn't have to recreate response object.  Something like:

```javascript
addEventListener("turbo:frame-missing", async ({ target, detail: { fetchResponse } }) => {
  // the details of `shouldRedirectOnMissingFrame(element: FrameElement)`
  // are up to the application to decide
  if (shouldRedirectOnMissingFrame(target)) {
    Turbo.visit(fetchResponse.location, { response: fetchResponse })
  }
})
```

Learned something new, thanks!

tobyzerner:
Working beautifully for the most part. But I think I'm running into a bug? I'm using the exact code sample suggested:

```js
addEventListener("turbo:frame-missing", async ({ target, detail: { fetchResponse } }) => {
    const { location, redirected, statusCode, responseHTML } = fetchResponse
    const response = { redirected, statusCode, responseHTML: await responseHTML }

    Turbo.visit(location, { response })
})
```

The first time this event is triggered by a form submission, it works great – the whole page is replaced with the new HTML. But if I repeat the submission a second time, it breaks. Turbo seems to revert to a cached snapshot of the whole page, and never completes the visit (`[aria-busy=true]` and `[data-turbo-preview]` attributes remain on body). Not sure if that's enough info for a reproduction?

seanpdoyle:
I'll investigate that. 

I'm not sure if it's related, but it's also worth mentioning that if you're using turbo-rails, you'll have to manually override the controller to render a layout so that the response is a fully formed HTML document.

tleish:
> you'll have to manually override the controller to render a layout so that the response is a fully formed HTML document

Assuming 'frame-missing' still logs a message to the console, should Turbo.visit also log a message of "responseHTML  not a fully formed HTML document"?

tleish:
@tobyzerner - this might be unrelated to these changes as I think I stumbled across a similar issue on 7.0.1.

1. Visit page with turbo-frame (first time) via redirect and the frame loads src.
2. Leave page and update data which impacts turbo-frame
3. Return to page via redirect with the turbo-frame, old data is still displayed in the turbo-frame
4. Hard refresh of the page shows the updated data.

What I observed is that turbo renders the page and fetches the turbo-frame source.  Inspecting the fetch request I can see in includes a full HTML document with the expected updates and the correct turbo-frame.  It just does not update the turbo-frame.

Adding `<meta name="turbo-cache-control" content="no-cache">` to this particular page resolves the issue, but unless I'm misunderstanding turbo caching, adding `<meta name="turbo-cache-control" content="no-cache">` to every page with a turbo-frame shouldn't be a permanent solution.

If needed, I can file a separate issue.

tleish:
Turns out this has nothing to do with turbo-frames, but overall turbo-caching and server redirects.

Created issue https://github.com/hotwired/turbo/issues/447

seanpdoyle:
@tleish I've extended the `turbo:frame-missing` to yield a callable `visit()` function that knows how to transform the `FetchResponse` into a visit. Callers can pass additional `VisitOptions` like `action:` to the function.

tleish:
Will this be in the next release?

seanpdoyle:
I'm not sure when the next release will be. I've marked this as a Draft PR to try and start up a discussion.

I think there might be some details to work out:

* Is `turbo:frame-missing` a good name? Should it be named `turbo:frame-not-found`, `:frame-error`, or something else?
* Does dispatching a reference to the `FetchResponse` make sense? It'd effectively make the shape of `FetchResponse` objects public API, and would represent a commitment to support that API in future releases
* Does treating `turbo-frame[id][disabled]` in the body as a "missing" frame deserve its own conversation or PR?

t27duck:
I'm a fan of `frame-missing` as it explicitly describes the issue (and less typing than `frame-not-found`) while `frame-error` implies some internal issue with displaying it.

Can't speak for point 2.

For the last point I think that's better saved for a separate PR/convo

tleish:
> Is turbo:frame-missing a good name? Should it be named turbo:frame-not-found, :frame-error, or something else?

`turbo:frame-not-found` does seem a familiar concept with `404 not found`

> Does dispatching a reference to the FetchResponse make sense? It'd effectively make the shape of FetchResponse objects public API, and would represent a commitment to support that API in future releases

Always a tradeoff

> Does treating turbo-frame[id][disabled] in the body as a "missing" frame deserve its own conversation or PR?

If you can add a disabled attribute to a turbo-frame in order for it to be ignored, could you not also just not render it in the first place of a turbo-frame request?

seanpdoyle:
> If you can add a disabled attribute to a turbo-frame in order for it to be ignored, could you not also just not render it in the first place of a turbo-frame request?

The desire to render the `[disabled]` attribute server side comes from https://github.com/hotwired/turbo/issues/257#issuecomment-968451458.  It's a mechanism to "break out" from the server while still server-generating the HTML.


The pattern there is that the frame is rendered (with server-generated attributes and contents) and sent to the client. If the request is made from a frame, you could toggle the `[disabled]` attribute and force a `turbo:frame-missing` (or `turbo:frame-not-found`) on the client side. If it _isn't_ a frame request, the `<turbo-frame>` element (along with the server-generated attributes and contents) is sent to the client's full-page request. Then the client has an opportunity to "enable" the frame by removing the `[disabled]` attribute through JS.

```html
<script>
window.Stimulus.register("auto-enable", class extends Controller {
  connect() { this.element.removeAttribute("disabled") }
})

addEventListener("turbo:frame-missing", (event) => {
  event.preventDefault()

  window.Turbo.visit(event.detail.fetchResponse)
})
</script>

<turbo-frame id="my_frame" data-controller="auto-enable" disabled></turbo-frame>
```

tleish:
> The desire to render the [disabled] attribute server side comes from #257 (comment). It's a mechanism to "break out" from the server while still server-generating the HTML.

Ah, that's right.

If the disabled behavior is of concern, a slight alternative to the above suggestion would be something like:

```html
<script>
  Stimulus.register("auto-id", class extends Controller {
    connect() { this.element.id = this.element.dataset.id; }
  })

  addEventListener("turbo:frame-missing", (event) => {
    event.preventDefault()

    Turbo.visit(event.detail.fetchResponse)
  })
</script>

<turbo-frame data-controller="auto-id" data-id="my_frame">
  <h1>My server-generated frame and its contents</h1>
</turbo-frame>
```

mollerhoj:
So `frame-missing` will be best practice for dealing iwth error responses from frames? Somebody asked how to do this on the forums https://discuss.hotwired.dev/t/error-handling-best-practices-patterns/3538 but as far as I know we will have to use this `frame-missing` event, right?

mollerhoj:
I think this PR will be a great stepping store for something more specific to errors in frames. I would love to see rails-turbo default to showing nice error pages with a turbo_frame returns 4xx responses. I our project, we actually currently catch errors in turbo controllers and render the rails error page:
```
<%= turbo_frame_tag model do %>
  <%= ActionDispatch::DebugExceptions.new(Rails.application).send(:render_exception, request, error)[2][0] %>
<% end %>
```
It's a hacky solution - but it's just so much nicer to display the error pages in development rather than just getting an error in the javascript console (who really notices those anyway? ;-) )

Looking forward for this PR to be merged so we can implement a less hacky solution though.

seanpdoyle:
@mollerhoj That makes a lot of sense in development, and might be a good improvement to make to `turbo-rails` (maybe behind a configuration flag).

Behavior like that isn't baked into `turbo` itself because transforming a failed Frame request for a portion of the page into page-wide error undercuts the value proposition of Frames that enables decomposing a page into chunks of asynchronously fetchable fragments.

For example, 500'ing the entire page because a lazily loaded menu's contents failed to load feels like an overreaction.

mollerhoj:
I agree. The current behaviour is very sensible in production - but in development we wanna catch this stuff!

swanson:
This will be super helpful for the use case of redirecting to a login page if a frame ties to load and the user session has expired for example -- and even better because it will encourage proper status codes when redirecting (e.g. using a `401` in an authorization-check redirect instead of default 3xx).

vitobotta:
> > If you can add a disabled attribute to a turbo-frame in order for it to be ignored, could you not also just not render it in the first place of a turbo-frame request?
> 
> The desire to render the `[disabled]` attribute server side comes from [#257 (comment)](https://github.com/hotwired/turbo/issues/257#issuecomment-968451458). It's a mechanism to "break out" from the server while still server-generating the HTML.
> 
> The pattern there is that the frame is rendered (with server-generated attributes and contents) and sent to the client. If the request is made from a frame, you could toggle the `[disabled]` attribute and force a `turbo:frame-missing` (or `turbo:frame-not-found`) on the client side. If it _isn't_ a frame request, the `<turbo-frame>` element (along with the server-generated attributes and contents) is sent to the client's full-page request. Then the client has an opportunity to "enable" the frame by removing the `[disabled]` attribute through JS.
> 
> ```
> <script>
> Stimulus.register("auto-enable", class extends Controller {
>   connect() { this.element.removeAttribute("disabled") }
> })
> 
> addEventListener("turbo:frame-missing", (event) => {
>   event.preventDefault()
> 
>   Turbo.visit(event.detail.fetchResponse)
> })
> </script>
> 
> <turbo-frame id="my_frame" data-controller="auto-enable" disabled>
>   <h1>My server-generated frame and its contents</h1>
> </turbo-frame>
> ```

Hi! I am trying to use this workaround but how do I add a "disabled" attribute to turbo_frame_tag? I tried `html: { disabled: true }` and `disabled: true` but they do not work. Thanks!

seanpdoyle:
Adding disabled: true to the Rails helper should render the element with the attribute. 

It's worth noting that if you're using that pseudo code example, the auto-enable controller would remove the attribute during the controllers connect() lifecycle. 

If you inspect the HTML respond payload, is the event rendered with the attribute?

vitobotta:
> Adding disabled: true to the Rails helper should render the element with the attribute.
> 
> It's worth noting that if you're using that pseudo code example, the auto-enable controller would remove the attribute during the controllers connect() lifecycle.
> 
> If you inspect the HTML respond payload, is the event rendered with the attribute?

I will try again, that is likely the reason why the attribute wasn't there. I didn't see it in the HTML either.

In the meantime I found another workaround. Instead of using a turbo_frame_tag, I do a normal redirect on success and I render a turbo stream template on failure which replaces the form. Seems to work fine.

lefermierperso:
Hey guys, any updates ?

tbeemster:
> This will be super helpful for the use case of redirecting to a login page if a frame tries to load and the user session has expired for example -- and even better because it will encourage proper status codes when redirecting (e.g. using a `401` in an authorization-check redirect instead of default 3xx).

I arrived here from the thread in https://github.com/hotwired/turbo/issues/257 and am looking a solution to the exact use case mentioned above and am wondering if there are plans for this event to be added to Turbo Frames (or if another solution is in the works)

pfeiffer:
Could this be a good candidate for inclusion in 7.2.0?

An alternative could be to combine the PR in https://github.com/hotwired/turbo/pull/560 introducing `turbo:error` with a change that returns a structured `Error` when a frame is missing instead of just an error message. This would at least allow developers to handle the missing frames, albeit not as elegant as this PR.

eladmarg:
this should be merged. 
I used this on production for long time, without any issues.

seanpdoyle:
I've opened https://github.com/hotwired/turbo-rails/pull/367 to explore an alternative for "breaking out" of a frame from the server.

As far as what happens when a frame is missing, that seems to be an ongoing discussion.

src/core/session.ts
```diff
@@ -305,6 +305,13 @@ export class Session
     this.notifyApplicationAfterFrameRender(fetchResponse, frame)
   }
 
+  async frameMissing(frame: FrameElement, fetchResponse: FetchResponse): Promise<void> {
+    const responseHTML = await fetchResponse.responseHTML
```
seanpdoyle, line 309: Since servers (like Rails) can [respond to Frame requests with incomplete HTML documents](https://github.com/hotwired/turbo-rails/pull/232), I'm unsure about how to best handle a response without a matching frame.

I'm not sure whether or not Drive should re-use the response HTML or issue a new request to fetch a _full_ page of content. My gut tells me to always make a full request, but I'm curious if there's a quick win I'm not considering.

dhh: I think the current default is correct, because the majority case here will be to deal with pages like 500 errors or authentication expired errors, where full pages will be returned. But I think we can help developers here by logging a warning to point them in the direction of what's going on.

dhh: Ah, I see you changed it to a full visit instead. I guess it doesn't matter that much. This behavior is for the exceptional case in any circumstance. Not for the normal, steady flow.

dhh:
Would be good to follow up with a doc PR explaining the new event and its power to prevent the default visiting fallback behavior.

Issue #446, opened 2021-11-15
GET forms targeting turbo farmes do not update the `FrameElement#src` on 7.1.0-rc.1


agrobbin:
Big fan of Turbo! I was in particular excited for 7.1.0-rc.1, specifically #398, so I tried it out with our app, and it appears there might be a small regression.

We have a `<form method="get" data-turbo-frame="foo" action="/search">`, and when that form is submitted, it's no longer updating the `src` attribute of our `<turbo-frame id="foo">`. The same code was working as expected on 7.0.1.

dhh:
cc @seanpdoyle

agrobbin:
My pleasure @seanpdoyle! It happens regardless of the new `[data-turbo-action]` attribute.

seanpdoyle:
Thank you for clarifying. My hunch is that it's related to the changes in https://github.com/hotwired/turbo/pull/424. I'll investigate.

agrobbin:
No problem! Let me know if I can be helpful in any other way to help nail down this issue!

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/449 to address this issue.

PR #449, opened 2021-11-16
Frames: handle `GET` form submissions

seanpdoyle:
Closes [446][].

When handling a `<form method="get" action="..."
data-turbo-frame="...">` submission targeting a `<turbo-frame>` element,
responses that don't redirect do not change the element's `[src]`
attribute.

Following the changes made in [424][], `<form>` submissions that result
in `GET` requests are handled the same as other form submissions in that
they fire `turbo:submit-start` and `turbo:submit-end` events. What
[424][] did not account for is navigations initiated by `<form
method="get">` submissions that _do not_ redirect.

In response to those requests, this commit adds two additional criteria
for updating the `<turbo-frame src="...">` attribute:

* the response's status code is [200 OK][], which is idempotent and does
  not indicate a server-side state change (for example, like a [201
  Created][] with a `Location:` header might)
* the response's `Content-Type:` is HTML (not JSON or even Turbo Stream)

Under those circumstances, the frame's `[src]` is updated.

This commit adds test coverage for this new behavior along with
additional coverage to guard against regressions.

[446]: https://github.com/hotwired/turbo/issues/446
[424]: https://github.com/hotwired/turbo/pull/424
[200 OK]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/200
[201 Created]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -99,7 +99,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   }
 
   async loadResponse(fetchResponse: FetchResponse) {
-    if (fetchResponse.redirected) {
+    if (fetchResponse.redirected || (fetchResponse.succeeded && fetchResponse.isHTML)) {
       this.sourceURL = fetchResponse.response.url
     }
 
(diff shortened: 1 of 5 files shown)


seanpdoyle:
While investigating this issue, I also uncovered some additional Frame-to-Visit quirks. I've opened https://github.com/hotwired/turbo/pull/448 to address those.

seanpdoyle:
@dhh if this is a good enough fix, I think that it'd be good to include in 7.1.0-rc2.

agrobbin:
Thanks @seanpdoyle @dhh! I just tried rc2 and everything is looking great.

PR #452, opened 2021-11-18
Export Type declarations for `turbo:` events

seanpdoyle:
Various `turbo:`-prefixed events are dispatched as [CustomEvent][]
instances with data encoded into the [detail][] property.

In TypeScript, that property is encoded as `any`, but the `CustomEvent`
type is generic (i.e. `CustomEvent<T>`) where the generic Type argument
describes the structure of the `detail` key.

This commit introduces types that extend from `CustomEvent` for each
event, and exports them from `/core/index.ts`, which is exported from
`/index.ts` in-turn.

In practice, there are no changes to the implementation. However,
TypeScript consumers of the package can import the types. At the same
time, the internal implementation can depend on the types to ensure
consistency throughout.

[CustomEvent]: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent
[detail]: https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/detail

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -12,6 +12,21 @@ import { FormSubmission } from "./drive/form_submission"
 const session = new Session()
 const { navigator } = session
 export { navigator, session, PageRenderer, PageSnapshot, FrameRenderer }
+export {
+  TurboBeforeCacheEvent,
+  TurboBeforeRenderEvent,
+  TurboBeforeVisitEvent,
+  TurboClickEvent,
+  TurboFrameLoadEvent,
+  TurboFrameRenderEvent,
+  TurboLoadEvent,
+  TurboRenderEvent,
+  TurboVisitEvent,
+} from "./session"
+
+export { TurboSubmitStartEvent, TurboSubmitEndEvent } from "./drive/form_submission"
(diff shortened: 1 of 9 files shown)


src/core/index.ts
```diff
@@ -11,6 +11,21 @@ import { FormSubmission } from "./drive/form_submission"
 const session = new Session
 const { navigator } = session
 export { navigator, session, PageRenderer, PageSnapshot }
+export {
+  TurboBeforeCacheEvent,
+  TurboBeforeRenderEvent,
+  TurboBeforeVisitEvent,
+  TurboClickEvent,
```
seanpdoyle, line 18: I'm not sure what would be better: `BeforeCacheEvent`, `BeforeRenderEvent`, etc internally and exporting them here as `Turbo`-prefixed versions, or sticking with the current style of qualifying them with `Turbo`.

tobyzerner: Or even export them without the `Turbo` prefix? `FrameElement` and `StreamElement` don't have it.

dhh:
Want to rebase this @seanpdoyle?

dhh:
Seems like a legit test failure.

Issue #453, opened 2021-11-18
turbo:click and turbo:before-visit event handlers accumulating on document


afcapel:
This happens when turbo frames are dynamically loaded. 

### Steps to reproduce

Given this html:

```html
<!-- index.html -->
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <link rel="icon" href="data:,">
    <script type="module">
      import hotwiredTurbo from 'https://cdn.skypack.dev/@hotwired/turbo';
    </script>
  </head>
  <h1>Index</h1>

  <turbo-frame id="frame1" src="frame1.html">
    Replaced content 1
  </turbo-frame>

  <turbo-frame id="frame2" src="frame2.html">
    Replaced content 2
  </turbo-frame>
</html>

<!-- frame1.html -->
<turbo-frame id="frame1">
  <p>Dynamically loaded content 1</p>
</turbo-frame>

<!-- frame2.html -->
<turbo-frame id="frame2">
  <p>Dynamically loaded content 2</p>
</turbo-frame>
```

When you turbo visit the first page, you see a new `turbo:click` and `turbo:before-visit` event handler added to the `document` for each dynamically loaded turbo frame, in each visit. These can grow pretty quickly if you have many frames in your app. Clearing the turbo cache doesn't help.

<img width="595" alt="Screenshot 2021-11-18 at 09 55 10" src="https://user-images.githubusercontent.com/150107/142392874-dcd8ea64-4535-407e-b42d-bb16a04b0d3a.png">

seanpdoyle:
Thank you for uncovering this! I've opened https://github.com/hotwired/turbo/pull/454 to prevent this accumulation.

afcapel:
Awesome, thanks!

PR #454, opened 2021-11-18
Disconnect loaded Frame Element while rendering

seanpdoyle:
Closes [#453]

---

While loading a matching `<turbo-frame>` element from a response body,
the `FrameController` "activates" it by invoking
[`FrameElement.connectedCallback()` after importing it into the document][activation].

During that activation, several event listeners are attached, including
some on the `document`.

Unfortunately, that `FrameElement.connectedCallback()` invocation is
never mirrored with a `FrameElement.disconnectedCallback()` invocation,
so the attached listeners are never detached, causing an accumulation.

This commit invokes the `disconnectedCallback()` after loaded element's
contents are extracted into the browser's document.

[#453]: https://github.com/hotwired/turbo/issues/453
[activation]: https://github.com/hotwired/turbo/blob/v7.1.0-rc.1/src/core/frames/frame_controller.ts#L441

### DIFF
--- a/src/core/frames/frame_renderer.ts
+++ b/src/core/frames/frame_renderer.ts
@@ -29,6 +29,7 @@ export class FrameRenderer extends Renderer<FrameElement> {
     if (sourceRange) {
       sourceRange.selectNodeContents(frameElement)
       this.currentElement.appendChild(sourceRange.extractContents())
+      frameElement.disconnectedCallback()
     }
   }
 


src/core/frames/frame_renderer.ts
```diff
@@ -29,6 +29,7 @@ export class FrameRenderer extends Renderer<FrameElement> {
     if (sourceRange) {
       sourceRange.selectNodeContents(frameElement)
       this.currentElement.appendChild(sourceRange.extractContents())
+      frameElement.disconnectedCallback()
```
seanpdoyle, line 32: I'm not sure of a way to add a test to guard against a regression.

The [getEventListeners()](https://developer.chrome.com/docs/devtools/console/utilities/#getEventListeners-function) function used to uncover this leak is defined in the console's suite of tools, and isn't available to the document.

I'm open to ideas!

seanpdoyle:
@dhh this feels like an important fix to ship as part of `7.1.0-rc2`.

src/core/frames/frame_renderer.ts
dhh: Maybe attach a listener that increments a global you can check?

seanpdoyle: I don't think that the code within the `connectedCallback()` and `disconnectedCallback()` fire any events directly. It's also worth noting that we're manually invoking these methods within Turbo itself, so they're kind of "synthetic" connections, which makes a `MutationObserver` similarly unpredictable, since the underlying element itself might not be attached to the browser's document.

Issue #447, opened 2021-11-16
Bug: Turbo.visit Render sequence race condition for Cache and Request


tleish:
I came across what I believe to be a bug.  

1. Visit page (saves to turbo cache)
2. Leave page
3. Return to same turbo-cached page via server 302 redirect
4. Quick flash of new network request content, but then replaced by old cached version.

Fetch requests show updated content in network request.

A few additional observations:
- Render via direct GET request (no redirect) of cached page works as expected
- Render via POST|PUT|DELETE and 302 redirect works as expected
- Render via GET and 302 redirect DOES NOT work as expected

Adding <meta name="turbo-cache-control" content="no-cache"> to the landing page page temporarily resolves the issue, but unless I'm misunderstanding turbo caching, adding <meta name="turbo-cache-control" content="no-cache"> to every page with a redirect is not permanent solution.

Small demo app to reproduce the error:
[GitHub - tleish/turbo-cache-test](https://github.com/tleish/turbo-cache-test)

- Turbo 7.0.1 and v7.1.0-rc.1
- Rails 6.1.4.1
- Chrome Latest
- Firefox Latest
- Safari Latests

Bug also reported in this MR (https://github.com/hotwired/turbo/pull/445#issuecomment-968787563)

tleish:
FYI, after additional troubleshooting I found the following sequence of events:

Events after POST, PUT, DELETE 302 Redirect

- proposeVisit
  - visitProposedToLocation
    - turbo:before-render
    - turbo:render
    - turbo:load

Events after GET 302 Redirect

- proposeVisit
  - visitProposedToLocation
    - turbo:before-render
    - turbo:render
    - turbo:load
- followRedirect
  - visitProposedToLocation
    - turbo:before-render
    - turbo:render

If I disable `#followRedirect, then POST, PUT, DELETE and GET all work as expected. 

```diff
complete() {
  if (this.state == VisitState.started) {
    this.recordTimingMetric(TimingMetric.visitEnd)
    this.state = VisitState.completed
    this.adapter.visitCompleted(this)
    this.delegate.visitCompleted(this)
-   this.followRedirect()
  }
}
```

Related to https://github.com/hotwired/turbo/pull/328.  Maybe @jaysson can take a look?

FYI, this may be also related to https://github.com/hotwired/turbo/issues/428

tleish:
[...]

PR #458, opened 2021-11-19
Do not render cache after preloaded response

tleish:
Fix a race condition where `BrowserAdapter#visitStarted` renders cache after preloaded content is rendered.

The BrowserAdapter has the following:

```javascript
export class BrowserAdapter implements Adapter {
  //...
  visitStarted(visit) {
    visit.issueRequest() //=> Async send request
    visit.changeHistory()
    visit.goToSamePageAnchor()
    visit.loadCachedSnapshot() //=> render cached snapshot while waiting for async request
  }
```
When `visit.issueRequest()` takes longer than 1ms, then `visit.loadCachedSnapshot()` renders the cache first.  When it renders immediately (as in a preloaded response) then `visit.loadCachedSnapshot()` renders last.

Fixes https://github.com/hotwired/turbo/issues/447

### DIFF
--- a/src/tests/functional/visit_tests.ts
+++ b/src/tests/functional/visit_tests.ts
@@ -98,16 +98,6 @@ export class VisitTests extends TurboDriveTestCase {
     this.assert.isTrue(fetchResponseResult.responseHTML.indexOf('An element with an ID') > -1)
   }
 
-  async "test cache does not override preloaded response"() {
-    this.assert(await this.hasSelector("#flash"))
-    this.clickSelector("#link")
-    await this.nextBody
-    await this.clickSelector("#redirection-link-to-cache")
-    await this.nextBody // redirect/cache preview
-    await this.nextBody // final
-    this.assert.ok(await this.hasSelector("#flash"))
-  }
-
   async visitLocation(location: string) {
     this.remote.execute((location: string) => window.Turbo.visit(location), [location])
   }
--- a/src/tests/functional/visit_tests.ts
+++ b/src/tests/functional/visit_tests.ts
@@ -98,6 +98,16 @@ export class VisitTests extends TurboDriveTestCase {
     this.assert.isTrue(fetchResponseResult.responseHTML.indexOf('An element with an ID') > -1)
   }
 
+  async "test cache does not override preloaded response"() {
+    this.assert(await this.hasSelector("#flash"))
+    this.clickSelector("#link")
+    await this.nextBody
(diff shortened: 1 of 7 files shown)


tleish:
@seanpdoyle - can you review this PR and let me know if I need to change anything?

src/tests/functional/visit_tests.ts
```diff
@@ -98,6 +98,21 @@ export class VisitTests extends TurboDriveTestCase {
     this.assert.isTrue(fetchResponseResult.responseHTML.indexOf('An element with an ID') > -1)
   }
 
+  async "test cache does not override response after redirect"() {
+    await this.remote.execute(() => {
+      const cachedElement = document.createElement("some-cached-element")
+      document.body.appendChild(cachedElement)
```
seanpdoyle, line 104: Is it significant that this is an appended element? Could an `<input>` with text typed into it suit this test as state to preserve?

tleish: In my experience, the browser stores `<input>` text in memory, not in the DOM.  Since turbo only caches DOM, testing cache through text input would not work.  

Alternatively, I could add (or remove) a attribute to an existing DOM item (e.g. `data-cached="true"`)

tleish: @seanpdoyle - thumbs up means you are good with the explanation or you want an update to the test with one of the suggestions?

seanpdoyle: The current version works. Thank you for uncovering and digging into this!

tleish:
@ghiculescu  Yes, we do use the iOS adapter.  What issue are you experiencing?

ghiculescu:
It was a screen flicker, but it ended up being unrelated and due to us not creating a new view controller on each visit - apparently yiu really need to do that. So not this issue at all. 

That said... does that code I linked look like it could cause the issue this PR addresses?

tleish:
> That said... does that code I linked look like it could cause the issue this PR addresses?

I haven't tested this particular bug on mobile, but it looks to follow the same older pattern and quite possibly has the same bug.  Are you going to submit a PR to suggest an update to follow the same pattern as this MR?

ghiculescu:
I can, but I'm not confident it's an issue yet. Do you have a test page I can use to replicate it?

tleish:
@ghiculescu - See the steps to reproduce and demo app linked in the original https://github.com/hotwired/turbo/issues/447 issue

Issue #243, opened 2021-04-04
Active Storage direct uploads


gahia:
Working with Active Storage direct uploads when the form is contained by a turbo frame seems not to working as expected.

```
<turbo-frame ...>
  <form ...>
    <input type="file" data-direct-upload-url="...">
    <input type="submit" ...>
  </form>
</turbo-frame>
```

Turbo intercepts the form submission and doesn't allow the direct upload run before do the submit request.

![imagen](https://user-images.githubusercontent.com/8942202/113497320-1ac33600-952d-11eb-899b-568fef8d2286.png)

But, when a form like this is NOT contained by a turbo-frame tag, it works as expected: the file is uploaded and then the form submitted by turbo. 

![imagen](https://user-images.githubusercontent.com/8942202/113497323-2878bb80-952d-11eb-907c-3a8ed706bdec.png)

Is there a workaround or possible solution for this?

_Using turbo-rails (0.5.9) by the assets pipeline_

javan:
[`<turbo-frame>`’s form interceptor](https://github.com/hotwired/turbo/blob/88e42cd0e8106fde4166618975683970d1ce7844/src/core/frames/form_interceptor.ts#L23-L33) needs to respect already canceled (`event.defaultPrevented`) `submit` events like the top-level submit observer: https://github.com/hotwired/turbo/blob/88e42cd0e8106fde4166618975683970d1ce7844/src/observers/form_submit_observer.ts#L33-L34

seanpdoyle:
Does this behavior occur on https://github.com/hotwired/turbo/pull/232? It reused FormSubmitObserver instead of FormInterceptor.

javan:
Almost! Active Storage [listens for `submit` on `document`](https://github.com/rails/rails/blob/58ce6907112af45467b9e003763044e634307b87/activestorage/app/javascript/activestorage/ujs.js#L12) so it cancels the event “too late”. Could fix that in Rails with:
```diff
--- a/activestorage/app/javascript/activestorage/ujs.js
+++ b/activestorage/app/javascript/activestorage/ujs.js
@@ -9,7 +9,7 @@ export function start() {
   if (!started) {
     started = true
     document.addEventListener("click", didClick, true)
-    document.addEventListener("submit", didSubmitForm)
+    document.addEventListener("submit", didSubmitForm, true)
     document.addEventListener("ajax:before", didSubmitRemoteElement)
   }
 }
```
Regardless, #232 seems like a good change to make.

[...]

PR #459, opened 2021-11-20
Frames: Ignore already cancelled `submit` events

seanpdoyle:
Closes [hotwired/turbo#243][]

In tandem with [rails/rails#41960][] (which shipped as part of
`7.0.0.alpha2`), this commit modifies the `FormInterceptor` to ignore
`submit` events that have been cancelled.

Although it doesn't test the integration directly, the cancellation of
the `submit` event is what powers the ActiveStorage.js direct upload
asynchrony. Following a direct upload, ActiveStorage dispatches a
synthetic `submit` event, which the `FormInterceptor` will then respond
to.

[hotwired/turbo#243]: https://github.com/hotwired/turbo/issues/243
[rails/rails#41960]: https://github.com/rails/rails/pull/41960

### DIFF
--- a/src/core/frames/form_interceptor.ts
+++ b/src/core/frames/form_interceptor.ts
@@ -22,7 +22,7 @@ export class FormInterceptor {
 
   submitBubbled = <EventListener>((event: SubmitEvent) => {
     const form = event.target
-    if (form instanceof HTMLFormElement && form.closest("turbo-frame, html") == this.element) {
+    if (!event.defaultPrevented && form instanceof HTMLFormElement && form.closest("turbo-frame, html") == this.element) {
       const submitter = event.submitter || undefined
       const method = submitter?.getAttribute("formmethod") || form.method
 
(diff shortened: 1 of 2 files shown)


CharlieIGG:
So what happens if we ever DO want to have Turbo handle a previously default-prevented form submission?

seanpdoyle:
@CharlieIGG what are the circumstances under which Turbo would be handling an event that was canceled?

ActiveStorage relies on cancelling the `submit`, handling direct uploads, then [programatically click an `input[type="submit"]` button](https://github.com/rails/rails/blob/30363d0b1da86772392ab315ba0afdc65f302817/activestorage/app/javascript/activestorage/ujs.js#L60-L78) to synthesize a separate `submit` event. Maybe applications could deploy the same pattern if they needed to.

CharlieIGG:
That actually makes perfect sense, thank you for the clarification @seanpdoyle

Issue #465, opened 2021-11-23
Upgrading to 7.1.rc3 doesn't pass parameters outside turboframe


acetinick:
Since upgrading from rc2 to rc3, it now strips out parameters from the pagy gem so it won't pass parameters to the server.  It removes the page parameter after clicking on it, and the server doesn't recieve any parameters in the request.

FYI , If I wrap my table inside a <turbo-frame>  it will work fine.

![image](https://user-images.githubusercontent.com/571649/142961999-0d3fec38-6d2b-4d45-a922-9427a4752aa7.png)

seanpdoyle:
I have a hunch at what's at the root of this, but I'm very worried this wasn't uncovered by failures in the test suite.

I'll open a PR to resolve the issue and add more thorough test coverage.

acetinick:
Yea looking at it more, it looks like its doing it for all standard links outside of turbo frames.  

Looking at a few of the tests in the PR looks like most assume submit tags inside a form tag, in ths scenario its just an <a> tag outside of any form/turbo-frame, and no data-turbo-method attribute.

seanpdoyle:
That's a good point, I'll expand that coverage to contain links.

What worries and surprises me is that the _rest_ of the suit passed with broken behavior in such an integral feature.

Intrepidd:
I have a similar issue where I have a page with links to itself with query params and clicking on those links don't change the page content anymore with RC3

tortuetorche:
Hi folks,

I've also some issues when opting out cache (`<meta name="turbo-cache-control" content="no-cache">`).
When I go to previous pages with back and forward buttons of my Web browser (Firefox or Chromium).

Have a good day,
Tortue Torche

seanpdoyle:
@acetinick @Intrepidd I've opened https://github.com/hotwired/turbo/pull/466 to try and resolve the searchParams issue.

@tortuetorche what you're describing sounds unrelated to the original issue outlined here. Could you open a new Issue?

dhh:
@acetinick @Intrepidd Lemme know when you've had a chance to test #466. Would like to move to a 7.1.0 final, if this is the last blocker!

Intrepidd:
@dhh works for me on `main` when it didn't on 7.1.rc3 :) Thanks @seanpdoyle

PR #466, opened 2021-11-23
Restore searchParams to Visit and Frame Navigation

seanpdoyle:
Closes [hotwired/turbo#456][]

The changes made in [hotwired/turbo#464][] resolved an issue in how
`GET` requests merged query parameters into their `URL` instance.
Unfortunately, the changes were too loosely scoped, and they affected
_all_ `GET` requests, including those initiated by `<a>` elements with
`[href]` attributes that declare search parameters directly (that *do
not* need to be "merged" into anything).

The [change][] made in [hotwired/turbo#464][] to conditionally include
the `FetchRequest.body` based on the request's verb enables the changes
made in this commit.

In response to that change, this commit moves the
`mergeFormDataEntries()` function out of the `FetchMethod` module and
into the `FormSubmission` class. Since merging _from_ `FormData`
instances _into_ a `URL` is a concern of `GET`-powered Form Submissions,
moving the function out of the `FetchRequest` further simplifies that
class. Move the `mergeFormDataEntries()` to the `FormSubmission`
constructor to re-write its `location` property's [searchParams][] prior
to constructing the instance's `FetchRequest`.

[hotwired/turbo#456]: https://github.com/hotwired/turbo/issues/465
[hotwired/turbo#464]: https://github.com/hotwired/turbo/pull/464
[change]: https://github.com/hotwired/turbo/pull/461/commits/6906e57769a79fcf39955c4c0d1037434ea0c579#diff-68b647dc2963716dc27c070f261d0b942ee9c00be7c4149ecb3a5acd94842d40R119
[searchParams]: https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -44,6 +44,7 @@ export class FormSubmission {
   readonly formElement: HTMLFormElement
   readonly submitter?: HTMLElement
   readonly formData: FormData
+  readonly location: URL
   readonly fetchRequest: FetchRequest
   readonly mustRedirect: boolean
   state = FormSubmissionState.initialized
@@ -58,6 +59,10 @@ export class FormSubmission {
     this.formElement = formElement
     this.submitter = submitter
     this.formData = buildFormData(formElement, submitter)
+    this.location = expandURL(this.action)
+    if (this.method == FetchMethod.get) {
+      mergeFormDataEntries(this.location, [ ...this.body.entries() ])
+    }
     this.fetchRequest = new FetchRequest(this, this.method, this.location, this.body, this.formElement)
     this.mustRedirect = mustRedirect
   }
@@ -70,10 +75,6 @@ export class FormSubmission {
   get action(): string {
     const formElementAction = typeof this.formElement.action === 'string' ? this.formElement.action : null
     return this.submitter?.getAttribute("formaction") || this.formElement.getAttribute("action") || formElementAction || ""
-  }
-
-  get location(): URL {
-    return expandURL(this.action)
   }
 
   get body() {
@@ -219,4 +220,18 @@ function getMetaContent(name: string) {
 
 function responseSucceededWithoutRedirect(response: FetchResponse) {
   return response.statusCode == 200 && !response.redirected
+}
+
+function mergeFormDataEntries(url: URL, entries: [string, FormDataEntryValue][]): URL {
+  const searchParams = new URLSearchParams
+
+  for (const [ name, value ] of entries) {
+    if (value instanceof File) continue
+
+    searchParams.append(name, value)
+  }
+
+  url.search = searchParams.toString()
+
+  return url
 }
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -236,7 +236,7 @@ export class FrameController implements AppearanceObserverDelegate, FetchRequest
   // Private
 
   private async visit(url: URL) {
-    const request = new FetchRequest(this, FetchMethod.get, url, url.searchParams, this.element)
+    const request = new FetchRequest(this, FetchMethod.get, url, new URLSearchParams, this.element)
 
     this.currentFetchRequest?.cancel()
     this.currentFetchRequest = request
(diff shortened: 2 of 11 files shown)


dhh:
Seems like this is still broken if there's a get-form with params in the URL. Like: `<form action="/show?var=1" method="get">`. The fix appears to only be working for ahrefs?

seanpdoyle:
https://github.com/hotwired/turbo/pull/461 and https://github.com/hotwired/turbo/pull/464 aimed to cover `<form method="get">` whose `[action]` encoded query parameters. 

They was guided by the to the HTML Specification's [§ 4.10.21.3 Form submission
algorithm][] section, submissions transmitted as `GET` requests [mutate
the `[action]` URL][mutate], overriding any search parameters already
encoded into the `[action]` value:

> [Mutate action URL][algorithm]
> ---
>
> Let <var>pairs</var> be the result of converting to a list of
> name-value pairs with <var>entry list</var>.
>
> Let <var>query</var> be the result of running the
> `application/x-www-form-urlencoded` serializer with <var>pairs</var>
> and <var>encoding</var>.
>
> Set <Var>parsed action</var>'s query component to <var>query</var>.
>
> Plan to navigate to <var>parsed action</var>.

Are the query parameters included, or ignored? 

[§ 4.10.21.3 Form submission algorithm]: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-algorithm
[algorithm]: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#submit-mutate-action
[mutate]: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#form-submission-algorithm:submit-mutate-action

dhh:
The query parameters are ignored. We just end up doing a visit to `/show` if the action is `/show?key=value` on a form with method `get`.

seanpdoyle:
If Turbo were removed or JavaScript were disabled, does the browser include or omit the `?key=value` from the next page's URL?

dhh:
It's included. And it used to be with Turbo as well. I found this as a regression in a system test that verified that this used to work.

seanpdoyle:
>  it used to be with Turbo as well

I can confirm this. The motivation behind #461, #464, and #466 (this PR) was to make Turbo's `<form method="get">` behavior align with built-in browser behavior.

In an environment without Turbo, the encoded values are ignored. Consider this code sample:

```html
<!-- index.html -->
<html>
  <head></head>
  <body>
    <form action="/destination.html">
      <button>form[action="destination.html"]</button>
    </form>
    <form action="/destination.html?from=index">
      <button>form[action="destination.html?from=index"]</button>
    </form>
    <form action="/destination.html?from=index">
      <button name="foo" value="bar">form[action="destination.html?from=index"] button[name="foo"][value="bar"]</button>
    </form>
    <form action="/destination.html">
      <input type="hidden" name="foo" value="bar">
      <button>form[action="destination.html"] input[type="hidden"][name="foo"][value="bar"]</button>
    </form>
    <script>document.body.insertAdjacentText("beforeend", window.location.toString())</script>
  </body>
</html>

<!-- destination.html -->
<html>
  <head></head>
  <body>
    <form action="/index.html">
      <button>form[action="index.html"]</button>
    </form>
    <form action="/index.html?from=destination">
      <button>form[action="index.html?from=destination"]</button>
    </form>
    <form action="/index.html?from=destination">
      <button name="foo" value="bar">form[action="index.html?from=destination"] button[name="foo"][value="bar"]</button>
    </form>
    <form action="/index.html">
      <input type="hidden" name="foo" value="bar">
      <button>form[action="index.html"] input[type="hidden"][name="foo"][value="bar"]</button>
    </form>
    <script>document.body.insertAdjacentText("beforeend", window.location.toString())</script>
  </body>
</html>
```

When navigating the `<form>` elements that encode values into their `[action]` as query strings, those values are omitted from the next page's URL. 

From what I'm experiencing, values are only included in the next page's URL when encoded onto the submitter as `[name]` and `[value]` pairs, or encoded as an `<input type="hidden" name="..." value="...">` element.

You can experiment with the sandbox by visiting https://fuschia-abyssinian-planarian.glitch.me.

Is that a proper recreation of your circumstances?

dhh:
Oh wow. Yeah, you're right. This was just a quirk of Turbo's. I guess it's a bit of regression on the Turbo side of things, but agree that it seems reasonable to match the native HTML experience. Thanks for the very thorough illumination ✌️. All good letting things stay as they are.

Issue #408, opened 2021-09-26
turbo:after-stream-render event


dwightwatson:
We've got `turbo:before-stream-render` but it would also be helpful to have an event fire after the rendering has occurred as well.

My use-case is that the new DOM element is rendered into a scrollable window, so I'd like to scroll the new element into view after it's been rendered. It doesn't appear as though there is any event I can currently hook into that will let me perform this action after the DOM update has occurred.

seanpdoyle:
Is this a duplicate of https://github.com/hotwired/turbo/issues/92 and https://discuss.hotwired.dev/t/event-to-know-a-turbo-stream-has-been-rendered/1554/6?

dwightwatson:
Apologies, I did search but nothing came up.

woto:
It helped me https://github.com/hotwired/turbo/pull/20#issue-773325276. Didn't found any drawbacks yet.

Issue #92, opened 2021-01-08
Add new event turbo:after-stream-render


lu4p:
Similar to the `turbo:before-stream-render` event, it would be useful to fire an event after rendering a Turbo Stream page update

sstephenson:
We don't plan on adding an after-stream-render event: https://discuss.hotwire.dev/t/event-to-know-a-turbo-stream-has-been-rendered/1554/6

woto:
It helped me https://github.com/hotwired/turbo/pull/20#issue-773325276. Didn't found any drawbacks yet.

modusss:
The big problem in my case is that after turbo_stream is loaded, the javacript library input_mask doesn't work anymore because the content that changes is the form itself. 

They brought to us a solution with Stimulus and "connect()" method b but it does not trigger after turbo stream is loaded. 

There is nothing that triggers after page changes.

jtoy:
ran into similar issue  and stimulus is not the answer for me.

searls:
Agree with @leewa, this would be of tremendous use to me. 

I have this situation:

* wrapping div (with a Stimulus controller connected)
    * a `select` input to change modes (where `change` triggers an action that makes a `responseKind: 'turbo-stream'` request using `'@rails/request.js'`
    * a `fieldset` containing some fields (that gets re-rendered by that turbo-stream update)
    * a `input[type=hidden]` that contains a JSON-serialized dump of the above fieldset (which is updated any time the user changes anything)

The stimulus controller just needs to know when that `fieldset` has been re-rendered so it can re-serialize its contents into that `input[type=hidden]`. Without this event, I'm not left with a lot of awesome options. Manual `MutationObserver`? Reaching upward to the wrapping form's `submit` event? This event would be great to have

ricvillagrana:
I did this and worked perfectly for what I needed. If you need to update the event listeners on a controller, you can do this:

```javascript
export default class extends Controller {
  static targets = ['button'];

  connect() {
    this.addEventListeners();

    const observer = new MutationObserver(() => {
      this.addEventListeners();
    });

    observer.observe(this.element, { childList: true });
  }

  addEventListeners() {
    this.buttonTarget.addEventListener('click' () => { /* Do something */ })
  }
}
```

tomashomey:
here is a solution: https://nts.strzibny.name/rails-turbo-after-stream-render/

```
// application.js
const afterRenderEvent = new Event("turbo:after-stream-render");
[...]

PR #474, opened 2021-11-24
Introducing the turbo:after-stream-render event

kapantzak:
This PR introduces the `turbo:after-stream-render event` event that gets fired right after any stream action has been performed.

**Why?**
Someone might need to perform an action based on an element that has been rendered by a stream response. We cannot rely on `turbo:before-stream-render` event for this kind of action because the requested element probably had not been rendered yet.


seanpdoyle:
Is this a duplicate of #92, #408, and https://discuss.hotwired.dev/t/event-to-know-a-turbo-stream-has-been-rendered/1554/6?

kapantzak:
Hi Sean, to be honest, I wasn't aware of the already closed issues.

What led me to open a PR about this event was the case where we wanted to call a function a soon as a stream was rendered. We thought of the stimulus solution that you propose here, but, as soon as there are quite a lot of partials that get rendered via this stream, we would have to bind the stimulus controller to many places inside the code. Instead, if we had the `turbo:after-stream-render` event available, we could add an event listener once, and handle any event occurrence accordingly. In other words, this would help us to minimize the interventions inside the codebase.

Now, regarding the fact that the event gets fired on the document, we could change this and fire it on the stream element itself.

acetinick:
I have some situations where this is very useful,  namly initializing UI components after the stream renders.  eg. Bootstrap tooltips etc.

samstickland:
Hi @kapantzak,

Did you get any further on this PR. We have the same use case as @acetinick, where we want to bind JS for UI controls in the returned HTML. This would be much more efficient if the event fired on the stream element, but it would also work for us if it was only on the document.

kapantzak:
Hi @samstickland ,

I think that it's better to fire this event on the stream element, as you said. I will try to continue with this asap!

kapantzak:
I have made some changes so that the event is triggered from the target element, but I see that some irrelevant tests fail

woto:
It helped me https://github.com/hotwired/turbo/pull/20#issue-773325276. Didn't found any drawbacks yet.

Issue #400, opened 2021-09-18
Anchor links appear at bottom of page instead of top


tao:
I found that on Firefox / Firefox mobile / Chrome that when I use an anchor link, sometimes it loads it at the bottom of the page instead of the top.

```
--------------
| #anchor    |   <-- expected position after page load
|            |
|            |
|            |
|            |
--------------
```

```
--------------
|            |
|            |
|            |
|            |
| #anchor    |   <-- the anchor appears at the bottom instead
--------------
```

I believe I've narrowed this down to the following css:

```css
        scroll-behavior: smooth;
```

If I remove this scroll-behavior then the issue disappears but my app requires smooth scroll so I don't disorientate the user when jumping between paragraphs.

My pages are long articles with anchors for each paragraph and they are numbered starting at anchor#1 up to anchor#40 for example.

If I am at anchor#3 and click on a higher number like anchor#15 it'll scroll down and put that anchor at the bottom of the page... so it must be scrolling until the anchor is in-screen but not to the top of the viewport.

If I am at anchor anchor#35 for example, and click a lower number like anchor#4 then it looks like it scrolls up the page and puts the anchor at the top of the page because it's higher up. This must be because the anchor only appears into view at the top of the page and then it stops. This bug happens when you load new pages, but not when jumping between paragraphs on the same page.

So I only found the problem going from lower numbers like anchor#3 to a higher one like anchor#15 where the next link has a much higher `Y` position on the new page. As soon as the anchor appears at the bottom of the page it must stop the scroll calculation and leave the anchor at the bottom instead of the top of viewport.

tao:
Here's a demo (https://github.com/tao/turbo-scroll-bug). 

1. Start a server: `php -S localhost:9000` and visit `http://localhost:9000/one.html` 
2. Scroll down to the bottom and click the link to `two.html#10` (works fine)
3. Click the link to go back to `one.html#25`  (the paragraph will be at the bottom of the page instead of the top)

This happens on Firefox. On Chrome, the turbo link behaviour seems to puts the link in the center of both pages.

seanpdoyle:
Thank you for providing such a simple example @tao!

I've opened https://github.com/hotwired/turbo/pull/476. Locally, I was unable to re-produce the issue with those changes.

For now, that PR is only the implementation changes.  If this resolves your issue, I'll try and transform your example repository into the Turbo test suite.

tao:
[...]

PR #476, opened 2021-11-25
Perform scrolling prior to Visit completion

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/400

When processing a `Visit`, invoke `Visit.performScroll()` while loading
a response, loading a Snapshot, or executing a same-page anchor
navigation. After in-lining those calls to more appropriate points in
the Visit lifecycle, this commit removes the `performScroll()` call from
the `Visit.render()` method.

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -258,6 +258,7 @@ export class Visit implements FetchRequestDelegate {
         if (this.view.renderPromise) await this.view.renderPromise
         if (isSuccessful(statusCode) && responseHTML != null) {
           await this.view.renderPage(PageSnapshot.fromHTMLString(responseHTML), false, this.willRender, this)
+          this.performScroll()
           this.adapter.visitRendered(this)
           this.complete()
         } else {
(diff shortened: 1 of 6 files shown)


seanpdoyle:
This is a draft in an attempt to resolve the issues causing https://github.com/hotwired/turbo/issues/400. If these changes fix the issue, I'll create a browser test and upgrade this to Ready for Review.

seanpdoyle:
@srt32 If this change is deemed viable, I'd appreciate if you could double-check to make sure it doesn't undo the fix you shipped in https://github.com/hotwired/turbo/pull/571.

seanpdoyle:
@srt32 since this follows-up the changes you made in https://github.com/hotwired/turbo/pull/571, would you be able to double-check that it doesn't reverse the behavior fixes or introduce new quirks?

petrsigut:
This is awesome!

We tested this branch and it fixes all problems we have in our chat app, where we need to scroll to the bottom after Turbo visit😇

PR #477, opened 2021-11-26
Adds an update-attribute stream action so you can update an attribute on targets with a value

yunggindigo:
This will be super useful for updating attributes on elements and save a lot of JavaScript used to do that. I will also do a pr to match on the hotwire-rails gem to add a new helper to use.
will be used like so `<turbo-stream action="update-attribute" target=".items" attribute="data-seen-value" value="true">
</turbo-stream>`


ParamagicDev:
I have no idea what the end-goal for streams is, but they seem heavily focused on CRUD based operations. I worry that if this gets introduced it opens up Pandora's box for a ton of additional operations and vastly increased the scope.

In the mean time, if you're looking for more fine-grained control of the DOM from the server, you could take a peek over at CableReady and see if that suits your needs.

https://cableready.stimulusreflex.com/reference/operations/element-mutations#set_attribute

seanpdoyle:
As an alternative to broadening official support for more `turbo-stream[action]` values, I've opened https://github.com/hotwired/turbo/pull/479 to explore other options.

yunggindigo:
> As an alternative to broadening official support for more `turbo-stream[action]` values, I've opened #479 to explore other options.

I like this I guess it makes sense to scope this to ones personal needs rather than expand the framework.

yunggindigo:
> I have no idea what the end-goal for streams is, but they seem heavily focused on CRUD based operations. I worry that if this gets introduced it opens up Pandora's box for a ton of additional operations and vastly increased the scope.
> 
> In the mean time, if you're looking for more fine-grained control of the DOM from the server, you could take a peek over at CableReady and see if that suits your needs.
> 
> https://cableready.stimulusreflex.com/reference/operations/element-mutations#set_attribute

It would do what I need perfectly. I was just hoping for an end goal where I can easily do something like turbo_stream.update_attribute("message_2", attribute: "data-seen-value", value: true) and make it really easy like it is to update html with hotwire.

julianrubisch:
> It would do what I need. I was just hoping for an end goal where I can easily do something like `turbo_stream.update_attribute("message_2", attribute: "data-seen-value", value: true)` and make it really easy like it is to update html with hotwire.

why some folks are striving for a complex solution when a simple one is already in place and battle proven in countless production apps is beyond me 🤷‍♂️

seanpdoyle:
> I have no idea what the end-goal for streams is, but they seem heavily focused on CRUD based operations. I worry that if this gets introduced it opens up Pandora's box for a ton of additional operations and vastly increased the scope.

Thematically, #479 shares similar aspirations to https://github.com/hotwired/turbo/pull/431. They aim to strike a balance between improving flexibility, limiting scope, and reducing any maintenance burden.

While they aren't exhaustive by any stretch of the imagination, the current collection of `turbo-stream[action]` values provide a collection of operations can provide applications with a sufficient level of dynamism and interactivity when supplemented with other building blocks like page-scoped redirects, `<turbo-frame>`-scoped redirects, and bespoke Stimulus controllers.

#479 and #431 aim to provide applications and other projects in the ecosystem with opportunities to integrate with and extend Streams and Frames in ways that aren't currently possible.

Could introducing those seams and customization touch-points provide enough flexibility to outsource particular bits to more specialized tools like CableReady or StimulusReflex?

yunggindigo:
closed because #479 adds the flexibility to add a custom action like this inside of our applications

PR #478, opened 2021-11-26
Adds append_unless_duplicate and prepend_unless_duplicate

dcyoung-dev:
When appending or prepending content we want to control what
the action should do if there are already matching elements
on the page.
This is especially true when responding and broadcasting streams
contain the same data.

Additional actions added to skip the action if elements are
already present. In this case it would be recommended to use
the `replace` action.


seanpdoyle:
As an alternative to broadening official support for more `turbo-stream[action]` values, I've opened https://github.com/hotwired/turbo/pull/479 to explore other options.

dcyoung-dev:
Closed in favour of #479

PR #479, opened 2021-11-26
Custom Actions: Export `StreamActions` module

seanpdoyle:
Closes https://github.com/hotwired/turbo/pull/478
Closes https://github.com/hotwired/turbo/pull/477

As an alternative to continuing to introduce support for more varied and
specialized `<turbo-stream action="...">` values, this commit exports
the `StreamActions` module from the root-module in order to provide
applications with direct access.

Applications can add new actions or override existing actions, so long
as they adhere to the `StreamAction` interface: a function whose `this`
context is an instance of `StreamElement`.

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -30,6 +30,8 @@ export { TurboSubmitStartEvent, TurboSubmitEndEvent } from "./drive/form_submiss
 export { TurboBeforeFetchRequestEvent, TurboBeforeFetchResponseEvent } from "../http/fetch_request"
 export { TurboBeforeStreamRenderEvent } from "../elements/stream_element"
 
+export { StreamActions } from "./streams/stream_actions"
+
 /**
  * Starts the main session.
  * This initialises any necessary observers such as those to monitor
--- a/src/tests/functional/stream_tests.ts
+++ b/src/tests/functional/stream_tests.ts
@@ -33,6 +33,24 @@ test("test receiving a stream message with css selector target", async ({ page }
   assert.equal(await element[1], "Hello CSS!")
 })
 
+test("test overriding with custom StreamActions", async ({ page }) => {
+  const html = "Rendered with Custom Action"
+
+  await page.evaluate((html) => {
+    window.Turbo.StreamActions.customUpdate = function () {
+      for (const target of this.targetElements) target.innerHTML = html
+    }
+    document.body.insertAdjacentHTML(
+      "afterbegin",
+      `<turbo-stream action="customUpdate" target="messages">
+        <template></template>
+      </turbo-stream>`
+    )
+  }, html)
+
+  assert.equal(await page.textContent("#messages"), html, "evaluates custom StreamAction")
+})
+
 test("test receiving a stream message asynchronously", async ({ page }) => {
   let messages = await page.locator("#messages > *").allTextContents()
 


seanpdoyle:
This is a draft because a feature like this is not entirely free. By opening up the innards of Turbo's support for `<turbo-stream>` elements, there will be an increased maintenance burden, since the module name will be part of the public API. This could have implications on work like https://github.com/hotwired/turbo/pull/146 that aims to make `StreamActions` more generally applicable outside the `<turbo-stream>` elements.

Similarly, since the `turbo:before-stream-render` event fires with the `document` as its target, the `StreamElement` interface is currently internal private API, and inaccessible from client applications. This change would change that by making instances of `StreamElement` available to custom rendering actions.

seanpdoyle:
Instead of exporting the module, a different variation of this change could involve dispatching the `turbo:before-stream-render` with an `event.detail.render` function that applications can override in the same way.

That would align with the pattern that https://github.com/hotwired/turbo/pull/431 is aiming to establish.

yunggindigo:
> Instead of exporting the module, a different variation of this change could involve dispatching the `turbo:before-stream-render` with an `event.detail.render` function that applications can override in the same way.
> 
> That would align with the pattern that #431 is aiming to establish.

would this be overriding by checkin

> Instead of exporting the module, a different variation of this change could involve dispatching the `turbo:before-stream-render` with an `event.detail.render` function that applications can override in the same way.
> 
> That would align with the pattern that #431 is aiming to establish.

I like the idea of having the stream actions being exported so that we can additional stream actions defined inside our applications. Then it would be okay to use `turbo_stream.action(our_custom_action)` to create a turbo_stream element in ruby

seanpdoyle:
> I like the idea of having the stream actions being exported so that we can have additional stream actions defined inside our applications.

That's good feedback! I'm thinking about the developer experience in terms of when would be best to intervene.

I'll share some pseudo-code to demonstrate the two styles.

First, consider manipulating the `StreamActions` object directly:

```js
import { Turbo } from "@hotwired/turbo"

const { update } = Turbo.StreamActions

const customActions = {
  // override the default `update`
  update() {
    doSomethingBeforeCallingUpdate(this.targets)
    update()
  },

  myCustomAction() {
    // ...
  }
}

Turbo.StreamActions = { ...Turbo.StreamActions, ...customActions }
```

Next, consider doing so within an event listener. This example assumes that `event.detail.render` is a reference to the default function that corresponds to the action, and `relatedTarget` is a reference to the `<turbo-stream>` element. The event is dispatched with `target` as the `document` since it's _before_ the `<turbo-stream>` element is inserted anywhere into the DOM:

```js
addEventListener("turbo:before-stream-render", ({ relatedTarget, detail }) => {
  const { render } = detail

  const customActions = {
    update() {
      doSomethingBeforeUpdate(this)
      render()
    },
    myCustomAction() {
      callAnotherLibraryWith(this)
    }
  }

  event.detail.render = customActions[target.action] || function() { console.error("Failed to handle action", this) }
})
```

Just spitballing here. The event listener version is more verbose, but grants more of an opportunity to incorporate contextual information like the state of the document, the `<turbo-stream>` element, etc. The overrides are scoped, and un-mounting the event listener would "reverse" any changes made. It doesn't commit to a name for the `StreamActions` object export, and it limits the "public" interface to a single `event.detail.render()` function.

It would still require committing to support some form of a `StreamElement` public interface.

>  Then it would be okay to use `turbo_stream.action(our_custom_action)` to create a turbo_stream element in ruby

Regardless of _when_ the client-side extension happens, changes to support something like `turbo_stream.action(our_custom_action)` would need to be made in the server-side library (either officially, or by mixing-in changes at the application level).

As far as I know, there isn't any coercion or validation  happening at the [`hotwired/turbo-rails` layer](https://github.com/hotwired/turbo-rails/blob/v0.9.0/app/helpers/turbo/streams/action_helper.rb#L12-L22) that would prevent a response or broadcast from transmitting a `<turbo-stream action="our_custom_action">` value. If that's the case, I don't see the client-side design decisions having an affect on the server-side construction of the element.

yunggindigo:
I think the first option would be better for defining reusable stream actions and being able to use it for something like #477.

seanpdoyle:
It's currently possible, without any internal or external changes to Turbo, to implement custom Turbo Stream actions.

As an alternative to extending `StreamActions`, it might be worth exploring and documenting how to integrate an application-specific action with the current plumbing.

Turbo Stream elements expect their first child to be a `<template>` element. There are no restrictions about the descendants of that `<template>` or their contents. In the case of an `[action="append"]`, the entirety of the `<template>` is appended into the element with an `[id]` that matches the `[target="..."]` value.

If that content were itself a `<template>` element that declared a `[data-controller="custom-stream"]` attribute, the corresponding `custom-stream` controller's `connect()` hook could implement any bespoke DOM operation code.

Consider the following example code (available as a JSFilddle: https://jsfiddle.net/o1kLsq6p/)

```html
<html>
  <head>
    <script type="module">
      import "https://unpkg.com/@hotwired/turbo@7.1.0"
      import { Application, Controller } from "https://unpkg.com/@hotwired/stimulus@3.0.1/dist/stimulus.js"

      const Stimulus = Application.start()

      Stimulus.register("custom-stream", class extends Controller {
        static get values() { return { action: String } }
       
        connect() {
        const { content, parentElement } = this.element
        switch (this.actionValue) {
          case "append": {
            parentElement.append(content)
            parentElement.style.color = "red"
            break
          }
          default: console.error(`Don't know how to handle ${this.actionValue}!`)
        }
      }
    })
  </script>
  </head>
  <body>
    <ul id="my_list">
      <li>First Element</li>
    </ul>
  </body>
</html>
```

Consider receiving the following `<turbo-stream>` (either in response to a Form Submission or from a broadcast over Action Cable):

```html
<turbo-stream action="append" target="my_list">
  <template>
    <template data-controller="custom-stream" data-custom-stream-action-value="append">
      <li>Another Element</li>
    </template>
  </template>
</turbo-stream>
```

When executed, the `[action="append"]` and `[target="my_list"]` pairing would direct the Stream to append the `<template>` element to the `<ul id="my_list">` document. Once appended, the `custom-stream` controller's `connect()` hook would fire. That connect reads from the controller's values to decide the details of the operation (in this case, the operation is a custom `"append"`).

The `connect()` hook executes the operation, extracting [`this.element.content`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTemplateElement/content), appending the content to the `<ul>` element, and marking the list to have `"red"` text (as a demonstration of the "custom" part of the action). The operation knows which element to "target" because the `<template>` was appending into the `<ul>`. This is another avenue of customization, since callers could append the `<template>` to the `<body>` element, then encode whatever selectors or element resolution logic they'd need into the `<template>` element's attributes.

The `<template>` element itself is inert unless acted upon. Appending or prepending it onto the page modifies the structure of the DOM, but doesn't "render" any visual changes. That behavior could be leveraged to encode additional information alongside the `[data-controller]` declaration. For example, its own `[data-custom-stream-action-value]`, `[data-custom-stream-target]`, etc. There's nothing preventing the element from encoding JSON objects of options that are ready to be forwarded along to another library like Morphdom.

yunggindigo:
This proves that stimulus + the default stream actions are enough to build any functionality that we desire even having custom dom actions.

seb-jean:
What remains to be done to finish this PR?

dhh:
After letting this marinate for a while, I like the simplicity of exporting the stream actions, and allowing the app to extend as they see fit. We should document it as a sharp knife. But I like [sharp knives](https://rubyonrails.org/doctrine#provide-sharp-knives) 😄.

dhh:
Should add a doc PR to explain how to actually extend the actions 👍

seanpdoyle:
> Should add a doc PR to explain how to actually extend the actions 

@dcyoung-dev @enoch-tamulonis since your original issues were the inspiration for this change, would either of you be interested in opening a pull request against https://github.com/hotwired/turbo-site/ outlining a use case or two that would motivate a customization? @marcoroth maybe you're interested as well?

The how-to is captured by the `page.evaluate` block in the test suite. Since the `function()` is evaluated in the context of the `<turbo-stream>` Custom element, it's important that it's declared as `function(){ ... }` and not `() => ...`.

dhh:
Rebase needed?

PR #483, opened 2021-11-30
Defensively create custom turbo elements

### LINKED ISSUE #188: 404 causes DOMException
I'm using the latest version of Turbo (beta 4 at the time of writing) and I've noticed that when you hit a 404 page it causes the following error in the console:
```js
Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "turbo-frame" has already been used with this registry
```
This only happens when hitting a 404 from another link... if you refresh the non existent page the error doesn't happen.

dphaener:
There are instances when turbo attempts to create the frame and stream
elements when they have already been registered. This is easily
reproducible in development when using webpacker by following these steps:

* Make a change to any file that webpacker compiles.
* Click a turbo enabled link.
* Boom.

When Turbo replaces the page it appears as if it reloads the Turbo
library and attempts to create the custom elements again. I have also
observed this behavior in my production app via my exception monitoring
service, but have been unable to reproduce this behavior in production
on my own.

This change simply checks for the existence of the custom element prior
to to attempting to define it. The `get` method used here explicitly
returns `undefined` when the element does not yet exist so it is safe to
use the strict equality operator here. Reference:
https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry/get

Fixes #188

### DIFF
--- a/src/elements/index.ts
+++ b/src/elements/index.ts
@@ -7,5 +7,10 @@ FrameElement.delegateConstructor = FrameController
 export * from "./frame_element"
 export * from "./stream_element"
 
-customElements.define("turbo-frame", FrameElement)
-customElements.define("turbo-stream", StreamElement)
+if (customElements.get("turbo-frame") === undefined) {
+  customElements.define("turbo-frame", FrameElement)
+}
+
+if (customElements.get("turbo-stream") === undefined) {
+  customElements.define("turbo-stream", StreamElement)
+}


dphaener:
Update: This happens reliably in production when the webpack bundle has changed (after a deploy) and a turbo enabled link is clicked.

dphaener:
Yet another update: After updating my application to use jsbundling-rails and esbuild I am no longer seeing this error on deploys (or at all for that matter).

t27duck:
Is that with or without this patch with your update to jsbundling-rails?

dphaener:
@t27duck Without the patch. I honestly don't know why. 🤷

mbardelmeijer:
We can also reproduce this in production where on deploy, the bundle changes and this error is triggered after clicking on any turbo link. Would love to see this merged.

Issue #429, opened 2021-10-25
Eager/lazy loaded turbo-frames are reloaded after snapshot restoration


tobyzerner:
Navigating away from a page, then going back, causes all eager/lazy loaded turbo frames to be reloaded on the original page after the snapshot is restored.

Simple example:

```html
<turbo-frame id="page_1" target="_top">
  <a href="/posts/1">Post 1</a>
  <turbo-frame id="page_2" src="?page=2" loading="lazy" target="_top"></turbo-frame>
</turbo-frame>
```

When you visit the page initially, the #page_2 frame lazy loads as expected. If you click on "Post 1" and then go back, after the page snapshot is restored, the #page_2 frame reloads, which is undesirable.

tleish:
One way to do this would be to remove `src="?page=2"` from the turbo-frame after it renders.  There might be some cases where you wouldn't want to do this by default.

As with many other solutions, you could easily control this with a simple stimulus controller.  Something like:

```javascript
class AttributeController extends Controller {
  remove({ params: {name}}) {
    this.element.removeAttribute(name);
  }
}
```

```html
<turbo-frame id="page-2020-12" src="page2.html" data-controller="attribute" data-action="turbo:frame-load->attribute#remove" data-attribute-name-param="src">
  <a href="page2.html">Page 2s</a>
</turbo-frame>
```

Or, if you wanted to do it for all your turbo-frames, you could just add a global event listener:

```javascript
document.addEventListener('turbo:frame-load', function({ srcElement }) {
  srcElement.removeAttribute('src');
});
```

tobyzerner:
@tleish Thanks, that's a great workaround!

seanpdoyle:
@tobyzerner I recognize that the example you shared is pseudo-code, but in the real scenario, is `<turbo-frame id="page_2">` visible in the viewport when you navigate back?

tobyzerner:
@seanpdoyle Yes that's correct – it's an infinite scroll use case. When you navigate back, the page scroll is restored and so the `page_2` frame is immediately in view, and Turbo reloads it.

The `[disabled]` attribute is a nice workaround too. However, I just note I still think of these as "workarounds" – intuitively I would expect the frame to not be reloaded on restoration, without having to intervene, hence why I didn't close this one just yet. Is there an easy/sensible path to preserving the frame's state across restoration visits by default (perhaps by adding something like a `[loaded]` attribute when the frame has loaded), or is it too much complexity for too little benefit?

[...]

PR #487, opened 2021-12-01
Expose Frame load state via `[complete]` attribute

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/429

---

Introduce the `<turbo-frame complete>` boolean attribute. The attribute's
absence indicates that the frame has not yet been loaded, and is ready
to be navigated. Its presence means that the contents of the frame have
been fetch from its `[src]` attribute.

Encoding the load state into the element's HTML aims to integrate with
Snapshot caching. Once a frame is loaded, navigating away and then
restoring a page's state from an Historical Snapshot should preserve the
fact that the contents are already loaded.

For both `eager` and `lazy` loaded frames, changing the element's
`[src]` attribute (directly via JavaScript, or by clicking an `<a>`
element or submitting a `<form>` element) will remove the `[loaded]`
attribute. Eager-loaded frames will immediately initiate a request to
fetch the contents, and Lazy-loaded frames will initiate the request
once they enter the viewport, or are changed to be eager-loading.

When the `[src]` attribute is changed, the `FrameController` will only
remove the `[complete]` attribute if the element [isConnected][] to the
document, so that the `[complete]` attribute is not modified prior to
Snapshot Caching or when re-mounting a Cached Snapshot.

The act of "reloading" involves the removal of the `[complete]` attribute,
which can be done either by `FrameElement.reload()` or
`document.getElementById("frame-element").removeAttribute("complete")`.

A side-effect of introducing the `[complete]` attribute is that the
`FrameController` no longer needs to internally track:

1. how the internal `currentURL` value compares to the external
  `sourceURL` value
2. whether or not the frame is "reloadable"

By no longer tracking the `sourceURL` and `currentURL` separately, the
implementation for the private `loadSourceURL` method can be simplified.
Since there is no longer a `currentURL` property to rollback, the `try {
... } catch (error) { ... }` can be omitted, and the `this.sourceURL`
presence check can be incorporated into the rest of the guard
conditional.

Finally, this commit introduce the `isIgnoringChangesTo()` and
`ignoringChangesToAttribute()` private methods to disable
FrameController observations for a given period of time. For example,
when setting the `<turbo-frame src="...">` attribute, previous
implementation would set, then check the value of a
`this.settingSourceURL` property to decide whether or not to fire
attribute change callback code. This commit refines that pattern to
support any property of the `FrameController`, including the
`"sourceURL"` or `"complete"` value. When making internal modifications to
those values, it's important to temporarily disable observation
callbacks to avoid unnecessary requests and to limit the potential for
infinitely recursing loops.

[isConnected]: https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -82,7 +82,7 @@ export class FrameController
     if (this.isIgnoringChangesTo("src")) return
 
     if (this.element.isConnected) {
-      this.loaded = false
+      this.complete = false
     }
 
     if (this.loadingStyle == FrameLoadingStyle.eager || this.hasBeenLoaded) {
@@ -90,8 +90,8 @@ export class FrameController
     }
   }
 
-  loadedChanged() {
-    if (this.isIgnoringChangesTo("loaded")) return
+  completeChanged() {
+    if (this.isIgnoringChangesTo("complete")) return
 
     this.loadSourceURL()
   }
@@ -106,7 +106,7 @@ export class FrameController
   }
 
   private async loadSourceURL() {
-    if (this.enabled && this.isActive && !this.loaded && this.sourceURL) {
+    if (this.enabled && this.isActive && !this.complete && this.sourceURL) {
       this.element.loaded = this.visit(expandURL(this.sourceURL))
       this.appearanceObserver.stop()
       await this.element.loaded
@@ -127,7 +127,7 @@ export class FrameController
         const renderer = new FrameRenderer(this.view.snapshot, snapshot, false, false)
         if (this.view.renderPromise) await this.view.renderPromise
         await this.view.render(renderer)
-        this.loaded = true
+        this.complete = true
         session.frameRendered(fetchResponse, this.element)
         session.frameLoaded(this.element)
         this.fetchResponseLoaded(fetchResponse)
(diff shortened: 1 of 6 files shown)


seanpdoyle:
It's worth highlighting that a `[loaded]` element attribute collides with the [`FrameElement.loaded`](https://turbo.hotwired.dev/reference/frames#properties) property that exposes the underlying `Promise` instance used during loading.

This is probably a bad collision for a Boolean attribute, since `myFrame.loaded` will never evaluate to `false`, even when the  `<turbo-frame>` element is missing the attribute _and_ when the frame isn't loaded (since [the default value for `loaded` is `Promise.resolve()`](https://github.com/hotwired/turbo/blob/aa9724dd23309a56b3e7c6ee62c673fa63799038/src/elements/frame_element.ts#L37).

We have two options:

1. Rename the `FrameElement.loaded` property and replace it with a `.loaded` boolean reader that wraps the `[loaded]` attribute
2. Rename the `[loaded]` attribute and introduce (or omit) a `FrameElement.thatProperty` reader and writer.

Options 1. involves breaking changes.

Option 2. is viable, but `[loaded]` is a really good attribute name! The front-runner competitor is probably `[complete]`, since it'd correspond to the [`FrameElement.complete`](https://turbo.hotwired.dev/reference/frames#properties) property, which itself draws inspiration from the [HTMLImageElement.complete](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/complete) property.

tobyzerner:
Hi @seanpdoyle, thanks for this PR. I can confirm that this does fix the issue I was having in #429.

I agree that `[loaded]` would be the best name for the attribute. I always found the `FrameElement.loaded` property referring to a `Promise` a bit counter-intuitive anyway – I would've expected it to return a boolean value before reading the docs. However if it's desirable to maintain BC in this case, that would be very understandable, and `[complete]` is a good runner-up.

dhh:
I think [complete], and it's parity to a similar attribute on image, is excellent. In some ways even preferable. Let's roll with that!

PR #501, opened 2021-12-18
Progress Bar Won't Render if CSP is Turned On

rience:
Progress Bar injects inline `<style />` element to page. However, when Content Security Policy is turned on and `nonce` is generated - Progress Bar element ignores that.

This PR fixes that and checks if `csp-nonce` is set in `meta` tags. Then it uses it to create inline style.

Update: I don't know why test is failing and is generating `<style />` without `nonce` being set. However, when testing manually - this fix works. I might need some help here from someone more experienced in writing JS tests.

### DIFF
--- a/src/core/drive/progress_bar.ts
+++ b/src/core/drive/progress_bar.ts
@@ -110,6 +110,9 @@ export class ProgressBar {
     const element = document.createElement("style")
     element.type = "text/css"
     element.textContent = ProgressBar.defaultCSS
+    if (this.cspNonce) {
+      element.nonce = this.cspNonce
+    }
     return element
   }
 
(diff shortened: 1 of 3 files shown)


DavidPetrasek:
> Update: I don't know why test is failing and is generating <style /> without nonce being set. However, when testing manually - this fix works.

@rience Even with meta tag containing the nonce in the head:
`<meta name="csp-nonce" content="EAcP7YZ0Z/753crufRT+Rg==">`
...it still doesn't work and on page refresh in production environment it gives me this error:
`Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self' https://www.gstatic.com https://fonts.googleapis.com". Either the 'unsafe-inline' keyword, a hash ('sha256-WAyOw4V+FqDc35lQPyRADLBWbuNK8ahvYEaQIYF1+Ps='), or a nonce ('nonce-...') is required to enable inline execution.`

Temporary solution is to use the hash which works.
I am using:
@hotwired/stimulus@3.2.2
@hotwired/turbo@8.0.5

PR #507, opened 2021-12-24
Add before callback to form link

condef5:
This pr adds a callback to the form link that is executed in `convertLinkWithMethodClickToFormSubmission`.

With this callback we could add some custom behavior like this:
```js
Turbo.setLinkToFormBeforeCallback((form) => {
  if (form.getAttribute('method') !== 'delete') return;
  
  const input = document.createElement("input");
  input.setAttribute("name", "_method")
  input.setAttribute("type", "hidden")
  input.setAttribute("value", "delete")
  
  form.method = 'post'
  form.appendChild(input)
})
``` 
This little script converts the links with the `delete` method to `post` and also adds a `_method` field with a `delete` value in the form link.

Notes: 
1. We also could add this script to turbo-rails in order to fix this https://github.com/hotwired/turbo-rails/issues/259
2. I'm open to suggestions to find a better name 😅

### DIFF
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -98,6 +98,12 @@ export function setProgressBarDelay(delay: number) {
   session.setProgressBarDelay(delay)
 }
 
+export function setLinkToFormBeforeCallback(
+  callback: (element: HTMLFormElement) => void
+) {
+  session.setLinkToFormBeforeCallback(callback);
+}
+
 export function setConfirmMethod(confirmMethod: (message: string, element: HTMLFormElement)=>boolean) {
   FormSubmission.confirmMethod = confirmMethod
 }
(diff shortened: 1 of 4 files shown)


dhh:
Given that this really is just to be used for the specific case of using _method field with delete for Rails, I think we should just bake that implementation in, exposed a flag to flip that setting, ala `Turbo.encodeHiddenLinkFormMethod = true`, which turbo-rails can then trigger. Other frameworks may indeed want to follow suit on this.

seanpdoyle:
Could `turbo-rails` support this directly, with an event listener like the one introduced in https://github.com/hotwired/turbo-rails/pull/239/files#diff-f438134433fd9897751d1b5cd219ea33ac29be8f3af3e4ac17e65475e4cadbeeR1-R7?

brunoprietog:
Hi @condef5, Thanks for the proposal, but I'm closing this for now. It's possible to do something similar with event listeners as well.

Issue #480, opened 2021-11-26
Turbo is ignored on any navigation that uses a dot in the URL


andrewjtait:
As of v7.1.0 form submissions do not use Turbo if the route includes a dot within its dynamic segment (https://guides.rubyonrails.org/routing.html#dynamic-segments)

For example, a form executing a POST to `/domains/example.com`

It looks as though the bug has been introduced as part of https://github.com/hotwired/turbo/pull/437, which introduces an additional check for `locationIsVisitable` in this code:

https://github.com/hotwired/turbo/blob/aa9724dd23309a56b3e7c6ee62c673fa63799038/src/core/session.ts#L202-L208

which is defined here:

https://github.com/hotwired/turbo/blob/18c93e9bfab790f65b447871f183b454c2ec1f2a/src/core/url.ts#L35-L37

and checks the extension on the URL: https://github.com/hotwired/turbo/blob/18c93e9bfab790f65b447871f183b454c2ec1f2a/src/core/url.ts#L26-L28

andrewjtait:
Actually, it seems like this is an issue on any navigation to a URL using a dot, not just form submissions.

tleish:
@andrewjtait — see https://github.com/hotwired/turbo/issues/385 discussion

andrewjtait:
@tleish ah, perfect - thanks!

Issue #385, opened 2021-09-08
Links with a  dot are not processed by turbo


silva96:
We have links in our platform that are user profiles like this:

domain.com/users/some.username

Whenever you click on those links, a normal request is made instead of a turbo request.

Is there any way to fix this? (I don't want to write a stimulus "link_controller#click" that calls Turbo.visit if it can be fixed from turbo side)

Thanks!

terracatta:
If anyone else was reading this issue and asked themselves "are periods even allowed in the path?", [apparently they are](https://stackoverflow.com/questions/7555553/can-period-be-part-of-the-path-part-of-an-url).

brunoprietog:
I have the same issue, it would be ideal if it could be solved directly from turbo.

silva96:
> If anyone else was reading this issue and asked themselves "are periods even allowed in the path?", [apparently they are](https://stackoverflow.com/questions/7555553/can-period-be-part-of-the-path-part-of-an-url).

Yes, instagram does it all day https://www.instagram.com/ruby.tuesday/

tleish:
TLDR: Turbo can handle periods, as long as the URL doesn't end in a way that suggests the URL is NOT an html page.

DETAILS:

Links with dots are processed with turbo:

- domain.com/users/some.username.html => fetched with turbo
- domain.com/users/some.username/ => fetched with turbo
- domain.com/users/some.png => NOT fetched with turbo
- domain.com/users/some.username => NOT fetched with turbo

Note in the instagram example, the URL ends in "/", indicating it's a directory

Turbo checks to see if a url is a directory or an html page

```javascript
function isHTML(url) {
  return !!getExtension(url).match(/^(?:|\.(?:htm|html|xhtml))$/)
}
```

Turbo skips non-html links (e.g. .json, .svg, .jpg, .png, etc).  

```javascript
  locationIsVisitable(location: URL) {
    return ... && isHTML(location)
  }
```

In the example of `domain.com/users/some.username`, how would you propose the javascript change to determine if the url is an html page vs a file with a `.username` extension?

silva96:
Thanks, we solved it adding trailing_slash to the route.rb

Issue #185, opened 2021-02-22
Override isHTML or locationIsVisitable?


warpr:
Hello!

With turbolinks I used to do this:

```js
Turbolinks.Location.prototype.isHTML = () => true;
```

Is this no longer possible with Turbo?

The default extension regex often trips me up, as a PHP developer I obviously expect `<turbo-frame id="messages" src="/messages.php">msg</turbo-frame>` to work.

In my case some of my URLs happen to have periods in them, which Turbo fails to load correctly (e.g. /music/prince/1989.batman) because it parses the ".batman" part as the extension. 

I also tried setting `data-turbo=true`, as setting `data-turbo=false` forces links to not be navigable or visitable, so I figured maybe `data-turbo=true` forces it in the other direction, but it seems to just give me the default behavior.

For now, the workaround I'm using is making sure my URLs end in a slash, as that seems to bypass the extension check, but it would be nice if there was some way to force Turbo to treat a  particular link as navigable/visitable.

msp-matt:
Just chiming in to say that this is an issue for my team also. I saw the PR that removed the Location class ([https://github.com/hotwired/turbo/pull/90](https://github.com/hotwired/turbo/pull/90)), and that seems like a solid engineering decision, but now that it's not possible to patch Location.isHTML, we're not able to upgrade. The comment [here](https://github.com/turbolinks/turbolinks/issues/301#issuecomment-699671429) is still applicable:

> Judging from the above, it looks like 5.3 will break PHP apps using the common `https://domain.com/endpoint.php` pattern, with routing handled directly by Apache (yes, we’re still here). We have a few not-great choices:
> 
> - freeze at 5.2
> - fork
> - major refactor
> 
> I think our team will probably freeze at 5.2 until we can refactor (it’s not like we’re thrilled with this kind of routing) but I wanted to leave this comment as evidence that the problem is not an edge case for older php apps and therefore might be worth a little bump in priority.

We've already upgraded to Stimulus 2.0 and we'd love to upgrade to Turbo as well. It seems a shame that there's no way to override the set of extensions that Turbo deems eligible for visiting. I really hope you'll consider making this possible again (in a better way than monkey-patching).

AndrewLivingston:
This would be a big boon to small teams that don't have the resources to refactor all their routing just to support turbo. Some of us are in a very dire situation with our legacy code in this area ;).

My team would be happy to submit a PR implementing this if the turbo team feels like this is worth supporting and if a design is settled on.

[...]

PR #519, opened 2022-01-24
Remove isHTML

f6p:
Hello,

Before **Turbo** will make a request it checks URL against regexp. The check passes for URLs without a file extension and those ending with `.htm`, `.html` and `.xhtml`. On success request is made using **Turbo** and on failure whole page is replaced.

I would like to propose removing `isHTML` function. Results of `isHTML` call are highly dubious and should not be trusted.

*Example 1*: Lets say there is a `download` action that maps to `/download` URL path. Response type of it would be `application/octet-stream` but **Turbo** will assume that it is HTML. 

*Example 2*: For URLs like `https://github.com/hotwired/turbo/blob/main/README.md` where parameter is part of path **Turbo** will fail to recognize browsable URL.

I believe a better approach would be to try to open everything by default and require to explicitly mark unbrowsable links with `data-turbo=”false”`. That would remove a bunch of unobvious URL problems like these mentioned above.

Moreover, currently its impossible to force **Turbo** visit as setting `data-turbo` to `true` triggers default behavior (that calls `isHTML`) and normal request instead of **Turbo** request is made (if `isHTML` will return `false`).

The change is a breaking one because it requires unbrowsable links to have `data-turbo=”false”` attribute added (not the case currently) but I believe it would result in more predictable less error prone behaviour in general.

Also after removing `isHTML` `locationIsVisitable` is reduced to return just the value of `isPrefixedBy` and is somewhat obsolete. So I replaced calls to `locationIsVisitable` with `isPrefixedBy`.

There are some disscussions regarding `isHTML` probems already open: #185, #385, #480.

Cheers
🍷

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -2,13 +2,14 @@ import { Action, isAction } from "../types"
 import { FetchMethod } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { FormSubmission } from "./form_submission"
-import { expandURL, getAnchor, getRequestURL, Locatable, locationIsVisitable } from "../url"
+import { expandURL, getAnchor, getRequestURL, Locatable } from "../url"
 import { getAttribute } from "../../util"
 import { Visit, VisitDelegate, VisitOptions } from "./visit"
 import { PageSnapshot } from "./page_snapshot"
 
 export type NavigatorDelegate = VisitDelegate & {
   allowsVisitingLocationWithAction(location: URL, action?: Action): boolean
+  locationIsVisitable(location: URL, rootLocation: URL): boolean
   visitProposedToLocation(location: URL, options: Partial<VisitOptions>): void
   notifyApplicationAfterVisitingSamePageLocation(oldURL: URL, newURL: URL): void
 }
@@ -25,7 +26,7 @@ export class Navigator {
 
   proposeVisit(location: URL, options: Partial<VisitOptions> = {}) {
     if (this.delegate.allowsVisitingLocationWithAction(location, options.action)) {
-      if (locationIsVisitable(location, this.view.snapshot.rootLocation)) {
+      if (this.delegate.locationIsVisitable(location, this.view.snapshot.rootLocation)) {
         this.delegate.visitProposedToLocation(location, options)
       } else {
         window.location.href = location.toString()
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -20,7 +20,7 @@ import {
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { ViewDelegate, ViewRenderOptions } from "../view"
-import { Locatable, getAction, expandURL, urlsAreEqual, locationIsVisitable } from "../url"
+import { Locatable, getAction, expandURL, urlsAreEqual } from "../url"
 import { FormSubmitObserver, FormSubmitObserverDelegate } from "../../observers/form_submit_observer"
 import { FrameView } from "./frame_view"
 import { LinkClickObserver, LinkClickObserverDelegate } from "../../observers/link_click_observer"
@@ -459,7 +459,7 @@ export class FrameController
   private formActionIsVisitable(form: HTMLFormElement, submitter?: HTMLElement) {
     const action = getAction(form, submitter)
 
-    return locationIsVisitable(expandURL(action), this.rootLocation)
+    return session.locationIsVisitable(expandURL(action), this.rootLocation)
   }
 
   private shouldInterceptNavigation(element: Element, submitter?: HTMLElement) {
--- a/src/core/frames/frame_redirector.ts
+++ b/src/core/frames/frame_redirector.ts
@@ -1,6 +1,6 @@
 import { FormSubmitObserver, FormSubmitObserverDelegate } from "../../observers/form_submit_observer"
 import { FrameElement } from "../../elements/frame_element"
-import { expandURL, getAction, locationIsVisitable } from "../url"
+import { expandURL, getAction } from "../url"
 import { LinkClickObserver, LinkClickObserverDelegate } from "../../observers/link_click_observer"
 import { Session, TurboClickEvent } from "../session"
 import { dispatch } from "../../util"
@@ -69,7 +69,7 @@ export class FrameRedirector implements LinkClickObserverDelegate, FormSubmitObs
     const meta = this.element.ownerDocument.querySelector<HTMLMetaElement>(`meta[name="turbo-root"]`)
     const rootLocation = expandURL(meta?.content ?? "/")
 
-    return this.shouldRedirect(form, submitter) && locationIsVisitable(action, rootLocation)
+    return this.shouldRedirect(form, submitter) && this.session.locationIsVisitable(action, rootLocation)
   }
 
   private shouldRedirect(element: Element, submitter?: HTMLElement) {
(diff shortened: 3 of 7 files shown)


packagethief:
> @packagethief I see you're in the `git blame` output around those lines. Do you have any recollection of what motivated introducing this check?

Hey @seanpdoyle! Yes, it was because we wanted to avoid visiting links to images and other media by default.

I agree that the heuristics are somewhat dubious. But I do think it's important that we not try to visit locations that are unlikely to return HTML. For example, it would be extremely unlikely for a path ending in ".jpg" to return an HTML document (but not impossible).

Perhaps it would be better to define the exclusions (likely also perilous), or to delegate the decision to the application.

packagethief:
Some additional context (from 2016!) regarding public, configurable API:

> so your application can define its own rules for what constitutes an “HTML URL.” That way it will work for link clicks and Turbolinks.visit without any annotations or extra arguments.

https://github.com/turbolinks/turbolinks/issues/15#issuecomment-186464668

packagethief:
@seanpdoyle, you have my memory gears turning now 😄  It seems we concluded that delegating the decision to applications via configuration was the way to go. Past me agrees:

- https://github.com/turbolinks/turbolinks/issues/99#issuecomment-219759894
- https://github.com/turbolinks/turbolinks/issues/256#issuecomment-296217525

f6p:
@seanpdoyle, @packagethief: Thank you very much for the input.

I guess it makes perfect sense to have some kind of configurable URL based HTML recognition. It would allow application wide defaults without being forced to annotate all links with `data-turbo=false`. And it also would be backward compatible. It would be much better.

I have some initial implementation that moves `locationIsVisitable` and `isHTML` from url to session. I will polish it up and force-push replacing current branch (I think it would be better to keep whole conversation here rather than closing this PR and opening another for configurable HTML check).

Cheers
🍷

f6p:
So that’s what I come with.

I moved `locationIsVisitible` and `isHTML` from url to session. As `Turbo.Location` doesn’t exists anymore I thought that it would be the best places for these to live in. Especially that some URL navigation helpers are already there.

After doing that I couldn’t stop thinking that `session.isHTML` sounds weird (how a session can be a HTML document?) so I renamed it to `session.isVisitable` (as `Location` doesn’t exists anymore it won’t be backward compatible anyway). I’m not sure about the name but that's the best what I was able to think of. My reasoning was that there are already functions/methods like `locationIsVisitable` and `formActionIsVisitable` so `isVisitable` as a shared name part would give a idea about relation between them (they call `isVisitable`). 

TLDR: `Turbo.Location.prototype.isHTML` become `Turbo.session.isVisitable`.

Other changes:

I added **php** extension to list of extensions that resolve as visitable by default. Probably it is as common as **html** extension and would almost certainly return some kind of HTML content type. I thought that this may be convenient addition as in issues it already starts to pop out that **Turbo** doesn’t work with **php** extensions by default. I think this would bring some peace :). 

Added tests that bind user defined functions to `session.isVisitable` to make sure we won’t lose it again.

Added `session` parameter to `FrameRedirector#constructor` to avoid circular dependency loop.

Cheers
🍷

ajlive:
As a php developer, I thank you!!

dmcge:
There must be an opportunity here to leverage the [`type` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-type) on `<a>`s. Turbo could simply require that links to non-HTML resources must be specified with a non-HTML type, with a check like:

```js
return link.type && link.type != "text/html"
```

This feels more in keeping with the current scope of Turbo, as a progressive enhancement that uses the HTML we already have. It also seems to meet the need of user configuration, but without requiring users to write Javascript.

dmcge:
(Incidentally, on the subject of `<a>` attributes, example 1 in the description of the original issue could be solved with [`[download]`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-download). [Turbo won’t attempt to visit those](https://github.com/hotwired/turbo/blob/main/src/observers/link_click_observer.ts#L63).)

f6p:
That's idea that could be considered but the problem here is that Turbo can't navigate certain URLs at all. For example now there is way to visit page with `.php` extension. To accomplish this one has to build his own version with modified `isHTML` function.

ajlive:
> Turbo could simply require that links to non-HTML resources must be specified with a non-HTML type

If I'm understanding correctly this means that I would need to add a `type` attribute to every `<a>` tag in my app? If so...

> This feels more in keeping with the current scope of Turbo, as a progressive enhancement that uses the HTML we already have. 

I don't have any HTML: I have PHP templates. So this would be the opposite of progressive for me: it would require a change to half the files in my codebase.

dmcge:
If you link to a `.php` URL (like `/show.php`), my thinking was that Turbo could simply assume it will ultimately return HTML unless told otherwise by the `type` attribute. We’d replace the existing `isHTML` check with something like the check I posted above. I’d assume that most apps using Turbo mostly link to HTML pages (whether they are, under the hood, PHP templates, Ruby templates, or whatever else). If that’s correct, you’d only add a `type` attribute on the rarer case when you don’t want to link to HTML (hopefully that’s not half your app!).

On GitHub, for example, you often encounter links that end with `README.md` but they are HTML resources (but a Ruby template underneath). Consider the [readme for this repo](https://github.com/hotwired/turbo/blob/main/README.md)). Turbo would treat that as a non-HTML page today and you’d get a full-page reload. With my proposal, Turbo would control the navigation and you wouldn’t have a full-page reload. If the page was actually markdown and not HTML (again, there would be Ruby underneath), you’d add `type="text/markdown"` and Turbo would let the browser do its thing.

This may well be a bad idea (and I don’t want to hijack this PR), but I’d personally rather write `type="whatever"` than a custom Javascript function. 🤷‍♂️

ajlive:
Ah, thanks for explaining further! Now that I understand it, I think I do like the idea, but you're right it's probably outside the scope of this PR

kasperkamperman:
Is there any way to override isHTML or locationIsVisitable with the current release? 
I could use this branch maybe, but it's already several commits behind the release. 

Still struggling to get Turbo to work with .php extension.

tleish:
I prefer `data-turbo` over `[download]` or `[type]`.  `[download]` is not an attribute of `form` and would only work with `a` elements. `data-turbo` already works with either element.

I also prefer exposing the isHTML method to a global level to allow override:

```javascript
Turbo.session.isHTML = () => true;
```

However, you could also provide a simpler on/off solution with a boolean variable.

```javascript
Turbo.session.strictVisitMimeType = true; // default
```

f6p:
Hey @dhh looks like github accidentally closed this one as string `resolve #519` was found in #629 comment.

Cheers
:wine:

f6p:
Hmmm, I'm not able to run tests locally anymore.

dhh:
Enabled tests to be run. Needs a rebase.

f6p:
@dhh Rebased it, but approval seems to expire after I force push rebased branch.

dhh:
Think it needs another rebase. Sorry!

f6p:
np, rebeased

f6p:
Hey @packagethief, any thoughts how we could move forward with this one?

Cheers
🍷

manuelpuyol:
@f6p could you fix the conflicts?

@dhh is there anything you feel like this PR is missing? This would be a great improvement to how turbo manages visitable URLs

f6p:
@manuelpuyol: I resolved conflicts but its hard to keep this one up to date. Usually it breaks after a few commits to mian. It seems that it touches the parts that change very often.

manuelpuyol:
@dhh could you enable the tests in this pr? 🙏

bennadel:
I just wanted to say that I arrived here because I am using `.cfm` (ColdFusion) file extensions. But, I will concede that a ColdFusion page can serve up _anything_ (HTML, CSV, Images, etc). As such, I understand how hard it would be have a blanket acceptance of all `.cfm` files. I'm new to Hotwire, so I have no advice -- just adding my voice to the discussion.

PR #525, opened 2022-01-27
Await for confirm method results

### LINKED ISSUE #522: Could you please add await keyword to FormSubmission.start method?
I would like to add a custom confirm dialog to my Rails project but for that to work I need the await keyword added to turbo.
In [core/drive/form_submission.ts:116](https://github.com/hotwired/turbo/blob/201f8b3260bfbc15635d24c9961ff027747fb046/src/core/drive/form_submission.ts#L116) there is this line of code:

```typescript
const answer = FormSubmission.confirmMethod(this.confirmationMessage!, this.formElement)
```

Could this please be changed to:

```typescript
const answer = await FormSubmission.confirmMethod(this.confirmationMessage!, this.formElement)
```

So I would be able to do something like this:

```javascript
const newConfirmMethod = async (message, formElement) => {

  let confirmModal = new bootstrap.Modal(document.getElementById('confirm_modal'));
  confirmModal.show();

  let titleElement = confirmModal._element.querySelector('.modal-title');
  titleElement.innerText = ''

  let p = confirmModal._element.querySelector('.modal-body p')
  p.innerText = message;


  return new Promise((resolve, reject) => {
    const confirmButton = confirmModal._element.querySelector('.confirm');
    const cancelButton = confirmModal._element.querySelector('.cancel');
    confirmButton.addEventListener('click', () => {
      resolve(true)
    });

    cancelButton.addEventListener('click', () => {
      resolve(false)
    });
  });
};

Turbo.setConfirmMethod(newConfirmMethod);
```

excid3:
The browser's `confirm` is synchronous and waits for a boolean. This is really convenient, but makes it painful to implement a custom confirm modal in Boostrap, Tailwind, etc.

This PR adds `await` so a custom `confirm` can return a promise that waits until the user confirms or cancels the operation.

The overridden confirm method simply needs to return a Promise that resolves to `true` or `false.
```js

Turbo.setConfirmMethod(() => {
  return new Promise((resolve, reject) => {
    // show modal
    // on confirm, resolve(true)
    // on cancel, resolve(false)
  })
})

```

Closes #522

Example of it in action:
https://user-images.githubusercontent.com/67093/151271318-5034bbff-6b24-462f-89f7-8e6e3f8d0466.mp4

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -51,8 +51,8 @@ export class FormSubmission {
   state = FormSubmissionState.initialized
   result?: FormSubmissionResult
 
-  static confirmMethod(message: string, _element: HTMLFormElement): boolean {
-    return confirm(message)
+  static confirmMethod(message: string, _element: HTMLFormElement): Promise<boolean> {
+    return Promise.resolve(confirm(message))
   }
 
   constructor(
@@ -121,7 +121,7 @@ export class FormSubmission {
     const { initialized, requesting } = FormSubmissionState
 
     if (this.needsConfirmation) {
-      const answer = FormSubmission.confirmMethod(this.confirmationMessage!, this.formElement)
+      const answer = await FormSubmission.confirmMethod(this.confirmationMessage!, this.formElement)
       if (!answer) {
         return
       }
--- a/src/core/index.ts
+++ b/src/core/index.ts
@@ -99,6 +99,6 @@ export function setProgressBarDelay(delay: number) {
   session.setProgressBarDelay(delay)
 }
 
-export function setConfirmMethod(confirmMethod: (message: string, element: HTMLFormElement) => boolean) {
+export function setConfirmMethod(confirmMethod: (message: string, element: HTMLFormElement) => Promise<boolean>) {
   FormSubmission.confirmMethod = confirmMethod
 }
--- a/src/tests/functional/form_submission_tests.ts
+++ b/src/tests/functional/form_submission_tests.ts
@@ -57,7 +57,7 @@ export class FormSubmissionTests extends TurboDriveTestCase {
   }
 
   async "test from submission with confirmation overriden"() {
-    await this.remote.execute(() => window.Turbo.setConfirmMethod(() => confirm("Overriden message")))
+    await this.remote.execute(() => window.Turbo.setConfirmMethod(() => Promise.resolve(confirm("Overriden message"))))
 
     await this.clickSelector("#standard form.confirm input[type=submit]")
 


excid3:
Tests are failing because `main` has failing tests.

src/core/drive/form_submission.ts
```diff
@@ -113,7 +113,7 @@ export class FormSubmission {
     const { initialized, requesting } = FormSubmissionState
 
     if (this.needsConfirmation) {
-      const answer = FormSubmission.confirmMethod(this.confirmationMessage!, this.formElement)
+      const answer = await FormSubmission.confirmMethod(this.confirmationMessage!, this.formElement)
```
seanpdoyle, line 116: Switching to an assumption that `FormSubmission.confirmMethod` is a promise, what happens when a user-provided Promise set by `Turbo.setConfirmMethod` throws an exception?

Before the `await`, the `async start()` (the method that wraps these lines) would have to deal with the exception.

After adding the `await`, is that exception swallowed?

excid3: Throwing an error in the promise raises the `Uncaught (in promise)` error in the `start()` method of FormSubmission: https://github.com/hotwired/turbo/blob/201f8b3260bfbc15635d24c9961ff027747fb046/src/core/drive/form_submission.ts#L122

It doesn't swallow the exception.

seanpdoyle: That's good. In that case, should the default function also be translated into a Promise?

https://github.com/hotwired/turbo/blob/201f8b3260bfbc15635d24c9961ff027747fb046/src/core/drive/form_submission.ts#L53-L55

```typescript
static confirmMethod(message: string, element: HTMLFormElement): Promise<boolean> {
  return Promise.resolve(confirm(message))
} 
```

excid3: It didn't affect the default method, but we could.

We would also want to change this to expect a Promise too?
```ts
export function setConfirmMethod(confirmMethod: (message: string, element: HTMLFormElement)=>Promise<boolean>) {
  FormSubmission.confirmMethod = confirmMethod
}
```

kstratis:
Great work @excid3! Thank you!
Seems there are a couple of duplicate commits though - Could you please have a look?

commits

excid3:
@kstratis I botched the rebase from main on accident. 😅

tabishiqbal:
any update on this?

MatheusRich:
Is there anything actionable here? How can we move this forward?

cc @dhh

src/core/drive/progress_bar.ts
```diff
@@ -12,7 +12,7 @@ export class ProgressBar {
         left: 0;
         height: 3px;
         background: #0076ff;
-        z-index: 9999;
+        z-index: 2147483647;
```
dhh, line 15: What's this change related to?

package.json
```diff
@@ -35,15 +35,15 @@
...
+    "@rollup/plugin-typescript": "8.3.1",
     "@types/multer": "^1.4.5",
+    "arg": "^5.0.1",
     "intern": "^4.9.0",
-    "arg": "^4.1.0",
     "multer": "^1.4.2",
     "rollup": "^2.35.1",
     "tslib": "^2.0.3",
-    "typescript": "^4.1.3"
+    "typescript": "^4.6.3"
```
dhh, line 46: Where these dependency changes needed for the code changes in the PR?

excid3:
@dhh fixed this, so it's ready for review now. 👍

excid3:
Tests that are failing are the same ones in `main`.

thewatts:
@excid3 congrats on the merge!

Does this mean that you're going to have a GoRails episode on implementing a custom confirm modal? 😇

thewatts:
@ivorpad YESSSSSSS! Thank you!


( And many thanks to you as well, @excid3 ! )

rapcal:
@excid3 how would you go about i18n buttons rendered in `insertConfirmModal`?

excid3:
@rapcal Lots of options.

You could pass the text as extra data attributes on the element with the `data-turbo-confirm`, like `data: { turbo_confirm_ok: I18n.t("ok"), turbo_confirm_cancel: I18n.t("cancel") }` and then use those when rendering your modal HTML.

Or have Rails render the HTML for the dialog ahead of time like we do in the screencast linked here. https://github.com/hotwired/turbo/pull/525#issuecomment-1154257056

rapcal:
> @rapcal Lots of options.
> 
> You could pass the text as extra data attributes on the element with the `data-turbo-confirm`, like `data: { turbo_confirm_ok: I18n.t("ok"), turbo_confirm_cancel: I18n.t("cancel") }` and then use those when rendering your modal HTML.
> 
> Or have Rails render the HTML for the dialog ahead of time like we do in the screencast linked here. https://github.com/hotwired/turbo/pull/525#issuecomment-1154257056

Thanks, Chris!

PR #527, opened 2022-02-02
Activate scripts from turbo streams

elShiaLabeouf:
Hi! I was building my new Rails app with Turbo Frames and Turbo Streams and bumped into the issue of script tags not executed after being appended via Turbo Streams. Found the same [issue(#186)](https://github.com/hotwired/turbo/issues/186) here.

There I found the @xtr3me's PR https://github.com/hotwired/turbo/pull/192/ with an approach for turbo frames, and decided to make the same-ish thing for streams.

So I've come up with this PR. Could you please take a look?

It works well for content with <script> tag, and it doesn't break code if the content doesn't have one. 

Unfortunately, I have no experience with TypeScript and testing Node applications. Spent an evening trying to add tests (by analogy with the existing ones), but failed miserably. The good thing is, the code doesn't fail more tests! 

I'm really excited to see such feature in main. It'd add so many possibilities to make user experience smoother! 
Looking forward to your reply!


geronimo:
I'm also having this problem. Specifically, I want to use turbo streams to replace one of the frames with a chartkick chart. The script tag generated by the rails helpers doesn't execute.

retsef:
Hi @geronimo, hi have your same issue and I have opted with an override of the chartkick helper and a stimulus controller for  render chart.js

https://gist.github.com/retsef/d6dffcc0b792bce66bd9038bd8646388

In our case we reload a turbo-frame containing charts on a date range picker.

dhh:
Would be happy to see this finished. Let's try to reuse the logic we use from frames to trigger script tags. Then we need tests.

elShiaLabeouf:
@dhh Happy to see your approval, it gave me extra motivation :)

Added a test to `stream_tests.ts`, which adds a blank `div.message` and then fills it with text and assigns a css class to it via a <script>.

src/core/streams/stream_message.ts
```diff
@@ -37,4 +38,32 @@ export class StreamMessage {
...
   }
+
+  activateScriptElements(parentElement: StreamElement) {
+    for (const inertScriptElement of ((parentElement.firstChild as HTMLMetaElement).content as unknown as Element).querySelectorAll("script")) {
+      const activatedScriptElement = this.createScriptElement(inertScriptElement)
+      inertScriptElement.replaceWith(activatedScriptElement)
+    }
+  }
+
+  createScriptElement(element: Element) {
```
jean-francois-labbe, line 49: Could `createScriptElement` from the [core/renderer.ts](https://github.com/hotwired/turbo/blob/e2d53054438fa16b4babf6a7afeba7a828764772/src/core/renderer.ts#L47) be reused ?

elShiaLabeouf: Hi @jean-francois-labbe ! Thanks for the advice! I wish I knew how to do it...
Learning and writing in TS is torture for me

elShiaLabeouf: should I create a new class similarly to the `FrameRenderer` ? Smth like
```
export class StreamRenderer extends Renderer<StreamElement> {
```

elShiaLabeouf: aaand tests are now refusing to start
```
➜  turbo git:(main) ✗ yarn test
yarn run v1.22.17
Debugger listening on ws://127.0.0.1:9229/767ccf04-609a-4577-8e35-0e2cb7ace1b5
Tunnel started
(ノಠ益ಠ)ノ彡┻━┻
SessionNotCreatedException: [POST http://localhost:4444/wd/hub/session / {"desiredCapabilities":{"name":"intern","idle-timeout":60,"browserName":"chrome","goog:chromeOptions":{"args":["headless","disable-gpu","no-sandbox"]},"browser":"chrome"}}] session not created: This version of ChromeDriver only supports Chrome version 99
Current browser version is 101.0.4951.54 with binary path /usr/bin/google-chrome
System info: host: 'localhost', ip: '127.0.0.1', os.name: 'Linux', os.arch: 'amd64', os.version: '5.17.0-5.1-liquorix-amd64', java.version: '1.8.0_131'
Driver info: driver.version: unknown
```
Jesus Christ, what an experience...

jean-francois-labbe: Oh, I'll try to help you.
I'd create a class such as `scriptActivator` or `scriptInterceptor` to give it a home.
I'd use FormSubmission as an example https://github.com/hotwired/turbo/blob/main/src/core/drive/form_submission.ts#L43
As an example.
It is used here https://github.com/hotwired/turbo/blob/e30d36bf56d020d63d87a8350d9d0361edd9b804/src/core/frames/frame_controller.ts#L173

I have no idea if that's what is expected.

elShiaLabeouf: hooray, fixed the tests execution error by adding this to intern.json:
```
"tunnelOptions": {
    "drivers": [
      { "name": "chrome", "version": "101.0.4951.41" }
    ]
  }
  ```
  not sure if I should add it to the PR though

Comment (elShiaLabeouf) [reply src/core/streams/stream_message.ts]: @jean-francois-labbe, Thank you for the help, but this is still too complicated for me, taking into account my little knowledge in statically typed languages, including TypeScript, and in the Turbo ecosystem. Implementing the new `class ScriptActivator` that would use a small part of the abstract `class Renderer`  following the example of `FormSubmission` (which doesn't exactly implement any abstract classes) is greek to me.

I suggest we make compromise: what if we make the `createScriptElement` function exportable in the `renderer.ts` . And then we import that function to `stream_message.ts` so it would be reused?

Please, check my solution out, just pushed a new commit to the PR. Looking forward to your reply!

File: src/core/renderer.ts [hunk 2]
```diff
@@ -45,18 +45,7 @@ export abstract class Renderer<E extends Element, S extends Snapshot<E> = Snapsh
...
-      const createdScriptElement = document.createElement("script")
-      if (this.cspNonce) {
-        createdScriptElement.nonce = this.cspNonce
-      }
-      createdScriptElement.textContent = element.textContent
-      createdScriptElement.async = false
-      copyElementAttributes(createdScriptElement, element)
-      return createdScriptElement
-    }
+    return createScriptElement(element)
```
jean-francois-labbe, line 48: It seems to me that this will create a loop. On line 83 you redefine a function with the same name.

I think that adding `export` on line 47 would have been enough.

elShiaLabeouf: no, this was done on purpose, it's `return createScriptElement(element)`, not `return this.createScriptElement(element)` 

I took the `createScriptElement` function out of the `abstract class Renderer`'s scope, so that I could import the function with a simple `import { createScriptElement } from "../renderer";` in `stream_message.ts`. And at the same time I left `createScriptElement` in the `Renderer` so that it wouldn't break other code that still use Renderer#createScriptElement

A loop won't be created because on line 48 I'm calling the line 83's `createScriptElement`.

If I return everything as it was and just add `export` on line 47, I won't be able to import the function `createScriptElement` with `import { createScriptElement } from "../renderer";` in `stream_message.ts`, because I would have to implement and create an object of `abstract class Renderer`, which I'm quite frankly unable to do due to my incompetence

I assure you, my solution works, the tests are passing. I wouldn't push the code and waste your time if tests were failing.

elShiaLabeouf: Hi @jean-francois-labbe ! How are things your end?
Can you please approve this PR?

jean-francois-labbe: Hi @elShiaLabeouf things are great.
I cannot approve this PR I'm not a maintainer on Hotwired.

elShiaLabeouf: Oh. Sorry to bother then. Thank you for the hints on the code improvements!

seanpdoyle:
Thank you @elShiaLabeouf for opening this original pull request.

I've created https://github.com/hotwired/turbo/pull/660 to catch this changeset up to the latest commits. I've added you as a co-author.

elShiaLabeouf:
Hi @seanpdoyle! Thanks, I appreciate it! 

I'm just happy this code has made it to the main! 

Sorry for "abandoning" this, been buried under the bloody work at my workplace lately.

cantin:
@seanpdoyle  The <script> would be run again while navigating back. Is it the desired behavior?

PR #529, opened 2022-02-08
Treat blank `[formaction]` like a blank `[action]`

seanpdoyle:
When handling form submissions, a `<form>` element that declares a
`[action]` to be the empty string (i.e. `""`) is treated as if it were
targeting the current path:

> The `action` IDL attribute must reflect the content attribute of the
> same name, except that on getting, when the content attribute is
> missing or its value is the empty string, the element's node
> document's URL must be returned instead.
>
> [§ 4.10.18.6 Form Submission attributes][action]

Similarly, when handling form submissions with a submitter that declares
a `[formaction]` to be the empty string (i.e. `""`) is treated as if it
were targetting the current path.

> The `formAction` IDL attribute must reflect the formaction content
> attribute, except that on getting, when the content attribute is missing
> or its value is the empty string, the element's node document's URL must
>be returned instead.
>
> [§ 4.10.18.6 Form Submission attributes][formaction]

[action]: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-fs-action
[formaction]: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-fs-formaction


seanpdoyle:
@dhh this is ready for review.

PR #541, opened 2022-02-22
Add Turbo Stream actions to add and remove CSS classes

percysnoodle:
A common UI pattern is to indicate whether a field is valid or invalid by adding a CSS class to invalid fields to turn them red. Another is to submit the input as the user enters the text rather than waiting for them to press a button to submit the form. Combining the two requires us to add or remove fields from a field as the user continues to type into it.  For example, the following field in our application is validated to be exactly 8 digits:
An example of a field which validates as the user types.  The field is initially red to show that it is invalid, but goes blue when the text reaches the required 8 digits.  When more text is entered, it goes back to red.

This is difficult to implement using Turbo Streams, because the only action that can change the CSS class of an element on a page is `update`, which removes the existing element and replaces it with a new one.  This interrupts the user's input and is likely to discard their latest changes.

It is possible to work around this, for example by updating a hidden element with [a Stimulus controller that adds or removes the class when it connects](https://gist.github.com/percysnoodle/aa843cb942f8be1b4fd51ec36bba8e8a) (as in the above example) or perhaps doing something with sibling selectors; but this is bespoke and unwieldy.

The Turbo handbook says that...

> Turbo Streams consciously restricts you to seven actions [...in order to...] keep you from turning individual responses in a jumble of behaviors that cannot be reused and which make the app hard to follow

...but I feel that the omission of an action which enables this behaviour has pushed us in that direction instead of away from it.

So, to enable these sorts of update using Turbo Streams out of the box, this PR adds two new stream actions: `add-class`, which adds the template's classes to the targets; and `remove-class`, which removes them.

Since it's also fairly common to enable and disable elements, I did consider adding more generic actions to handle element attributes, but since disabling a field would also interrupt user input I think it is sufficiently covered by existing actions, and so it's not worth the extra complexity.

### DIFF
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -32,5 +32,17 @@ export const StreamActions: { [action: string]: (this: StreamElement) => void } 
       e.innerHTML = ""
       e.append(this.templateContent)
     })
+  },
+
+  "add-class"() {
+    this.targetElements.forEach(e => {
+      this.templateElement.classList.forEach(c => e.classList.add(c))
+    })
+  },
+
+  "remove-class"() {
+    this.targetElements.forEach(e => {
+      this.templateElement.classList.forEach(c => e.classList.remove(c))
+    })
   }
 }
--- a/src/tests/unit/stream_element_tests.ts
+++ b/src/tests/unit/stream_element_tests.ts
@@ -4,7 +4,7 @@ import { DOMTestCase } from "../helpers/dom_test_case"
 
 export class StreamElementTests extends DOMTestCase {
   async beforeTest() {
-    this.fixtureHTML = `<div><div id="hello">Hello Turbo</div></div>`
+    this.fixtureHTML = `<div><div id="hello" class="hello1 hello2">Hello Turbo</div></div>`
   }
 
   async "test action=append"() {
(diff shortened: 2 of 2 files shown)


tleish:
I would prefer to refactor this for any attribute (css, data-controller, data-action, etc).  Something like:


```javascript
function tokenize(value) {
  return value.match(/[^\s]+/g) || []
}

export const AttributeStreamActions {
  // ....

  appendAttributeValue() {
    this.targetElements.forEach(e => {
      const values = [...new Set(tokenize(`${e.getAttribute(this.attributeName) || ''} ${this.attributeValue}`))]
      e.setAttribute(this.attributeName, values.join(' '))
    })
  },

  prependAttributeValue() {
    this.targetElements.forEach(e => {
      const values = [...new Set(tokenize(`${this.attributeValue} ${e.getAttribute(this.attributeName) || ''}`))]
      e.setAttribute(this.attributeName, values.join(' '))
    })
  },

  removeAttributeValue() {
    this.targetElements.forEach(e => {
      const values = tokenize(e.getAttribute(this.attributeName) || '').filter(v => v !== this.attributeValue)
      e.setAttribute(this.attributeName, values.join(' '))
    })
  },

  updateAttributeValue() {
    this.targetElements.forEach(e => e.setAttribute(this.attributeName, this.attributeValue))
  }
}
```

```html
<turbo-stream action="appendAttributeValue" target="message_1" data-attribute-name="data-action" data-attribute-value="click->gallery#next" />

<turbo-stream action="removeAttributeValue" target="message_1" data-attribute-name="class" data-attribute-value="error" />
```

brunoprietog:
I really like the second proposed approach, especially for accessibility reasons. It is very common to change aria attributes after an action.

dhh:
I don't think turbo streams is the right option here. You're better off creating a custom stimulus controller, if you can't include the value that's relevant in the HTML that's returned.

PR #546, opened 2022-03-01
Autofocus avoids inert elements

keithamus:
When loading a page containing `<dialog>` elements or `<div hidden>` elements, which themselves contain `autofocus` elements, the browser is able to determine the first "active" `autofocus` element, and focus that.

Right now Turbo does not do the same thing. As a reduced test case, consider the following two pages:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <script type="module" src="https://cdn.skypack.dev/@hotwired/turbo@7.1.0"></script>
  </head>
  <body>
    <a href="/two.html">Click</a>
  </body>
</html>
```

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="UTF-8">
    <script type="module" src="https://cdn.skypack.dev/@hotwired/turbo@7.1.0"></script>
  </head>
  <body>
    <dialog> <!-- this could be also <div hidden> -->
      <input autofocus>
    </dialog>
    <input autofocus>
  </body>
</html>
```

Navigating to the second page with `Turbo.session.drive = false` will result in the `<input autofocus>` _outside_ of the dialog to be focussed. Refreshing the page will also cause the outer `<input autofocus>` to be focussed. However with Turbo drive on, the element does not autofocus.

### DIFF
--- a/src/tests/fixtures/autofocus-inert.html
+++ b/src/tests/fixtures/autofocus-inert.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Autofocus</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+  </head>
+  <body>
+    <h1>Autofocus With Inert Elements</h1>
+    <dialog>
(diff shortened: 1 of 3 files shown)


dhh:
Please do investigate a fix for this!

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/654 to cover this case. I'd really appreciate a review!

PR #551, opened 2022-03-14
Upgrade all packages

keithamus:
This upgrades known dependencies to their latest versions, and makes a few small changes to the `tsconfig` and `fetch_request` files to upgrade the failing TypeScript build.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,15 @@ jobs:
     - run: yarn install
     - run: yarn build
 
+    - name: Set Chrome Version
+      run: |
+        CHROMEVER="$(chromedriver --version | cut -d' ' -f2)"
+        echo "Actions ChromeDriver is $CHROMEVER"
+        CONTENTS="$(jq '.tunnelOptions.drivers[0].name = "chrome"' < intern.json)"
+        CONTENTS="$(echo ${CONTENTS} | jq --arg chromever "$CHROMEVER" '.tunnelOptions.drivers[0].version = $chromever')"
+        echo "${CONTENTS}" > intern.json
+        cat intern.json
+
     - name: Test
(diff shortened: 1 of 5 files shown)


keithamus:
The tests are failing because `intern` comes pre-baked with its own copy of ChromeDriver, the ChromeDriver it uses only works with Chrome 99.

GitHub Actions runners use Chrome 98 right now (https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#browsers-and-drivers). They also come with ChromeDriver which works with 98. It should be possible to set an environment flag to get `intern` to use the built in ChromeDriver... somehow :joy:

dhh:
Could we lock intern to a version that has the prebaked 98 for now?

.github/workflows/ci.yml
```diff
@@ -20,6 +20,15 @@ jobs:
...
 
+    - name: Set Chrome Version
+      run: |
+        CHROMEVER="$(chromedriver --version | cut -d' ' -f2)"
+        echo "Actions ChromeDriver is $CHROMEVER"
+        CONTENTS="$(jq '.tunnelOptions.drivers[0].name = "chrome"' < intern.json)"
+        CONTENTS="$(echo ${CONTENTS} | jq --arg chromever "$CHROMEVER" '.tunnelOptions.drivers[0].version = $chromever')"
+        echo "${CONTENTS}" > intern.json
+        cat intern.json
+
```
keithamus, line 31: This script uses the actions environment to determine the correct Chrome version, and writes into Intern.json the version to use. I think intern still downloads it's own copy of ChromeDriver, but at least it gets the tests _running_ :joy:

dhh: Yak riders, saddle up!

Issue #174, opened 2021-02-12
Prefetching


fleck:
First, thanks for turbo! I'm working on a prefetcher for turbo https://github.com/fleck/nitrous-oxide. I'm taking advantage of the browsers built in http caching and max-age header. The issue I encountered was when a user hovers over a link and clicks before that fetch is finished resulting in a double fetch. I've fixed this by tracking inflight requests.

When a user clicks a link and there's in inflight request I start the visit (I check whether this is targeting a frame or a general visit):
https://github.com/fleck/nitrous-oxide/blob/master/src/index.ts#L73

After the promise is fulfilled I then trigger a frame or regular visit:
https://github.com/fleck/nitrous-oxide/blob/master/src/index.ts#L82

This seems to work well enough, but the code feels a little janky, just wondering if you had suggestions for a better way to handoff a url and `Promise<Response>` to programmatically trigger a turbo visit/TurboFrame visit?

Thanks again for the work on Turbo, It's been really nice to work with!

zejji:
+1 Would love to see this. It would be really helpful to combine Turbo with the functionality of something like [FlyingPages](https://github.com/gijo-varghese/flying-pages) or [instant.page}(https://instant.page/).

silentjay:
If like me you don't want to add npm/another toolchain to their projects someones shared prefetch.js gist https://gist.github.com/vitobotta/8ac3c6f65633b5edb2949aeff0dec69b

fleck:
Cross posting a comment for feedback on non bundler setup: 

https://github.com/fleck/nitrous-oxide/issues/2#issuecomment-812955150

> This can be used without build tooling via https://www.skypack.dev/view/nitrous-oxide. But if you're looking for a more traditional UMD build that you can vendor into your source control and include via a regular <script tag (e.g. NOT type="module") then you can checkout turbo and our UMD builds: https://unpkg.com/nitrous-oxide@0.15.0/dist/nitrous-oxide.umd.production.min.js and https://unpkg.com/@hotwired/turbo@7.0.0-beta.4/dist/turbo.es2017-umd.js I think the script tag for turbo would need to be before the script tag for nitrous oxide. Let me know if this works for you or if you run into any issues. Would like to add docs on how to use without build tooling, just haven't had time to test that approach myself.

dhh:
I'd very much like to see this built into Turbo. If anyone wants to work on this, please do start a PR.

PR #552, opened 2022-03-17
pre populate cache with views based on links with `turbo_preload` attribute

hey-leon:
First off, we ❤️  developing with hotwire and turbo at my workplace. So I want to thank all the contributors and the team.

Description
------------
At my workplace we have released a new mobile application using hotwire and coming from a react native application one part of the UX that felt a bit less refined was the first 0-300ms of navigation. When navigating to a page for a second time turbo loads a snapshot from cache which makes for a pretty snappy (pun intended) experience. However for hot paths we wanted to make this the reality on first visits.

Prior Art
---------
Before investigating this solution we did look at leveraging browser based prefetching. This can be achieved via the usage of `<link rel="prefetch" href="/app/path/here" />`. However our experience showed a couple ergonomic challenges using this for app screens:
* It does not appear to follow redirects in major browsers (expectations do not seem to be defined in specs).
* It relies on usage of the `Cache-control` header.
  * To leverage it you essentially need to completely hand over cache control to the browser which is not ideal for highly dynamic web apps.

The implementation in this PR makes it possible to pre populate the cache by marking links with `data-turbo-preload`. which makes for better user experience without needing to worry about stale views.


seanpdoyle:
Thank you for opening this PR, the introduction of the `Preloader` class lots of potential!

I wonder, had you considered reading from an [`<a>` element's `[rel]` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-rel), instead of introducing a bespoke `[data-turbo-preload]` attribute?

There are existing semantics for values like [rel="prefetch"](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/prefetch) and [rel="preload"](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload), which could pose opportunities for progressive enhancement while gracefully degrading to built-in browser support when JS is disabled or Turbo fails to load.

src/core/drive/preloader.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  preloadOnLoadLinksForView(element: Element) {
+    const links = element.querySelectorAll('a[data-turbo-preload="load"]')
+
+    for (const link of links) {
+      this.preloadURL(link)
+    }
+  }
+
+  preloadURL(link: Element) {
```
seanpdoyle, line 35: If this method accepted [HTMLAnchorElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement) as a type, we could read the [link.href](https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/href) directly, without the need to call `expandURL` or `|| ""`.

src/core/drive/preloader.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    if (this.snapshotCache.has(url)) {
+      return
+    }
+
+    fetch(url.toString(), { headers: { 'VND.PREFETCH': 'true', 'Accept': 'text/html' } })
+      .then(res => res.text())
+      .then(responseText => PageSnapshot.fromHTMLString(responseText))
+      .then(snapshot => this.snapshotCache.put(url, snapshot))
+      .catch(() => {})
```
seanpdoyle, line 46: This codebase utilizes [`async`/`await` syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function), which enables this to be implemented without chaining:

```js
const response = await fetch(url.toString(), { headers: { 'VND.PREFETCH': 'true', 'Accept': 'text/html' } })
const responseText = await response.text()
const snapshot = PageSnapshot.fromHTMLString(responseText)

this.snapshotCache.put(url, snapshot)
```

src/core/drive/preloader.ts
hey-leon: Thanks for the feedback @seanpdoyle I'll have a go at refining the types.

src/core/drive/preloader.ts
hey-leon: Yes that does seem to be the convention. I will tidy this up 👍. Thank you.

hey-leon:
> Thank you for opening this PR, the introduction of the `Preloader` class has a lot of potential!
> 
> I wonder, had you considered reading from an [`<a>` element's `[rel]` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-rel), instead of introducing a bespoke `[data-turbo-preload]` attribute?
> 
> There are existing semantics for values like [rel="prefetch"](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/prefetch) and [rel="preload"](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload), which could pose opportunities for progressive enhancement while gracefully degrading to built-in browser support when JS is disabled or Turbo fails to load.

Good point 

> 

This is a great thought. The "prefetch" and "preload" types are not currently supported by anchor tags (link only) at this point. See [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types) table for rel type support. But based on moz docs it appears it is at least being considered [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ#are_anchor_a_tags_prefetched).

What is your thoughts on adopting this pre-emptively. I am open to the refactor if we think this is a good approach?

src/core/drive/preloader.ts
hey-leon: cfd830d0df4c46a4383e620117005e113dd77472

src/core/drive/preloader.ts
hey-leon: https://github.com/hotwired/turbo/pull/552/commits/8e353855278743a3585fffc13b2bf0d27fbd51eb

src/core/drive/preloader.ts
```diff
@@ -28,23 +27,25 @@ export class Preloader {
     const links = element.querySelectorAll('a[data-turbo-preload="load"]')
```
seanpdoyle, line 27: Does executing this query with additional type information help avoid the `instances` check?

```js
for (const link of element.querySelectorAll<HTMLAnchorElement>('a[data-turbo-preload="load"]')) {
  this.preloadURL(link)
}
```

hey-leon: Resolved in https://github.com/hotwired/turbo/pull/552/commits/15fedf0af59c68e77325c5c60bf45949437d73ac

rik:
I'm concerned about Turbo re-purposing existing HTML attributes. Currently, `rel=preload` and `rel=prefetch` aren't allowed on `<a>` by [the HTML spec](https://html.spec.whatwg.org/multipage/links.html#linkTypes:link-type-prefetch-2). If the HTML spec wants to change to allow it, Turbo might get in the way.

On top of that, I'm not sure this feature is needed in Turbo. You can add `<link rel=prefetch>` in your HTML (or in JS) and supporting browsers will prefetch them. So the fetch call from Turbo when visiting should be instant. Maybe just a polyfill for Safari [until they implement prefetching](https://bugs.webkit.org/show_bug.cgi?id=194539)?

hey-leon:
> I'm concerned about Turbo re-purposing existing HTML attributes. Currently, `rel=preload` and `rel=prefetch` aren't allowed on `<a>` by [the HTML spec](https://html.spec.whatwg.org/multipage/links.html#linkTypes:link-type-prefetch-2). If the HTML spec wants to change to allow it, Turbo might get in the way.

@rik you are correct that they are not supported on anchors at this point. However mozilla's [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Link_prefetching_FAQ#are_anchor_a_tags_prefetched) demonstrates an openness to to accept it. Aside from this, the more I get in the weeds on this PR I am realizing there are some fundamental differences between the changes in this PR and `link[rel="prefetch"]` and I think that is probably a good enough reason to change to a custom attributes.

> On top of that, I'm not sure this feature is needed in Turbo. You can add `<link rel=prefetch>` in your HTML (or in JS) and supporting browsers will prefetch them. So the fetch call from Turbo when visiting should be instant. Maybe just a polyfill for Safari [until they implement prefetching](https://bugs.webkit.org/show_bug.cgi?id=194539)?

After my exploration I am not sure if `link[rel="prefetch"]` is an awesome time for the use case outlined in this PR. Leveraging `link[rel="prefetch"]` was the first approach I looked at before authoring this PR. I too would prefer to leverage the native browser functionality. However the ergonomics seem to be quite challenging. There are two main problems:
- It doesn't appear to follow redirects (however this seems to be undefined in spec)
- It sensibly but painfully abides the cache-control header.

So personally I feel there is a benefit with being able to preload a preview while not having to hand cache control to the browser.

src/core/drive/preloader.ts
```diff
@@ -0,0 +1,54 @@
...
+import { PageSnapshot } from "./page_snapshot"
+import { SnapshotCache } from "./snapshot_cache"
+
+export interface PreloaderDelegate {
+  readonly navigator: Navigator
+}
+
+export class Preloader {
+  readonly delegate: PreloaderDelegate
+  readonly selector: string = 'a[data-turbo-preload="true"]'
```
ghiculescu, line 11:
```suggestion
  readonly selector: string = 'a[data-turbo-preload]'
```

I think this is the preferred style here when a truthy attribute is expected, for example see https://github.com/hotwired/turbo/blob/8221c1e5f2021c41c3ecada0dde55bae9ee18380/src/core/snapshot.ts#L29

hey-leon:
Hey turbo team, I wanted to circle back here as I have not stayed on top of this.

I can say that we have been using this in production to preload our hottest app paths with success. I am wondering if there is interest in merging this feature or a revision of this?

@seanpdoyle do you have any opinions?

ghiculescu:
@dhh do you have any feedback on the concept here?

silentjay:
Hoping this gets merged. Some of us have already implemented pre-fetching via 3rd party scripts mentioned in this old issue https://github.com/hotwired/turbo/issues/174 would be great to have it built in.

dhh:
This is great. Agree it would be nice to lean on a href's support when that's there, but don't think we should get ahead of the spec. If it ends up happening, it wouldn't be a big deal to go from data-turbo-preload to whatever that is. But as with the link preloads, it might well be that the cache-control dynamics don't match with what we're trying to do here.

ghiculescu:
Should we update https://github.com/hotwired/turbo-site to note this feature? Maybe as a new item in https://turbo.hotwired.dev/handbook/drive ?

hey-leon:
> This is great. Agree it would be nice to lean on a href's support when that's there, but don't think we should get ahead of the spec. If it ends up happening, it wouldn't be a big deal to go from data-turbo-preload to whatever that is. But as with the link preloads, it might well be that the cache-control dynamics don't match with what we're trying to do here.

Thanks for taking the time to understand the PR @dhh. I think your analysis pretty spot on as well.

> Should we update https://github.com/hotwired/turbo-site to note this feature? Maybe as a new item in https://turbo.hotwired.dev/handbook/drive ?
> Yes please do add a PR we can ship with 7.2.0 👍

Good suggestion. I will put something together 👍

clementmas:
Are the linked pages all preloaded on page load or only when each link becomes visible?

If not, I think that could be a great addition in a new PR. Using Intersection Observers. I think Next.js and Nuxt.js do that automatically.

hey-leon:
> Are the linked pages all preloaded on page load or only when each link becomes visible?
> 
> If not, I think that could be a great addition in a new PR. Using Intersection Observers. I think Next.js and Nuxt.js do that automatically.

That could be a pretty cool idea @clementmas

pfeiffer:
> > Are the linked pages all preloaded on page load or only when each link becomes visible?
> > If not, I think that could be a great addition in a new PR. Using Intersection Observers. I think Next.js and Nuxt.js do that automatically.
> 
> That could be a pretty cool idea @clementmas

In that case I'd prefer to have links annotated with `data-turbo-preload="lazy"` to opt-in to the lazy-preloading behavior. I believe that links explicitly hinted with `data-turbo-preload="true"` should still be preloaded, even outside of the viewport, as those links could be visited programatically.

A different approach to this could be to implement the functionality in this PR using MutationObserver instead of preloading links after a frame visit or on document load. This would allow developers to define their own behavior on top, eg. by implementing an Interaction Observer that sets the `data-turbo-preload` as links scroll into view.

It could also be used to add `data-turbo-preload` when the user has a "hover intent" (ie. like https://www.npmjs.com/package/hoverintent).

hey-leon:
> > > Are the linked pages all preloaded on page load or only when each link becomes visible?
> > > If not, I think that could be a great addition in a new PR. Using Intersection Observers. I think Next.js and Nuxt.js do that automatically.
> > 
> > 
> > That could be a pretty cool idea @clementmas
> 
> In that case I'd prefer to have links annotated with `data-turbo-preload="lazy"` to opt-in to the lazy-preloading behavior. I believe that links explicitly hinted with `data-turbo-preload="true"` should still be preloaded, even outside of the viewport, as those links could be visited programatically.
> 
> A different approach to this could be to implement the functionality in this PR using MutationObserver instead of preloading links after a frame visit or on document load. This would allow developers to define their own behavior on top, eg. by implementing an Interaction Observer that sets the `data-turbo-preload` as links scroll into view.
> 
> It could also be used to add `data-turbo-preload` when the user has a "hover intent" (ie. like https://www.npmjs.com/package/hoverintent).

Yes I don't think there is any reason to remove the current functionality for our purpose I'm not sure we would need the intersection observer but I cannot see why we can't have both approaches. Having said that if we wanted to keep the API simple I don't see any downside to the intersection observer approach for us. E.g. most things we are preloading are defined in fixed navbars and they usually load in such a small time frame the time between pressing the link and it coming into frame it would likely be loaded.

PR #553, opened 2022-03-18
Update <html> element attributes when navigating

manuelpuyol:
Turbo only replaces the `<body>` and updates the `<head>` when navigating, ignoring possible changes to the `<html>` element, like classes or `data-*` attributes.
With this change, Turbo will update the `<html>` when navigating through pages.

### DIFF
--- a/src/core/drive/page_renderer.ts
+++ b/src/core/drive/page_renderer.ts
@@ -49,7 +49,9 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
 
   updateHtmlElementAttributes() {
     for(const attr of document.documentElement.attributes) {
-      document.documentElement.removeAttribute(attr.nodeName)
+      if (!this.newElement.hasAttribute(attr.nodeName)) {
+        document.documentElement.removeAttribute(attr.nodeName)
+      }
     }
 
     for(const attr of this.newHtmlElement.attributes) {
--- a/src/core/drive/page_renderer.ts
+++ b/src/core/drive/page_renderer.ts
@@ -1,6 +1,8 @@
 import { Renderer } from "../renderer"
 import { PageSnapshot } from "./page_snapshot"
 
+const INTERNAL_ATTRIBUTES = ["aria-busy", "data-turbo-preview"]
+
 export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
   get shouldRender() {
     return this.newSnapshot.isVisitable && this.trackedElementsAreIdentical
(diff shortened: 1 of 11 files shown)


dhh:
Bunch of failures. But seem unlikely that they're directly related to this.

manuelpuyol:
👋  @dhh this _should_ be ready now! Mind approving the workflow?

dhh:
There's just one failure now, but that's definitely not related to this change, so good to go:

Screen Shot 2022-03-31 at 10 48 25

(But would be great to have a look at this too, so we can have a clean build!)

packagethief:
I'd argue that not changing the document element during navigation is a feature. At Basecamp we use `<html>` as a permanent element to hold state that we don't want to change between pages.

It's useful for state that can't be obtained on the server and that won't change during a session. For example, data about the current theme, client capabilities, client permissions, or other client preferences that only need to be obtained once.

@manuelpuyol, was this motivated by a particular use case?

dhh:
Yeah, looking at how we use HTML in Basecamp, this is going to be a regression. Can't ship this as is. Best we can do if we actually want to change this is to put it behind a configuration option.

manuelpuyol:
> @manuelpuyol, was this motivated by a particular use case?

Sorry for the delay!

We also set some state data in the `<html>`, which may change when navigating pages. So pages that were setting new attributes or even classes weren't working since the new `<html>` is discarded

packagethief:
> So pages that were setting new attributes or even classes weren't working since the new <html> was discarded

I see, thanks for the explanation @manuelpuyol!

It's intentional that we're only replacing the `<body>` element, and it's not something we can change without ceremony. We're calling out its persistence explicitly in the documentation:

> During rendering, Turbo Drive replaces the current `<body>` element outright and merges the contents of the `<head>` element. The JavaScript window and document objects, and the `<html>` element, persist from one rendering to the next.
> 
> https://turbo.hotwired.dev/handbook/introduction#turbo-drive%3A-navigate-within-a-persistent-process

If we did decide to change the behavior I think it would need to be more nuanced, such that changes to the document element are merged, as is done for `<head>`.

In any case, I think we should revert on the grounds that this is a breaking API change.

dhh:
Concur. I missed that originally. @manuelpuyol can you do a revert PR? Then we can give it another try behind a configuration option or something. But we can't break the API as-is.

manuelpuyol:
Working on a revert here https://github.com/hotwired/turbo/pull/596
I'm gonna keep the test setup changes to use localstorage though, I think they are still valuable :)

PR #557, opened 2022-03-30
Update intern and dependencies to fix test suite

dhh:
Resolve "BUG: suiteEnd was received for invalid session".


dhh:
Will use #551 instead once we've solved the chrome issue.

PR #558, opened 2022-03-30
Smaller dependency update just to resolve chrome driver issue

dhh:


dhh:
Will use #551 instead once we've solved the chrome issue.

PR #560, opened 2022-03-31
Raise turbo:error event on error

salomvary:
This pr continues abandoned work on https://github.com/hotwired/turbo/pull/357

I am struggling a bit with adding event firing tests as test cases for failure-scenario are missing for most of the affected functionality and I have not found any documentation on he expected behavior either. For example what do we expect  if a request in a Turbo frame fails? Seems like the frame is not updated at all. What do we expect if a Drive link fails? Seems like the page gets updated with the error response.

There is still a test failing and I'm not entirely sure what to expect here:
```
DriveTests - link click network error fires turbo:error event (0.53s)
    AssertionError: expected 'load' to equal 'advance'
    
    A load
    E advance
    
      at DriveTests.test link click network error fires turbo:error event @ src/tests/functional/drive_tests.ts:36:16
      at processTicksAndRejections @ internal/process/task_queues.js:95:5
      at async DriveTests.runTest @ src/tests/helpers/intern_test_case.ts:43:6
```

And lastly, I'm not entirely sure what the semantics of `turbo:error` should be. Is this for any fetch error response or exception? Should it be limited to fetch errors only? If that's the case, would `turbo:fetch-error` be a more appropriate name? I'm personally only interested in fetch error but #357 contains more.

Anyway, feedback welcome.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -1,3 +1,5 @@
+import { FetchResponse } from "./http/fetch_response";
+
 export type DispatchOptions = { target: EventTarget, cancelable: boolean, detail: any }
 
 export function dispatch(eventName: string, { target, cancelable, detail }: Partial<DispatchOptions> = {}) {
@@ -83,7 +85,7 @@ export function clearBusyState(...elements: Element[]) {
   }
 }
 
-export function reportError(error) {
+export function reportError(error: FetchResponse | unknown) {
   console.warn(error)
   dispatch('turbo:error', { detail: { error } })
 }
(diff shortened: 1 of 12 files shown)


src/util.ts
```diff
@@ -82,3 +84,8 @@ export function clearBusyState(...elements: Element[]) {
     element.removeAttribute("aria-busy")
   }
 }
+
+export function reportError(error: FetchResponse | unknown) {
```
salomvary, line 88: We might want a better name for this as `reportError` is an existing global: https://developer.mozilla.org/en-US/docs/Web/API/reportError

dhh: Can go with `processError` instead?

dhh:
Needs a rebase.

nate-at-gusto:
Thanks for resurrecting this @salomvary! I was just looking for this functionality today 😄

PR #561, opened 2022-03-31
Update Github actions to use Node 16

t27duck:
Node 12 will reach EOL at the end of April. Many Github actions have recently updated to use Node 16 (the Active LTS) and moved off of 12.

Updated `actions/checkout`, `actions/setup-node` and `actions/cache` in ci.yml to their most recent versions which are ran/built against Node 16. Also, chnged the node-version used in Github actions to 16.

The dev container appears to already be set up to use Ndoe 16.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


t27duck:
The failing test is the same as what's in `main` currently.

Issue #526, opened 2022-02-02
Turbo frame loads twice when it is in the response of a redirected request


karthikkasturitg:
Using turbo-rails with turbo-frame :

Assuming there is a page with an anchor that links to "/page" and the /page action redirects to "/main" where we have a response of .html.erb as such: Frame here: <turbo-frame id="data" src="/test/frame"></turbo-frame>

Expected behavior: On clicking the link to /page, main.html.erb should be loaded and since the response has a turbo frame, "/test/frame" should be called once to load the data for the frame

Actual behavior: "/test/frame" is called twice.

Can be replicated on: https://github.com/karthikkasturi/turbo-frame-loads-twice-on-redirect

dhh:
Solved via https://github.com/hotwired/turbo/pull/516

PR #516, opened 2022-01-20
Use 'will render' false for action 'replace' after GET redirect

blrB:
Todo list:
- [X] fix issue
- [X] pass previous tests
- [X] add tests

Closes #515

PR #563, opened 2022-04-03
Fix `turbo:load` firing twice after a redirect and avoid double render

### LINKED ISSUE #428: trubo:load gets triggered twice when GET request is a redirect 302
Hi, is this on purpose and if so, how can I make sure that the first call doesn't do anything?

[I created a test where you can test this problem](https://github.com/joker-777/TestAppTurbo)

Start there the server `bundle exec rails s` and open the page `localhost:3000`. You
will see a link that points to page 1 but then gets redirected to page 2. On the
first page load, you will see an alert which you then will see twice after
clicking on this link.

### LINKED ISSUE #537: When following a redirect, the resulting HTML is rendered to the page twice
I want to first caveat this issue with the fact that it is only causing me problems with flakiness using capybara and selenium based tests. As far as I'm aware, this hasn't caused any actual humans using this any problems.

The problem I'm seeing specifically relates to the following general circumstances:
- I start on a page with a form that submits via a POST.
- The endpoint for the POST then redirects to another URL with a form on it
- Using rspec and capybara to have a `fill_in('Label', with: 'value')` for this next form, I sometimes see an error where the first input isn't filled in properly.

My understanding of things is that Turbo renders the form once, capybara sees the input on the page and fills it, and then turbo re-renders the form clearing the input entry.

If I'm tracing things properly, I think this is the current set of actions:
1. Request is made and [the browser is told to follow the redirect](https://github.com/hotwired/turbo/blob/b419c0de0b82025b1c7ba644efcb12c57c23d05d/src/http/fetch_request.ts#L116)
2. The response that actually [makes it to the Visit class](https://github.com/hotwired/turbo/blob/b88bb245652d1b8b7cfe89d58bdff46344ca14f1/src/core/drive/visit.ts#L219-L222) is the final, redirected response with a 200 status code and HTML to render. This is the first rendering of the page/form.
3. In the cleanup, there is a check if there was a redirect, and if so it [triggers a visit to that new URL using the same response](https://github.com/hotwired/turbo/blob/539b249d82f44f5a4fd6bd245fe1e45718b31508/src/core/drive/visit.ts#L272-L280). I believe this is required to update the browser history properly.
4. That recursion to a new Visit doesn't just update the history though, it will also update the DOM again, which resets anything that is happened (in the milliseconds since the last render).

I'm not sure if there are any use cases where you'd want the second DOM update, but it seems like it would be best to just update the history on the recursion and not do a full update?

feliperaul:
Hi there! This is my first contribution to turbo.

This fixes #428, #526 and may fix #537 as well.

Turbo is firing `turbo:load` twice when the server responds with a redirect (regardless of status code, it happens with 302, 303). 

This is harmful because [per the official docs](https://turbo.hotwired.dev/handbook/building#observing-navigation-events), `turbo:load` replaces the native `DOMContentLoaded`, which **never** fires twice on a redirect. The docs clearly state `turbo:load` should fire only once. 

Our team had to halt the migration to Turbo because of this; not only analytics events listening to `turbo:load` were registering two visits while there actually just one, but many pages had `GET` and even `POST` javascript events fired on `turbo:load`, some of them that even updated a Rails model's `lock_version` on the backend, making it so that when the second render was actually mounted on the screen it couldn't save the model because it was already stale (due to the previous "ghost" turbo:load inadvertently firing). 

Our code _is_ idempotent; the fact is Turbo is firing `turbo:load` twice and it's impossible to distinguish one from another (let alone this shouldn't be a extra burden on the developer). 

I'm aware of the pull request at #516 , but it appears to only solve the double render, but the event will still fire twice;
I tested it and it does not solve the double `turbo:load` firing. This pull request solves both issues.

PS: I added a test that is passing on Firefox, but after hours of attempts I couldn't make it pass on Chrome (it fails with a `StaleElementReference`, and I couldn't find any test on the suite that asserted events after a redirect). 

You will also see `NavigationTests - skip link with hash-only path moves focus and changes tab order` failing, but it is already failing on `main` so it didn't break due to this pull request.


dhh:
Looks like we have some test failures?

dhh:
Screen Shot 2022-06-19 at 10 55 57

Want to take a look at this @feliperaul?

feliperaul:
@dhh Hi David, thanks for bumping me on this.

I wish I could help further... the fact is that when I issued the PR I added a test for my change, and the test is passing on Firefox, but after literally hours of attempts I couldn't make it pass on Chrome (it always fails with this `StaleElementReference` you are seeing, and I couldn't find any test on the suite that asserted events after a redirect as reference).

Sorry for issuing a PR with a failing test, but I mentioned this on the comment above hoping someone could chip in and get this test passing on Chrome so this could be merged. I hit a wall here. Maybe you could cc someone that could give it a try?

domchristie:
`turbo:visit` is also triggered twice on a redirect, which can result in double animations when animating page exits.

I wonder if visits should handle redirects internally, or have some way of determining a visit that originated from a redirect e.g. via a property on the `turbo:visit` `event.detail`?

PR #564, opened 2022-04-04
Make data-turbo-confirm work with multiple submitters inside a form

feliperaul:
This pull request makes so that you can use different `data-turbo-confirm` on each submitter inside your form, instead of allowing only one `data-confirm` per form.

Problem: If you're nested inside a form, you can't use `button_to` because you'd end up with a nested form, which is invalid and discarded by the browser.

So, how one would accomplish a UI like the following?

image

That delete button is inside a form. A `button_to` there isn't allowed. 

However, as @seanpdoyle greatly points [here](https://thoughtbot.com/blog/dynamic-forms-with-turbo), the browser spec allows you to change the default `action` and `method` of the form **on the submitter**, by using `formaction` and `formmethod` attributes **on the button itself**.

So you could go for this:

```erb
<%= f.button formaction: destroy_all_empty_path, formmethod: :delete, class: "btn btn-default pull-right", data: { turbo_frame: :_top, turbo_confirm: "Confirm destroying all empty records?" } do %>
  <span class="when-enabled"><i class='fa fa-trash-alt' ></i> Destroy all empty</span>
  <span class="when-disabled"><i class="fa fa-spinner fa-spin"></i>  Destroying</span>
<% end %>
```

And you're done. Turbo will submit the form with the correct DELETE verb, to the correct `destroy_all_empty_path`, will automatically disable the submitter and, with the clever CSS we already know, we can show a loading state on the button.

There was only one piece missing, which this pull request addresses: the `data-turbo-confirm` currently doesn't work on that button, because the code only looks at the encompassing `form` `data-turbo-confirm`. 

We now will look first to the `data-turbo-confirm` **of the form submitter** and show it instead of the form's if it's present.

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -100,7 +100,7 @@ export class FormSubmission {
   }
 
   get confirmationMessage() {
-    return this.formElement.getAttribute("data-turbo-confirm")
+    return this.submitter?.getAttribute("data-turbo-confirm") || this.formElement.getAttribute("data-turbo-confirm")
   }
 
   get needsConfirmation() {
--- a/src/tests/fixtures/form.html
+++ b/src/tests/fixtures/form.html
@@ -75,6 +75,7 @@
       </form>
       <form action="/__turbo/redirect" method="post" class="confirm" data-turbo-confirm="Are you sure?">
         <input type="submit">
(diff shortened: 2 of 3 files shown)


feliperaul:
For future readers, this worked on 7.1.0, but broke in 7.3.0 with Rails 7. 

Check `https://github.com/hotwired/turbo-rails/pull/445` for an open PR fixing the regression. 

In the meantime, a simple fix is change your `<%= f.button formaction: whatever_delete_path, formmethod: :delete } do %>` `<%= f.button name: :whatever, formaction: whatever_delete_path, formmethod: :delete } do %>`. 

Simply by setting a `name` will prevent the default override of `formmethod` and it will make it work as before.

Issue #387, opened 2021-09-10
Turbo scrolls to Top of current page before rendering new visited page (related to data-turbo-track="reload")


chipairon:
I have a demo of the issue here https://prairie-swamp-observatory.glitch.me/index.html
`second_page.html` has a script with `data-turbo-track="reload"` that the first page does NOT have.

When clicking the link at the bottom of the page (from either page):
1- The scroll is moved to the top of the current page.
2- Then the new page is rendered.

My expectation would be that the current page is not scrolled to top before the next page is rendered.

If the `turbo-track:reload` attribute is removed from the script tag, the scroll does not happen:
```html
<script src="/second_page_script.js" defer data-turbo-track="reload"></script>
```

The source code of the demo is here: https://glitch.com/edit/#!/remix/prairie-swamp-observatory

The demo loads Turbo from skypack (currently `@hotwired/turbo@7.0.0-rc.3`).

PS: it might be related to the behaviour described in [this issue](https://github.com/hotwired/turbo/issues/114) but I'm not sure.

srt32:
Thanks for reporting this! We've bumped into a similar issue. I've been debugging for a bit but no success yet. I can report that I can repro the issue by running the server and going to http://localhost:9000/src/tests/fixtures/rendering.html.

Then, scroll down (you may need to shrink the height of your window), and click on "Visit control: reload". The URL will update, the page scroll up, and then navigate.

srt32:
I've determined the problem is in https://github.com/hotwired/turbo/blob/da35d5b440e226300927b90d12c02e0cade30931/src/core/drive/visit.ts#L411 and in the "reload" case we shouldn't scroll but I'm not sure how to properly gate this action without breaking other behavior like properly scrolling back to anchors on the same page.

PR #571, opened 2022-04-18
Don't scroll to top of page on reload

srt32:
Potential solution for https://github.com/hotwired/turbo/issues/387

This change updates how we handle scrolling when the page gets reloaded due to a `reload` directive.

I'm not sure if this is the proper way to gate the behavior but I was able to confirm the fix manual using the fixtures html files.

I'd also love to hear if there is a good way write a test for this fix.

## Before:

https://user-images.githubusercontent.com/2181356/163845443-0c025ea8-5c5c-465e-9c6e-c886164bea72.mov

## After:

https://user-images.githubusercontent.com/2181356/163845321-6354af64-b160-47ae-86fc-922e112614c1.mov

### DIFF
--- a/src/core/drive/page_view.ts
+++ b/src/core/drive/page_view.ts
@@ -18,7 +18,7 @@ export class PageView extends View<Element, PageSnapshot, PageViewRenderer, Page
 
   renderPage(snapshot: PageSnapshot, isPreview = false, willRender = true) {
     const renderer = new PageRenderer(this.snapshot, snapshot, isPreview, willRender)
-    if (!renderer.trackedElementsAreIdentical) { this.forceReloaded = true }
+    if (!renderer.shouldRender) { this.forceReloaded = true }
     return this.render(renderer)
   }
 
--- a/src/core/drive/page_view.ts
+++ b/src/core/drive/page_view.ts
@@ -18,8 +18,7 @@ export class PageView extends View<Element, PageSnapshot, PageViewRenderer, Page
 
   renderPage(snapshot: PageSnapshot, isPreview = false, willRender = true) {
     const renderer = new PageRenderer(this.snapshot, snapshot, isPreview, willRender)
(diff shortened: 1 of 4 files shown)


srt32:
@seanpdoyle 👋 could I bug you to give me some guidance on next steps here?

seanpdoyle:
Thank you for taking this on @srt32!

> I'd also love to hear if there is a good way write a test for this fix.

They aren't perfect, but there is a [scrollToSelector](https://github.com/hotwired/turbo/blob/main/src/tests/helpers/functional_test_case.ts#L65-L68) and [isScrolledToSelector](https://github.com/hotwired/turbo/blob/main/src/tests/helpers/functional_test_case.ts#L105-L110) methods available to the test suite. When interacting with them, make sure that the content being scrolled to or away from is far enough above or below the fold for the headless browsing session to change the scroll depth, otherwise the delta in scroll `y` won't be significant enough for the assertion.

It isn't exactly the same, but https://github.com/hotwired/turbo/pull/476/files#diff-b7c9a06f6b41512fee24c53728cacc1b98c47e90ac291d047e576a686ee86c60 attempts to add coverage for a scroll-related issue.

srt32:
Thank you! I've started in on a spec but I'm struggling to reproduce a test where I can assert the scroll position _prior_ to the page transitioning to the new URL. Do you have a recommendation on somehow pausing the transition at the right time?

srt32:
@manuelpuyol and I got a spec added that went red, green. It gets creative with localStorage and we saw it is a bit flaky with chrome (ie - it passes when it should not have). If we remove `headless`, it's consistent. We're very open to any feedback. Thanks! ❇️

srt32:
👋 I wonder if y'all could advise on next best steps here. Thanks!

dhh:
Great work on this!

PR #575, opened 2022-05-02
Generated: execute `yarn lint --fix`

seanpdoyle:
Generated: execute `yarn lint --fix`
===

Follow-up to [hotwired/turbo#573][].

ESLint support was recently merged and integrated in to the Continuous
Integration checks, but excluded several existing linting violations.

This commit's diff was generated by executing `yarn lint --fix`.

These violations were discovered when rebasing [hotwired/turbo#436][].
Once this is passing and merged, these changes will be rebased into that
branch's changeset.

[hotwired/turbo#573]: https://github.com/hotwired/turbo/pull/573
[hotwired/turbo#436]: https://github.com/hotwired/turbo/pull/436/files

Resolve failing `FormSubmissionTest`
===

There is a race condition in the `FormSubmissionTest` file's coverage
for confirmation. If evaluated quickly enough, the test's assertions
might be checked before the subsequent Turbo Driver navigation occurs,
which would fail the check for the Form Submission's change to the
current path.

This commit adds an additional `await` expression to wait until the next
[turbo:load][] event to fire, so that the entire navigation can complete
before the assertions are evaluated.

[turbo:load]: https://turbo.hotwired.dev/reference/events

### DIFF
--- a/src/core/drive/page_renderer.ts
+++ b/src/core/drive/page_renderer.ts
@@ -12,13 +12,13 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
   get reloadReason(): ReloadReason {
     if (!this.newSnapshot.isVisitable) {
       return {
-        reason: "turbo_visit_control_is_reload"
+        reason: "turbo_visit_control_is_reload",
       }
     }
 
     if (!this.trackedElementsAreIdentical) {
       return {
-        reason: "tracked_element_mismatch"
+        reason: "tracked_element_mismatch",
       }
     }
   }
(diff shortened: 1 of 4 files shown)


dhh:
Wonder if we could tweak the lint rule here to skip that trailing comma on hashes with just a single pair. Looks bizarre to me 😄

Good to go on the test fix. Probably easier to separate stuff like this into separate PRs going forward.

seanpdoyle:
The latest commits on `main` have failing CI builds due to the ESLint changes:

* https://github.com/hotwired/turbo/runs/6249193108?check_suite_focus=true
* https://github.com/hotwired/turbo/runs/6239964262?check_suite_focus=true

I was hoping to restore a passing build with this PR. I feel like it'd be more appropriate to configure ESLint to match preferences with a passing suite.

I was hoping that the ESLint changes could be made in isolation, but the test flakiness addressed in the second commit was also failing CI.

Unfortunately in this case, each change depends on the other change to pass CI.

dhh:
Ah, I see. Could we change the eslint config to drop the trailing commas? Anyway, let's get this passing first.

PR #579, opened 2022-05-04
Add failed test

inkstak:


src/core/frames/frame_controller.ts
```diff
@@ -172,14 +172,15 @@ export class FrameController
...
 
   // Fetch request delegate
 
-  prepareHeadersForRequest(headers: FetchRequestHeaders, _request: FetchRequest) {
-    headers["Turbo-Frame"] = this.id
+  prepareHeadersForRequest(headers: FetchRequestHeaders, _request: FetchRequest, id?: string | null) {
+    headers["Turbo-Frame"] = id || this.id
   }
 
   requestStarted(_request: FetchRequest) {
```
tleish, line 186: I suggest using the existing method `this.findFrameElement` to locate the turbo-frame.  Perhaps something like:

```diff
  formSubmissionIntercepted(element: HTMLFormElement, submitter?: HTMLElement) {
    if (this.formSubmission) {
      this.formSubmission.stop()
    }
  
    this.reloadable = false
    this.formSubmission = new FormSubmission(this, element, submitter)
    const { fetchRequest } = this.formSubmission
-   this.prepareHeadersForRequest(fetchRequest.headers, fetchRequest)
+   const frame = this.findFrameElement(element, submitter)
+   this.prepareHeadersForRequest(fetchRequest.headers, fetchRequest, frame);
    this.formSubmission.start()
  }

// Fetch request delegate

- prepareHeadersForRequest(headers: FetchRequestHeaders, _request: FetchRequest) {
+ prepareHeadersForRequest(headers: FetchRequestHeaders, _request: FetchRequest, frame: FrameElement = this.element) {
-   headers["Turbo-Frame"] = this.id
+   headers["Turbo-Frame"] = frame.id
  }
```

inkstak: This was my first approach but... if the target frame is missing, -- which throws an error but doesn't interrupt the form submission -- it won't send the correct header.

tleish: The diff I show above does exactly what you suggest.  Do have a sample where it's not working?

inkstak:
````html
<turbo-frame id="frame">
  <form data-turbo-frame="hello">
    (...)
  </form>
</turbo-frame>
````

`findFrameElement` always return a FrameElement.
If there is no `<turbo-frame id="hello">` in the current page, it will return the parent frame `<turbo-frame id="frame">` instead.

For reference:
````js
private findFrameElement(element: Element, submitter?: HTMLElement) {
  const id = getAttribute("data-turbo-frame", submitter, element) || this.element.getAttribute("target")
  return getFrameElementById(id) ?? this.element
}
````

Given so, the form submission will send a wrong header `Turbo-Frame: frame` if the target frame (hello) is missing.

With the solution I provided, we just try to extract the target ID to send the correct header, nevermind the target frame is present or not. Throwing an exception is another concern.

However, we can refactor the ID extraction into a new function `findFrameId`.
It's also used in [shouldInterceptNavigation](https://github.com/hotwired/turbo/blob/main/src/core/frames/frame_controller.ts#L333)


````diff
   formSubmissionIntercepted(element: HTMLFormElement, submitter?: HTMLElement) {
     [...]
-    this.prepareHeadersForRequest(fetchRequest.headers, fetchRequest)
+    const id = findFrameId(element, submitter)
+    this.prepareHeadersForRequest(fetchRequest.headers, fetchRequest, id)
     this.formSubmission.start()
   }

+  private findFrameId(element: Element, submitter?: HTMLElement) {
+    const id = getAttribute("data-turbo-frame", submitter, element) || this.element.getAttribute("target")
+    return id ?? this.id
+  }

   private findFrameElement(element: Element, submitter?: HTMLElement) {
-    return getFrameElementById(id) ?? this.element
+    const id = getAttribute("data-turbo-frame", submitter, element) || this.element.getAttribute("target")
+    return getFrameElementById(findFrameId(element, submitter)
   }

[...]

   private shouldInterceptNavigation(element: Element, submitter?: HTMLElement) {
-    const id = getAttribute("data-turbo-frame", submitter, element) || this.element.getAttribute("target")
+    const id = findFrameId(element, submitter)
````

Comment (inkstak) [reply src/core/frames/frame_controller.ts:186]: I've pushed two more commits :
- [one test to verify header sent when target frame is missing](https://github.com/hotwired/turbo/pull/579/commits/9a3a3654e1e06cad6c58adda856388fcb7428111)
- [refactor of ID extraction](https://github.com/hotwired/turbo/pull/579/commits/7c4657375f294ab36fe6171155d8948734375d0f)

Comment (tleish) [reply src/core/frames/frame_controller.ts:186]: The original implementation of `data-turbo-frame` by @seanpdoyle purposely falls back to parent turbo-frame.  The comment from the commit states:

> When navigating via an `<a>` click within a `<turbo-frame>`, determine the target of the navigation by checking for `[data-turbo-frame]` overrides on the `<a>` element itself, and then check for frames referenced by `turbo-frame[target]`, then fall back to the frame itself.

see: https://github.com/hotwired/turbo/commit/f62ad3b3069597c60334b1d0623c53d6432252ff

The proposed changes creates inconsistent behavior for links and form submissions when the target for `data-turbo-frame` is missing.

- link: falls back to parent turbo-frame and sends request
- form: fails request

Comment (inkstak) [reply src/core/frames/frame_controller.ts:186]: You're right, it should be the same behavior for links and forms.
I also agree with behavior that fall back to the parent frame to performs DOM updates.

But IMHO, the header we sent should be the frame it was meant to be, no matter the frame exists or not.

I understand that this may be open to debate and may lead to breaking changes, so I'll refactor this PR to use `findFrameElement` and open another issue to debate discuss this last behavior.

**Comment by dhh** [comment]
cc @seanpdoyle

**Comment by tleish** [comment]
FYI, we just ran into this issue.  Trying to figure out a workaround for now.

**Comment by sukei** [comment]
Just ran into this issue too. The default behavior is counterintuitive IMO.

**Comment by rmeoc** [comment]
I also encountered this issue.
I would like to use the Turbo-Frame header as a way for the server to leave out html that is expensive to generate. For this to work the Turbo-Frame header needs to indicate what turbo-frame will be updated with the returned html.

**Comment by tleish** [comment]
Perhaps a monkeypatch like following might work in the meantime.

```javascript
// Monkeypatch to fix turbo issue of wrong turbo-frame
// see: https://github.com/hotwired/turbo/pull/579
document.addEventListener('turbo:before-fetch-request', (event) => {
  const targetTurboFrame = event.target.getAttribute('data-turbo-frame');
  const fetchTurboFrame = event.detail.fetchOptions.headers['Turbo-Frame'];
  if (targetTurboFrame && targetTurboFrame != fetchTurboFrame && document.querySelector(`turbo-frame#${targetTurboFrame}`)) {
    event.detail.fetchOptions.headers['Turbo-Frame'] = targetTurboFrame;
  }
});
```

rmeoc:
Thanks! This monkeypatch does not correct the case where the submit button has an data-turbo-frame attribute, but I can certainly use it.

tvongaza:
We've run into this too, the monkey patch does the job though would be great to get this working upstream - I've taken a look at the PR but need to spend more time with it to fully understand what is going on. I've created the following spec to document it in our code base, unsure if it would be helpful to others:

turbo_monkeypatch.ts
```typescript
// Monkeypatch to fix turbo issue of wrong turbo-frame
// see: https://github.com/hotwired/turbo/pull/579
export async function addTurboFramePatch() {
  document.addEventListener("turbo:before-fetch-request", (event: any) => {
    const targetTurboFrame = event.target.getAttribute("data-turbo-frame");
    const fetchTurboFrame = event.detail.fetchOptions.headers["Turbo-Frame"];
    if (
      targetTurboFrame &&
      targetTurboFrame != fetchTurboFrame &&
      document.querySelector(`turbo-frame#${targetTurboFrame}`)
    ) {
      event.detail.fetchOptions.headers["Turbo-Frame"] = targetTurboFrame;
    }
  });
}
```

turbo_monkeypatch.spec.ts
```typescript
import { addTurboFramePatch } from "@/lib/turbo_monkeypatch";
import { fireEvent } from '@testing-library/dom';
import { Turbo } from "@hotwired/turbo-rails"
import MockIntersectionObserver from '@/spec/support/mockIntersectionObserver';
import createFetchMock from 'vitest-fetch-mock';
import { vi } from 'vitest';

// See: https://github.com/hotwired/turbo/pull/579
describe("addTurboFramePatch", () => {
  const fetchMocker = createFetchMock(vi);
  let originalIntersectionObserver: any;

  beforeEach(() => {
    originalIntersectionObserver = window.IntersectionObserver;
    window.IntersectionObserver = MockIntersectionObserver;
    fetchMocker.enableMocks();
  });

  afterEach(() => {
    window.IntersectionObserver = originalIntersectionObserver;
    fetchMocker.disableMocks();
  });

  // This spec ensures that when Turbo version changes that our links behaviour works as expected
  // This also provides a reference for how we *want* our forms to behave without the monkey patch
  it("updates Turbo-Frame header based on link data-turbo-frame attribute", async () => {
    // Set up the DOM and event
    document.body.innerHTML = `
      <turbo-frame id="frame1">
        <turbo-frame id="frame2">
          <a href="/default" id="default">Default (parents frame - frame2)</a>
          <a href="/frame1" data-turbo-frame="frame1">Frame 1</a>
          <a href="/frame2" data-turbo-frame="frame2">Frame 2</a>
          <a href="/frame3" data-turbo-frame="frame3">Frame 3</a>
        </turbo-frame>
        <turbo-frame id="frame3">
        </turbo-frame>
      </turbo-frame>
    `;

    Turbo.start();

    async function submitFormAndCheckHeader(formSelector: string, expectedTurboFrame: string) {
      return new Promise<void>(async (resolve) => {
        fireEvent.click(document.querySelector(formSelector));
        fetchMocker.mockResponse((req) => {
          expect(req.headers.get("Turbo-Frame")).toBe(expectedTurboFrame);
          resolve();
          return "";
        });
      });
    }

    await (async () => {
      await submitFormAndCheckHeader("a#default", "frame2");
      await submitFormAndCheckHeader("a[data-turbo-frame='frame1']", "frame1");
      await submitFormAndCheckHeader("a[data-turbo-frame='frame2']", "frame2");
      await submitFormAndCheckHeader("a[data-turbo-frame='frame3']", "frame3");
    })();
  });

  // This spec ensures that when Turbo version changes that our forms behaviour works as expected
  // Currently this is broken, so we introducde a monkey patch to fix it
  it("updates Turbo-Frame header based on form data-turbo-frame attribute", async () => {
    // Set up the DOM and event
    document.body.innerHTML = `
      <turbo-frame id="frame1">
        <turbo-frame id="frame2">
          <form id="default">
            <button type="submit">Submit to default (parents frame - frame2)</button>
          </form>
          <form data-turbo-frame="frame1">
            <button type="submit">Submit to frame 1</button>
          </form>
          <form data-turbo-frame="frame2">
            <button type="submit">Submit to frame 2</button>
          </form>
          <form data-turbo-frame="frame3">
            <button type="submit">Submit to frame 3</button>
          </form>
        </turbo-frame>
        <turbo-frame id="frame3">
        </turbo-frame>
      </turbo-frame>
    `;

    Turbo.start();

    async function submitFormAndCheckHeader(formSelector: string, expectedTurboFrame: string) {
      return new Promise<void>(async (resolve) => {
        fireEvent.submit(document.querySelector(formSelector));
        fetchMocker.mockResponse((req) => {
          expect(req.headers.get("Turbo-Frame")).toBe(expectedTurboFrame);
          resolve();
          return "";
        });
      });
    }

    // Test the default (broken) behavior, let's us validate if we can remove our patch on version upgrades
    await (async () => {
      await submitFormAndCheckHeader("form#default", "frame2");
      await submitFormAndCheckHeader("form[data-turbo-frame='frame1']", "frame2");
      await submitFormAndCheckHeader("form[data-turbo-frame='frame2']", "frame2");
      await submitFormAndCheckHeader("form[data-turbo-frame='frame3']", "frame2");
    })();

    // Install our patch
    await addTurboFramePatch();

    // Test the patched behavior
    await (async () => {
      await submitFormAndCheckHeader("form#default", "frame2");
      await submitFormAndCheckHeader("form[data-turbo-frame='frame1']", "frame1");
      await submitFormAndCheckHeader("form[data-turbo-frame='frame2']", "frame2");
      await submitFormAndCheckHeader("form[data-turbo-frame='frame3']", "frame3");
    })();
  });
});
```

mockIntersectionObserver.ts
```typescript
// Minimal mock implementation of IntersectionObserver for testing purposes.
// This is required because JSDOM does not implement IntersectionObserver and it is needed for Turbo to work.
export default class {
  readonly root: Element | null;
  readonly rootMargin: string;
  readonly thresholds: ReadonlyArray<number>;

  constructor() {
    this.root = null;
    this.rootMargin = '';
    this.thresholds = [];
  }

  disconnect() {}
  observe() {}
  takeRecords(): IntersectionObserverEntry[] { return [];}
  unobserve() {}
}
```

marcelolx:
We also ran into this today, I wonder if we should remove this section 👇  from the website docs https://turbo.hotwired.dev/reference/frames since it doesn't reflect reality 
```html
<turbo-frame id="messages">
  ...

  <form action="/messages" data-turbo-frame="navigation">
    Submitting form will replace the navigation frame.
  </form>
</turbo-frame>
```

packagethief:
Thanks @inkstak. This bit me recently, too. Your change looks good. I took the liberty of updating and rebasing against `main`.

packagethief (approved):

JoshAntBrown:
Nice! We ran into this issue today, too. I almost ended up creating a very similar PR until I found this one. Will be good to get this merged. 🙏

Issue #587, opened 2022-05-16
data-turbo-method ignored when data-turbo-frame is set


botandrose:
Hi folks, thanks for Turbo! I'm feeling the joy of server-side webdev like its 2005 again.

However, I've hit a snag with using data-turbo-method and data-turbo-frame at the same time:
+ I can use `<a data-turbo-method="post">` to POST within a turbo frame, as expected
+ I can use `<a data-turbo-frame="results">` to update the turbo frame from outside
- I CANNOT use `<a data-turbo-method="post" data-turbo-frame="results">`

Here is a JSFiddle demonstrating the issue:

https://jsfiddle.net/6kvyxLmf/

As you can see, when data-turbo-frame is set, all the fetch methods are made with GET, not the method specified in data-turbo-method. Shall I proceed with a PR?

dhh:
cc @seanpdoyle

seanpdoyle:
@botandrose was this resolved by https://github.com/hotwired/turbo/pull/631#issuecomment-1186256417? This test case in particular seems to reproduce your issue:

https://github.com/hotwired/turbo/blob/f8a94c5f761c4f080d68a87459d3916eb4fb6993/src/tests/fixtures/form.html#L302

https://github.com/hotwired/turbo/blob/f8a94c5f761c4f080d68a87459d3916eb4fb6993/src/tests/functional/form_submission_tests.ts#L901-L907

botandrose:
@seanpdoyle Confirmed! I just switched from my hacked fork to 7.2.0-beta.2, and everything works!

Thanks again for your efforts here, and on Turbo in general.Turbo and import maps reaffirm that my job doesn't have to be Node and Webpack and React, so that means I'll continue to enjoy it :)

PR #588, opened 2022-05-16
Can use both data-turbo-method and data-turbo-frame in a link

botandrose:
Attempt to fix https://github.com/hotwired/turbo/issues/587

Here's a failing test. Will push a proposed fix next.


dhh:
cc @seanpdoyle

seanpdoyle:
> I'm guessing there's a better way to solve this at a more holistic, conceptual level, but I honestly don't grok the turbo internals enough to know what that might be!

I've opened https://github.com/hotwired/turbo/pull/631 to introduce an abstraction that happens to resolve this underlying issue.

botandrose:
@seanpdoyle Ah, there's that conceptual improvement I was imagining! Beautiful work

PR #596, opened 2022-06-02
Revert https://github.com/hotwired/turbo/pull/553

manuelpuyol:
Reverting https://github.com/hotwired/turbo/pull/553 since it introduced a breaking change. We can revisit this and implement it behind a configuration


manuelpuyol:
@dhh I think this is ready for some 👀

dhh:
@packagethief can you verify that we are good?

packagethief:
Looking good. Thanks @manuelpuyol 🙏

Issue #547, opened 2022-03-04
Custom elements aren't treated as turbo frames


dvisockas:
Hi! 👋 

There is an issue between turbo and tables - turbo-frames cannot wrap `tr, tbody` and similar elements without breaking the styles.
One of the workarounds is to use custom-elements and just write `<tr is="turbo-frame">` - solves most of the problems.

However, if you do a request from within this fake `turbo-frame` element and the response returns a template, turbo cannot find a matching `turbo-frame` element and returns an error: `Response has no matching <turbo-frame...`

Personally I fixed it with patching `turbo` query selecting.
Currently it's:
```
querySelector(`turbo-frame#${id}`)
```

Changing it to this worked:
```
querySelector(`turbo-frame#${id}, [is="turbo-frame"]#${id}`)
```

Would this change be acceptable as a PR or this (custom elements with is="" attribute) is not the way to go with turbo?

jacobdaddario:
Hey! I have no authority as to what is and isn’t PR worthy, but I was reading through issues and wanted to see if I can help.

Your solution seems reasonable, but it definitely kind of reads as a workaround as opposed to actually addressing the issue with Turbo. I think the issue at the root of this bug is that Turbo breaks the formatting of table elements. Could you describe this in more detail? What about the styles break when you wrap them in a Turbo Frame? Are you able to spin this up in a reproducible example on the current version of Turbo?

seanpdoyle:
Is this related to https://github.com/hotwired/turbo/issues/48?

dvisockas:
> Is this related to #48?

It is in a way that it talks about how turbo and tables don't go well together. However #48 ends with "Rails scaffold does not use tables" and I believe that turbo still could support them.

francesco-loreti:
Hi

> It is in a way that it talks about how turbo and tables don't go well together. However #48 ends with "Rails scaffold does not use tables" and I believe that turbo still could support them.

I don't understand why a working "one shot" function has been implemented. If the tables don't have to exist in turbo, why keep this attribute?

tleish:
FYI, in 2020 they officially added the `is` attribute to the HTML spec in the [Custom Elements specification](https://html.spec.whatwg.org/multipage/custom-elements.html#attr-is).  Edge just recently added support for it, but as of today Safari and IE do not support it.

![image](https://user-images.githubusercontent.com/1099993/168382290-f73622fd-345c-4ff5-9ae8-32f9739c22cd.png)

[...]

PR #599, opened 2022-06-08
Fix issue #547

francesco-loreti:


dhh:
Please present this PR in a way that encapsulates the problem, the limitations of the current solution (i.e. that it doesn't work in Safari), includes doc changes, and also a test. Thanks for contributing!

PR #601, opened 2022-06-14
Improve scroll position when navigating

manuelpuyol:


src/core/drive/page_view.ts
```diff
@@ -16,15 +17,18 @@ export class PageView extends View<Element, PageSnapshot, PageViewRenderer, Page
   lastRenderedLocation = new URL(location.href)
   forceReloaded = false
 
-  renderPage(snapshot: PageSnapshot, isPreview = false, willRender = true) {
+  renderPage(snapshot: PageSnapshot, isPreview = false, willRender = true, visit?: Visit) {
```
koddsson, line 20: Probably in a follow-up refactor but it's starting to look like this function should take in a "options" object rather than individual arguments.

src/tests/fixtures/visit_control_reload.html
```diff
@@ -6,8 +6,17 @@
...
+    </style>
+   </head>
+   <body>
+     <h1>Visit control: reload</h1>
+
+    <hr class="push-off-screen">
+
+    <h2 id="middle">Middle the page</h2>
+    <hr class="push-off-screen">
+    <h2>Down the page</h2>
```
koddsson, line 20: Very nit-pick but:

```suggestion
     <hr class="push-off-screen">

     <h2 id="middle">Middle the page</h2>
     <hr class="push-off-screen">
     <h2>Down the page</h2>
```

koddsson (approved):
I'm not on the Turbo team and have a bias towards getting this merged but this looks good to me 😄

dhh:
cc @packagethief

src/core/native/browser_adapter.ts
```diff
@@ -27,9 +28,9 @@ export class BrowserAdapter implements Adapter {
   }
 
   visitStarted(visit: Visit) {
+    this.location = visit.location
     visit.loadCachedSnapshot()
     visit.issueRequest()
-    visit.changeHistory()
```
packagethief, line 32: I'm not positive, but moving the history changing to `PageView` might have implications for the native adapters for iOS and Android. The adapters will still be calling `visit.changeHistory` themselves, so it will be called twice:

- https://github.com/hotwired/turbo-ios/blob/2277f7d/Source/WebView/turbo.js#L71
- https://github.com/hotwired/turbo-android/blob/14ad418/turbo/src/main/assets/js/turbo_bridge.js#L70

The second should be a no-op because history will have been flagged as changed, but I think we'll still have the same problem.

manuelpuyol: how do we avoid calling it twice now?

packagethief: It's only called once by each adapter now. If we move it out of the browser adapter, we may need to move it out of the other adapters as well. Perhaps we don't (it's been a while since I've looked at the native frameworks), but we need to make sure it behaves as expected on iOS and Android.

manuelpuyol: Any tips on how to test that properly? Looking at https://github.com/manuelpuyol/turbo/blob/2eb9ea2c7282f1ec2b41c75705d7e43af7e2076a/src/core/drive/visit.ts#L173-L174 I'd expect it to **not** do a double update

packagethief: You're right, it will be a no-op because we've already flagged history as having changed. My concern is that we're changing the behavior by moving this out of the adapter, so we just need to ensure it's all good on iOS and Android 👍

dhh: @jayohms Can you verify whether this would bring any complication on native?

jayohms: I don't foresee any issues on the mobile side, since as mentioned, the second `visit.changeHistory` call will be a no-op, so the mobile adapters will continue to work as before. It'd be nice to update the mobile adapters to reflect this change, however, we've maintained backwards/forwards compatibility for all iterations of `turbo/links` API changes. 

I don't think we'll be able to reasonably make corresponding changes in the mobile adapters without exposing an additional API, so the adapters can know to conditionally change history (or not). This shouldn't be a problem in the short-term, but it may be a nuance that's difficult for others to understand as time goes on.

src/core/native/browser_adapter.ts
```diff
@@ -121,7 +122,10 @@ export class BrowserAdapter implements Adapter {
 
   reload(reason: ReloadReason) {
     dispatch("turbo:reload", { detail: reason })
-    window.location.reload()
+
+    if (!this.location) return
+
+    window.location.href = this.location.toString()
```
jayohms, line 128: Is there any potential that this causes issues with or prevents reloads for same-page anchors? I'm not clear if that situation is possible here.

manuelpuyol: I can't imagine a case where a same-page anchor causes a full page reload

PR #606, opened 2022-06-20
Support development ChromeDriver version overrides

seanpdoyle:
Refines the changes introduced to the CI configuration made in
[b1cbe12][].

Recent Chrome upgrades are causing errors like the following while
executing the test suite locally:

```
SessionNotCreatedException: [POST http://localhost:4444/wd/hub/session / {"desiredCapabilities":{"name":"intern","idle-timeout":60,"browserName":"chrome","goog:chromeOptions":{"args":["headless","disable-gpu","no-sandbox"]},"browser":"chrome"}}] session not created: This version of ChromeDriver only supports Chrome version 99
Current browser version is 102.0.5005.115 with binary path
```

The **only supports Chrome version 99** portion is due to `digdug`'s
locked-support for Chrome version 99 declared in its [webdrivers.json][]
file.

This commit checks for the presence of the `CHROMEWEBDRIVER` environment
variable, and overrides the [intern.json](./intern.json) configuration
to incorporate that into its `tunnelOptions.drivers` value.

[webdrivers.json]: https://github.com/theintern/digdug/blob/806dcf29c2265d3cb1a26a09ef5b43e93fc2a739/src/webdrivers.json#L8-L11
[b1cbe12]: https://github.com/hotwired/turbo/pull/551/commits/b1cbe123895259408d35b1c918ab2e9a51ba6683

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,10 +24,7 @@ jobs:
       run: |
         CHROMEVER="$(chromedriver --version | cut -d' ' -f2)"
         echo "Actions ChromeDriver is $CHROMEVER"
-        CONTENTS="$(jq '.tunnelOptions.drivers[0].name = "chrome"' < intern.json)"
-        CONTENTS="$(echo ${CONTENTS} | jq --arg chromever "$CHROMEVER" '.tunnelOptions.drivers[0].version = $chromever')"
-        echo "${CONTENTS}" > intern.json
-        cat intern.json
+        echo "CHROMEVER=${CHROMEVER}" >> $GITHUB_ENV
 
     - name: Lint
       run: yarn lint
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -22,12 +22,8 @@ jobs:
 
     - name: Set Chrome Version
       run: |
-        CHROMEVER="$(chromedriver --version | cut -d' ' -f2)"
+        export CHROMEVER="$(chromedriver --version | cut -d' ' -f2)"
```
seanpdoyle, line 25: @keithamus and @koddsson this incorporates the CI configuration you two introduced in https://github.com/hotwired/turbo/pull/551/commits/b1cbe123895259408d35b1c918ab2e9a51ba6683 to support development-environment configurations.

It's still an incomplete developer experience and requires an awareness of your local chrome driver version.

On top of that, I'm not sure how to programmatically set a GitHub actions environment variable in this way, `export` doesn't seem to cut it.

seanpdoyle: > On top of that, I'm not sure how to programmatically set a GitHub actions environment variable in this way, export doesn't seem to cut it.

Evidently, piping an assignment into `$GITHUB_ENV` does the trick!

keithamus (approved):
LGTM

PR #607, opened 2022-06-21
Allow frames to scroll smoothly into view

npezza93:
The default behavior can be pretty jarring depending on your use case but this allows people to specify the scroll behavior when autoscrolling.


dhh:
Great! Could you add a doc PR as well to hotwired/turbo-site we can merge on 7.2.0 release?

PR #609, opened 2022-06-23
Drive Browser tests with `playwright`

seanpdoyle:
Replaces the [intern][]-powered functional test suite with one powered
by [playwright][].

The majority of the changes made aim to preserve the structure and
method names of the original `intern` suite. There are some
out-of-the-box Playwright features that could replace many of our
bespoke helpers, but that work should be done after the migration.

As a result, the majority of the changes involve a combination of:

* replacing `async "test ..."() { ... }` with `test("test ...", async
  ({ page }) => { ... }`
* replacing calls in the style of `this.methodName` with calls in the
  style of `methodName(page)`

In some cases, exceptions were made. For example, most calls to
`this.clickSelector` were replaced with calls to [page.click][].

The unit test suite
---

Playwright's focus on a browser-powered test harness make it an odd fit
for our small suite of "unit" tests. Unlike their browser-test
counterparts, the unit tests do not have a history of flakiness.

This changeset defers migrating those tests off of `intern` for a later
batch of work.

[intern]: https://theintern.io
[playwright]: https://playwright.dev
[page.click]: https://playwright.dev/docs/api/class-page#page-click


seanpdoyle:
cc: @keithamus

src/tests/helpers/page.ts
```diff
@@ -0,0 +1,241 @@
...
+    return false
+  }
+}
+
+export function nextBeat() {
+  return sleep(100)
+}
+
+export function nextBody(_page: Page, timeout = 500) {
+  return sleep(timeout)
```
seanpdoyle, line 66: @keithamus this is one of the bigger hurdles for the migration. The old version tracked the `<body>` element with instance state. Since playwright test blocks have no context, I'm not sure of the best way to synchronize on a change.

PR #611, opened 2022-06-23
Add original click event to 'turbo:click' details

manuelpuyol:
Right now, if you `preventDefault` a `turbo:click`, it will still fallback to the original click, causing a full-page reload. Sending the original click event will allow users to completely cancel a click if they need/want to.


src/core/session.ts
```diff
@@ -296,8 +296,8 @@ export class Session
 
   // Application events
 
-  applicationAllowsFollowingLinkToLocation(link: Element, location: URL) {
-    const event = this.notifyApplicationAfterClickingLinkToLocation(link, location)
+  applicationAllowsFollowingLinkToLocation(link: Element, location: URL, ev: MouseEvent) {
+    const event = this.notifyApplicationAfterClickingLinkToLocation(link, location, ev)
     return !event.defaultPrevented
```
seanpdoyle, line 301: As an alternative to dispatching the event with a reference to the original, would explicitly preventing the default if `event.defaultPrevented` serve the same purpose without expanding the surface area of the `CustomEvent.detail`?

```typescript
applicationAllowsFollowingLinkToLocation(link: Element, location: URL, mouseEvent: MouseEvent) {
  const event = this.notifyApplicationAfterClickingLinkToLocation(link, location)

  if (event.defaultPrevented) {
    mouseEvent.preventDefault()
    return false
  } else {
    return true
  }
}
```

seanpdoyle:
Is this related to https://github.com/hotwired/turbo/issues/610?

If it is, could we add a test case that exercises a call to `event.preventDefault()` preventing a frame navigation?

src/core/session.ts
manuelpuyol: the use cases would be:

1. I want to prevent Turbo from navigating and let the browser do it -> prevent `turbo:click`
2. I want to prevent all navigations -> prevent both events

manuelpuyol:
> Is this related to #610?
> 
> If it is, could we add a test case that exercises a call to `event.preventDefault()` preventing a frame navigation?

No, this is just something I've been thinking about for some time 😅

src/core/session.ts
seanpdoyle: With that in mind, could we make Frame navigation internally aware of the cancellation?

manuelpuyol: from my understanding, this is where we trigger the Frame navigation:

https://github.com/hotwired/turbo/blob/98cdc4037788f078a741c7fb50412db0101bcb97/src/core/frames/link_interceptor.ts#L37-L43

this is listening to the `turbo:click` event, so it _should_ be possible for it to know if the event is `defaultPrevented`. The problem is that the interceptor runs **before** any `addEventListener` I add, so the it always evaluates the event first, and it is never prevented.

maybe we could update 

https://github.com/hotwired/turbo/blob/98cdc4037788f078a741c7fb50412db0101bcb97/src/core/session.ts#L153-L159

to dispatch another event (if the click isn't prevented) that the LinkInterceptor listens to

manuelpuyol: @seanpdoyle would adding another _internal_ event make sense here?

seanpdoyle (approved):

dhh:
Let's get a doc PR going to explain this as well for the 7.2.0 release 👍

PR #612, opened 2022-06-27
Allow Turbo Streams with GET via `data-turbo-stream`

kevinmcconnell:
Turbo Streams are normally supported only for [non-GET requests][0]. However there are cases where Turbo Streams responses to GET requests are useful.

This commit adds the ability to use Turbo Streams with specific GET requests by setting `data-turbo-stream="true"` on a form or link.

[0]: https://github.com/hotwired/turbo/pull/52

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -1,7 +1,7 @@
 import { FetchRequest, FetchMethod, fetchMethodFromString, FetchRequestHeaders } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { expandURL } from "../url"
-import { dispatch } from "../../util"
+import { attributeTrue, dispatch } from "../../util"
 import { StreamMessage } from "../streams/stream_message"
 
 export interface FormSubmissionDelegate {
@@ -153,6 +153,9 @@ export class FormSubmission {
       if (token) {
         headers["X-CSRF-Token"] = token
       }
+    }
+
+    if (this.requestAcceptsTurboStreamResponse(request)) {
       headers["Accept"] = [StreamMessage.contentType, headers["Accept"]].join(", ")
     }
   }
@@ -204,8 +207,14 @@ export class FormSubmission {
     this.delegate.formSubmissionFinished(this)
   }
 
+  // Private
+
   requestMustRedirect(request: FetchRequest) {
     return !request.isIdempotent && this.mustRedirect
+  }
+
+  requestAcceptsTurboStreamResponse(request: FetchRequest) {
+    return !request.isIdempotent || attributeTrue(this.formElement, "data-turbo-stream")
   }
 }
 
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -7,7 +7,7 @@ import {
 import { FetchMethod, FetchRequest, FetchRequestDelegate, FetchRequestHeaders } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { AppearanceObserver, AppearanceObserverDelegate } from "../../observers/appearance_observer"
-import { clearBusyState, getAttribute, parseHTMLDocument, markAsBusy } from "../../util"
+import { clearBusyState, getAttribute, parseHTMLDocument, markAsBusy, attributeTrue } from "../../util"
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { ViewDelegate } from "../view"
@@ -149,7 +149,7 @@ export class FrameController
   // Link interceptor delegate
 
   shouldInterceptLinkClick(element: Element, _url: string) {
-    if (element.hasAttribute("data-turbo-method")) {
+    if (element.hasAttribute("data-turbo-method") || attributeTrue(element, "data-turbo-stream")) {
       return false
     } else {
       return this.shouldInterceptNavigation(element)
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -12,7 +12,7 @@ import { ScrollObserver } from "../observers/scroll_observer"
 import { StreamMessage } from "./streams/stream_message"
 import { StreamObserver } from "../observers/stream_observer"
 import { Action, Position, StreamSource, isAction } from "./types"
-import { clearBusyState, dispatch, markAsBusy } from "../util"
+import { attributeTrue, clearBusyState, dispatch, markAsBusy } from "../util"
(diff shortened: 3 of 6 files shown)


dhh:
Is the annotation required? What consequences would we have by simply always allowing streams in response to GET?

kevinmcconnell:
The annotation is only required if we want to keep the behaviour introduced in #52. If we're happy to revisit that decision, then always allowing streams in response to GET is a simpler and (imho) preferred option.

Reasons being:
- Form submissions and Turbo Stream responses are orthogonal concepts. Turbo Streams provide a fantastic way to send targeted DOM changes from the server, avoiding the need for custom JS in many cases. But form submissions aren't the only time you might want to target the DOM in that way. If Turbo processes stream responses in all cases, then folks are free to send whatever response suits their use case, without having to alter their requests to do it.
- Not being able to stream back a response unless you're submitting a form is surprising behaviour. I expected to be able to GET a stream, and anecdotally from chatting with others, that seems to be a common expectation. (e.g. "it works with Turbo Frames, so why not with Turbo Streams?"). Using annotations solves the functionality side of this, but it doesn't stop the default behaviour from being surprising.
- The example Rails pattern in #52 is definitely useful, but I'd argue that there are clearer ways to do it than switching on the mime type of the request. In the specific example given, moving `new.turbo_stream.erb` to `create.turbo_stream.erb` would be clearer.

The only downside I can think of to always allowing streams over GET is that it would be a breaking change for anyone relying on that aspect of the current behaviour. I think it's worth it for the simplicity, but others might feel differently?

kevinmcconnell:
Added an alternative to this in #621, which always allows streams in response to `GET` without needing the annotation.

dhh:
Need a PR to document this as well.

kevinmcconnell:
Documentation is in https://github.com/hotwired/turbo-site/pull/103

seanabrahams:
@kevinmcconnell thank you for this. It was merged at an opportune time as I'm looking to implement a "Load More"-style of pagination using Turbo Streams via GET request.

On testing though I'm experiencing Turbo not including query params in `data-turbo-stream` GET requests generated from a link:
```html
<a href="/posts?page=2" data-turbo-stream="true">
  Load More
</a>
```
Will end up sending only `/posts` to the backend with the `page=2` param discarded.

I'll take a look to see if this is an easy fix but thought I'd pass it along in case the fix obvious to you. Cheers.

kevinmcconnell:
@seanabrahams oh good catch! I'm guessing this is because links with `data-turbo-stream` are being submitted as forms, and forms with `method="GET"` [don't submit the query string part of their action](https://www.w3.org/TR/2011/WD-html5-20110525/association-of-controls-and-forms.html#submit-mutate-action). I expect the same thing would happen if you had a link with `data-method="GET"`. We probably need to prevent the link from being converted into a form in this case.

I'll try to look into this today if I can. Thanks for letting me know about it.

BakiVernes:
## Some issues I encountered

Not sure if this is the right place to post this but here we go. Decided to update my app and remove UJS altogether. Fetched the latest beta.1 release specifically to try this feature. Couple of things I noticed:

- Works properly only if wrapped inside a turbo-frame element

### Working
image
image

-----------------------------------------
### Not working
image
image

-----------------------------------------

- Fires request twice. Once with turbo_stream request format and once with html. And since there are no .html responds in my backend throws an error both on client and backend. Video below:


https://user-images.githubusercontent.com/31761693/181233430-07ff5b3d-bb9e-4750-92b5-3bfa75d05e28.mov

kevinmcconnell:
Hi @BakiVernes. The requests firing twice is, I think, a separate issue that's being addressed in #645.

I've not been able to reproduce your first issue though. I'm using this outside of Turbo Frames here and it's working as expected for me. Could you share a code example that reproduces it?

thibaudgg:
I tried this feature with the `7.2.0-beta.2` release and added the `data-turbo-stream="true"` attribute to my link. I got the proper `TURBO_STREAM` response as expected but it seems like the `aria-busy="true"` attribute in the `html` tag isn't removed like when clicking on any other Turbo enabled link. Is it expected? For us it's problematics as we are showing a spinner when `aria-busy` is `true`, which is now never going away.

kevinmcconnell:
@thibaudgg I'll look into that (hopefully today) and will get back to you. Thanks for letting me know!

kevinmcconnell:
@thibaudgg I've opened #681 to fix this. Hopefully that will fix the issue for you, but let me know if you have any questions or concerns.

Thanks again for reporting!

thibaudgg:
@kevinmcconnell thanks a lot for the quick fix, I will give it a try next week. 👍🏻

bilogic:
What should the default URL behavior be when using `GET` with `data-turbo-stream`?

Currently, it stays unchanged, I was expecting it to `advance`. 

My use case is that I tag my navigation `<a href=...` with `data-turbo-stream=true` so that the server can return a `turbo-stream` to `update` another part of the HTML. URL needs to update as part of navigation.

bilogic:
@luctus no I did not

kevinmcconnell:
> What should the default URL behavior be when using GET with data-turbo-stream?

@bilogic the behaviour is to load the new content, but not trigger a visit, which matches what you're seeing. It's consistent with Turbo Stream responses over other HTTP methods (like `POST`ing a form). I think it's a good default, since a Turbo Stream response could be making any sort of change to the page, so we don't really know whether it represents a navigation or not. That also matches the default behaviour of navigations inside a `<turbo_frame>`, which are conceptually quite similar.

Perhaps we should add the ability to override this with `data-turbo-action`, similar to how that [can be done with frames](https://turbo.hotwired.dev/handbook/frames#promoting-a-frame-navigation-to-a-page-visit). That seems like it would be useful?

luctus:
Hello @kevinmcconnell,

Let's say we have a to-do list. In the index view we will have all the tasks listed, you click on one (with `data-turbo-stream=true`) and it triggers a slideover with the details of that task, so you can close it quickly and continue scrolling your list.

Once the slideover is open, I want to copy the url and share it with my friend, but the url will remain being the one for the index view.

Do you think this is not the right use case for this new feature? I may be super lost, tbh.

Thanks a lot for your work, anyway!

bilogic:
https://discuss.hotwired.dev/t/simpler-way-to-think-about-drive-vs-frame-vs-stream/4406

Because hotwire is quite "new", I think this is a good way to think about our use cases.

@kevinmcconnell in your case, it seems to be *click here*, *don't change here* + *change elsewhere* (details), then copy and paste (which is not part of hotwire)

pySilver:
@luctus in that case you should manage the URL state yourself. Here is how: https://github.com/hotwired/turbo/issues/792

Issue #613, opened 2022-06-28
Turbo should wait for new CSS to be loaded before rendering new pages/frames


manuelpuyol:
[CSS is a render blocking resource](https://web.dev/critical-rendering-path-render-blocking-css/), so normally the browser will wait for it to be loaded before continuing rendering the page.

When navigating using either Turbo Drive, Turbo will add new Stylesheets coming from the response to the head, but won't wait until they are loaded to render the new body, causing jitter in the page in case the new CSS changes the layout.''

In this (extreme) example, I'm throttling the request to a CSS file by 1s to showcase the bug:

https://user-images.githubusercontent.com/11280312/176277479-75aa0f5a-277f-4344-ba5d-58aa549bc4ad.mov

 Using the same throttle, but with `data-turbo="false"`:

https://user-images.githubusercontent.com/11280312/176277743-e82f81f6-1a9e-4f40-b1ae-de9e0078c25d.mov

seanpdoyle:
Thank you for opening this issue!

I wonder what our options are for synchronizing their load times.

Maybe we could attach [`load` and `error` listeners to `<link>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#stylesheet_load_events) when a `<turbo-frame>` response contains a new element, transform those into `Promise` instances, then incorporate them into the Frame resolution.

```typescript
function waitForLoad(element: HTMLLinkElement): Promise<void> {
  return new Promise((resolve, reject) => {
    const onLoad = () => {
      element.removeEventListener("error", onError)
      resolve()
    }
    const onError = () => {
      element.removeEventListener("load", onLoad)
      reject()
    }

    element.addEventListener("load", onLoad, { once: true })
    element.addEventListener("error", onError, { once: true })
  })
}

const linkElementsLoadedInFrameResponse = ...

await Promise.all(linkElementsLoadedInFrameResponse.map(waitForLoad)
```

manuelpuyol:
I have https://github.com/hotwired/turbo/pull/614 for visits, but I couldn't find how Frames deal with new CSS. I see https://github.com/hotwired/turbo/blob/98cdc4037788f078a741c7fb50412db0101bcb97/src/core/frames/frame_renderer.ts#L48-L53

but nothing for stylesheets

jody90:
I encountered the same problem. Is there already a solution for the behavior in frames?

PR #614, opened 2022-06-28
Wait for CSS to be loaded before replacing the Body

manuelpuyol:
Related to https://github.com/hotwired/turbo/issues/613

As suggested in https://github.com/hotwired/turbo/issues/613#issuecomment-1169208644, we can listen for `load` and `error` events in stylesheets to make sure they are loaded before continuing a render.

Here I'm updating how `PageRenderer` adds new stylesheets to the document head, adding listeners to those events and waiting them to fire before replacing the body.


src/util.ts
```diff
@@ -92,3 +92,19 @@ export function clearBusyState(...elements: Element[]) {
...
+
+export function waitForLoad(element: HTMLLinkElement): Promise<void> {
+  return new Promise((resolve, reject) => {
+    const onLoad = () => {
+      element.removeEventListener("error", onError)
+      resolve()
+    }
+    const onError = () => {
+      element.removeEventListener("load", onLoad)
+      reject()
```
seanpdoyle, line 104: I had suggested `reject()` here to handle an `error`. After thinking about it a little longer, I wonder if `resolve()` is a better response. I suspect that `reject()` here would halt the other Promises, which feels like an overreaction to an external resource failing to load.

If our intent here is only to synchronize, I think `resolve()` would be less brittle:

```typescript
return new Promise((resolve) => {
  const onComplete = () => {
    element.removeEventListener("load", onComplete)
    element.removeEventListener("error", onComplete)
    resolve()
  }
  element.addEventListener("load", onComplete)
  element.addEventListener("error", onComplete)
})
```

dhh:
What would the failure state be if those assets fail to load (bad CDN or whatever)? How long would we wait for them to load? Do we need a timeout?

src/util.ts
```diff
@@ -92,3 +92,16 @@ export function clearBusyState(...elements: Element[]) {
...
 }
+
+export function waitForLoad(element: HTMLLinkElement): Promise<void> {
+  return new Promise((resolve) => {
+    const onComplete = () => {
+      element.removeEventListener("error", onComplete)
+      element.removeEventListener("load", onComplete)
+      resolve()
+    }
+
```
seanpdoyle, line 103: Could `setTimeout(resolve, resourceLoadingTimeoutInMilliseconds)` serve as a guard against remote resources that take forever to load?

manuelpuyol:
> What would the failure state be if those assets fail to load (bad CDN or whatever)? How long would we wait for them to load? Do we need a timeout?

good shout, I've added a timeout as @seanpdoyle suggested

src/core/drive/page_renderer.ts
```diff
@@ -50,8 +51,8 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
     return this.newSnapshot.element
   }
 
-  mergeHead() {
-    this.copyNewHeadStylesheetElements()
+  async mergeHead() {
+    await this.copyNewHeadStylesheetElements()
     this.copyNewHeadScriptElements()
     this.removeCurrentHeadProvisionalElements()
     this.copyNewHeadProvisionalElements()
```
rik, line 58: This will wait for all stylesheets to be fetched before starting to fetch new scripts or fetch `<link rel=preload>`. This is not how browsers behave so I would suggest awaiting later, maybe this way:
```suggestion
    const newStylesheetElements = this.copyNewHeadStylesheetElements()
    this.copyNewHeadScriptElements()
    this.removeCurrentHeadProvisionalElements()
    this.copyNewHeadProvisionalElements()
    await newStylesheetElements
```

manuelpuyol:
> Do we have adequate test coverage for this?

added tests in https://github.com/hotwired/turbo/pull/614/commits/3171b0ac1c21e83341eaa355c9c7f47526438827

manuelpuyol:
@dhh this should be ready for some 👀

PR #615, opened 2022-06-29
Allow isHTML to be overridden

bteller:
This exposes isHTML as a session variable to allow overriding the function used in the same manner as ```Turbo.session.drive = false``` can be used to disable turbo functionality by default and placing it in opt-in mode. To modify the default isHTML check one can simply do the following. 

```
Turbo.session.isHTML = (url) => { return true; }
```

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -52,6 +52,7 @@ export class Session
   progressBarDelay = 500
   started = false
   formMode = "on"
+  isHTML:any = null
 
   start() {
     if (!this.started) {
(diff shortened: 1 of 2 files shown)


bteller:
I attempted to run the tests locally before submitting this, was able to resolve the chromedriver error by installing the chromedriver cask on my Mac and setting an environment variable for CHROMEVER that pointed at it, and there were a number of failures. It appears that might be an issue on the main branch as well, but do let me know if this PR is worth consideration and I will dig in and attempt to sort out my local environment test issues if not global.

dhh:
We're exploring this on https://github.com/hotwired/turbo/pull/519.

Issue #617, opened 2022-06-30
Turbo should update the URL before frames render


manuelpuyol:
When promoting a frame navigation with `data-turbo-action`, Turbo will render the Frame and then update the URL:

https://user-images.githubusercontent.com/11280312/176727319-e34c98d2-d3e5-449b-9742-3464c52dcbee.mov

Promoted frames should act like visits, where we update the URL first:

https://user-images.githubusercontent.com/11280312/176727237-cae41974-412a-4f8c-a241-eb48e8a96301.mov

PR #618, opened 2022-06-30
Update the history before rendering promoted frames

manuelpuyol:
Closes https://github.com/hotwired/turbo/issues/617

I'm extracting the history method selection to `getHistoryMethodForAction` in `util and using it in both Frames and Visits.

In case the frame needs to change the history, it update it **before** rendering the frame's content, and will call the proposed visit with `updateHistory = false`, which will avoid a second history update.


src/tests/fixtures/tabs.html
```diff
@@ -0,0 +1,22 @@
...
+    <script src="/src/tests/fixtures/test.js"></script>
+  </head>
+  <body>
+    <h1>Tabs</h1>
+
+    <turbo-frame id="tab-content" data-turbo-action="advance">
+      <div>
+        <a id="drive_enabled" href="/src/tests/fixtures/tabs.html">Tab 1</a>
+        <a id="drive_enabled" href="/src/tests/fixtures/tabs/two.html">Tab 2</a>
+        <a id="drive_enabled" href="/src/tests/fixtures/tabs/three.html">Tab 3</a>
```
seanpdoyle, line 16: Is it necessary to declare these anchors with an `[id]` attribute? If it is, could we make sure their values don't collide with other elements on the page?

Also, are there plans to add tests to cover the new behavior and to guard against regressions?

src/core/frames/frame_controller.ts
```diff
@@ -288,16 +301,25 @@ export class FrameController
...
           })
         }
       }
     }
   }
 
+  changeHistory() {
+    if (this.action && this.frame) {
+      const method = getHistoryMethodForAction(this.action)
+      session.history.update(method, expandURL(this.frame.src as string), this.restorationIdentifier)
```
seanpdoyle, line 319: What makes the `as string` casting necessary?

Is it possible that `this.frame.src` is `null`? Could `|| ""` cover that case without circumventing the compiler?

src/core/frames/frame_controller.ts
```diff
@@ -277,9 +289,10 @@ export class FrameController
   }
 
   private proposeVisitIfNavigatedWithAction(frame: FrameElement, element: Element, submitter?: HTMLElement) {
-    const action = getAttribute("data-turbo-action", submitter, element, frame)
+    this.action = getAttribute("data-turbo-action", submitter, element, frame) as Action
```
seanpdoyle, line 292: Would a check with [isAction](https://github.com/hotwired/turbo/blob/main/src/core/types.ts#L3-L5) here help assure the compiler that the returned value is of type `Action`?

seanpdoyle: https://github.com/hotwired/turbo/pull/430 extracts a [`getVisitAction` utility function](https://github.com/hotwired/turbo/pull/430/files#diff-78d8451f964182fd51330bac500ba6e71234f81aad9e8a57e682e723d0f517b3R422-R427). It might be useful here:

```typescript
export function getVisitAction(...elements: (Element|undefined)[]): Action | null {
  const action = getAttribute("data-turbo-action", ...elements)

  return isAction(action) ? action : null
}
```

seanpdoyle (commented):
Thank you for opening this issue _and_ following it up with a PR!

src/tests/fixtures/tabs.html
manuelpuyol: yeah, I wanted to validate the idea and implementation and will follow up with some tests

manuelpuyol: @seanpdoyle I added some tests in e99e69e...16b3bdd, but it's really hard to know the state of tests already in place since there's a lot of flakiness. I'm different seeing test pass and fail each run

manuelpuyol:
@seanpdoyle anything you'd like me to do to move this forward?

src/tests/fixtures/tabs/three.html
```diff
@@ -0,0 +1,9 @@
+<turbo-frame id="tab-frame" data-turbo-action="advance">
+  <div>
+    <a id="drive_enabled" href="/src/tests/fixtures/tabs.html">Tab 1</a>
+    <a id="drive_enabled" href="/src/tests/fixtures/tabs/two.html">Tab 2</a>
+    <a id="drive_enabled" href="/src/tests/fixtures/tabs/three.html">Tab 3</a>
```
seanpdoyle, line 5: Could we make sure that each of these elements's `[id]` is unique?

src/globals.d.ts
```diff
@@ -11,3 +16,58 @@ interface Window {
   Turbo: typeof import("./core/index")
   SubmitEvent: typeof Event
 }
+
+interface CustomEventMap {
```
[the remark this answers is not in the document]
manuelpuyol: what do you think is the best path forward? I can leave only the BeforeFrameRender as a type here, or I could do something similar to #452, exporting the type in `session.ts` and typing the `dispatch` call.
or we can roll this back and continue with the test casting the event manually

manuelpuyol:
@seanpdoyle I'm rolling back the type changes in favor of #452. I'm bringing the `as CustomEvent` back while that's not merged. I don't think the casting is harmful especially because it's only in a test

dhh:
Can you rebase/update the tests?

manuelpuyol:
done!

PR #622, opened 2022-07-06
Introduce `turbo:before-permanent-element-render`

seanpdoyle:
While transposing a [permanent element][] from page to page, dispatch a
`turbo:before-permanent-element-render` event on the document with a
`(currentPermanentElement: Element, newPermanentElement: Element) =>
void` available as the event's `detail.render` property.

The `turbo:before-permanent-element-render` is named to follow the
pattern established by its predecessors (`turbo:before-render`,
`turbo:before-frame-render`, `turbo:before-stream-render`), and its
`CustomEvent.detail.render` property name mirrors the
`turbo:before-render` event's `CustomEvent.detail.render` property.

This opportunity to modify a permanent element might be useful if it's
possible for the preserved element to become partially stale. For
example, consider an invalid `<form>` submission with an `<input
type="file">`.

Since the server's response cannot fully encode the submitted file into
the response's element, it might be a useful to mark it with
`[data-turbo-permanent]` to preserve whatever client-side state precedes
the submission:

```html
<label for="profile_image">Profile image</label>
<input id="profile_image" type="file" data-turbo-permanent>
```

Suppose the uploaded file is too large, or doesn't match the server's
expectations for file type. The server's response might contain
a fragment like:

```html
<label for="profile_image">Profile image</label>
<input id="profile_image" type="file" data-turbo-permanent
       class="invalid" aria-describedby="profile_image_error">
<p id="profile_image_error">Profile image is too large (5GB). Try a smaller file (<5KB)</p>
```

Prior to this change, the fact that the `<input>` is marked with
`[data-turbo-permanent]` would ignore the `[class]` and
`[aria-describedby]` attributes from the response, and the `<input>`
would remain unchanged (with the attached file still in-memory).

With a new `turbo:before-permanent-element-render` available, there's an
opportunity to do application-specific merging:

```js
addEventListener("turbo:before-permanent-element-render", ({ target, { detail } }) => {
    currentElement.setAttribute("class", newElement.getAttribute("class"))
    currentElement.setAttribute("aria-describedby", newElement.getAttribute("aria-describedby"))
  }
})
```

[permanent element]: https://turbo.hotwired.dev/handbook/building#persisting-elements-across-page-loads

### DIFF
--- a/src/core/bardo.js
+++ b/src/core/bardo.js
@@ -1,3 +1,5 @@
+import { dispatch } from "../util"
+
 export class Bardo {
   static async preservingPermanentElements(delegate, permanentElementMap, callback) {
     const bardo = new this(delegate, permanentElementMap)
@@ -21,8 +23,12 @@ export class Bardo {
 
   leave() {
     for (const id in this.permanentElementMap) {
-      const [currentPermanentElement] = this.permanentElementMap[id]
-      this.replaceCurrentPermanentElementWithClone(currentPermanentElement)
+      const [currentPermanentElement, newPermanentElement] = this.permanentElementMap[id]
+      const event = dispatch("turbo:before-permanent-element-render", {
+        detail: { render: this.replaceCurrentPermanentElementWithClone }
+      })
+
+      event.detail.render(currentPermanentElement, newPermanentElement)
       this.replacePlaceholderWithPermanentElement(currentPermanentElement)
       this.delegate.leavingBardo(currentPermanentElement)
     }
@@ -33,7 +39,7 @@ export class Bardo {
     permanentElement.replaceWith(placeholder)
   }
 
-  replaceCurrentPermanentElementWithClone(permanentElement) {
+  replaceCurrentPermanentElementWithClone = (permanentElement, _newElement) => {
     const clone = permanentElement.cloneNode(true)
     permanentElement.replaceWith(clone)
   }
--- a/src/tests/fixtures/permanent_element.html
+++ b/src/tests/fixtures/permanent_element.html
@@ -10,7 +10,7 @@
     <section>
       <h1>Permanent element</h1>
     </section>
-    <div id="permanent" data-turbo-permanent>Permanent element</div>
+    <div id="permanent" class="loaded" data-turbo-permanent>Permanent element</div>
 
     <turbo-frame id="frame">
       <div id="permanent-in-frame" data-turbo-permanent>Permanent element</div>
--- a/src/tests/fixtures/rendering.html
+++ b/src/tests/fixtures/rendering.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html data-skip-event-details="turbo:before-permanent-element-render">
   <head>
     <meta charset="utf-8">
     <title>Turbo</title>
--- a/src/tests/fixtures/test.js
+++ b/src/tests/fixtures/test.js
@@ -70,6 +70,7 @@
   "turbo:before-stream-render",
   "turbo:before-cache",
   "turbo:before-render",
+  "turbo:before-permanent-element-render",
   "turbo:before-visit",
   "turbo:load",
   "turbo:render",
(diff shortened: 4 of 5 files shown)


dhh:
Wouldn't you be able to get the same done using `turbo:before-render` and digging through the existing page body and the new page body? What's the additional utility of having this narrowed down?

seanpdoyle:
I considered a permanent-specific event to separate the idea from the details of the rendering (was it a full Drive visit? a Frame navigation? a Stream?) so that the code could be more general purpose.

Currently, the `newBody` is only available to a `turbo:before-render` event. Frame navigations only dispatch `turbo:frame-render` events, and those don't provide access to the new fragment of HTML.

I'm also not sure if Turbo Stream rendering currently accounts for permanent elements correctly. If they do, that would be another point in time and event (`turbo:before-stream-render`) where callers would need access to the new element ahead of time.

If a new event specific to permanent elements isn't warranted, then a `turbo:before-frame-render` with access to the `newFrame` would be an appropriate replacement, along with an extension to the `turbo:before-stream-render` event to provide a handle to the `<turbo-stream>` element's `<template>` as well as the elements referenced by the `[target]` and `[targets]` attributes.

seanpdoyle:
> I'm also not sure if Turbo Stream rendering currently accounts for permanent elements correctly. If they do, that would be another point in time and event (turbo:before-stream-render) where callers would need access to the new element ahead of time.

I've just tried this out, and it appears that Turbo Stream operations ignore `[data-turbo-permanent]`.

You can [experiment with the code on JSFiddle](
https://jsfiddle.net/yf8e6zm0/2/):

```html
<html>                                        
  <head>                                  
    <script type="importmap">                                                                                      
      {
        "imports": {
          "@hotwired/turbo": "https://cdn.skypack.dev/@hotwired/turbo"
        }
      }
    </script>
    <script type="module">
      import "@hotwired/turbo"
      
      addEventListener("click", ({ target }) => {
        if (target.matches("button")) {
          const { content } = target.querySelector("template")
          target.append(content.cloneNode(true))
        }
      })
    </script>
    <script type="module" src="https://unpkg.com/es-module-shims@0.10.1/dist/es-module-shims.js"></script>
  </head>
  <body>
    <p id="permanent" data-turbo-permanent>This should not change when you click the button</p>
      
    <button type="button">
      Click me, and hopefully nothing changes

      <template>
        <turbo-stream action="replace" target="permanent">
          <template><p id="permanent" data-turbo-permanent>Uh oh, this changed when you clicked the button!</p></template>
        </turbo-stream>
      </template>
    </button>
  </body>
</html>
```

My gut tells me that's surprising behavior, but maybe it isn't.

Turbo Streams aside, the gap between `turbo:before-render` and `turbo:frame-render` still poses a problem.

kstratis:
Does this mean that we get the chance to rid the menu "flashing" when navigating with turbo drive?

seanpdoyle:
> Does this mean that we get the chance to rid the menu "flashing" when navigating with turbo drive?

Maybe! Could you share some more details of what you're experiencing?

kstratis:
Sure.

The "flashing" effect I'm talking about usually happens when clicking around a menu or top nav.
This thing often comes in the form of a partial fixed somewhere in the layout. 

When a user clicks any of the menu options, Turbo Drive replaces the entire body. The side-effect is that since the "nav" partial is part of the layout which is part of the body, it gets replaced too. This is totally expected cause the "active" menu link has to somehow be updated. However is it suboptimal cause the user sees a "flash" as a result of the "rerender". 

At the moment one could set the entire menu as `permanent` but then the "active" link would never update. 
Looking at this PR I was wondering if I could somehow use the `permanent` attribute along with the new hook `turbo:before-permanent-element-render` and manually update the "active" menu link in javascript. 

I'm also attaching a couple of videos below further demonstrating the issue:




https://user-images.githubusercontent.com/1859806/183642480-21b574b5-64b5-42bd-a09c-62a4bb70f130.mp4

https://user-images.githubusercontent.com/1859806/183642509-dd2c802e-8193-4fae-8183-8503ebf47991.mp4

kstratis:
cc @seanpdoyle

Issue #625, opened 2022-07-13
Frame with `[loading="lazy"]` cannot load when styled with `display: contents`


seanpdoyle:
When a `<turbo-frame>` element is styled as [`display: contents`](https://developer.mozilla.org/en-US/docs/Web/CSS/display#box), it derives its box model from its descendants, and browsers treat it as if the element itself is not present in the document.

Unfortunately, [since it's "omitted" from the document](https://stackoverflow.com/q/72741599), the underlying [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) instance that powers the [`[loading="lazy"]`](https://turbo.hotwired.dev/reference/frames#html-attributes) behavior never detects when the `<turbo-frame>` contents enter the viewport.

Would it be possible to detect when a `FrameElement` instance has [`style.display == "contents"`]https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/style), then support special treatment of the element's `IntersectionObserver`?

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/626 to reproduce the issue with a failing test.

dhh:
So this is a frame that's hidden, and therefore does not autoload when scrolled into view? That seems like the behavior I'd expect? If you want to overwrite this, and force eager loading, don't we already have control attribute for that? (If not, sounds like that should be the fix, rather than try to work around the box model)?

seanpdoyle:
A Frame marked with `[loading="lazy"]` is not necessarily marked with `[hidden]` or obfuscated by a `<details>` or `<dialog>` that is not `[open]`.

For example, imagine a `<turbo-frame id="comments" loading="lazy" src="...">` that's rendered below an Article that pushes it below the fold.

If that element had `display: contents`, it would never intersect with the viewport, which means it never loads.

seanpdoyle:
It's possible for applications to work around the quirky behavior. They could nest the `<turbo-frame>` within another element that manages its own `IntersectionObserver` instance, then toggle `[loading="eager"]`. 

Unfortunately, I think this issue is more about surprise than technical limitations.

dhh:
It just seems quite invasive for us to try to work around the box model as the browsers intended it. I think starting with documentation is probably the better way to go?

seanpdoyle:
Would `console.warn` output in combination with a documentation change be enough of a balance between hands-off and informative?

dhh:
[...]

PR #626, opened 2022-07-13
Lazy load Frame styled with `display: contents`

seanpdoyle:
Reproduces [hotwired/turbo#625][]

When a `<turbo-frame>` element is styled as [`display:
contents`](https://developer.mozilla.org/en-US/docs/Web/CSS/display#box),
it derives its box model from its descendants, and browsers treat it as
if the element itself is not present in the document.

Unfortunately, [since it's "omitted" from the
document](https://stackoverflow.com/q/72741599), the underlying
[IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)
instance that powers the
[`[loading="lazy"]`](https://turbo.hotwired.dev/reference/frames#html-attributes)
behavior never detects when the `<turbo-frame>` contents enter the
viewport.

[hotwired/turbo#625]: https://github.com/hotwired/turbo/issues/625

### DIFF
--- a/src/tests/fixtures/loading.html
+++ b/src/tests/fixtures/loading.html
@@ -7,6 +7,8 @@
     <script src="/src/tests/fixtures/test.js"></script>
   </head>
   <body>
+    <turbo-frame style="display: contents;" id="eager-loaded-frame" src="/src/tests/fixtures/frames/frame_for-eager.html" loading="lazy"></turbo-frame>
+
     <details id="loading-lazy">
       <summary>Lazy-loaded</summary>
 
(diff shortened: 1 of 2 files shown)


seanpdoyle:
Would `console.warn` output in combination with a documentation change be enough of a balance between hands-off and informative?

brunoprietog:
It would be great to see this merged, especially to make infinitely scroll using only nested turbo frames and Tailwind's contents class, for example. Currently, a temporary option that can be used in Tailwind is to add a variant that detects when a turbo frame has not been loaded, so that in that variant there is no contents class. Once loaded, the contents class would be changed automatically.

seb-jean:
I have the same problem as @seanpdoyle

setineos:
Nice feature.

PR #627, opened 2022-07-14
Allow customizing the element that Drive replaces

agrobbin:
While most of the time, replacing `<body>` makes perfect sense, when working with 3rd party integrations (i.e. Stripe), there are elements injected just before the closing `</body>` tag that should not be
removed between page visits. There is more detail on this issue, particularly with injected `<iframe>` elements in https://github.com/hotwired/turbo/pull/305#issuecomment-879282187.

Now, if someone wants to customize the element that is replaced by Drive, they can add `data-turbo-drive-body` to an element, and only that element will be replaced between visits.

Note that I have not added tests yet, because I wanted to get some clarity on the approach before doing so (I was also running into an issue running tests on Codespaces and hadn't installed Java locally)!


src/core/drive/page_renderer.ts
```diff
@@ -104,9 +105,14 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
     }
   }
 
-  assignNewBody() {
+  async assignNewBody() {
+    await nextEventLoopTick()
```
agrobbin, line 109: There are some more details on why this is necessary in https://github.com/hotwired/turbo/pull/305#issuecomment-1183505367, but in short, without this, snapshot caches don't happen at the right time, as [`cacheSnapshot`](https://github.com/hotwired/turbo/blob/98cdc4037788f078a741c7fb50412db0101bcb97/src/core/drive/page_view.ts#L40-L49) waits *just* long enough to end up caching the new content, rather than the original. While not an issue when replacing the entire `<body>` element (since that element has not changed), it is an issue when replacing content *within* the `<body>`.

seanpdoyle:
How do you feel about declaring a `<meta>` element in the document's `<head>` (in the same style as Turbo's other [`<meta>` element configurations](https://turbo.hotwired.dev/reference/attributes#meta-tags) that declares the element's `[id]` attribute?

When the `<meta>` is absent or an element with the `meta[content]` cannot be found, Drive can still fall back to the `<body>`.

agrobbin:
@seanpdoyle I had a feeling someone would suggest doing that. 😆 I am happy to make that change. My only followup question before I do is if it should look for the `<meta>` tag in the originally loaded document every time, or if it should look in the new document? Based on how the other `<meta>` tags are read, I'm inclined to have it look at the new document to see what *it* should replace, but I'll defer to you.

agrobbin:
@seanpdoyle I just pushed up a version that leverages a `<meta name="turbo-body">` tag. Full disclosure, I don't love the meta tag name, and part of me really wants it to be `turbo-root-element-id`, but since `turbo-root` already exists (and I imagine you wouldn't want to replace that with `turbo-root-path`), I'm curious what your recommendation is.

dhh:
Does the permanent marker not work for these cases? (Not that I'm necessarily against scoping down what gets replaced, it's an interesting idea).

agrobbin:
@dhh unfortunately not, mainly because of the behavior of `<iframe>`. @mrrooijen goes into some good detail on the reasons in https://github.com/hotwired/turbo/pull/305#issuecomment-879282187, but in short, when an `<iframe>` is detached and reattached, its `src` is reloaded, which doesn't play well with these 3rd party integrations. Since `data-turbo-permanent` moves an element from one document to another, it results in the same behavior.

seanpdoyle:
Could custom render support introduced by https://github.com/hotwired/turbo/pull/431 help achieve the same outcome?

agrobbin:
@seanpdoyle I'd imagine that would work, the downside is each application would need to define its own `turbo:before-render` event, and implement its own `render` function. Considering this is a relatively common behavior to require if you're interacting with 3rd party tools, it'd be great if it was built in, but I'll defer to you and the other maintainers!

dhh:
Seems like we have some legit failures here? (In addition to some timeouts)

seanpdoyle:
> ...each application would need to define its own `turbo:before-render` event, and implement its own render function. Considering this is a relatively common behavior to require if you're interacting with 3rd party tools...

Is replacing the `<body>` swapping with another element a one-size-fits-all solution for any and every 3rd party tool that modifies the document? Would the `<meta>` element's presence and absence vary page to page? What would a transition from a `<body>`-swap to a `<main id="app">`-swap behave like? What about vice versa?

It seems like it would largely depend app-to-app, so a `turbo:before-render` override feels like enough flexibility without adding another Turbo feature.

I don't have a firm grasp on the motivating use case here, so I might be missing some crucial information. Does that still seem like an unreasonable burden for consumer applications?

agrobbin:
@seanpdoyle those are all excellent points! In this case, I think the equivalent `detail.render` would looks something like:

```js
const findAppElement = (element) => element.querySelector('#app')

    findAppElement(currentElement).replaceWith(findAppElement(newElement))
  }
})
```

Does that seem right to you (assuming `#app` is the element within `<body>` that we want to replace)?

agrobbin:
@dhh @seanpdoyle based on that, and the changes made in #431, I've made some further updates here. I went a slightly different route, to hook into `PageRenderer#shouldRender` as you suggested @dhh.

I have a feeling that tests will still fail, but I wanted to get this rebased to see how different the failures were. I imagine the issue is the addition of `await nextEventLoopTick()`, which changes the waiting behavior that might be required in tests, but I'm new to this test suite and am curious what comes to mind to you both.

dhh:
Needs a rebase and we have some failing tests.

agrobbin:
Sorry it took me a couple of weeks to get back to this, @dhh. I've done a rebase, and spent some time looking at the test failures, and while I know the cause, I'm at a bit of a loss for how to fix them.

In short, the addition of `await nextEventLoopTick()` is the cause of the test failures. Removing it allows tests to pass, but causes issues with the custom "body" element as I mentioned in https://github.com/hotwired/turbo/pull/627#discussion_r921455484.

I'm sure I'm missing something pretty obvious here, so I'm hoping someone who knows the intricacies of the library and test suite might be able to provide some guidance anyone has a free moment! Thanks in advance. ❤️

dhh:
cc @seanpdoyle

src/core/drive/page_renderer.ts
```diff
@@ -78,6 +91,12 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
     return this.currentHeadSnapshot.trackedElementSignature == this.newHeadSnapshot.trackedElementSignature
   }
 
+  get bodyElementMatches() {
+    const elementId = getBodyElementId()
+
+    return !elementId || this.newElement.querySelector(`#${elementId}`) !== null
+  }
+
```
marcoroth, line 99:
```suggestion
  get bodyElementMatches() {
    return PageRenderer.getBodyElement(this.newElement) !== null
  }

  static get bodySelector() {
    const bodyId = getBodyElementId()

    return bodyId ? `#${bodyId}` : 'body'
  }

  static getBodyElement(element: HTMLElement): HTMLElement {
    return element.querySelector(this.bodySelector) || element
  }
```

src/core/drive/page_renderer.ts
```diff
@@ -1,19 +1,26 @@
...
-  static renderElement(currentElement: HTMLBodyElement, newElement: HTMLBodyElement) {
+  static async renderElement(currentElement: HTMLBodyElement, newElement: HTMLBodyElement) {
+    await nextEventLoopTick()
+
     if (document.body && newElement instanceof HTMLBodyElement) {
-      document.body.replaceWith(newElement)
+      const bodyElementId = getBodyElementId()
+
+      const currentBody = (bodyElementId && document.querySelector(`#${bodyElementId}`)) || document.body
+      const newBody = (bodyElementId && newElement.querySelector(`#${bodyElementId}`)) || newElement
```
marcoroth, line 14:
```suggestion
      const currentBody = PageRenderer.getBodyElement(currentElement)
      const newBody = PageRenderer.getBodyElement(newElement)
```

src/core/drive/page_renderer.ts
```diff
@@ -1,19 +1,26 @@
...
 export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
-  static renderElement(currentElement: HTMLBodyElement, newElement: HTMLBodyElement) {
+  static async renderElement(currentElement: HTMLBodyElement, newElement: HTMLBodyElement) {
+    await nextEventLoopTick()
+
     if (document.body && newElement instanceof HTMLBodyElement) {
-      document.body.replaceWith(newElement)
+      const bodyElementId = getBodyElementId()
+
+      const currentBody = (bodyElementId && document.querySelector(`#${bodyElementId}`)) || document.body
```
marcoroth, line 13: @agrobbin Is there a reason why we are using `document` here instead of `currentElement`?

marcoroth (commented):
I think this looks good to me implementation-wise! 

I left two small suggestions which might help to clean-up the code a bit.

src/core/drive/page_renderer.ts
agrobbin: @marcoroth I wish I could tell you. I followed the existing implementation, which used `document.body` rather than `currentElement`, for reasons I'm not aware of.

agrobbin:
@marcoroth thanks for those suggestions! I've implemented them both, rebased, and force-pushed. I still think tests will fail, but the implementation should be in good shape now.

src/core/drive/page_renderer.ts
```diff
@@ -1,19 +1,26 @@
...
 
 export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
-  static renderElement(currentElement: HTMLBodyElement, newElement: HTMLBodyElement) {
+  static async renderElement(currentElement: HTMLBodyElement, newElement: HTMLBodyElement) {
+    await nextEventLoopTick()
+
     if (document.body && newElement instanceof HTMLBodyElement) {
-      document.body.replaceWith(newElement)
+      const bodyElementId = getBodyElementId()
+
```
marcoroth, line 12: Thanks for the response @agrobbin! I think we can even get rid of this line here
```suggestion

```

dhh (commented):
Lint fix

agrobbin:
Thanks for the merge, @dhh! Looking forward to this being released.

PR #630, opened 2022-07-16
Treat `[data-turbo-stream]` as boolean attribute

seanpdoyle:
Replace equality checks for `[data-turbo-stream="true"]` with presence
checks for `[data-turbo-stream]` to more closely match the canonical
behavior for [boolean HTML attributes][].

Turbo already supports other boolean attributes like
`[data-turbo-permanent]` and `[data-turbo-confirm]`, so
`[data-turbo-stream]` should match the precedent.

[boolean HTML attributes]: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#boolean_attributes

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -1,7 +1,7 @@
 import { FetchRequest, FetchMethod, fetchMethodFromString, FetchRequestHeaders } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { expandURL } from "../url"
-import { attributeTrue, dispatch, getMetaContent } from "../../util"
+import { dispatch, getMetaContent } from "../../util"
 import { StreamMessage } from "../streams/stream_message"
 
 export interface FormSubmissionDelegate {
@@ -219,7 +219,7 @@ export class FormSubmission {
   }
 
   requestAcceptsTurboStreamResponse(request: FetchRequest) {
-    return !request.isIdempotent || attributeTrue(this.formElement, "data-turbo-stream")
+    return !request.isIdempotent || this.formElement.hasAttribute("data-turbo-stream")
   }
 }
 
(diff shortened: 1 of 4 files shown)


seanpdoyle:
@kevinmcconnell this is a follow-up to https://github.com/hotwired/turbo/pull/612.

If this ships, could you update https://github.com/hotwired/turbo-site/pull/103 to reflect the boolean nature of the attribute?

dhh:
Looks good to me. Can you rebase?

evgenyneu:
I want "data-turbo-stream" attribute in my form with POST to be "false", so it does not request turbo stream response, but still uses turbo (without page reload). Is it possible?

marcoroth:
> I want "data-turbo-stream" attribute in my form with POST to be "false"

In that case you can just omit the `data-turbo-stream` attribute on your form.

kevinmcconnell:
@evgenyneu I don't think that's possible right now, unfortunately. We don't have a way to opt out of stream responses for POST forms, as that's the default behaviour.

Could you handle this on the server side instead, by serving a turbo stream response for that case?

evgenyneu:
@kevinmcconnell, do we want ability to opt out of turbo streams for POST forms? If yes I can create a PR, so we can use  `data-turbo-stream="false"` for forms. But this will basically revert changes in this PR? I'm confused.

> Could you handle this on the server side instead, by not serving a turbo stream response for that case?

Thanks, good idea, I can add a "no-turbo-stream" parameter to my form, and on server side render a standard `html.erb view (instead `turbo_stream.erb`) if it's present.

evgenyneu:
@kevinmcconnell sounds like a good plan, I'll make a PR. What's the purpose of the `auto` value, if it's the same as the default (missing `data-turbo-stream` attribute)?

MatheusRich:
@evgenyneu any updates on that PR? I'm interested in the same feature and wondering if I should work on that.

evgenyneu:
@MatheusRich no sorry I gradually became disinterested in it for some reason :D

PR #631, opened 2022-07-16
Introduce `FormLinkInterceptor`

seanpdoyle:
In an effort to match the patterns established by `LinkInterceptor`
and `FormInterceptor`, this commit introduces a
`FormLinkInterceptor` and `FormLinkInterceptorDelegate`.

Behind the scenes, the `FormLinkInterceptor` relies on an instance of
the `LinkInterceptor` to intervene in `<a>` element clicks when
`[data-turbo-method]` or `[data-turbo-stream]` are present. When those
clicks are detected, it creates a `<form hidden>` element, attaches it
to the document, delegates to a `FormLinkInterceptorDelegate` to map
the `<a>` element's attributes to the `<form>` element, submits the form
through the polyfilled [HTMLFormElement.requestSubmit][] method, then
removes the `<form>` from the document.

The `Session` serves as a `FormLinkInterceptorDelegate`, making sure
to start and stop the observer _before_ its `LinkInterceptor`
instance, so that clicks that are intercepted by the
`FormLinkInterceptor` are not also intercepted by the
`LinkInterceptor`.

[HTMLFormElement.requestSubmit]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit

### DIFF
--- a/src/observers/form_link_interceptor.ts
+++ b/src/observers/form_link_interceptor.ts
@@ -31,8 +31,9 @@ export class FormLinkInterceptor implements LinkInterceptorDelegate {
 
   linkClickIntercepted(link: Element, action: string): void {
     const form = document.createElement("form")
+    form.setAttribute("data-turbo", "true")
     form.setAttribute("action", action)
-    form.hidden = true
+    form.setAttribute("hidden", "")
 
     const method = link.getAttribute("data-turbo-method")
     if (method) form.setAttribute("method", method)
--- a/src/observers/form_link_interceptor.ts
+++ b/src/observers/form_link_interceptor.ts
@@ -1,0 +1,52 @@
+import { LinkInterceptor, LinkInterceptorDelegate } from "../core/frames/link_interceptor"
+
+export type FormLinkInterceptorDelegate = {
+  shouldInterceptFormLinkClick(link: Element): boolean
+  formLinkClickIntercepted(link: Element, form: HTMLFormElement): void
+}
+
+export class FormLinkInterceptor implements LinkInterceptorDelegate {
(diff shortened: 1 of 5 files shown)


seanpdoyle:
This might resolve some `<a>`-related issues like https://github.com/hotwired/turbo/issues/587 & https://github.com/hotwired/turbo/pull/588.

src/observers/form_link_interceptor.ts
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      this.delegate.shouldInterceptFormLinkClick(link) &&
+      (link.hasAttribute("data-turbo-method") || link.hasAttribute("data-turbo-stream"))
+    )
+  }
+
+  linkClickIntercepted(link: Element, action: string): void {
+    const form = document.createElement("form")
+    form.setAttribute("action", action)
+    form.hidden = true
```
willcosgrove, line 35: Apologies if this edge case is already handled, I don't have a full understanding of all of the internals of Turbo.

But in the case that `Turbo.session.drive = false`, but on this link `data-turbo="true"` is specified, then wouldn't the form created by this also need to set `data-turbo="true"` on the form element for it to be handled by Turbo?

I have a PR open to address this (#500) but the addition of this FormLinkInterceptor would change where that behavior should be implemented.

seanpdoyle: Thank you for raising these concerns! I've pushed up https://github.com/hotwired/turbo/pull/631/commits/d317319dadb5825f710009666cf6b27785839618 to add test coverage for the cases you've outlined.

dhh:
Nice cleanup!

PR #632, opened 2022-07-16
Introduce client-side Turbo Cache Control API

marcoroth:
This Pull Request introduces a `Cache` class which exposes an API as `Turbo.cache` that can be used to control the value of the `turbo-cache-control` meta element form the client-side.

Most of the times you just need to the set the value of the `turbo-cache-control` meta element from the server-side when you are responding to your regular HTML request, based on if you want that page to be cached or not.

I encountered a few use-cases in a couple of apps I worked on where you dynamically want to set the cache control behavior after a certain action happened on the client (i.e. you don't want to show remaining artifacts of a finished animation in a later visit, but if the animation wasn't triggered you still want that page to be properly cached).

You can achieve a similar behavior by dynamically prepending a `<meta>` element to the `<head>` via JavaScript, like:
```javascript
document.head.insertAdjacentHTML(
  'beforeend',
  '<meta name="turbo-cache-control" content="no-cache">'
)
```

Though, this feels very error-prone and doesn't account for duplicate `<meta>` elements in the `<head>`.

This Pull Request aims to simplify this by exposing a client-side API to control the value of the meta element with just the values Turbo actually knows about. With that, the above example becomes:
```javascript
Turbo.cache.exemptPageFromCache()
```

The name of the exposed functions are kept in-line with the helpers in:
https://github.com/hotwired/turbo-rails/blob/43bf84a9b1d2cc78da6810b82f92be2c32c9dbfd/app/helpers/turbo/drive_helper.rb#L15-L23

**Open Points / Design Decisions:** 

- [ ] Introducing a `Cache` class and exposing an instance of said class as `Turbo.cache` might be overkill. It might be just fine to instead add the helpers as top-level functions, similar as with the `Turbo.clearCache()` helper.
  - `Turbo.cache.exemptPageFromCache()`  => `Turbo.exemptPageFromCache()`
  - `Turbo.cache.exemptPageFromPreview()` =>  `Turbo.exemptPageFromPreview()`
  
- [ ] The helper names `exemptPageFromCache()` and `exemptPageFromPreview()` don't exactly line up the with attribute values.
  - An alternative would be to do something like `Turbo.cache.noCache()` / `Turbo.cache.noPreview()`

### DIFF
--- a/src/core/cache.ts
+++ b/src/core/cache.ts
@@ -1,0 +1,30 @@
+import { Session } from "./session"
+import { setMetaContent } from "../util"
+
+export class Cache {
+  readonly session: Session
+
+  constructor(session: Session) {
+    this.session = session
+  }
+
+  clear() {
+    this.session.clearCache()
+  }
+
+  resetCacheControl() {
+    this.setCacheControl("")
+  }
+
+  exemptPageFromCache() {
+    this.setCacheControl("no-cache")
+  }
+
+  exemptPageFromPreview() {
+    this.setCacheControl("no-preview")
+  }
+
+  private setCacheControl(value: string) {
+    setMetaContent("turbo-cache-control", value)
+  }
+}
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -1,7 +1,7 @@
 import { FetchRequest, FetchMethod, fetchMethodFromString, FetchRequestHeaders } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { expandURL } from "../url"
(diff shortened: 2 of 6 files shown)


dhh:
I think what you've presented here is the right trade-offs. Gathering all this in a class provides a nice namespace, and the clearer names that include page are easier to understand the meta matching ones.

dhh:
Should get a doc PR going for 7.2.0 release too.

marcoroth:
I will follow up with a doc PR 👍🏼

Should we deprecate `Turbo.clearCache()` in favor of `Turbo.cache.clear()`?

dhh:
Yes please 👍

PR #636, opened 2022-07-18
Add support for data-turbo-action to link_to 

ondrej-zadnik:
As I described in https://github.com/hotwired/turbo-rails/issues/359 `link_to` ignores `data-turbo-action` in combination with `data-turbo-method`.

Original error was found in `convertLinkWithMethodClickToFormSubmission` but it was overwritten by https://github.com/hotwired/turbo/pull/631. Unfortunately the `data-turbo-action` for `link_to` is still missing.

I added the `data-turbo-action` setting to `FormLinkInterceptor` in the same way as `data-turbo-confirm` and added a test  for it.

### DIFF
--- a/src/observers/form_link_click_observer.ts
+++ b/src/observers/form_link_click_observer.ts
@@ -42,6 +42,9 @@ export class FormLinkClickObserver implements LinkClickObserverDelegate {
     const turboFrame = link.getAttribute("data-turbo-frame")
     if (turboFrame) form.setAttribute("data-turbo-frame", turboFrame)
 
+    const turboAction = link.getAttribute("data-turbo-action")
+    if (turboAction) form.setAttribute("data-turbo-action", turboAction)
+
     const turboConfirm = link.getAttribute("data-turbo-confirm")
     if (turboConfirm) form.setAttribute("data-turbo-confirm", turboConfirm)
 
--- a/src/tests/fixtures/navigation.html
+++ b/src/tests/fixtures/navigation.html
@@ -37,6 +37,7 @@
       <p><a id="same-origin-unannotated-link-search-params" href="/src/tests/fixtures/one.html?key=value">Same-origin unannotated link ?key=value</a></p>
       <p><form id="same-origin-unannotated-form" method="get" action="/src/tests/fixtures/one.html"><button>Same-origin unannotated form</button></form></p>
       <p><a id="same-origin-replace-link" href="/src/tests/fixtures/one.html" data-turbo-action="replace">Same-origin data-turbo-action=replace link</a></p>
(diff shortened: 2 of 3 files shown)


seanpdoyle (approved):

src/observers/form_link_interceptor.ts
```diff
@@ -38,6 +38,9 @@ export class FormLinkInterceptor implements LinkInterceptorDelegate {
     const method = link.getAttribute("data-turbo-method")
     if (method) form.setAttribute("method", method)
 
+    const turboAction = link.getAttribute("data-turbo-action")
+    if (turboAction) form.setAttribute("data-turbo-action", turboAction)
```
seanpdoyle, line 42: Thank you for catching this missing attribute!

This method is becoming unwieldy. I'd considered an implementation that looped over the `<a>` element's attributes and mapped them to attributes on the `<form>` element, but decided against it.

As this list continues to grow, I wonder whether or not its time.

What do you think about implementing this method in terms of [Element.attributes](https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes) and [Attr.localName](https://developer.mozilla.org/en-US/docs/Web/API/Attr/localName) and [Attr.value](https://developer.mozilla.org/en-US/docs/Web/API/Attr/value), special-casing for `[data-turbo-method]`?

```diff
diff --git a/src/observers/form_link_interceptor.ts b/src/observers/form_link_interceptor.ts
index 6df0203..a2a309f 100644
--- a/src/observers/form_link_interceptor.ts
+++ b/src/observers/form_link_interceptor.ts
@@ -35,14 +35,13 @@ export class FormLinkInterceptor implements LinkInterceptorDelegate {
     form.setAttribute("action", action)
     form.setAttribute("hidden", "")
 
-    const method = link.getAttribute("data-turbo-method")
-    if (method) form.setAttribute("method", method)
-
-    const turboConfirm = link.getAttribute("data-turbo-confirm")
-    if (turboConfirm) form.setAttribute("data-turbo-confirm", turboConfirm)
-
-    const turboStream = link.hasAttribute("data-turbo-stream")
-    if (turboStream) form.setAttribute("data-turbo-stream", "")
+    for (const { localName, value } of link.attributes) {
+      if (localName == "data-turbo-method") {
+        form.setAttribute("method", value)
+      } else if (localName.startsWith("data-turbo")) {
+        form.setAttribute(localName, value)
+      }
+    }
 
     this.delegate.formLinkClickIntercepted(link, form)
 ```

seanpdoyle: That works too! Let's address the issue then we can assess after the fact

ondrej-zadnik:
@seanpdoyle @dhh Do you think we could add this to the turbo release 7.2? It would be helpful because now we have to use `button_to` everywhere and it may not fit in all situations.

dhh:
Yes. Please do rebase and we'll get it in.

ondrej-zadnik:
@dhh rebased

Issue #462, opened 2021-11-22
Provide event to handle `turbo-frame` hitting a `NetworkError` when fetching


romaricpascal:
At the moment, if a `turbo-frame` cannot make its request to the server due to a network issue, there is no way to provide feedback to users as the error gets only logged to the console and not surfaced anywhere. No `before-fetch-response` gets triggered as, well..., there was indeed no response. 

It'd be great if the frame would dispatch a `turbo:fetch-error` event (or similarly named) to let the page add some rendering.

For now, this can be achieved through the following, but requires some very far reaching overriding:
```js
const frame = document.createElement("turbo-frame");
const controllerPrototype = Object.getPrototypeOf(frame.delegate);
const originalRequestErrored = controllerPrototype.requestErrored;
controllerPrototype.requestErrored = function (request, error) {
  console.log('Network error', this.element, 'Dispatching event');
  this.element.dispatchEvent(new CustomEvent('turbo:fetch-error', {
    bubbles: true,
    cancelable: true,
    detail: {request,error}
  }));
  return originalRequestErrored.apply(this, arguments);
};
```

To harmonize error handling, it may also be worth dispatching this event when the response is not a success, though that could be easily achieved by library users through:

```js
document.addEventListener('turbo:before-fetch-response', (event) => {
  if (!event.detail.fetchResponse.response.ok) {
    event.target.dispatchEvent(new CustomEvent('turbo:fetch-error', {
      bubbles: true,
      cancelable: true,
      detail: event.detail
    }))
  }
});
```

I'm not familiar if this would translate to Turbo drive as well, I'd guess there must be some place it needs to handle fetch errors too.

corporealfunk:
I agree that this would be useful. It would be nice to be able access the fact that the network failed when we set the ```src``` attribute on a frame. Homegonizing errors would be nice too instead of having to listen for ```submit-end``` and also ```before-fetch-response```.

Issue #460, opened 2021-11-21
Feature Request: Ability to access fetch errors when the network or server is unreachable during GET requests


corporealfunk:
In the case where a fetch request fails because the network or remote server is unreachable, a NetworkError is thrown and logged in the console (as well as the uncaught in promise error). If the fetch request was created because of a form POST, turbo:submit-end is also dispatched and developers can access the submit state and error message via the dispatched event.

However, other actions that may cause Turbo to create a fetch GET request that then fails because the network is unreachable (for example, programmatically setting a turbo frame src attribute, or using a form with GET instead of POST) do not dispatch any turbo:* events at all and developers do not have the ability to programmatically respond to the network error.

Could these types of failures always dispatch a turbo event, not just in the case of a form submission?

seanpdoyle:
As of https://github.com/hotwired/turbo/pull/424, `<form method="get">` submissions will fire a `turbo:submit-start` and `turbo:submit-end` event. The `turbo:submit-end` has HTTP response information in the `event.detail`.

Outside of that, would attaching listeners for [`turbo:before-fetch-request`](https://turbo.hotwired.dev/reference/events) events suit your needs?

corporealfunk:
#424 looks like it will work for my use case.

```turbo:before-fetch-request``` dispatches an event that only contains the fetchOptions regarding what is _about to be fetched_, and happens too early in the fetch lifecycle to contain any information about a network failure or unavailable server.

seanpdoyle:
@corporealfunk you're correct in that. What I meant to share was the `turbo:before-fetch-response` event that's a companion to the request. That should fire for a Visit, a Form Submission, and event a Frame navigation.

tleish:
@seanpdoyle - I think `turbo:before-fetch-response` is the appropriate location, as long as the listener can call `event.preventDefault()` to stop the application from rendering an error page so you can show a custom message.

As of right now, this does not fire on a network connection is lost for Visit using the following steps:
- Add Turbo event listener for `turbo:before-fetch-response`
- Load page with turbo
- Chrome > Developer Tools > Network Tab > Change Throttling to "Offline"
- Click on link

`turbo:before-fetch-response` is not called. 

FYI, similar request in [#460](https://github.com/hotwired/turbo/issues/462)

corporealfunk:
I'm actually going to reopen this. My previous use case was satisfied with the merge of PR #424, but I have another use case now where Turbo doesn't seem to expose fetch error states like NetworkError:

[...]

PR #640, opened 2022-07-20
Add turbo:fetch-request-error event on frame and form network errors

srt32:
closes https://github.com/hotwired/turbo/issues/462 and possible https://github.com/hotwired/turbo/issues/460

This PR adds a new event called `turbo:fetch-error` that dispatches
when a form or frame fetch request errors.

This event will let developers respond appropriately when a fetch request
fails due network errors (such as on flaky wifi). A developer could reload the page or show an error message of some kind.

At the moment, if a frame navigation fetch request fails due to a network
error an error is thrown so https://github.com/hotwired/turbo/blob/2d5cdda4c030658da21965cb20d2885ca7c3e127/src/http/fetch_request.ts#L107 never runs
and therefore the `turbo:before-fetch-response` event is not dispatched.

Here is a demo using http://127.0.0.1:9000/src/tests/fixtures/tabs.html as a fixture:

https://user-images.githubusercontent.com/2181356/179891552-86e46b86-6d2e-4e3e-a1da-1066dea44550.mp4

I'm very open to feedback or other ideas on a good direction to go here.


src/core/drive/form_submission.ts
```diff
@@ -199,6 +200,11 @@ export class FormSubmission {
 
   requestErrored(request: FetchRequest, error: Error) {
     this.result = { success: false, error }
+    dispatch<TurboFetchErrorEvent>("turbo:fetch-error", {
+      target: this.formElement,
+      detail: { request, error },
+      cancelable: true,
```
keithamus, line 206: This probably shouldn't be cancelable. 

```suggestion
```

src/core/frames/frame_controller.ts
```diff
@@ -241,6 +241,11 @@ export class FrameController
 
   requestErrored(request: FetchRequest, error: Error) {
     console.error(error)
+    dispatch<TurboFetchErrorEvent>("turbo:fetch-error", {
+      target: this.element,
+      detail: { request, error },
+      cancelable: true,
```
keithamus, line 247: This probably shouldn't be cancelable. 

```suggestion
```

seanpdoyle:
Thank you for proposing this! Is it possible to add test coverage? Does playwright have the ability to force the Chrome or Firefox browser running behind the scenes into Offline mode?

srt32:
> Thank you for proposing this! Is it possible to add test coverage? Does playwright have the ability to force the Chrome or Firefox browser running behind the scenes into Offline mode?

I've got something ugly in https://github.com/hotwired/turbo/pull/640/commits/ecdd1c9a054914f19e69f013c94941f46a4c423a and curious if you know of a good way to determine which browser we need to pull in.

src/tests/functional/frame_navigation_tests.ts
```diff
@@ -24,6 +25,25 @@ test("test frame navigation with exterior link", async ({ page }) => {
...
+  if (testInfo.project.name === "chrome") {
+    browser = await chromium.launch()
+  } else if (testInfo.project.name === "firefox") {
+    browser = await firefox.launch()
+  } else {
+    throw new Error("unsupported browser")
+  }
+  const context = await browser.newContext()
+  page = await context.newPage()
+  await context.setOffline(true)
```
seanpdoyle, line 40: Could we get at this `Context` instance through [Page.context()](https://playwright.dev/docs/api/class-page#page-context)?

src/tests/functional/frame_navigation_tests.ts
```diff
@@ -24,6 +24,13 @@ test("test frame navigation with exterior link", async ({ page }) => {
   await nextEventOnTarget(page, "frame", "turbo:frame-load")
 })
 
+test("test frame navigation emits fetch-error event when offline", async ({ page }) => {
+  await page.goto("/src/tests/fixtures/tabs.html")
```
seanpdoyle, line 28: While I appreciate that you're recreating the failure you experienced with the page that you experienced it on, is there a link in the `/src/tests/fixtures/frame_navigation.html` page that we could exercise for the same outcome? That way we could cut down on the number of `page.goto` calls.

seanpdoyle (approved):

src/core/frames/frame_controller.ts
```diff
@@ -247,6 +247,10 @@ export class FrameController
 
   requestErrored(request: FetchRequest, error: Error) {
     console.error(error)
+    dispatch<TurboFetchErrorEvent>("turbo:fetch-error", {
```
seanpdoyle, line 250: How do you feel about naming the event `turbo:fetch-request-error`? Other [fetch-related events](https://turbo.hotwired.dev/reference/events) are named to describe which side of they pipe they're dispatched on: `turbo:before-fetch-request` and `turbo:before-fetch-response`.

seanpdoyle (approved):

src/core/frames/frame_controller.ts
seanpdoyle: Also, could you add the type to the [exports from `src/core/index.ts`](https://github.com/hotwired/turbo/blob/60c646383b3c0355242d2e3c69f80742efb266be/src/core/index.ts#L17-L31)?

srt32: Done in db678ec176d74eff9c82876e05b979b1fddc16c2! I'm still fighting tests a bit.

src/tests/functional/frame_navigation_tests.ts
```diff
@@ -23,6 +23,13 @@ test("test frame navigation with exterior link", async ({ page }) => {
   await nextEventOnTarget(page, "frame", "turbo:frame-load")
 })
 
+test("test frame navigation emits fetch-request-error event when offline", async ({ page }) => {
+  await page.goto("/src/tests/fixtures/tabs.html")
+  await page.context().setOffline(true)
+  await page.click("#tab-2")
+  await nextEventNamed(page, "turbo:fetch-request-error")
```
seanpdoyle, line 30: @srt32 to use this helper method, you need to add an entry for the event name to the src/tests/test.js file.

srt32:
@seanpdoyle thanks for your help here! This change should be green and ready to go. I bumped into some recursion issues with the test helper so I opted to not serialize the `detail` for that specific event. I'm not sure exactly why this event is behaving differently than the others but the build is 🟢  now ;)

seanpdoyle (approved):

srt32:
Thanks for the thumb! What do we need to do to get this PR merged?

PR #641, opened 2022-07-20
Convert search parameters to form inputs when using `data-turbo-stream`

LVMBDV:
Fixes a problem pointed out recently on #612 by @seanabrahams and @kevinmcconnell.

Links with `data-turbo-stream` are converted to forms, and forms with method="GET" don't submit the query string part of their action. This can be circumvented by iterating through any query string parameters and adding them to the generated form as hidden inputs.

Note: MDN refers to what we usually call "query string parameters" as "search parameters". This is a bit confusing, but it's the correct terminology I guess :)

### DIFF
--- a/src/observers/form_link_interceptor.ts
+++ b/src/observers/form_link_interceptor.ts
@@ -35,6 +35,16 @@ export class FormLinkInterceptor implements LinkInterceptorDelegate {
     form.setAttribute("action", action)
     form.setAttribute("hidden", "")
 
+    const searchParameters = new URLSearchParams(new URL(action, window.location.toString()).search)
+    for (const [name, value] of searchParameters) {
+      const input = document.createElement("input")
+      input.setAttribute("type", "hidden")
+      input.setAttribute("name", name)
+      input.setAttribute("value", value)
+
+      form.appendChild(input)
+    }
+
     const method = link.getAttribute("data-turbo-method")
     if (method) form.setAttribute("method", method)
 


kevinmcconnell:
I've been looking at a different approach to this problem as well. Originally it seemed like submitting `data-turbo-stream` links as forms would be a straightforward & safe way to do this (especially as we already have `data-turbo-method` doing something similar), but seeing that issue with parameters come up it's making me doubt that approach.

I've got a branch going where I'm trying a different tack -- rather than submit these stream links as `GET` forms, the navigation code can simply add the stream MIME type to its usual request if the link has `data-turbo-stream` (in much the way way as the `FormSubmission` adds it according to its own logic). I think that will actually be a less invasive change, and less likely to introduce any unexpected behaviour, since the only change to the request will be that extra item in the `Accept` header.

I think it would be worth looking at that approach before we settle on the right fix. I'll try to get a PR open as soon as I can (but it will probably be a few days, sorry; busy week at work!)

@seanpdoyle does that sound reasonable to you?

seanpdoyle:
@kevinmcconnell I'm in favor of that change.

> I'll try to get a PR open as soon as I can (but it will probably be a few days, sorry; busy week at work!)

I could draft something up today, if that's OK with you.

@LVMBDV could you create a test or two exercising the implementation change in this pull request? If you're able to add them here, I'll make sure to get them passing on the alternative PR.

kevinmcconnell:
> I could draft something up today, if that's OK with you.

Absolutely, that would be great 👍

FWIW, the approach I was looking at is to add an `acceptsStreamResponse` option to `VisitOptions`, which `Visit` can refer to in a `prepareHeadersForRequest`. That way, the `Session` can set that option appropriately when handling clicks. Seems to work pretty well, and is a less intrusive change. The part I hadn't tackled yet is that the `FrameController` will (I think) also want something similar, for the cases where it's issuing the request. That's where I ran out of time :)

Of course you may have an entirely different approach in mind; just wanted to mention where I'd reached, in case it saves you any time.

LVMBDV:
Hi, I needed this for a specific use case I stopped pursuing to allocate more time to releasing an MVP. If the other approaches don't work out, I can refine this PR i.e. add tests and documentation but let's shelve this one until that's decided.

dhh:
Solved via https://github.com/hotwired/turbo/pull/647 instead.

PR #647, opened 2022-07-25
Don't convert `data-turbo-stream` links to forms

kevinmcconnell:
The initial implementation of the `data-turbo-stream` attribute worked
by treating requests with that attribute as form submissions (which
would be `GET` form submissions, unless a different method type was
indicated).

This allowed us to include the `data-turbo-stream` logic in
`FormSubmission`, which was responsible for properly setting the
`Accept` header.

However, there are some downsides to submitting the requests as form
submissions. For example, a `GET` form submission does not include the
search portion of the URL. Additionally, servers are free to respond to
`data-turbo-stream` requests with plain HTML responses, and in that case
we don't want Turbo's handling of the response to differ from what it
would have been if the `data-turbo-stream` attribute wasn't present.

This commit takes a different approach. In this version, elements that
have `data-turbo-stream` continue to be processed by the same object
that they would otherwise, whether that's a `FormSubmission`, a `Visit`
or a `FrameController`. However each of these objects is now aware of
the `data-turbo-stream` attribute, and each will include the Turbo
Stream MIME type when appropriate.

This minimizes the impact of `data-turbo-stream` so that the only effect
it has is on the inclusion of that MIME type.

### DIFF
--- a/src/tests/functional/form_submission_tests.ts
+++ b/src/tests/functional/form_submission_tests.ts
@@ -886,17 +886,10 @@ test("test stream link GET method form submission inside frame", async ({ page }
 test("test stream link inside frame", async ({ page }) => {
   await page.click("#stream-link-inside-frame")
 
-  const { fetchOptions } = await nextEventNamed(page, "turbo:before-fetch-request")
+  const { fetchOptions, url } = await nextEventNamed(page, "turbo:before-fetch-request")
 
   assert.ok(fetchOptions.headers["Accept"].includes("text/vnd.turbo-stream.html"))
-})
-
-test("test stream link outside frame", async ({ page }) => {
-  await page.click("#stream-link-outside-frame")
-
-  const { fetchOptions } = await nextEventNamed(page, "turbo:before-fetch-request")
-
-  assert.ok(fetchOptions.headers["Accept"].includes("text/vnd.turbo-stream.html"))
+  assert.equal(getSearchParam(url, "content"), "Link!")
 })
 
 test("test link method form submission within form inside frame", async ({ page }) => {
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -161,7 +161,7 @@ export class FormSubmission {
     }
 
     if (this.requestAcceptsTurboStreamResponse(request)) {
-      headers["Accept"] = [StreamMessage.contentType, headers["Accept"]].join(", ")
+      request.acceptResponseType(StreamMessage.contentType)
     }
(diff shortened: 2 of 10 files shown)


src/observers/form_link_interceptor.ts
```diff
@@ -23,10 +23,7 @@ export class FormLinkInterceptor implements LinkInterceptorDelegate {
   }
 
   shouldInterceptLinkClick(link: Element): boolean {
-    return (
-      this.delegate.shouldInterceptFormLinkClick(link) &&
-      (link.hasAttribute("data-turbo-method") || link.hasAttribute("data-turbo-stream"))
-    )
+    return this.delegate.shouldInterceptFormLinkClick(link) && link.hasAttribute("data-turbo-method")
```
kevinmcconnell, line 26: Removing the check for `data-turbo-stream` here is what removes the previous behaviour of turning `data-turbo-stream` links into `GET` form submissions.

src/tests/functional/form_submission_tests.ts
```diff
@@ -843,17 +843,10 @@ test("test stream link GET method form submission inside frame", async ({ page }
 test("test stream link inside frame", async ({ page }) => {
   await page.click("#stream-link-inside-frame")
 
-  const { fetchOptions } = await nextEventNamed(page, "turbo:before-fetch-request")
-
-  assert.ok(fetchOptions.headers["Accept"].includes("text/vnd.turbo-stream.html"))
-})
-
-test("test stream link outside frame", async ({ page }) => {
```
kevinmcconnell, line 851: As links like this are no longer treated as form submissions, I've moved this to the visit tests.

kevinmcconnell:
This is an alternative approach to fixing the issue in #641 (as [discussed](https://github.com/hotwired/turbo/pull/641#issuecomment-1190328737) on that PR), and also generally improves on #612.

@seanpdoyle I finally found some time to finish this up after we discussed it last week. Would you mind taking a look if you get a chance? Thanks!

kevinmcconnell:
@dhh could we include this in the 7.2.0 milestone? I think this a better implementation of `data-turbo-stream`, and it would be nice to get the extra testing on this version before shipping.

dhh:
Can you rebase?

kevinmcconnell:
@dhh this has been rebased now.

kevinmcconnell:
You can ignore the extra couple of force pushes here -- was just triggering CI to get a clean test run. It looks like we have a couple of flaky tests on `main` at the moment.

PR #648, opened 2022-07-25
Disabling Drive keeps Frames default behaviour

rik:
https://github.com/hotwired/turbo/pull/196 introduced the ability to disable Drive by default. An unintentional side-effect was that it also disabled Frame by default. Every Frame needed the data-turbo attribute to be enabled: `<turbo-frame data-turbo="true">`. This makes it so that `Turbo.session.drive = false` only affects Drive.

On top of the logic change, we rename the Session methods to remove the Drive vocabulary.

---

@gregschmit Ping as the author of #196. Am I right to assume that this was an un unintentional side-effect?

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -431,11 +431,11 @@ export class FrameController
       }
     }
 
-    if (!session.elementDriveEnabled(element)) {
+    if (!session.elementIsNavigatable(element)) {
       return false
     }
 
-    if (submitter && !session.elementDriveEnabled(submitter)) {
+    if (submitter && !session.elementIsNavigatable(submitter)) {
       return false
     }
 
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -175,7 +175,7 @@ export class Session
 
   willFollowLinkToLocation(link: Element, location: URL, event: MouseEvent) {
     return (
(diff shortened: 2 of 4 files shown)


gregschmit:
Yes, that was definitely unintentional. I thought that since frames are opt-in already (in the sense that you have to explicitly use the `turbo-frame` element), that they wouldn't be impacted by the `data-turbo` element property to determine whether they should function as frames (but I see now that they are looking at `elementIsNavigatable`).

This change makes a lot of sense to me, because it essentially says that if you've inserted a `turbo-frame` element, then turbo should be enabled in that context, even if you disabled it by default.

guillaumewrobel:
Breaking change in our codebase, but we'll add `target="_top"` or `data-turbo="false"` to all links inside `<turbo-frame-tag>`

hidr0:
This broke something for me too.

@gregschmit could you please humour me and explain how:

> This change makes a lot of sense to me, because it essentially says that if you've inserted a turbo-frame element, then turbo should be enabled in that context, even if you disabled it by default.

`should be enabled in that context, even if you disabled it by default.` isn't default needed to be default?

Thinking out loud here, I kind of get your point, but in this way there is not easy one switch to disabled Drive application wide. Do we need that one switch, hmm.. Not sure.

Maybe we need a flag to disable Drive + Frames?

Just raising a flag here.

PR #649, opened 2022-07-27
`Turbo.visit(..., { frame: "frame" })`

seanpdoyle:
Add support for driving a `<turbo-frame>` element from a
`Turbo.visit(...)` call.

Internally, Turbo has some constraints for finding and driving a
`<turbo-frame>` element, namely that it has an `[id]` attribute and that
it is not marked with a `[disabled]` attribute.

Supporting `Turbo.visit` calls that specify a `frame:` option
provide an opportunity to encapsulate those details from callers while
still providing element finding utility.

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -48,6 +48,7 @@ export type VisitOptions = {
   updateHistory: boolean
   restorationIdentifier?: string
   shouldCacheSnapshot: boolean
+  frame?: string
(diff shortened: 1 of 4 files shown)


dhh:
`Turbo.visit("/pathname.html", { frame: "x" })` strikes me as easier on the eye. I know that doesn't mesh fully with the destruct, but imo that's ok.

PR #654, opened 2022-07-29
Do not autofocus inert or hidden elements

seanpdoyle:
Closes [hotwired/turbo#546][]

According to the specification, there is an [algorithm][] for
determining the [focusable area][], that is, the elements with the
`[autofocus]` attribute that can receive focus:

> Elements that meet all the following criteria:
> * the element's [tabindex][] value is non-null, or the element is determined by the user agent to be focusable;
> * the element is either not a shadow host, or has a shadow root whose delegates focus is false;
> * the element is not actually [disabled][];
> * the element is not [inert][];
> * the element is either being rendered or being used as relevant canvas fallback content.

This commit extends the `Snapshot.firstAutofocusableElement` to
programmatically determine the focusable area based on the available
attributes, including checks for `<details>` and `<dialog>` ancestors
that do not declare `[open]`.

[hotwired/turbo#546]: https://github.com/hotwired/turbo/pull/546
[tabindex]: https://html.spec.whatwg.org/multipage/interaction.html#tabindex-value
[algorithm]: https://html.spec.whatwg.org/multipage/interaction.html#get-the-focusable-area
[disabled]: https://html.spec.whatwg.org/multipage/semantics-other.html#concept-element-disabled
[inert]: https://html.spec.whatwg.org/multipage/interaction.html#inert
[focusable area]: https://html.spec.whatwg.org/multipage/interaction.html#focusable-area
[hidden]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden

### DIFF
--- a/src/core/snapshot.ts
+++ b/src/core/snapshot.ts
@@ -26,7 +26,14 @@ export class Snapshot<E extends Element = Element> {
   }
 
   get firstAutofocusableElement() {
-    return this.element.querySelector("[autofocus]")
+    const inertDisabledOrHidden = "[inert], :disabled, [hidden], details:not([open]), dialog:not([open])"
+
+    for (const element of this.element.querySelectorAll("[autofocus]")) {
+      if (element.closest(inertDisabledOrHidden) == null) return element
+      else continue
+    }
+
+    return null
   }
 
   get permanentElements() {
--- a/src/tests/fixtures/autofocus-inert.html
+++ b/src/tests/fixtures/autofocus-inert.html
@@ -1,0 +1,28 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Autofocus</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+  </head>
+  <body>
+    <h1>Autofocus With Inert Elements</h1>
+    <dialog>
+      <button id="dialog-autofocus-element" autofocus>dialog[autofocus]</button>
+    </dialog>
+    <details>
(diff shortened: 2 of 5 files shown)


src/core/snapshot.ts
```diff
@@ -26,7 +26,14 @@ export class Snapshot<E extends Element = Element> {
   }
 
   get firstAutofocusableElement() {
-    return this.element.querySelector("[autofocus]")
+    const inertDisabledOrHidden = "[inert], [disabled], [hidden], details:not([open]), dialog:not([open])"
```
keithamus, line 29: Might be worth adding `:disabled`?

```suggestion
    const inertDisabledOrHidden = "[inert], [disabled], [hidden], details:not([open]), dialog:not([open]), :disabled"
```

seanpdoyle: Interesting catch!

What's the semantic difference between the attribute and the pseudo class?

keithamus: `[disabled]` would only select if the attribute is present. Inputs can be disabled if they're part of a disabled fieldset:

```html
<fieldset disabled>
  <input type="text" autofocus>
</fieldset>
```

```js
console.log(document.querySelectorAll('input[disabled]').length) // 0
console.log(document.querySelectorAll('input:disabled').length) // 1
```

Having said that you're using `.closest`, so maybe this isn't an issue... although this might need to be verified but I think `<form disabled>` _won't_ disable child inputs:

```html
<form disabled>
  <input type="text" autofocus>
</form>
```

^ I think this _might_ need to be autofocused.

seanpdoyle: Thank you for clarifying! I've removed the check for `[disabled]`, and replaced it with `:disabled`.

PR #656, opened 2022-07-29
Change `elementDriveEnabled()` to `elementIsNavigatable()`

marcoroth:
The Pull Request #648 refactored the function name of
`session.elementDriveEnabled()` to `session.elementIsNavigatable()`.

This Pull Request fixes an old function reference which kept the library
from building successfully.


rik:
Ah thanks for fixing this. Looks like https://github.com/hotwired/turbo/pull/645 being merged before https://github.com/hotwired/turbo/pull/648 introduced the issue.

PR #657, opened 2022-07-30
Extract `DriveDelegate` interface

seanpdoyle:
The `FrameController` depends on the application's singleton `Session`
instance in several ways:

* it drives the `History`
* it determines whether or not an `Element` is "enabled" to navigate a
  page or frame
* it preloads links
* it dispatchs `turbo:frame-render` and `turbo:frame-load` events
* it can initiate a `Visit`

This commit codifies those dependent behaviors by creating the
`DriveDelegate` interface and declaring the `Session` class as an
extension of that interface.

The `FrameController` still directly imports the singleton instance of
the `Session` from the `src/core/index` module, but assigns it to a
`delegate: DriveDelegate` property for use throughout the class.

The next step would be to find a way to forward the `session` instance
into the [FrameElement.delegateConstructor][] assignment.

[FrameElement.delegateConstructor]: https://github.com/hotwired/turbo/blob/3a18111412cb4190b818b96c9c6aad7264cc4441/src/elements/index.ts#L6


marcoroth (approved):
I really appreciate the work being done here, this is a really good extraction and refactoring!

I just have one question: what was your thought around having `frameMissingFromResponse()` as part of the `DriveDelegate` interface?

Would why other potential classes (that have a `DriveDelegate`) care about dispatching a warning about missing frames? Isn't that the sole responsibility of the `FrameController`?

seanpdoyle:
@marcoroth these changes were proposed in the early stages of missing frame handling. I've rebased and moved all frame-related checks to the `FrameController`.

marcoroth (approved):
Ah, that would explain it!

This looks good from my side! 🥳

seanpdoyle:
@manuelpuyol @kevinmcconnell I'd really appreciate a review of this changeset, if you're able to provide some.

manuelpuyol (approved):

PR #660, opened 2022-07-31
Activate `<script>` in Turbo Streams

seanpdoyle:
Closes https://github.com/hotwired/turbo/pull/527

When a `<turbo-stream>` element connects to the document, "activate"
each `<script>` element by re-creating it as an element owned by the
`document` instance.

To re-use the existing `Renderer.createdScriptElement`, this commit
extracts that logic to the `src/util.ts` module, and renames it to
`activateScriptElement`. Then, we replace each
`this.createdScriptElement` call site with a call to
`activateScriptElement`.

Along with that change, this commit changes the signature to expect an
[HTMLScriptElement][] instance,

Finally, as part of adding test coverage for the new behavior, this
commit re-structures existing tests to be more particular about their
expectations, and to throw failure messages that provide more clarity.

[HTMLScriptElement]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement

Co-authored-by: Shia Labeouf <felixfearest@gmail.com>

### DIFF
--- a/src/core/streams/stream_message.ts
+++ b/src/core/streams/stream_message.ts
@@ -1,40 +1,33 @@
 import { StreamElement } from "../../elements/stream_element"
+import { activateScriptElement, createDocumentFragment } from "../../util"
 
 export class StreamMessage {
   static readonly contentType = "text/vnd.turbo-stream.html"
-  readonly templateElement = document.createElement("template")
+  readonly fragment: DocumentFragment
 
   static wrap(message: StreamMessage | string) {
     if (typeof message == "string") {
-      return new this(message)
+      return new this(createDocumentFragment(message))
     } else {
       return message
     }
   }
 
-  constructor(html: string) {
-    this.templateElement.innerHTML = html
+  constructor(fragment: DocumentFragment) {
+    this.fragment = importStreamElements(fragment)
   }
+}
 
-  get fragment() {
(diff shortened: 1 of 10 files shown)


src/core/streams/stream_message.ts
```diff
@@ -1,40 +1,33 @@
...
+
+function importStreamElements(fragment: DocumentFragment): DocumentFragment {
+  for (const element of fragment.querySelectorAll<StreamElement>("turbo-stream")) {
+    const streamElement = document.importNode(element, true)
 
-  get fragment() {
-    const fragment = document.createDocumentFragment()
-    for (const element of this.foreignElements) {
-      fragment.appendChild(document.importNode(element, true))
+    for (const inertScriptElement of streamElement.templateElement.content.querySelectorAll("script")) {
```
marcelolx, line 25: @seanpdoyle I think this breaks the current behavior we have with the `remove` action https://turbo.hotwired.dev/reference/streams#remove because the [templateElement](https://github.com/hotwired/turbo/blob/main/src/elements/stream_element.ts#L111-L114) requires that the stream element contains a child element `<template>` which for this specific action is not the case

seanpdoyle: Thank you for raising this issue! I've opened https://github.com/hotwired/turbo/pull/665 to resolve it.

Issue #664, opened 2022-08-04
v7.2.0-beta.2 doesn't accept turbo streams of `remove` action without a `template` child element


marcelolx:
In the latest beta release, we have seen the following error when we send a turbo-stream targeting an element to be removed from the DOM

```
Uncaught Error: <turbo-stream action="remove" target="target_element_here">: first child element must be a <template> element
    at HTMLElement.raise (turbo.es2017-esm.js:6517:1)
    at HTMLElement.get (turbo.es2017-esm.js:6497:1)
    at importStreamElements (turbo.es2017-esm.js:1036:1)
    at new StreamMessage (turbo.es2017-esm.js:1008:1)
    at Function.wrap (turbo.es2017-esm.js:1015:1)
    at StreamObserver.receiveMessageHTML (turbo.es2017-esm.js:4248:1)
    at StreamObserver.receiveMessageEvent (turbo.es2017-esm.js:4168:1)
    at TurboCableStreamSourceElement.dispatchMessageEvent (cable_stream_source_element.js:99:1)
    at subscriptions.js:128:1
    at Array.map (<anonymous>)
```

It seems that it is caused by the changes made in https://github.com/hotwired/turbo/pull/660 because now [here](https://github.com/hotwired/turbo/blob/main/src/core/streams/stream_message.ts#L25) we do access the `.templateElement` which does [raise an error](https://github.com/hotwired/turbo/blob/main/src/elements/stream_element.ts#L111-L116) if the child isn't a `template` element, but as we know the `remove` action doesn't require a child element https://turbo.hotwired.dev/reference/streams#remove

dhh:
Solved by #665

PR #665, opened 2022-08-04
Make `StreamElement.templateElement` more flexible

seanpdoyle:
Prior to this commit, the `StreamElement.templateElement` property would
raise an error if it were called without a `<template>` descendant.

However, a `<turbo-stream action="remove">` element without a descendant
`<template>` element is valid and supported.

When called on a `<turbo-stream>` that has a `firstElementChild` that
_is not_ a `<template>` element will continue to raise an exception.

When called without _any_ child elements,
`StreamElement.templateElement` will create and append a `<template>`
element.

Since listeners to the `turbo:before-stream-render` event now have
access to the `StreamElement` creating, appending, and returning a
`<template>` element will provide them with an opportunity to manipulate
the element from their event listener if they choose to.

### DIFF
--- a/src/elements/stream_element.ts
+++ b/src/elements/stream_element.ts
@@ -109,7 +109,11 @@ export class StreamElement extends HTMLElement {
    * Gets the main `<template>` used for rendering
    */
   get templateElement() {
-    if (this.firstElementChild instanceof HTMLTemplateElement) {
+    if (this.firstElementChild === null) {
+      const template = this.ownerDocument.createElement("template")
+      this.appendChild(template)
+      return template
+    } else if (this.firstElementChild instanceof HTMLTemplateElement) {
       return this.firstElementChild
     }
     this.raise("first child element must be a <template> element")
--- a/src/tests/functional/stream_tests.ts
+++ b/src/tests/functional/stream_tests.ts
@@ -40,6 +40,16 @@ test("test receiving a stream message with css selector target", async ({ page }
   assert.deepEqual(await messages3.allTextContents(), ["Third", "Hello CSS!"])
 })
 
+test("test receiving a message without a template", async ({ page }) => {
+  await page.evaluate(() =>
+    window.Turbo.renderStreamMessage(`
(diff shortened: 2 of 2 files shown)


davekaro:
This fixes https://github.com/hotwired/turbo/issues/664

marcelolx (approved):
You rock @seanpdoyle!

excid3:
Just ran into this as well. Any chance we could get a beta3 release with this?

seanpdoyle:
In the short-term, are you able to target https://github.com/hotwired/dev-builds/releases/tag/%40hotwired%2Fturbo%2Flatest or https://github.com/hotwired/dev-builds/releases/tag/%40hotwired%2Fturbo%2Fc4e0aba in your `package.json`?

excid3:
Sure can. 👍

marcelolx:
@excid3 if you're using turbo-rails see this message https://discord.com/channels/988103760050012160/988104775079968849/1006524274540023808

⬇️ (if you don't want to open discord)

> For those that want to test the beta 2 but are blocked because of this problem and may be waiting for a release (and depend on turbo-rails), we have a branch with Turbo updated to the latest release that is available in the dev-builds repo

> https://github.com/marcelolx/turbo-rails/tree/%40hotwired/turbo/c4e0aba

> Change your package.json to (if node env) 
> "@hotwired/turbo-rails": "marcelolx/turbo-rails.git#@hotwired\/turbo\/c4e0aba"

> And your Gemfile to 
> gem "turbo-rails", github: "marcelolx/turbo-rails", branch: "@hotwired/turbo/c4e0aba"

excid3:
Awesome, thanks @marcelolx 👍

rbclark:
Just as a note, if anyone is using importmaps and wants to target a version that includes this fix, I've done so using 
`https://cdn.jsdelivr.net/gh/hotwired/dev-builds@35faf826645b4f57f62f8386904a0bc8d6969dc6/dist/turbo.es2017-esm.js`

It turns out that targeting files directly on Github doesn't work however jdselivr.net provides a proxy.

PR #666, opened 2022-08-04
Preserve input values in cache

seanpdoyle:
Re-submission of https://github.com/hotwired/turbo/pull/70

This change will save the values of `<input>`s, `<textarea>`s, and
`<select>`s when creating a snapshot so that when you navigate back or
forward in history, form fields will not be reset.

This behavior was already present prior to [5dfc79e][] for `<input>`s
and `<textarea>`s but not `<select>`s (see
[turbolinks/turbolinks#238][]).

[5dfc79e]: https://github.com/hotwired/turbo/commit/5dfc79ee2feba441ea9c9b486009d08336e78d76
[turbolinks/turbolinks#238]: https://github.com/turbolinks/turbolinks/issues/238

### DIFF
--- a/src/core/drive/page_snapshot.ts
+++ b/src/core/drive/page_snapshot.ts
@@ -24,7 +24,22 @@ export class PageSnapshot extends Snapshot<HTMLBodyElement> {
   }
 
   clone() {
-    return new PageSnapshot(this.element.cloneNode(true), this.headSnapshot)
+    const clonedElement = this.element.cloneNode(true)
+
+    const selectElements = this.element.querySelectorAll("select")
+    const clonedSelectElements = clonedElement.querySelectorAll("select")
+
+    for (const [index, source] of selectElements.entries()) {
+      for (const [optionIndex, option] of Array.from(source.options).entries()) {
+        clonedSelectElements[index].options[optionIndex].selected = option.selected
+      }
+    }
+
+    for (const clonedPasswordInput of clonedElement.querySelectorAll<HTMLInputElement>('input[type="password"]')) {
+      clonedPasswordInput.value = ""
+    }
+
+    return new PageSnapshot(clonedElement, this.headSnapshot)
   }
 
   get headElement() {
--- a/src/tests/fixtures/rendering.html
+++ b/src/tests/fixtures/rendering.html
@@ -44,6 +44,24 @@
       <p><a id="permanent-in-frame-element-link" href="/src/tests/fixtures/permanent_element.html" data-turbo-frame="frame">Permanent element in frame</a></p>
       <p><a id="permanent-in-frame-without-layout-element-link" href="/src/tests/fixtures/frames/without_layout.html" data-turbo-frame="frame">Permanent element in frame without layout</a></p>
       <p><a id="delayed-link" href="/__turbo/delayed_response">Delayed link</a></p>
+      <form>
+        <input type="text" id="text-input">
+        <input type="radio" id="radio-input">
+        <input type="checkbox" id="checkbox-input">
+        <textarea id="textarea"></textarea>
+        <select id="select">
+          <option value="1">1</option>
+          <option value="2">2</option>
+        </select>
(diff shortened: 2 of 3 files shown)


src/core/drive/page_snapshot.ts
```diff
@@ -24,7 +24,27 @@ export class PageSnapshot extends Snapshot<HTMLBodyElement> {
...
+      clonedSingleSelects[i].value = singleSelects[i].value
+    }
+
+    const multipleSelects = this.element.querySelectorAll<HTMLSelectElement>("select[multiple]")
+    const clonedMultipleSelects = clonedElement.querySelectorAll<HTMLSelectElement>("select[multiple]")
+    for (let i = 0; i < multipleSelects.length; i++) {
+      for (let j = 0; j < multipleSelects[i].options.length; j++) {
+        clonedMultipleSelects[i].options[j].selected = multipleSelects[i].options[j].selected
+      }
+    }
```
marcoroth, line 41:
```suggestion
    multipleSelects.forEach((mulipleSelect, i) => {
      Array.from(multipleSelect.options).forEach(option, j) => {
        clonedMultipleSelects[i].options[j].selected = option.selected
      })
    }
```

src/core/drive/page_snapshot.ts
```diff
@@ -24,7 +24,27 @@ export class PageSnapshot extends Snapshot<HTMLBodyElement> {
...
 
   clone() {
-    return new PageSnapshot(this.element.cloneNode(true), this.headSnapshot)
+    const clonedElement = this.element.cloneNode(true)
+
+    const singleSelects = this.element.querySelectorAll<HTMLSelectElement>("select:not([multiple])")
+    const clonedSingleSelects = clonedElement.querySelectorAll<HTMLSelectElement>("select:not([multiple])")
+    for (let i = 0; i < singleSelects.length; i++) {
+      clonedSingleSelects[i].value = singleSelects[i].value
+    }
```
marcoroth, line 33:
```suggestion
    singleSelects.forEach((select, i) => {
      clonedSingleSelects[i].value = select.value
    })
```

marcoroth (commented):
Personally, I feel like this is more readable

seanpdoyle:
Thank you for the review @marcoroth. I've re-structured the `<select>` cloning to avoid doubly index-based iteration _and_ [Array.from(...).forEach](https://github.com/github/eslint-plugin-github/blob/main/docs/rules/array-foreach.md#array-foreach).

marcoroth:
Beautiful, that's even better!

src/core/drive/page_snapshot.ts
```diff
@@ -24,7 +24,22 @@ export class PageSnapshot extends Snapshot<HTMLBodyElement> {
...
+    const clonedElement = this.element.cloneNode(true)
+
+    const selectElements = this.element.querySelectorAll("select")
+    const clonedSelectElements = clonedElement.querySelectorAll("select")
+
+    for (const [index, source] of selectElements.entries()) {
+      for (const [optionIndex, option] of Array.from(source.options).entries()) {
+        clonedSelectElements[index].options[optionIndex].selected = option.selected
+      }
+    }
```
tleish, line 36: FYI, the following approach is around 3x to 4x faster.  Not an issue for small selects, but if the page contains large selects (e.g. Country Selector) it adds up.

```javascript
for (const [index, source] of selectElements.entries()) {
  [...clonedSelectElements[index].selectedOptions].forEach(option => { option.selected = false });
  [...source.selectedOptions].forEach(option => clonedSelectElements[index].options[option.index].selected = true)
}
```

seanpdoyle: Would it be even faster if the `[...]` syntax were replaced with a `for ... of` loop? I can open a follow-up PR.

seanpdoyle: I've opened https://github.com/hotwired/turbo/pull/669. Thank you!

PR #672, opened 2022-08-08
`turbo:frame-missing`: Re-use `FetchResponse` HTML

seanpdoyle:
After re-considering a comment on [hotwired/turbo#445][], this commit
re-purposes the response HTML from the `FetchResponse` instance, instead
of issuing a follow-up HTTP request.

[hotwired/turbo#445]: https://github.com/hotwired/turbo/pull/445#discussion_r934776811

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -309,9 +309,13 @@ export class Session
     this.notifyApplicationAfterFrameRender(fetchResponse, frame)
   }
 
-  frameMissing(frame: FrameElement, fetchResponse: FetchResponse): Promise<void> {
-    console.warn(`Completing full-page visit as matching frame for #${frame.id} was missing from the response`)
-    return this.visit(fetchResponse.location)
+  async frameMissing(frame: FrameElement, fetchResponse: FetchResponse): Promise<void> {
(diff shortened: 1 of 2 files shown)


dhh:
Nice. This will save that extra request 👍

willcosgrove:
Doesn't the response not include a layout? I know there was talk of always including the layout for frame requests, but it sounds like that was dismissed (https://github.com/hotwired/turbo-rails/pull/232#issuecomment-925620618).

I'm experiencing an issue where I have a form contained within a Turbo Frame, and on submit I want to redirect to a different page that does not contain a matching frame. What happens is:

1. The frame handles the form submit
2. It gets a redirect response
3. It follows the redirect (still including the Turbo-Frame header)
4. The document fetched as a result of the redirect has no matching frame, and no layout
5. This response document replaces the body because it is being promoted to a full page visit. But because it has no layout, it breaks the page.

It seems like the two options to fix this are either to 1. remove the `layout: false` optimization for frame requests, or 2. issue two requests in this specific scenario. The first one follows the redirect with the Turbo-Frame header, sees it has no matching frame, and then performs a visit to the same URL but this time with no Turbo-Frame header so that the full layout is included.

PR #673, opened 2022-08-08
Add `[data-turbo-enforced]` to disable `data-turbo=false`

ghiculescu:
Developers can use [`data-turbo=false`](https://turbo.hotwired.dev/handbook/drive#disabling-turbo-drive-on-specific-links-or-forms) to disable Turbo Drive on specific links or form elements. But this introduces a problem in Turbo Native apps, where using `data-turbo=false` will exit the app and open the link in a web browser.

One of the best things about Turbo is that you can build features that will probably work natively without ever having to test natively. But `data-turbo=false` introduces an easy for developers to break this pact. A recent example I noticed in code review was a developer using `data-turbo=false` on a link to delete a record, because that would cause a full page reload and thus show an empty state again. (A better approach would be to use a Turbo Stream, or the CSS pattern described [here](https://boringrails.com/articles/css-tips-and-tricks-hotwire/).) On the native app, this link switched over the browser, so it didn't work at all.

In https://github.com/hotwired/turbo-site/pull/104 I added a warning about this, but warnings require people to read docs. So in this PR I'm proposing a way of controlling this using Turbo. This example shows what I propose:

```html
<div data-turbo="false">
  <a href="foo.html">Internal link. Full page navigation in browser; exits app in native.</a>
  <a href="https://www.example.com">External link. Full page navigation in browser; exits app in native.</a>
</div>

<div data-turbo-enforced>
  <div data-turbo="false">
    <a href="foo.html">Internal link. Link will be followed using Turbo! (data-turbo="false" is ignored)</a>
    <a href="https://www.example.com">External link. Full page navigation in browser; exits app in native.</a>
  </div>
</div>
```

The most common usage would be adding `data-turbo-enforced` to the `<body>` ([or relevant element](https://github.com/hotwired/turbo/pull/627)), either always or just when rendering in a native app (via user agent). I'd suggest doing it always so that you get the same behaviour in browser development as you do in app production.

### DIFF
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -407,12 +407,15 @@ export class Session
 
   elementIsNavigatable(element: Element): boolean {
     const container = element.closest("[data-turbo]")
-    const withinFrame = element.closest("turbo-frame")
 
     // Check if Drive is enabled on the session or we're within a Frame.
-    if (this.drive || withinFrame) {
-      // Element is navigatable by default, unless `data-turbo="false"`.
-      if (container) {
+    if (this.drive || element.closest("turbo-frame")) {
+      // Element is navigatable by default
+      // Use `data-turbo="false"` to disable navigation; add `data-turbo-enforce` to a parent to force Turbo navigation.
+      const enforcedByParent = element.closest("[data-turbo-enforce]")
+      if (enforcedByParent) {
+        return true
+      } else if (container) {
         return container.getAttribute("data-turbo") != "false"
       } else {
         return true
--- a/src/tests/fixtures/drive_enforced.html
+++ b/src/tests/fixtures/drive_enforced.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Drive</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+  </head>
+  <body>
+    <h1>Drive with <code>data-turbo-enforce</code></h1>
+
+    <div data-turbo-enforce>
+      <a id="drive_enabled" href="/src/tests/fixtures/drive_enforced.html">Drive enabled link</a>
+      <a id="drive_enabled_external" href="https://example.com">Drive enabled external link</a>
+      
+      <div data-turbo="false">
+        <a id="drive_disabled" href="/src/tests/fixtures/drive_enforced.html">Drive disabled link</a>
+      </div>
+    </div>
+  </body>
+</html>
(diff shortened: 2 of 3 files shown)


seanpdoyle:
Thank you for opening this issue, and for providing a diff to review!

At the moment, if a developer wanted to configure their instance of Turbo, there are currently:

* [data-* attributes](https://turbo.hotwired.dev/reference/attributes#data-attributes) declared on particular elements (including `[data-turbo="true"]` and `[data-turbo="false"]`, like you mentioned)
* [`<meta>` elements declared in the document `<head>`](https://turbo.hotwired.dev/reference/attributes#meta-tags)
* [`Turbo.session.drive`](https://turbo.hotwired.dev/handbook/drive#disabling-turbo-drive-on-specific-links-or-forms) boolean property (assigned in JavaScript)
* [`Turbo.setFormMode`](https://github.com/hotwired/turbo/pull/419) to opt-into or opt-out of support for `<form>` submissions to be combined with `Turbo.session.drive = true` (assigned in JavaScript)
* events like [`turbo:click`](https://turbo.hotwired.dev/reference/events) to cancel to fallback to default browser behavior

I don't quite understand the ins and outs of the case you've outlined, so I'm excited to see where this discussion takes us, but I want to highlight that sprawling options that have evolved over time.

I wonder if there's some synthesis to be done to unify these competing options into something more cohesive.

In this case, could Turbo attach a mobile-only `turbo:click` listener, and remove all `[data-turbo="false"]` attributes from its ancestors at click-time, instead of adding another `[data-*]` with its own semantics and combinatorial explosions?

Could the existing `Turbo.session.drive` and `Turbo.setFormMode` configuration seams be combined to limit the amount of differing semantics, or maybe incorporate a web-only or mobile-only value?

Is there a `<meta>` element that could be expanded to incorporate a mobile-only request header or Action Dispatch variant?

I don't have all (or any) of the answers, but I want to call attention to the fact that all of these competing options have potential to complicate documentation, interplay off one another in the internal implementation, and need to be accounted for in our growing test suite.

ghiculescu:
Yeah, that's a good point. There are a _lot_ of options.

The reason I didn't use `Turbo.session.something` for this one is because I think there's value in being able to apply this enforcement only to specific DOM elements. Same goes for the click handler suggestion. But I can compromise on that if needed.

ryankopf:
Why not instead:
data-turbo="false-except-native"
Or something similar but less wordy.

brunoprietog:
Hi @ghiculescu, thanks for your proposal, but I'm closing this for now. We don't plan to introduce a new API considering the options we already have.

ghiculescu:
Is this your suggested approach @brunoprietog ?

> In this case, could [turbo-ios](https://github.com/hotwired/turbo-ios/blob/main/Source/WebView/turbo.js) attach a mobile-only `turbo:click` listener, and remove all `[data-turbo="false"]` attributes from its ancestors at click-time, instead of adding another `[data-*]` with its own semantics and combinatorial explosions?

brunoprietog:
Yes, you can listen to the `turbo:click` event and cancel it when necessary. That's the equivalent of not using Turbo on that link.

PR #677, opened 2022-08-10
`turbo:frame-missing`: Do not `Turbo.visit` by default

seanpdoyle:
Restore the existing default behavior when a matching frame is missing
from a response: log an error and blank the frame.

Alongside that original behavior, yield the [Response][] instance and a
`Turbo.visit`-like callback that can transform the `Response` instance
into a `Visit`. It also accepts all the arguments that the `Turbo.visit`
call normally does.

[Response]: https://developer.mozilla.org/en-US/docs/Web/API/Response

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -20,7 +20,7 @@ import {
 import { FormSubmission, FormSubmissionDelegate } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
 import { ViewDelegate, ViewRenderOptions } from "../view"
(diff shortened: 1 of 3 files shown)


jon-sully:
Curious why you want to roll this default behavior back, Sean? I know consuming apps could catch the event and do the visit themselves manually (e.g. this could all still work the same way very easily after this PR) but having it as a default feels easier. Looks like there was a little chatter in [these comments](https://github.com/hotwired/turbo/pull/445#discussion_r937152425) with you and DHH but curious for your thoughts as always 🙂 cheers!

dhh:
I remain confident that the fallback to visit on missing frame is the better default. We allow you to catch the frame missing event, so you can always override that default if you see fit.

PR #679, opened 2022-08-12
Stream responses to GET should clear busy state

kevinmcconnell:
We set the `aria-busy` state on the document element whenever a link is
clicked that initiates a visit. After the visit completes, we clear this
state.

However, Turbo Stream responses are not visits, and so in the case where
the server chooses to respond with a stream response that busy state
wasn't being cleared.

To fix this, we'll now also clear the busy state when a stream response
is received in response to an HTTP request.

### DIFF
--- a/src/observers/stream_observer.ts
+++ b/src/observers/stream_observer.ts
@@ -3,8 +3,10 @@ import { FetchResponse } from "../http/fetch_response"
 import { StreamMessage } from "../core/streams/stream_message"
 import { StreamSource } from "../core/types"
 
+export type StreamDelivery = "http" | "event"
+
 export interface StreamObserverDelegate {
-  receivedMessageFromStream(message: StreamMessage): void
+  receivedMessageFromStream(message: StreamMessage, delivery: StreamDelivery): void
 }
 
 export class StreamObserver {
@@ -58,19 +60,19 @@ export class StreamObserver {
 
   receiveMessageEvent = (event: MessageEvent) => {
     if (this.started && typeof event.data == "string") {
-      this.receiveMessageHTML(event.data)
+      this.receiveMessageHTML(event.data, "event")
     }
   }
 
   async receiveMessageResponse(response: FetchResponse) {
     const html = await response.responseHTML
     if (html) {
-      this.receiveMessageHTML(html)
+      this.receiveMessageHTML(html, "http")
     }
   }
 
-  receiveMessageHTML(html: string) {
-    this.delegate.receivedMessageFromStream(StreamMessage.wrap(html))
+  receiveMessageHTML(html: string, delivery: StreamDelivery) {
+    this.delegate.receivedMessageFromStream(StreamMessage.wrap(html), delivery)
   }
 }
 
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -11,7 +11,7 @@ import { Navigator, NavigatorDelegate } from "./drive/navigator"
 import { PageObserver, PageObserverDelegate } from "../observers/page_observer"
 import { ScrollObserver } from "../observers/scroll_observer"
 import { StreamMessage } from "./streams/stream_message"
-import { StreamObserver } from "../observers/stream_observer"
+import { StreamObserver, StreamDelivery } from "../observers/stream_observer"
 import { Action, Position, StreamSource, isAction } from "./types"
 import { clearBusyState, dispatch, markAsBusy } from "../util"
 import { PageView, PageViewDelegate, PageViewRenderOptions } from "./drive/page_view"
@@ -260,7 +260,10 @@ export class Session
 
   // Stream observer delegate
 
-  receivedMessageFromStream(message: StreamMessage) {
+  receivedMessageFromStream(message: StreamMessage, delivery: StreamDelivery) {
+    if (delivery === "http") {
+      clearBusyState(document.documentElement)
+    }
     this.renderStreamMessage(message)
   }
 
--- a/src/tests/fixtures/visit.html
+++ b/src/tests/fixtures/visit.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html id="html">
   <head>
     <meta charset="utf-8">
     <title>Turbo</title>
@@ -19,7 +19,9 @@
       <hr class="push-below-fold">
       <p><a id="below-the-fold-link" href="/src/tests/fixtures/one.html">one.html</a></p>
       <hr class="push-below-fold">
-      <p><a id="stream-link" href="/src/tests/fixtures/one.html?key=value" data-turbo-stream>Stream link with ?key=value</a></p>
+      <p><a id="stream-link" href="/__turbo/stream-response?content=link&type=stream&key=value" data-turbo-stream>Stream link with ?key=value</a></p>
     </section>
+
+    <div id="messages"></div>
   </body>
 </html>
--- a/src/tests/functional/visit_tests.ts
+++ b/src/tests/functional/visit_tests.ts
@@ -5,6 +5,7 @@ import {
   getSearchParam,
   isScrolledToSelector,
   isScrolledToTop,
+  nextAttributeMutationNamed,
   nextBeat,
   nextEventNamed,
   readEventLogs,
@@ -185,6 +186,21 @@ test("test visits with data-turbo-stream include MIME type & search params", asy
 
   assert.ok(fetchOptions.headers["Accept"].includes("text/vnd.turbo-stream.html"))
   assert.equal(getSearchParam(url, "key"), "value")
+})
+
+test("test visits with data-turbo-stream set and clear aria-busy", async ({ page }) => {
+  await page.click("#stream-link")
+
+  assert.equal(
+    await nextAttributeMutationNamed(page, "html", "aria-busy"),
+    "true",
+    "sets [aria-busy] on the document element"
+  )
+  assert.equal(
+    await nextAttributeMutationNamed(page, "html", "aria-busy"),
+    null,
+    "removes [aria-busy] from the document element"
+  )
 })
 
 test("test cache does not override response after redirect", async ({ page }) => {
--- a/src/tests/server.ts
+++ b/src/tests/server.ts
@@ -86,6 +86,17 @@ router.post("/messages", (request, response) => {
   }
 })
 
+router.get("/stream-response", (request, response) => {
+  const params = { ...request.body, ...request.query }
+  const { content, target, targets } = params
+  if (acceptsStreams(request)) {
+    response.type("text/vnd.turbo-stream.html; charset=utf-8")
+    response.send(targets ? renderMessageForTargets(content, targets) : renderMessage(content, target))
+  } else {
+    response.sendStatus(422)
+  }
+})
+
 router.put("/messages/:id", (request, response) => {
   const { content, type } = request.body
   const { id } = request.params


src/observers/stream_observer.ts
```diff
@@ -58,19 +60,19 @@ export class StreamObserver {
 
   receiveMessageEvent = (event: MessageEvent) => {
     if (this.started && typeof event.data == "string") {
-      this.receiveMessageHTML(event.data)
+      this.receiveMessageHTML(event.data, "event")
```
seanpdoyle, line 63: As an alternative to making the Streams aware of their origin, how do you feel about adding or extending the `Session` behavior for dispatching `Request` instances or receiving `Response` instances?

Something like a `BusyStateObserver` that listens for `turbo:before-fetch-request` and `turbo:before-fetch-response` events. The `Session` could be the delegate for one instance, and our existing code for Form and Frame `[aria-busy]` toggling could be abstracted into their own instances.

kevinmcconnell: Yeah, I like that idea. I'll give that a shot and see how it looks 👍

seanpdoyle: @kevinmcconnell after suggesting that and sitting with the idea for a while, I've realized that the current behavior might be the most correct.

When a Frame is navigating, it's marked as `[aria-busy="true"]`. When a `<form>` is submitting, it's marked as `[aria-busy="true"]`. When the entire page is redirected, the `<body>` is marked as `[aria-busy="true"]`.

For a Turbo Stream request that's targeted at a small fragment of the document, nesting the entire page within an `[aria-busy="true"]` element would pose accessibility issues, since it'd communicate to assistive technologies that the full document is in a state of flux, which isn't the case.

With that in mind, I think we're better off with the current behavior. Am I misunderstanding your intent?

seanpdoyle: Ok, I'm sorry to misunderstand _again_. I think a more correct resolution to this issue would be to _prevent_ the document from ever setting `[aria-busy="true"]` in the first place, instead of finding a way to properly clear that attribute.

Does that sound like a reasonable alternative?

seanpdoyle:
@kevinmcconnell I've experimented with a solution on my own. What do you think about something like:

```diff
diff --git a/src/core/session.ts b/src/core/session.ts
index b6c1940..12c6274 100644
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -210,6 +210,9 @@ export class Session
   }
 
   visitStarted(visit: Visit) {
+    if (!visit.acceptsStreamResponse) {
+      markAsBusy(document.documentElement)
+    }
     extendURLWithDeprecatedProperties(visit.location)
     if (!visit.silent) {
       this.notifyApplicationAfterVisitingLocation(visit.location, visit.action)
@@ -217,6 +220,7 @@ export class Session
   }
 
   visitCompleted(visit: Visit) {
+    clearBusyState(document.documentElement)
     this.notifyApplicationAfterPageLoad(visit.getTimingMetrics())
   }
 
@@ -346,7 +350,6 @@ export class Session
   }
 
   notifyApplicationAfterVisitingLocation(location: URL, action: Action) {
-    markAsBusy(document.documentElement)
     return dispatch<TurboVisitEvent>("turbo:visit", { detail: { url: location.href, action } })
   }
 
@@ -366,7 +369,6 @@ export class Session
   }
 
   notifyApplicationAfterPageLoad(timing: TimingData = {}) {
-    clearBusyState(document.documentElement)
     return dispatch<TurboLoadEvent>("turbo:load", {
       detail: { url: this.location.href, timing },
     })
diff --git a/src/tests/fixtures/visit.html b/src/tests/fixtures/visit.html
index d138fd3..cbeae69 100644
--- a/src/tests/fixtures/visit.html
+++ b/src/tests/fixtures/visit.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html id="html">
   <head>
     <meta charset="utf-8">
     <title>Turbo</title>
@@ -19,7 +19,9 @@
       <hr class="push-below-fold">
       <p><a id="below-the-fold-link" href="/src/tests/fixtures/one.html">one.html</a></p>
       <hr class="push-below-fold">
-      <p><a id="stream-link" href="/src/tests/fixtures/one.html?key=value" data-turbo-stream>Stream link with ?key=value</a></p>
+      <p><a id="stream-link" href="/__turbo/stream-response?content=link&type=stream&key=value" data-turbo-stream>Stream link with ?key=value</a></p>
     </section>
+
+    <div id="messages"></div>
   </body>
 </html>
diff --git a/src/tests/functional/visit_tests.ts b/src/tests/functional/visit_tests.ts
index e03fbc6..995e640 100644
--- a/src/tests/functional/visit_tests.ts
+++ b/src/tests/functional/visit_tests.ts
@@ -5,6 +5,7 @@ import {
   getSearchParam,
   isScrolledToSelector,
   isScrolledToTop,
+  noNextAttributeMutationNamed,
   nextBeat,
   nextEventNamed,
   readEventLogs,
@@ -187,6 +188,12 @@ test("test visits with data-turbo-stream include MIME type & search params", asy
   assert.equal(getSearchParam(url, "key"), "value")
 })
 
+test("test visits with data-turbo-stream set and clear aria-busy", async ({ page }) => {
+  await page.click("#stream-link")
+
+  assert.ok(await noNextAttributeMutationNamed(page, "html", "aria-busy"), "never sets [aria-busy] on the document element")
+})
+
 test("test cache does not override response after redirect", async ({ page }) => {
   await page.evaluate(() => {
     const cachedElement = document.createElement("some-cached-element")
diff --git a/src/tests/helpers/page.ts b/src/tests/helpers/page.ts
index 347e3eb..498ab5d 100644
--- a/src/tests/helpers/page.ts
+++ b/src/tests/helpers/page.ts
@@ -98,6 +98,15 @@ export async function nextAttributeMutationNamed(
   return attributeValue
 }
 
+export async function noNextAttributeMutationNamed(
+  page: Page,
+  elementId: string,
+  attributeName: string
+): Promise<boolean> {
+  const records = await readMutationLogs(page, 1)
+  return !records.some(([name]) => name == attributeName)
+}
+
 export async function noNextEventNamed(page: Page, eventName: string): Promise<boolean> {
   const records = await readEventLogs(page, 1)
   return !records.some(([name]) => name == eventName)
diff --git a/src/tests/server.ts b/src/tests/server.ts
index 49a292e..46610e0 100644
--- a/src/tests/server.ts
+++ b/src/tests/server.ts
@@ -86,6 +86,17 @@ router.post("/messages", (request, response) => {
   }
 })
 
+router.get("/stream-response", (request, response) => {
+  const params = { ...request.body, ...request.query }
+  const { content, target, targets } = params
+  if (acceptsStreams(request)) {
+    response.type("text/vnd.turbo-stream.html; charset=utf-8")
+    response.send(targets ? renderMessageForTargets(content, targets) : renderMessage(content, target))
+  } else {
+    response.sendStatus(422)
+  }
+})
+
 router.put("/messages/:id", (request, response) => {
   const { content, type } = request.body
   const { id } = request.params
```

src/observers/stream_observer.ts
kevinmcconnell: We can certainly avoid setting the `aria-busy` attribute in this case, if that's preferred. I actually wondered about that initially, but I didn't go that route for a couple of reasons:

1. If you have a link with `data-turbo-stream`, it's still valid to get an HTML response. In which case the behaviour is a regular visit, and so it might be expected that the busy state behaviour matches that of a regular visit.
2. It can still be useful to show a busy state even if you're updating part of the page (as we do with frame, forms, etc). When the link is outside of a frame or form it's harder to know what the scope of that change is going to be, so setting the attribute on the document element seems like a reasonable thing to do.

I agree with your point that the attribute is communicating that the document is in a state of flux, but without having a defined boundary of the part that's changing, then I'm not sure where better to indicate this.

If we want to say that `data-turbo-stream` doesn't set busy state because of this, I think that's OK, but it might leave people without such an easy way to show a loading spinner or similar during that request.

kevinmcconnell:
@seanpdoyle your solution looks good if we want to take the approach of not setting the attribute for `data-turbo-stream` links. I'm happy to see it solved either way, so if you feel that's better, I'll update this PR to use that.

src/observers/stream_observer.ts
seanpdoyle: > When the link is outside of a frame or form it's harder to know what the scope of that change is going to be, so setting the attribute on the document element seems like a reasonable thing to do.

In the context of full page navigations, marking the entire document as busy is reasonable, since all content is likely to be swapped out.

From the specification for [aria-busy](https://www.w3.org/TR/wai-aria/#aria-busy):

> Indicates an element is being modified and that **assistive technologies MAY want to wait until the modifications are complete before exposing them to the user**.

From the section on [Live Regions](https://www.w3.org/TR/wai-aria/#dfn-live-region):

> Since these asynchronous areas are expected to update outside the user's area of focus, assistive technologies such as screen readers have either been unaware of their existence or unable to process them for the user. WAI-ARIA has provided a collection of properties that allow the author to identify these live regions and process them: aria-live, aria-relevant, aria-atomic, and aria-busy.

From the section on [Required Owned Elements]()

> When a widget is missing required owned elements due to script execution or loading, authors MUST mark a containing element with [aria-busy](https://www.w3.org/TR/wai-aria/#aria-busy) equal to true. For example, **until a page is fully initialized and complete, an author could mark the document element as busy**.

Marking the entire document as busy would make its descendant content inaccessible to assistive tech, which has the potential to be very jarring unless a Turbo Stream response targets the entire `<body>` element.

> If you have a link with `data-turbo-stream`, it's still valid to get an HTML response. In which case the behaviour is a regular visit, and so it might be expected that the busy state behaviour matches that of a regular visit.

This is certainly an issue.

When considering `<form>` submissions that are Stream-enabled, the operation is scoped to the `<form>` element itself, which sets `[aria-busy=]` throughout the lifecycle of the request. It doesn't face the same uncertainty, since a submission response is either `2xx` or `4xx` (operates on the current page) or a `3xx` redirect, which is _for certain_ a `Visit`. The `Visit` lifecycle kicks off, and its activity bubbles up throughout the document, and is reflected in the `html[aria-busy]` attribute.

Given the nature of `<a data-turbo-stream>` elements, that ambiguity is baked-in.

I've historically been strongly opposed to `<a data-turbo-stream>`, so that bias might be obscuring me from a solution for that ambiguity. With that being said, I am hopeful that there is a solution, and am happy to help implement if we're able to discover it.

kevinmcconnell:
Closing this is favour of #681.

src/observers/stream_observer.ts
kevinmcconnell: Yes, I agree it would be good to find a solution. Appreciate your feedback on this. (And I also appreciate we may have different opinions on whether `<a data-turbo-stream>` is a good idea ;))

One idea could be to allow marking a boundary where a busy state should apply (and default to not setting `aria-busy` if no such boundary is present). That way people could define behaviour that's appropriate for their use case, without much effort.

I just started looking at `aria-live` to see if that would be a valid way to approach this. Something like: when requesting a stream update from within a Live Region we'd mark that region as busy. Otherwise we'd fall back to current behaviour, only setting it on the `form`, `frame` etc where appropriate). But I'm not yet very familiar with `aria-live` so I don't know if that's actually a misguided approach.

seanpdoyle: I think the nature of `<turbo-steam>` responses makes any kind of "geographic" limitation on the client-side moot, since a response can target any part of the document, regardless of where the request was initiated from.

kevinmcconnell: I think that's just an issue with stream responses in general (not with `data-turbo-stream` specifically). Responses to form submissions can target areas outside the form (and might only do so); actions that are initiated from inside a `turbo-frame` can similarly update areas outside of that frame, etc. Anything can reach outside its own scope, and since `aria-busy` is communicating what's being modified, it seems like the only "safe" place to set it would be the document element. But as you point out, it may not always be desirable to do that.

I think that's why it could be useful to have a simple way for people to override the default behaviour, since the default behaviour won't always be right.

Another consideration is that `aria-busy` is really about what's being modified, but often what people want to show in their UI (and what they use the attribute for) is which operation is pending. e.g. if you have a button that causes another section of the UI to change, you may want to show some sort of spinner or pending state on the button, rather than on that part of the UI that will be mutated.

PR #684, opened 2022-08-13
Add `render` to `turbo:before-stream-render` event

seanpdoyle:
Follow-up to https://github.com/hotwired/turbo/pull/479
Related to https://github.com/hotwired/turbo-site/pull/107

As an alternative to exposing an otherwise private and internal
`StreamActions` "class", support custom `<turbo-stream action="...">`
values the same way as custom `<turbo-frame>` rendering or `<body>`
rendering: as part of its `turbo:before-render-stream` event.

For example, if Turbo were to rename `StreamActions` to something else,
it would need to continue to export `StreamActions` publicly for the
sake of backwards compatibility. By depending on an event listener
instead, client applications are insulated from any internal restructuring.

To change how Turbo renders the document during page rendering, client
applications declare a `turbo:before-render` event listener that
overrides its `CustomEvent.detail.render` function from its
[default][PageRenderer.renderElement].

Similarly, to change how Turbo renders a frame, client applications
declare a `turbo:before-frame-render` and override its
`CustomEvent.detail.render` function from its [default][FrameRenderer.renderElement].

This commit introduces the `StreamElement.renderElement` function, and
extends the existing `turbo:before-stream-render` event to support the
same pattern with `StreamElement.renderElement` serving as the default
`CustomEvent.detail.render` value.

With those changes in place, callers can declare a document-wide event
listener and override based on the value of `StreamElement.action`:

```javascript
const CustomActions = {
  customUpdate: (stream) => { /* ... */ },
  customReplace: (stream) => { /* ... */ },
  // ...
}

document.addEventListener("turbo:before-stream-render", (({ target, detail }) => {
  const defaultRender = detail.render

  detail.render = CustomActions[target.action] || defaultRender
}))
```

[PageRenderer.renderElement]: https://github.com/hotwired/turbo/blob/256418fee0178ee483d82cd9bb579bd5df5a151f/src/core/drive/page_renderer.ts#L7-L13
[FrameRenderer.renderElement]: https://github.com/hotwired/turbo/blob/256418fee0178ee483d82cd9bb579bd5df5a151f/src/core/frames/frame_renderer.ts#L13-L24

### DIFF
--- a/src/elements/stream_element.ts
+++ b/src/elements/stream_element.ts
@@ -1,7 +1,9 @@
 import { StreamActions } from "../core/streams/stream_actions"
 import { nextAnimationFrame } from "../util"
 
-export type TurboBeforeStreamRenderEvent = CustomEvent<{ newStream: StreamElement }>
+type Render = (currentElement: StreamElement) => Promise<void>
+
+export type TurboBeforeStreamRenderEvent = CustomEvent<{ newStream: StreamElement; render: Render }>
 
 // <turbo-stream action=replace target=id><template>...
 
@@ -26,6 +28,10 @@ export type TurboBeforeStreamRenderEvent = CustomEvent<{ newStream: StreamElemen
  *   </turbo-stream>
  */
 export class StreamElement extends HTMLElement {
+  static async renderElement(newElement: StreamElement): Promise<void> {
+    await newElement.performAction()
+  }
+
   async connectedCallback() {
     try {
       await this.render()
@@ -40,9 +46,11 @@ export class StreamElement extends HTMLElement {
 
   async render() {
     return (this.renderPromise ??= (async () => {
-      if (this.dispatchEvent(this.beforeRenderEvent)) {
+      const event = this.beforeRenderEvent
+
+      if (this.dispatchEvent(event)) {
         await nextAnimationFrame()
-        this.performAction()
+        await event.detail.render(this)
       }
     })())
   }
@@ -153,7 +161,7 @@ export class StreamElement extends HTMLElement {
     return new CustomEvent("turbo:before-stream-render", {
       bubbles: true,
       cancelable: true,
-      detail: { newStream: this },
+      detail: { newStream: this, render: StreamElement.renderElement },
     })
   }
 
--- a/src/tests/fixtures/stream.html
+++ b/src/tests/fixtures/stream.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html data-skip-event-details="turbo:submit-start turbo:submit-end">
   <head>
     <meta charset="utf-8">
     <title>Turbo Streams</title>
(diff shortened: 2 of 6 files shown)


marcoroth:
Nice, I like where this is heading! This also brings some consistency over which parts we have control over via event-listeners.

I like that you can intercept the rendering of a stream element, this allows the host application to intercept non-custom actions (like `update` or `replace`) but also custom actions introduced by third-party packages.

I can also see this feature useful when you conditionally want to "discard" a stream action or if you want to handle "the rendering" of certain actions by yourself. 

In order to properly support a new custom action (like a custom `morph` action via `morphdom`)  you also need to teach your server-side about it, which is not always desirable. 

With this feature however, you would be able to intercept the rendering of the default actions and **a)** use morphdom without the need for a new custom action and **b)** you don't need to have any specific logic on the server-side for when to use `morph` over `update/replace`. You can simplify delegate this decision to the client.

Using the same example from above: I can also see a use-case where you have a server-side part which emits  `update` / `replace` actions over which you don't have control over (like a part of the app which is served via a third-party Rails engine). With this feature you would be able to intercept and override the render function on the client-side without having to touch the emit-logic on the server-side. 

With all of that said, I still think the export of the `StreamActions` module is beneficial. It feels more straightforward to let Turbo know about all the available Custom Actions. If you just rely on event-listeners to "register" and support custom actions you are never letting Turbo (or other packages) know that such custom actions even exist. 

It feels more appropriate for packages to register new actions via `StreamActions`. Host applications can then intercept actions via the proposed event-listeners.

Versus the package needs to add an event-listener to override the render function in order to even "support and add" a new custom action.  If the host application then wants to intercept the action it would need to add another event listener and override the function again, which feels kinda hacky (not to the mention the uncertainty that the two event-listener need to fire in the right order).

---

**TL;DR:** In my opinion your proposed changes have neat use-cases on their own and or would be worth to add by themselves. In my opinion they don't interfere with the concept of defining Custom Actions via the exported `StreamActions` module, on the contrary, it even enhances them. 

If it's about the public API / leaking internals, we could also add some helper functions on the exported `Turbo` object instead.

hopsoft:
Agree that a `turbo:before-stream-render` event enhances the extensibility provided by `StreamActions`. The approaches aren't mutually exclusive. For example, I'd like to extend Turbo via `StreamActions` while allowing consumers of my extension(s) to change behavior via the event. 

Hoping to see both of these features make it into the library. 🤞🏻

dhh:
Agree with @marcoroth's assessment that this seems a nice AND but needn't be an OR. If all you want to do is add a custom action, the currently exposed approach seems more intuitive. But this then allows you to go even further. Happy to see the AND go in.

seanpdoyle:
I've pushed up changes to restore the `StreamActions` symbol's `export` declaration.

PR #685, opened 2022-08-16
Dispatch `turbo:fetch-request-error` during Visits

seanpdoyle:
Follow-up to https://github.com/hotwired/turbo/pull/640
Related to https://github.com/hotwired/turbo-site/pull/110

When a `Visit` results in a `fetch` error (like when the browser is
offline), dispatch a `turbo:fetch-request-error` event in the same style
as `<turbo-frame>`- and `<form>`-initiated `fetch` errors.

For the sake of consistency, also make the `TurboFetchRequestErrorEvent`
cancelable.

Along with the implementation change, this commit also adds test
coverage to ensure that the `Event.target` is correct for
`<turbo-frame>` and `<form>` error events.


srt32 (approved):
Yay!

PR #686, opened 2022-08-16
Turbo Streams: Manage element focus

seanpdoyle:
When a `<turbo-stream>` modifies the document, it has the potential to
affect which element has focus.

For example, consider an element with an `[id]` that has focus:

```html
<label for="an-input">
<input id="an-input" value="an invalid value"> <!-- this element has focus -->
```

Next, consider a `<turbo-stream>` element to replace it:

```html
<turbo-stream action="replace" target="an-input">
  <template>
    <input id="an-input" value="an invalid value" class="invalid-input">
  </template>
</turbo-stream>
```

Prior to this commit, rendering that `<turbo-stream>` would remove the
element with focus, and never restore it.

After this commit, the `Session` will capture the `[id]` value of the
element with focus (if there is any), then "restore" focus to an element
in the document with a matching `[id]` attribute _after_ the render.

Similarly, consider a `<turbo-stream>` that appends an element with
`[autofocus]`:

```html
<turbo-stream action="append" targets="body">
  <template>
    <input autofocus>
  </template>
</turbo-stream>
```

Prior to this commit, inserting an `[autofocus]` into the document with
a `<turbo-stream>` had no effect.

After this commit, the `Session` will scan any `<turbo-stream>` elements
its about to render, extracting the first focusable element that
declares an `[autofocus]` attribute.

Once the rendering is complete, it will attempt to autofocus that
element. Several scenarios will prevent that, including:

* there aren't any `[autofocus]` elements in the collection of
  `<turbo-stream>` elements
* the `[autofocus]` element does not exist in the document after the
  rendering is complete
* the document already has an element with focus

### DIFF
--- a/src/core/streams/stream_message_renderer.js
+++ b/src/core/streams/stream_message_renderer.js
@@ -1,11 +1,16 @@
 import { Bardo } from "../bardo"
 import { getPermanentElementById, queryPermanentElementsAll } from "../snapshot"
+import { around, elementIsFocusable, nextAnimationFrame, queryAutofocusableElement, uuid } from "../../util"
 
 export class StreamMessageRenderer {
   render({ fragment }) {
-    Bardo.preservingPermanentElements(this, getPermanentElementMapForFragment(fragment), () =>
-      document.documentElement.appendChild(fragment)
-    )
+    Bardo.preservingPermanentElements(this, getPermanentElementMapForFragment(fragment), () => {
+      withAutofocusFromFragment(fragment, () => {
+        withPreservedFocus(() => {
+          document.documentElement.appendChild(fragment)
+        })
+      })
+    })
   }
 
   // Bardo delegate
@@ -33,4 +38,61 @@ function getPermanentElementMapForFragment(fragment) {
   }
 
   return permanentElementMap
+}
+
+async function withAutofocusFromFragment(fragment, callback) {
+  const generatedID = `turbo-stream-autofocus-${uuid()}`
+  const turboStreams = fragment.querySelectorAll("turbo-stream")
+  const elementWithAutofocus = firstAutofocusableElementInStreams(turboStreams)
+  let willAutofocusId = null
+
+  if (elementWithAutofocus) {
+    if (elementWithAutofocus.id) {
+      willAutofocusId = elementWithAutofocus.id
+    } else {
+      willAutofocusId = generatedID
+    }
+
+    elementWithAutofocus.id = willAutofocusId
+  }
+
+  callback()
+  await nextAnimationFrame()
+
+  const hasNoActiveElement = document.activeElement == null || document.activeElement == document.body
+
+  if (hasNoActiveElement && willAutofocusId) {
+    const elementToAutofocus = document.getElementById(willAutofocusId)
+
+    if (elementIsFocusable(elementToAutofocus)) {
+      elementToAutofocus.focus()
+    }
+    if (elementToAutofocus && elementToAutofocus.id == generatedID) {
+      elementToAutofocus.removeAttribute("id")
+    }
+  }
+}
+
+async function withPreservedFocus(callback) {
+  const [activeElementBeforeRender, activeElementAfterRender] = await around(callback, () => document.activeElement)
+
+  const restoreFocusTo = activeElementBeforeRender && activeElementBeforeRender.id
+
+  if (restoreFocusTo) {
+    const elementToFocus = document.getElementById(restoreFocusTo)
+
+    if (elementIsFocusable(elementToFocus) && elementToFocus != activeElementAfterRender) {
+      elementToFocus.focus()
+    }
+  }
+}
+
+function firstAutofocusableElementInStreams(nodeListOfStreamElements) {
+  for (const streamElement of nodeListOfStreamElements) {
+    const elementWithAutofocus = queryAutofocusableElement(streamElement.templateElement.content)
+
+    if (elementWithAutofocus) return elementWithAutofocus
+  }
+
+  return null
 }
--- a/src/util.js
+++ b/src/util.js
@@ -185,3 +185,25 @@ export function findClosestRecursively(element, selector) {
     )
   }
 }
+
+export function elementIsFocusable(element) {
+  const inertDisabledOrHidden = "[inert], :disabled, [hidden], details:not([open]), dialog:not([open])"
+
+  return !!element && element.closest(inertDisabledOrHidden) == null && typeof element.focus == "function"
+}
+
+export function queryAutofocusableElement(elementOrDocumentFragment) {
+  return Array.from(elementOrDocumentFragment.querySelectorAll("[autofocus]")).find(elementIsFocusable)
+}
+
+export async function around(callback, reader) {
+  const before = reader()
+
+  callback()
+
+  await nextAnimationFrame()
+
+  const after = reader()
+
+  return [before, after]
+}
--- a/src/core/renderer.js
+++ b/src/core/renderer.js
@@ -41,7 +41,7 @@ export class Renderer {
 
   focusFirstAutofocusableElement() {
     const element = this.connectedSnapshot.firstAutofocusableElement
-    if (elementIsFocusable(element)) {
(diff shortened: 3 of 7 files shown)


seanpdoyle:
@kevinmcconnell if you're available, this is ready for review.

seanpdoyle:
@manuelpuyol @marcoroth if you're available and interested in this feature, I'd really appreciate some review!

marcoroth (approved):
+1 from me for this change.

The only thing to consider, which I think is currently not supported, is something like.

```html
<input class="input" value="a value"> <!-- this element has focus -->
```

and a Turbo Stream like:
```html
<turbo-stream action="replace" targets=".input">
  <template>
    <input class="input" value="a new value">
  </template>
</turbo-stream>
```

Unless I'm missing something.

seanpdoyle:
> +1 from me for this change.
> 
> 
> 
> The only thing to consider, which I think is currently not supported, is something like.
> 
> 
> 
> ```html
> 
> <input class="input" value="a value"> 
> 
> ```
> 
> 
> 
> and a Turbo Stream like:
> 
> ```html
> 
> <turbo-stream action="replace" targets=".input">
> 
>   <template>
> 
>     <input class="input" value="a new value">
> 
>   </template>
> 
> </turbo-stream>
> 
> ```
> 
> 
> 
> Unless I'm missing something.

I _think_ this accounts for that with the timestamp `[id]` generation, marking, and teardown.

seanpdoyle:
Re-reading the code, those time stamps only mark autofocus elements. Without some kind of css path matching, tracking focus across anonymous elements isn't supported.

marcoroth:
Doesn't the timestamp generation just account for the case in Turbo Streams where you have the `autofocus` attribute?

marcoroth:
> Re-reading the code, those time stamps only mark autofocus elements. Without some kind of css path matching, tracking focus across anonymous elements isn't supported.

Okay yeah, that makes sense. I guess it's fine to have the requirement for an id attribute to be present.

seanpdoyle:
> Okay yeah, that makes sense. I guess it's fine to have the requirement for an id attribute to be present.

👍

Just to clarify, tracking focus for anonymous elements isn't currently supported either.

brunoprietog:
Hi, why are you making this exception?

> the document already has an element with focus

I imagine a case like the following. A user presses a button, and the focus stays on that button. In that case, as the document has a focused element, the autofocus attribute of the element coming from the Turbo stream would not be respected. Is this the case?

In my opinion, if there is an autofocus attribute, it should be focused even if another element has the current focus.

seanpdoyle:
@afcapel @kevinmcconnell I've rebased this changeset to account for the migration from TypeScript. It's ready for re-review.

src/util.js
```diff
@@ -185,3 +185,32 @@ export function findClosestRecursively(element, selector) {
...
+  return !!element && element.closest(inertDisabledOrHidden) == null && typeof element.focus == "function"
+}
+
+export function queryAutofocusableElement(elementOrDocumentFragment) {
+  for (const element of elementOrDocumentFragment.querySelectorAll("[autofocus]")) {
+    if (elementIsFocusable(element)) return element
+    else continue
+  }
+
+  return null
```
afcapel, line 201:
```suggestion
Array.from(elementOrDocumentFragment.querySelectorAll("[autofocus]")).find(e => 
  elementIsFocusable(e))
```

src/util.js
```diff
@@ -185,3 +185,32 @@ export function findClosestRecursively(element, selector) {
...
+  const after = reader()
+
+  return [before, after]
+}
+
+export function waitForCallback(callback) {
+  callback()
+
+  return nextAnimationFrame()
+}
```
afcapel, line 216: This is not awaiting for `callback` to finish. I think it works anyway because at the moment we're only using it with a non-async callbacks. But the net effect of this the same as:

```js
callback()
await nextAnimationFrame()

// carry on after nextAnimationFrame...
```

Is that what we are trying to achieve? If that's the case, I think we can inline those two lines. That removes some code indirection and makes more explicit what's going on.

src/core/streams/stream_message_renderer.js
```diff
@@ -34,3 +39,55 @@ function getPermanentElementMapForFragment(fragment) {
...
+  const hasNoActiveElement = document.activeElement == null || document.activeElement == document.body
+
+  if (hasNoActiveElement && willAutofocus) {
+    const elementToAutofocus = document.getElementById(willAutofocus)
+
+    if (elementIsFocusable(elementToAutofocus)) {
+      elementToAutofocus.focus()
+    }
+    if (elementToAutofocus && elementToAutofocus.id == generatedID) {
+      elementToAutofocus.removeAttribute("id")
```
afcapel, line 66: Should we restore the original `id` here?

src/core/streams/stream_message_renderer.js
```diff
@@ -34,3 +39,55 @@ function getPermanentElementMapForFragment(fragment) {
 
   return permanentElementMap
 }
+
+async function withAutofocusFromFragment(fragment, callback) {
+  const generatedID = `turbo-stream-autofocus-${Date.now()}`
+  const turboStreams = fragment.querySelectorAll("turbo-stream")
+  const elementWithAutofocus = firstAutofocusableElementInStreams(turboStreams)
+  let willAutofocus = generatedID
```
afcapel, line 47: Should `willAutofocus` be falsy if there's no `elementWithAutofocus`? We could move the assignment to the next conditional.

src/core/streams/stream_message_renderer.js
```diff
@@ -34,3 +39,55 @@ function getPermanentElementMapForFragment(fragment) {
...
+}
+
+function firstAutofocusableElementInStreams(nodeListOfStreamElements) {
+  for (const streamElement of nodeListOfStreamElements) {
+    const elementWithAutofocus = queryAutofocusableElement(streamElement.templateElement.content)
+
+    if (elementWithAutofocus) return elementWithAutofocus
+  }
+
+  return null
```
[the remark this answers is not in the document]
afcapel: I don't think there's any performance concern here. Using `Array.from` is still O(n) and `nodeListOfStreamElements` are the `turbo-stream` tags inside a stream message, which are only going to be a handful, at most.

It's great to care about performance, but that also has to be weighted against code expressiveness and readability.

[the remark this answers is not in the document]
seanpdoyle: > that also has to be weighted against code expressiveness and readability.

In this case, `Array.find` would return the item in `nodeListOfStreamElements` (i.e. a `StreamElement` instance) that _has_ the autofocusable element. The original version of this method returns the _autofocusable_ element itself.

If we were to convert to an array, we'd need to combine `Array.map` and `Array.find`:

```js
Array.from(nodeListOfStreamElements).map((streamElement => queryAutofocusableElement(streamElement.templateElement.content)).find((element) => !!element)
```

Even broken across multiple lines/blocks, it still seems a bit wordy:

```js

Array.from(nodeListOfStreamElements)
  .map((streamElement => queryAutofocusableElement(streamElement.templateElement.content))
  .find((element) => !!element)
```

src/util.js
seanpdoyle: Since `elementIsFocusable` accepts an element argument, we can omit the block and pass the function itself!

```js
return Array.from(elementOrDocumentFragment.querySelectorAll("[autofocus]")).find(elementIsFocusable)
```

src/core/streams/stream_message_renderer.js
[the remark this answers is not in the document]
afcapel: Fair enough, it's true that when you concatenate `map` and `find` then the difference in readability is not clear anymore 👍

afcapel:
Nice one, thanks @seanpdoyle 🙏

michelson:
Hi @seanpdoyle could happen that the processing of new messages is blocked by this addition?
I'm I diving into the code because we have found an issue when receiving stream messages on tabs that are not visible/inactive, please see this issue:

https://github.com/hotwired/turbo-rails/issues/338

PR #690, opened 2022-08-17
Opt-into Turbo Stream `GET` from Form submitter

seanpdoyle:
Prior to this commit, declaring `[data-turbo-stream]` on a `<form>`
element with `[method=get]` would opt-into Turbo Stream support for the
response.

This commit extends that same support to `<form>` elements declared
_without_ the `[data-turbo-stream]` attribute that are submitted by
`<button>` or `<input type="submit">` elements that declare the
`[data-turbo-stream]` attribute.


dhh:
Curious how/where this has come up for you?

seanpdoyle:
Every other attribute that modifies a `<form>` submission (`formmethod`, `formaction`, `data-turbo-frame`, `data-turbo-method`, `data-turbo-confirm`) supports overrides from the submitter.

The motivation behind this change is to maintain consistency.

PR #691, opened 2022-08-17
Override `FetchOptions` from event listeners

seanpdoyle:
The problem
---

Prior to this commit, instances of `FormSubmission` and `FetchRequest`
referenced in dispatched `turbo:submit-start` and
`turbo:before-fetch-request` events were tedious to modify from a host
application.

For example, the `FormSubmission` instance made available through
`event.detail.formSubmission` exposed properties that _seemed_ writable.

Unfortunately, any mutations to `formSubmission.method` or
`formSubmission.location` from within host application event listeners
(like `turbo:before-fetch-request`) occur _after_ the underlying 
`FetchRequest` instance is built, which has subtle and not obvious 
implications. In practice, modifications to these properties are 
unobserved by the underlying `FetchRequest` instance, and have no effect
on the ensuing `fetch` call.

Luckily, if host applications are aware of these issues and limitations,
they can work around them by mutating the
`event.detail.formSubission.fetchRequest` instance.

The proposal
---

After this commit, the `FormSubmission.constructor` will infer the
required data (like `[method]` and `[action]` values) from its 
`HTMLFormElement` and optional `HTMLElement` arguments, then use those
values to construct the `FetchRequest` instance.

After that construction, any changes to the `FormSubmission` properties
will _delegate_ to the `FetchRequest` instance.

In tandem with those changes, the `FetchRequest` will follow a similar
pattern: extract as much information up-front during construction, then
delegate property reads and writes to properties. In most
cases, it will delegate to its [`fetchOptions:
RequestInit`][RequestInit] instance.

To achieve this behavior, several module-local functions and type
definitions were shuffled around between the `FormSubmission` and
`FetchRequest` modules. The order of arguments for the `FetchRequest`
constructor were also re-arranged to support omitting values that are
made redundant by default arguments.

[RequestInit]: https://developer.mozilla.org/en-US/docs/Web/API/fetch#parameters

### DIFF
--- a/src/core/drive/form_submission.js
+++ b/src/core/drive/form_submission.js
@@ -1,4 +1,4 @@
-import { FetchRequest, FetchMethod, fetchMethodFromString } from "../../http/fetch_request"
+import { FetchRequest, FetchMethod, fetchMethodFromString, fetchEnctypeFromString, isSafe } from "../../http/fetch_request"
 import { expandURL } from "../url"
 import { dispatch, getAttribute, getMetaContent, hasAttribute } from "../../util"
 import { StreamMessage } from "../streams/stream_message"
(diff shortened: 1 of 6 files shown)


seanpdoyle:
@afcapel @kevinmcconnell I've rebased this branch to account for the removal of types. It's ready for review.

afcapel:
@seanpdoyle looks good, clever refactor! 🙏

PR #695, opened 2022-08-19
Dispatch visit events on the initiator link or form

### LINKED ISSUE #99: Events: Accessing the link or form element in dispatched events
Most events fired by Turbo fire on `document`, irrespective of if the navigation was triggered programatically or through a link click or form submission. The `event.details` for these events does not contain a reference to the original interactive event (click / form) nor a reference to the original element in case a navigation was fired interactively.

I'd be very handy if these events (eg. `turbo:before-fetch-request`) would provide details with a reference to the link or form element that triggered the navigation. This would allow customisation based on the element's attributes, as well as updating the element during the lifecycle of the visit.

A few common examples that are very hard to achieve today:

* Add a `loading` class to the link or form or adding a spinner while the turbo request is in-flight (listening to `turbo:before-fetch-request`)

* Customize the fetch options based the element attributes, eg. `<a href=".." data-prefer-modal> ..</a>` could customize the `Accept` headers to allow the backend to return a specific variant.

* Handle network and server-side errors by adding an error class to the form in case the network fetch failed

* Customize cache behavior based on the `action`

A suggestion could be to provide `originalEvent` in `event.details` if a visit was triggered interactively, eg:

```js
// visit triggered via click on a link
event.details.originalEvent // => the original event, with event.target pointing to the link

// visit triggered via Turbo.visit(...)
event.details.originalEvent // => null
```

It could be there are ways to do this today that I've missed. If not, what do you think? I'll be happy to wrap up a PR supporting this, as I believe it would be very useful for customizing the behavior when doing navigations.

domchristie:
In a similar vein to #367, this pull request dispatches `turbo:visit`, `turbo:before-visit`, `turbo:before-fetch-request`, and `turbo:before-fetch-response` on the initiator link or form (where applicable), rather than solely on the `html` element.

**Why?** When developing libraries that build on top of Turbo, it'd be nice to customise behaviour based on HTML attributes of the clicked link or submitted form. For example, if I were building an animation library and wished to disable a transition when clicking a particular element, I could do:

```html
<a href="…" data-animate="false">Do not animate me</a>
```

I could then use `event.target.dataset.animate` in the `turbo:visit` event to conditionally animate.
Other examples are mentioned in #99.

Closes #99

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -51,7 +51,7 @@ export type VisitOptions = {
   shouldCacheSnapshot: boolean
   frame?: string
   acceptsStreamResponse: boolean,
-  initiator?: Element
+  initiator?: HTMLAnchorElement | HTMLFormElement
 }
 
 const defaultOptions: VisitOptions = {
@@ -87,6 +87,7 @@ export class Visit implements FetchRequestDelegate {
   readonly willRender: boolean
   readonly updateHistory: boolean
   readonly promise: Promise<void>
+  readonly initiator?: HTMLAnchorElement | HTMLFormElement
 
   private resolvingFunctions!: ResolvingFunctions<void>
 
@@ -127,6 +128,7 @@ export class Visit implements FetchRequestDelegate {
       updateHistory,
       shouldCacheSnapshot,
       acceptsStreamResponse,
+      initiator
     } = {
       ...defaultOptions,
       ...options,
@@ -143,6 +145,7 @@ export class Visit implements FetchRequestDelegate {
     this.scrolled = !willRender
     this.shouldCacheSnapshot = shouldCacheSnapshot
     this.acceptsStreamResponse = acceptsStreamResponse
+    this.initiator = initiator
   }
 
   get adapter() {
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -51,7 +51,7 @@ export type VisitOptions = {
   shouldCacheSnapshot: boolean
   frame?: string
   acceptsStreamResponse: boolean
-  initiator?: HTMLAnchorElement | HTMLFormElement
+  initiator?: Element
 }
 
 const defaultOptions: VisitOptions = {
@@ -87,7 +87,7 @@ export class Visit implements FetchRequestDelegate {
   readonly willRender: boolean
   readonly updateHistory: boolean
   readonly promise: Promise<void>
-  readonly initiator?: HTMLAnchorElement | HTMLFormElement
+  readonly initiator?: Element
 
   private resolvingFunctions!: ResolvingFunctions<void>
 
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -226,7 +226,7 @@ export class Visit implements FetchRequestDelegate {
     if (this.hasPreloadedResponse()) {
       this.simulateRequest()
     } else if (this.shouldIssueRequest() && !this.request) {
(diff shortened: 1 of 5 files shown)


src/core/drive/visit.ts
```diff
@@ -50,7 +50,8 @@ export type VisitOptions = {
   restorationIdentifier?: string
   shouldCacheSnapshot: boolean
   frame?: string
-  acceptsStreamResponse: boolean
+  acceptsStreamResponse: boolean,
+  initiator?: HTMLAnchorElement | HTMLFormElement
```
seanpdoyle, line 54: Could we declare this property without the `?` optional modifier? `VisitOption` instances are typically passed around as `Partial<VisitOption>` types, so the `Partial` makes it optional.

src/core/session.ts
```diff
@@ -191,17 +191,17 @@ export class Session
     )
   }
 
-  followedLinkToLocation(link: Element, location: URL) {
+  followedLinkToLocation(link: HTMLAnchorElement, location: URL) {
```
seanpdoyle, line 194: While I appreciate the type narrowing, so long as the `initiator` can invoke `dispatchEvent`, we might be able to achieve the outcome without disrupting these interfaces. 

I've been tempted to narrow this delegate method before, but I have a hunch that there is some Custom Element support reason that it's still `Element`.

src/core/drive/visit.ts
domchristie: Without `?` TypeScript moans because it's not present in `defaultOptions`:

```
Property 'initiator' is missing in type '{ action: "advance"; historyChanged: false; visitCachedSnapshot: () => void; willRender: true; updateHistory: true; shouldCacheSnapshot: true; acceptsStreamResponse: false; }' but required in type 'VisitOptions'.
```

seanpdoyle: Could its default value be `document.documentElement`, since that's what it's currently dispatched from?

domchristie: I'm not strongly opposed to this, but just wonder if having an undefined `initiator` usefully signals that the `Visit` is a programmatic one?

src/core/session.ts
domchristie: What would your advice be when updating the types here?:
https://github.com/hotwired/turbo/pull/695/files#diff-68b647dc2963716dc27c070f261d0b942ee9c00be7c4149ecb3a5acd94842d40R65

src/core/drive/visit.ts
domchristie: > I'm not strongly opposed to this, but just wonder if having an undefined initiator usefully signals that the Visit is a programmatic one?

I suppose it doesn't _really_ matter since a programmatic visit could be determined if `visit.initiator === document.documentElement`

domchristie: Or if we want a default, perhaps `initiator` is the wrong name for this? i.e. for programmatic visits the HTML element doesn't "initiate" the visit, the event just gets dispatched on it. It's not as clear, but perhaps `Visit#element` would work?

src/core/session.ts
domchristie: https://github.com/hotwired/turbo/pull/695/commits/95590c442526fc27fd1542a7a8dc3d0478827bb8 uses `Element` for all references

src/core/drive/visit.ts
domchristie: Default initiator implemented in: 36dd97bdb653dbc2c600729ad1364907d94455dc

seanpdoyle (approved):

PR #702, opened 2022-08-30
Dispatch `turbo:frame-render` when Frame connects

seanpdoyle:
Closes https://github.com/hotwired/turbo-rails/issues/379

Currently, a `<turbo-frame>` element only dispatches events when its
contents are fetched over HTTP. The `turbo:frame-render` and
`turbo:frame-load` events are dispatched sequentially and immediately
following one another. The `turbo:frame-render` detail exposes a
reference to the `FetchResponse`, while the `turbo:frame-load` event has
no details.

Consumer applications might expect a `<turbo-frame>` to dispatch a
`turbo:frame-render` event immediately whenever a `<turbo-frame>` is
connected to the document (similar to the way that `turbo:render` is
dispatched prior to `turbo:load`, even on restoration Visits).

For the sake of consistency and distinguishing between two events that
are more or less equivalent and temporally interchangeable, this commit
proposes that a `<turbo-frame>` should dispatch a `turbo:frame-render`
_whenever its contents change_, including its initial render.

As a complementary change to that behavior, the `turbo:frame-load`
events will be dispatched with `detail.fetchResponse` as a reference to
the `FetchResponse`.

Ideally, the `turbo:frame-render` would not expose the
`detail.fetchResponse` reference. However, removing it would be a
breaking change. To that end, this commit dispatches a `detail` that
prints a console warning (encouraging a migration to `turbo:frame-load`)
whenever a `turbo:frame-render` listener reads the
`detail.fetchResponse` property. Future major releases can remove the
property and warning.

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -59,6 +59,10 @@ export class FrameController {
       this.formLinkClickObserver.start()
       this.linkInterceptor.start()
       this.formSubmitObserver.start()
+
+      if (this.element.isConnected) {
+        session.frameRendered(this.element)
+      }
     }
   }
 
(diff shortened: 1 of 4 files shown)


seanpdoyle:
The commit message mentions a deprecation warning for `turbo:frame-render` listeners that attempt to access the `detail.fetchResponse` property.

As an alternative to that, we could remove that property entirely, and encourage applications upgrading from `7.1.0` to `7.2.0` to do a global search and replace, swapping `turbo:frame-render` with `turbo:frame-load`. Prior to this commit, there is absolutely zero difference in timing between the two, and after this commit, the `turbo:frame-load` event details will be interchangeable for pre-`7.2.0` shaped `turbo:frame-render` events.

brunoprietog:
@seanpdoyle do you plan to continue with this? It would be awesome to see it merged

PR #711, opened 2022-09-13
Re-introduce #627

seanpdoyle:
Restore #627
===

The changes originally introduced in [hotwired/turbo#627][] had some
negative side effects, including a batch of the [current Chrome
failures][ci].

To resolve the issues, the diff was reverted by [hotwired/turbo#715][].

This commit re-introduces the changes proposed in [hotwired/turbo#627][]
with additional test coverage to ensure that the rendering timing plays
nice with Snapshot caching and other existing rendering behavior.

[hotwired/turbo#627]: https://github.com/hotwired/turbo/pull/627
[ci]: https://github.com/hotwired/turbo/actions/runs/3045476866/jobs/4907085095
[hotwired/turbo#715]: https://github.com/hotwired/turbo/pull/715

Co-authored-by: Alex Robbin <agrobbin@gmail.com>

### DIFF
--- a/src/core/drive/page_renderer.ts
+++ b/src/core/drive/page_renderer.ts
@@ -1,19 +1,22 @@
 import { Renderer } from "../renderer"
 import { PageSnapshot } from "./page_snapshot"
 import { ReloadReason } from "../native/browser_adapter"
-import { activateScriptElement, waitForLoad } from "../../util"
+import { activateScriptElement, waitForLoad, getBodyElementId } from "../../util"
 
 export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
-  static renderElement(currentElement: HTMLBodyElement, newElement: HTMLBodyElement) {
-    if (document.body && newElement instanceof HTMLBodyElement) {
-      document.body.replaceWith(newElement)
+  static renderElement(currentBody: HTMLBodyElement, newBody: HTMLBodyElement) {
+    if (document.body && newBody instanceof HTMLBodyElement) {
+      const currentElement = PageRenderer.getRenderedElement(currentBody) || currentBody
+      const newElement = PageRenderer.getRenderedElement(newBody) || newBody
+
+      currentElement.replaceWith(newElement)
     } else {
-      document.documentElement.appendChild(newElement)
+      document.documentElement.appendChild(newBody)
     }
   }
 
   get shouldRender() {
-    return this.newSnapshot.isVisitable && this.trackedElementsAreIdentical
+    return this.newSnapshot.isVisitable && this.trackedElementsAreIdentical && this.renderedElementMatches
   }
 
   get reloadReason(): ReloadReason {
@@ -26,6 +29,12 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
     if (!this.trackedElementsAreIdentical) {
       return {
         reason: "tracked_element_mismatch",
+      }
+    }
+
+    if (!this.renderedElementMatches) {
+      return {
+        reason: "rendered_element_mismatch",
       }
     }
   }
@@ -76,6 +85,24 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
 
   get trackedElementsAreIdentical() {
     return this.currentHeadSnapshot.trackedElementSignature == this.newHeadSnapshot.trackedElementSignature
+  }
+
+  get renderedElementMatches() {
+    return PageRenderer.getRenderedElement(this.newElement) !== null
+  }
+
+  static get bodySelector() {
+    const bodyId = getBodyElementId()
+
+    return bodyId ? `#${bodyId}` : "body"
+  }
+
+  static getRenderedElement(element: HTMLElement): HTMLElement | null {
+    if (element.matches(this.bodySelector)) {
+      return element
+    } else {
+      return element.querySelector(this.bodySelector)
+    }
   }
 
   async copyNewHeadStylesheetElements() {
--- a/src/core/drive/page_view.ts
+++ b/src/core/drive/page_view.ts
@@ -1,4 +1,3 @@
-import { nextEventLoopTick } from "../../util"
 import { View, ViewDelegate, ViewRenderOptions } from "../view"
 import { ErrorRenderer } from "./error_renderer"
 import { PageRenderer } from "./page_renderer"
@@ -39,11 +38,10 @@ export class PageView extends View<HTMLBodyElement, PageSnapshot, PageViewRender
     this.snapshotCache.clear()
   }
 
-  async cacheSnapshot() {
+  cacheSnapshot() {
     if (this.shouldCacheSnapshot) {
       this.delegate.viewWillCacheSnapshot()
       const { snapshot, lastRenderedLocation: location } = this
-      await nextEventLoopTick()
       const cachedSnapshot = snapshot.clone()
       this.snapshotCache.put(location, cachedSnapshot)
       return cachedSnapshot
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -465,7 +465,8 @@ export class Visit implements FetchRequestDelegate {
 
   cacheSnapshot() {
     if (!this.snapshotCached) {
-      this.view.cacheSnapshot().then((snapshot) => snapshot && this.visitCachedSnapshot(snapshot))
+      const snapshot = this.view.cacheSnapshot()
+      if (snapshot) this.visitCachedSnapshot(snapshot)
       this.snapshotCached = true
     }
   }
--- a/src/tests/fixtures/drive_custom_body.html
+++ b/src/tests/fixtures/drive_custom_body.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="turbo-body" content="app">
+    <title>Drive (with custom body)</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+  </head>
+  <body>
+    <h1>Drive (with custom body)</h1>
+    <div id="app">
+      <div>
+        <a id="drive" href="/src/tests/fixtures/drive_custom_body_2.html">Drive enabled link</a>
+        <a id="mismatch" href="/src/tests/fixtures/drive_custom_body_3.html">Drive enabled link to page with mismatched custom body</a>
+      </div>
+
+      <p id="different-content">Drive 1</p>
+    </div>
+  </body>
+</html>
--- a/src/tests/fixtures/drive_custom_body_2.html
+++ b/src/tests/fixtures/drive_custom_body_2.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <meta name="turbo-body" content="app">
+    <title>Drive (with custom body)</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+  </head>
+  <body>
+    <h1>Drive (with custom body 2)</h1>
+
+    <div id="app">
+      <div>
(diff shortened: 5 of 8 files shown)


src/core/drive/page_renderer.ts
```diff
@@ -1,12 +1,10 @@
 import { Renderer } from "../renderer"
 import { PageSnapshot } from "./page_snapshot"
 import { ReloadReason } from "../native/browser_adapter"
-import { activateScriptElement, waitForLoad, nextEventLoopTick, getBodyElementId } from "../../util"
+import { activateScriptElement, waitForLoad, getBodyElementId } from "../../util"
 
 export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
-  static async renderElement(currentElement: HTMLBodyElement, newElement: HTMLBodyElement) {
-    await nextEventLoopTick()
```
seanpdoyle, line 8: @agrobbin reverting these lines was necessary to pass CI. If they were crucial parts of the feature, could you try re-introducing them with test coverage that passes in their presence but fails in their absence?

agrobbin: @seanpdoyle this was where I was struggling in #627. The `await nextEventLoopTick()` was necessary for reasons I explained in https://github.com/hotwired/turbo/pull/627#discussion_r921455484 (and are further described in https://github.com/hotwired/turbo/pull/305#issuecomment-1183505367), but for reasons unknown to me, it caused a bunch of other test failures that I could not understand.

seanpdoyle: Thank you for sharing that context.

This might leave the release candidate in an awkward place: the CI failures introduced by the original changeset appear to be genuine, so the fact that they're failing signals that Real rendering regressions might have been released.

On the flip side, making CI pass (including the failing rendering tests and the test coverage introduced by #627) _without_ the `nextEventLoopTick()` signals that passing CI has the potential to break the feature introduced by #627 in a way that isn't covered by the test suite.

@agrobbin are you able to investigate an alternative pull request that maintains the changes you've introduced, exercises them with text coverage, and passes the rest of the suite?

seanpdoyle: @agrobbin alternatively, if you can introduce a failing test that exhibits the behavior you've outlined in https://github.com/hotwired/turbo/pull/305#issuecomment-1183505367 and https://github.com/hotwired/turbo/pull/627#discussion_r921455484, I could troubleshoot on my own.

agrobbin: I was looking for some help with that exact thing in https://github.com/hotwired/turbo/pull/627#issuecomment-1212647667, as I was not able to figure out why that particular addition failed tests, when things _did_ seem to work as expected when testing it locally.

seanpdoyle: @agrobbin I've pushed up a second commit to try and restore the asynchronous rendering by moving the call to the general purpose renderer. Since this isn't covered by tests, it's difficult to verify on my own that it works as expected. Could you pull these changes down locally and attempt to reproduce the buggy behavior you've described in https://github.com/hotwired/turbo/pull/627#discussion_r921455484?

agrobbin: @seanpdoyle, on it, give me a few minutes and I'll confirm one way or the other!

agrobbin: It seems to work @seanpdoyle!

agrobbin: Your change also seems to be way more obvious and less magical than mine, so thank you for that.

seanpdoyle: awesome! how did you verify that it worked? Could you share an outline of the steps you took, or share some pseudocode to exercise it?

src/core/drive/page_renderer.ts
```diff
@@ -18,7 +16,7 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
   }
 
   get shouldRender() {
-    return this.newSnapshot.isVisitable && this.trackedElementsAreIdentical && this.bodyElementMatches
```
seanpdoyle, line 21: @agrobbin could you also share some context around this change and the `{ reason: "body_element_mismatch" }` below? If they're crucial, we should include coverage for them.

src/core/drive/page_renderer.ts
agrobbin: It was a bit cumbersome, since there is no single compiled distrubtion of your changes, so what I did was add this resolution to my `package.json`:

```json
"@hotwired/turbo": "7.2.0-rc.1"
```

And then I explicitly edited the `dist/turbo.es2017-esm.js` file to match these changes. That allowed me to run this exact code on my application that was previously leveraging a patched version of the changes in #627. From there I could try it with `await nextAnimationFrame()` and without, and confirm that it does not work when it's missing, and that it does work when it's present. It *not* working manifests as the page caches being out of sync w/ the URL that is shown in the browser address bar when navigating forward and backward in browser history (after clicking on a few different links to navigate through several pages).

seanpdoyle: Could you share more details specific about your application code, or could you generalize it to pseudocode or elements that could be added directly to the codebase's test suite?

src/core/drive/page_renderer.ts
agrobbin: This was done at the request of @dhh in https://github.com/hotwired/turbo/pull/627#issuecomment-1188307421, to trigger a full page reload when there is a mismatch of the configured body element, and provide a `reason` for the reload that is passed along through the invalidation code path, and eventually to the `turbo:reload` event `detail` [here](https://github.com/hotwired/turbo/blob/eda47178e57a5f666c268c54ab459a482faf020b/src/core/native/browser_adapter.ts#L124).

seanpdoyle: Could you outline a scenario where that might occur? Having a test case to cover that code would guard against future regressions.

agrobbin: Imagine these 2 HTML documents (abbreviated to convey the intent):

document 1:

```html
<html>
  <head>
    <meta name="turbo-body" content="app">
  </head>
  <body>
    <div id="app"></div>
  </body>
</html>
```

document 2:

```html
<html>
  <head>
    <meta name="turbo-body" content="application">
  </head>
  <body>
    <div id="application"></div>
  </body>
</html>
```

Transitioning between these 2 documents (in either direction) via Turbo will hit this scenario, since the `turbo-body` configurations do not match (`app` vs. `application`).

src/core/drive/page_renderer.ts
agrobbin: Let me try and boil it down to the basics that I can share. It might take me a little bit while I juggle some family stuff this afternoon.

src/core/drive/page_renderer.ts
seanpdoyle: I've added test coverage to exercise these lines. Thank you for sharing an example.

src/core/drive/page_renderer.ts
agrobbin: OK, here goes...

My application has a layout that looks like what I shared below:

```html
<html>
  <head>
    <meta name="turbo-body" content="app">
  </head>
  <body>
    <div id="app"></div>
  </body>
</html>
```

Within the `#app` element, the vast majority of links run through Turbo Drive. Imagine a few navigational links like this:

```html
<a href="/users">Users</a>
<a href="/projects">Projects</a>
<a href="/billing">Billing</a>
```

Here is some sample content for each of those pages, wrapped inside of the `#app` element:

```html
<!-- /users -->
<h1>Users</h1>
<p>You are the only user.</p>

<!-- /projects -->
<h1>Projects</h1>
<p>There are currently no projects.</p>

<!-- /billing -->
<h1>Billing</h1>
<p>You owe us $1.</p>
```

Clicking on each navigational link works as you'd expect through Turbo Drive, replacing the content within `#app`, updating `<head>` appropriately, etc.

If you start at Users, go to Projects, and then to Billing, everything is A-OK. When you start clicking the browser back/forward buttons however, the content within `#app` that is presented to you does not match the URL you see. Clicking the back button once will update the URL in the address bar from `/billing` to `/projects`, but continue to show the content from `/billing`. Clicking back again updates the URL to `/users`, but shows the content from `/projects`. Essentially, everything is off by 1.

That's the minimal set of changes that I think represent the situation that I experience when the `await nextAnimationFrame()` is not included.

I hope that's helpful!

seanpdoyle: I tried to reproduce that locally with this diff:

```diff
diff --git a/src/core/view.ts b/src/core/view.ts
index e02835b..0c34160 100644
--- a/src/core/view.ts
+++ b/src/core/view.ts
@@ -3,7 +3,6 @@ import { Renderer, Render } from "./renderer"
 import { Snapshot } from "./snapshot"
 import { Position } from "./types"
 import { getAnchor } from "./url"
-import { nextAnimationFrame } from "../util"
 
 export interface ViewRenderOptions<E> {
   resume: (value: any) => void
@@ -127,7 +126,6 @@ export abstract class View<
   }
 
   async renderSnapshot(renderer: R) {
-    await nextAnimationFrame()
     await renderer.render()
   }
 
diff --git a/src/tests/functional/drive_custom_body_tests.ts b/src/tests/functional/drive_custom_body_tests.ts
index 54fe5a5..c418a7e 100644
--- a/src/tests/functional/drive_custom_body_tests.ts
+++ b/src/tests/functional/drive_custom_body_tests.ts
@@ -1,6 +1,6 @@
 import { test } from "@playwright/test"
 import { assert } from "chai"
-import { nextBody, pathname, getFromLocalStorage, setLocalStorageFromEvent } from "../helpers/page"
+import { nextEventNamed, pathname, getFromLocalStorage, setLocalStorageFromEvent } from "../helpers/page"
 
 const path = "/src/tests/fixtures/drive_custom_body.html"
 
@@ -9,15 +9,26 @@ test.beforeEach(async ({ page }) => {
 })
 
 test("test drive with a custom body element", async ({ page }) => {
-  page.click("#drive")
-  await nextBody(page)
+  await page.click("#drive")
+  await nextEventNamed(page, "turbo:load")
 
-  const h1 = await page.locator("h1")
-  const differentContent = await page.locator("#different-content")
+  assert.equal(pathname(page.url()), "/src/tests/fixtures/drive_custom_body_2.html")
+  assert.equal(await page.textContent("h1"), "Drive (with custom body)")
+  assert.equal(await page.textContent("#different-content"), "Drive 2")
+
+  await page.goBack()
+  await nextEventNamed(page, "turbo:load")
+
+  assert.equal(pathname(page.url()), "/src/tests/fixtures/drive_custom_body.html")
+  assert.equal(await page.textContent("h1"), "Drive (with custom body)")
+  assert.equal(await page.textContent("#different-content"), "Drive 1")
+
+  await page.goForward()
+  await nextEventNamed(page, "turbo:load")
 
   assert.equal(pathname(page.url()), "/src/tests/fixtures/drive_custom_body_2.html")
-  assert.equal(await h1.textContent(), "Drive (with custom body)")
-  assert.equal(await differentContent.textContent(), "Drive 2")
+  assert.equal(await page.textContent("h1"), "Drive (with custom body)")
+  assert.equal(await page.textContent("#different-content"), "Drive 2")
 })
 
 test("test drive with mismatched custom body elements", async ({ page }) => {
```

Unfortunately, the test passes without the `await nextAnimationFrame()`. Is it possible that there are other factors involved in your application that might affect the timing in an application-specific way?

Is there something missing from the test that would go along with the `page.goBack()` and `page.goForward()` calls to more faithfully reproduce your use case?

seanpdoyle: I was able to add test coverage that fails with the removal of the `nextAnimationFrame()` call.

agrobbin: @seanpdoyle great! What changed in the last ~5 minutes? 😆

manuelpuyol:
@seanpdoyle what are we missing to reintroduce this change?

seanpdoyle:
> @seanpdoyle what are we missing to reintroduce this change?

The implementation needs to be changed so that the test suite passes.

The failures in CI are genuine.

joonty:
Is this still feasible? My application uses a Spotify audio player iframe, meaning that even with the `data-turbo-permanent` attribute the iframe is reloaded on page navigation and the user's playback is interrupted. If turbo could use a non-body element as the root node that would solve my problem, since I could move the iframe out of that node.

PR #718, opened 2022-09-16
Reintroduce `Visit` `identifier` property

kevinmcconnell:
This property is required by turbo-ios, so removing it causes crashes within that library.

E.g. https://github.com/hotwired/turbo-ios/blob/main/Source/WebView/WebViewBridge.swift#L129

This reverts commit d81e24db77da05f53c6ab3c66e2620574c9e08c9.

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -79,6 +79,7 @@ export enum SystemStatusCode {
 
 export class Visit implements FetchRequestDelegate {
   readonly delegate: VisitDelegate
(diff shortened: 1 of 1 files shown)


marcoroth (commented):
Thanks @kevinmcconnell! That also explains why it was unused throughout the `@hotwired/turbo` codebase...


Would you mind adding a comment for this, so that we don't run into the same issue again?

kevinmcconnell:
@marcoroth yeah, same thing just occurred to me after pushing 👍 Added now.

Issue #720, opened 2022-09-16
Navigations failing in Turbo iOS apps, due to `initiator` visit option


kevinmcconnell:
In #695, an `initiator: Element` property is added to `VisitOptions`, in order to track the element that initiates the visit. This works great in the browser 👍

However it causes a problem for native apps that use the Turbo iOS adapter. The adapter's implementation of `visitProposedToLocation` [passes the options through to native code using `postMessage`](https://github.com/hotwired/turbo-ios/blob/main/Source/WebView/turbo.js#L109), and the presence of the `Element` causes it to fail with a `DataCloneError` exception. (Presumably because the `Element` type isn't supported by the underlying [structuredClone](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) operation.)

The result is that tapping links in Turbo iOS apps silently fails (aside from the console error that you can see in development).

This could be fixed in Turbo iOS, but In order to not break backwards compatibility with existing apps, it would be better to take the `initiator` property back out of the options.

One option could be to extend `visitProposedToLocation` with an additional argument (`EventOptions` perhaps?) and move the `initiator` property into it. That extra argument would be ignored by the existing Turbo iOS code, so it would be a backwards-compatible way to evolve that part of the adapter interface. Alternatively, we could try to sidestep `visitProposedToLocation` entirely with this, and set the initiator property on the `Visit` after creation.

What do you think @domchristie @seanpdoyle ?

kevinmcconnell:
Hi @domchristie

> How about sanitising the options before [calling adapter#visitProposedToLocation](https://github.com/hotwired/turbo/blob/e592476087bea3acbd60bbde65b9717e7455cb4b/src/core/session.ts#L212)?

It's a good idea, but (unless I'm overlooking something here) the problem is that `visitProposedToLocation` is the route used by all adapters to initiate a visit (including [the browser adapter](https://github.com/hotwired/turbo/blob/e592476087bea3acbd60bbde65b9717e7455cb4b/src/core/native/browser_adapter.ts#L27)). So sanitising the options in this way would stop the feature working.

If you only sanitise it for a specific adapter, that could stop the crash, but it also means that the feature won't be working for clients using that adapter. Browser clients will then have events fired on the initiating element, but in mobile clients those events will continue to fire on the document element as before. I think that would lead to confusion for anyone adding listeners to those events.

> Ideally this would be done on the Turbo iOS side before calling postMessage, but as you mention: backwards compatibility :/

[...]

PR #723, opened 2022-09-19
Revert "Dispatch visit events on the initiator link or form (#695)"

kevinmcconnell:
The inclusion of the `initiator` in `VisitOptions` causes a JS crash in the Turbo iOS adapter.

This is because Turbo iOS's implementation of `visitProposedToLocation` relies on being able to pass the entire options struct to native code. Initiator is an `Element`, which can't be passed in this way, resulting in a `DataCloneError` exception.

There is some discussion on this [GitHub issue][]. This PR reverts the change for the moment, until we have an implementation of it that works with all adapters.

[GitHub issue]: https://github.com/hotwired/turbo/issues/720

This reverts commit 887181796e8d5586de89c70327f31e9a81dcfec2.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -8,7 +8,7 @@ import { Visit, VisitDelegate, VisitOptions } from "./visit"
 import { PageSnapshot } from "./page_snapshot"
 
 export type NavigatorDelegate = VisitDelegate & {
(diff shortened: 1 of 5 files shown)


seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/725 to resolve this PR's [CI test failures](https://github.com/hotwired/turbo/actions/runs/3082072410/jobs/4981335608#step:11:28).

PR #725, opened 2022-09-19
CI: Restore passing tests

seanpdoyle:
With the reverting of [hotwired/turbo#695][] (merged as [hotwired/turbo#723][]), the "Visit with network error" test coverage was missed in the diff.

Prior to the revert, the test was listening for
`turbo:fetch-request-error` events dispatched on the _link_ that was clicked. Since that's no longer the case, this commit changes the test to listen for `turbo:fetch-request-error` that bubble up to the document.

[hotwired/turbo#695]: https://github.com/hotwired/turbo/pull/695
[hotwired/turbo#723]: https://github.com/hotwired/turbo/pull/723

### DIFF
--- a/src/tests/functional/visit_tests.ts
+++ b/src/tests/functional/visit_tests.ts
@@ -7,7 +7,6 @@ import {
   isScrolledToTop,
   nextBeat,
   nextEventNamed,
-  nextEventOnTarget,
   noNextAttributeMutationNamed,
   readEventLogs,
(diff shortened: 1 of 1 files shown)


kevinmcconnell:
Thanks @seanpdoyle, I missed that one. There is quite a lot of flakiness in the tests currently!

Just to clarify, this was actually from the revert of #695 I think, rather than #569?

PR #732, opened 2022-09-23
Update README to include install/confirm info

asheroto:
- Add install as a package instructions
- Add install using CDN instructions
- Add confirmation instructions


brunoprietog:
Appreciate the contribution, but installation instructions are already provided [here](https://turbo.hotwired.dev/handbook/installing). Closing so as not to duplicate.

PR #733, opened 2022-09-23
Prevent errors when communicating with iOS webview

domchristie:
To successfully communicate with the iOS view, messages need to be of a type supported by the structuredClone algorithm. The new TransferableVisitOptions type ensures that VisitOptions passed to the adapter conform to this type. The StructuredCloneValue type has been created based on https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm#supported_types

This should hopefully prevent future errors such as https://github.com/hotwired/turbo/issues/720

---

An alternative to this might be that we become stricter with the `VisitOptions` we pass around, and instead of using `Partial` use `Pick` and `Omit`?

/cc @kevinmcconnell


src/core/types.ts
```diff
@@ -18,3 +18,37 @@ export type StreamSource = {
     options?: boolean | EventListenerOptions
   ): void
 }
+
+export type StructuredCloneValue =
```
kevinmcconnell, line 22: It would've been great if there was an existing TypeScript definition for compatible types. But I guess there isn't.

kevinmcconnell (commented):
This is a great idea! 👏 I really like the idea of using the compiler to enforce the safety of the options fields, rather than rely on a comment (or our memory).

My only thought is that I'd be tempted to tweak it so that adding a non-transferable option would cause a compiler error, rather than it being automatically sanitised. It just makes it more obvious to the person adding that field that that's happening.

src/core/session.ts
```diff
@@ -424,6 +424,18 @@ export class Session
...
+    const sanitized: { [key: string]: StructuredCloneValue } = {}
+    Object.entries(options).forEach(([key, value]) => {
+      try {
+        sanitized[key] = structuredClone(value)
+      } catch (_) {
+        // Ignore non-transferable values
+      }
+    })
+    return sanitized
+  }
```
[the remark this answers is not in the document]
domchristie: > I think that would be nice because it makes it clear that those options are going to be lost when handed to the adapter.

Love it. Done in https://github.com/hotwired/turbo/pull/733/commits/495973c33bbeb34327328756bf3e2708320ed651

src/core/types.ts
domchristie: I couldn't easily find one. I was hopeful on `Transferable`, but that's just `ArrayBuffer | ImageBitmap | MessagePort`, and unfortunately the VSCode `structuredClone` tooltips don't give any clues either 😞

kevinmcconnell: Yeah, it looks like TS's lib.dom defines the argument as `any`, which is really not helpful! :(

src/core/drive/visit.ts
```diff
@@ -53,6 +53,8 @@ export type VisitOptions = {
   acceptsStreamResponse: boolean
 }
 
+export type TransferableVisitOptions = Partial<VisitOptions> & { [key: string]: StructuredCloneValue }
```
domchristie, line 56: Perhaps don't use `Partial<…` here And use when annotating?

domchristie:
Just to add: this doesn't change the existing behaviour; it just adds type checking, and removes any properties that would cause a `DataCloneError`

domchristie:
(And I'm not sure why the tests are failing 🤷‍♂️ They also fail for me locally on the `main` branch`)

domchristie:
> Just to add: this doesn't change the existing behaviour; it just adds type checking, and removes any properties that would cause a DataCloneError
> (And I'm not sure why the tests are failing 🤷‍♂️ They also fail for me locally on the main branch`)

On second thoughts, this _does_ alter the behaviour, because the options passed to `Navigator#startVisit` will be without `referrer` and `visitCachedSnapshot`. This is amended in https://github.com/hotwired/turbo/pull/750/commits/122fd26d787567bde72010bdd8d8056dc9dc1204

domchristie:
Superseded by https://github.com/hotwired/turbo/pull/750

Issue #743, opened 2022-09-28
7.2 data-remote inside turbo-frame produces double requests


acetinick:
Since upgrading to turbo 7.2 which we use extensively throughout our project, we have some introduced issues when we have some legacy data-remote="true" UJS being triggered within a turbo-frame.  

When we have a situation where we have a data-remote inside a turbo-frame, things break down.  Example:

```
<turbo-frame id="example">
   <a data-remote="true" href="/customers/new">click</a>
</turbo-frame>
```

What ends up happening is that the remote gets triggered, but for some reason a double request occurs issueing the turbo as well along with the UJS call to the URL.   Attached example from console (from a real situation not above example)

![image](https://user-images.githubusercontent.com/571649/192798644-041f4249-1331-4e8f-8836-78ec9c527b53.png)

- the 406 is the turbo request, which is NOT expected resulting in an UnkownFormat exception which trying to render a .html format, but we only have a .js.erb view
- the 200 is the UJS request, which is expected to return JS to open a modal window

I can conifrm that when we downgrade to 7.1 everything works fine,  but as soon as we upgrade issue is reproduced.  VERY strange tho, for some reason it only happens in non-local environments.   Locally it works, but in staging it breaks.

It also doesn't seem to matter if its data-method="post" or data-method="get",  it occurs always with data-remote when inside a turbo-frame.

My gut feel it must be related to #612 or #645, which is a huge benefit we already making use of this, but somehow its conflicting with data-remote somehow.   @kevinmcconnell @seanpdoyle perhaps you got some ideas?

seanpdoyle:
Thank you for opening this issue @acetinick. I've opened https://github.com/hotwired/turbo/pull/744 to try and resolve it.

PR #744, opened 2022-09-28
Ignore UJS `<a>` clicks and `<form>` submissions

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/743

The change in behavior can be traced back to [hotwired/turbo#412][].
When the overlap between the `LinkInterceptor` and the
`LinkClickObserver` were unified, the technique used by the
`LinkInterceptor` to prevent duplicate event handlers from responding to
the same `click` event was changed in a subtle way.

In its place, this commit changes the `LinkClickObserver` and
`FormSubmitObserver` to ignore `<a>` clicks and `<form>` submissions if
they're annotated with `[data-remote="true"]`.

To exercise these edge cases, this commit also adds a `ujs.html` fixture
file along with a `ujs_tests.ts` module to cover situations when
`@rails/ujs` and `@hotwired/turbo` co-exist.

[hotwired/turbo#412]: https://github.com/hotwired/turbo/pull/412

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -23,7 +23,7 @@ import { ViewDelegate, ViewRenderOptions } from "../view"
 import { Locatable, getAction, expandURL, urlsAreEqual, locationIsVisitable } from "../url"
 import { FormSubmitObserver, FormSubmitObserverDelegate } from "../../observers/form_submit_observer"
 import { FrameView } from "./frame_view"
-import { LinkClickObserver, LinkClickObserverDelegate } from "../../observers/link_click_observer"
+import { LinkInterceptor, LinkInterceptorDelegate } from "./link_interceptor"
 import { FormLinkClickObserver, FormLinkClickObserverDelegate } from "../../observers/form_link_click_observer"
 import { FrameRenderer } from "./frame_renderer"
 import { TurboClickEvent, session } from "../index"
@@ -43,14 +43,14 @@ export class FrameController
     FormSubmissionDelegate,
     FrameElementDelegate,
     FormLinkClickObserverDelegate,
-    LinkClickObserverDelegate,
+    LinkInterceptorDelegate,
     ViewDelegate<FrameElement, Snapshot<FrameElement>>
 {
   readonly element: FrameElement
   readonly view: FrameView
   readonly appearanceObserver: AppearanceObserver
   readonly formLinkClickObserver: FormLinkClickObserver
-  readonly linkClickObserver: LinkClickObserver
+  readonly linkInterceptor: LinkInterceptor
   readonly formSubmitObserver: FormSubmitObserver
   formSubmission?: FormSubmission
   fetchResponseLoaded = (_fetchResponse: FetchResponse) => {}
@@ -70,7 +70,7 @@ export class FrameController
     this.view = new FrameView(this, this.element)
     this.appearanceObserver = new AppearanceObserver(this, this.element)
     this.formLinkClickObserver = new FormLinkClickObserver(this, this.element)
-    this.linkClickObserver = new LinkClickObserver(this, this.element)
+    this.linkInterceptor = new LinkInterceptor(this, this.element)
     this.restorationIdentifier = uuid()
     this.formSubmitObserver = new FormSubmitObserver(this, this.element)
   }
@@ -84,7 +84,7 @@ export class FrameController
         this.loadSourceURL()
       }
       this.formLinkClickObserver.start()
-      this.linkClickObserver.start()
+      this.linkInterceptor.start()
       this.formSubmitObserver.start()
     }
   }
@@ -94,7 +94,7 @@ export class FrameController
       this.connected = false
       this.appearanceObserver.stop()
       this.formLinkClickObserver.stop()
-      this.linkClickObserver.stop()
+      this.linkInterceptor.stop()
       this.formSubmitObserver.stop()
     }
   }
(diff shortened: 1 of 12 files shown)


seanpdoyle:
I've converted this to a draft, since I'm not sure about the intended behavior.

While the test coverage ensures that there is only a single request issued, I'm not entirely sure about how `@rails/ujs` and `@hotwired/turbo` should co-exist in this scenario.

The test asserts that clicking `<a data-remote="true">` would drive the frame. From a turbo-centric perspective, that _feels_ correct, but the fact that the Turbo event listeners respond to the `click` event before the UJS handlers feels incidental instead of deliberate.

This covers an edge-case scenario for an application in the midst of a migration from UJS to Turbo. Given that rendering an `<a data-remote="true">` element requires opt-in, the intention behind the element are likely at odds with driving a `<turbo-frame>` element.

To quote the [original issue](https://github.com/hotwired/turbo/issues/743) opened by @acetinick:

> *the [...] turbo request, **which is NOT expected** resulting in an UnknownFormat exception which trying to render a .html format, but we only have a .js.erb view
> the [...] UJS request, **which is expected** to return JS to open a modal window

Given those expectations, Turbo **should lose the race** against UJS for handling the event.

Since `7.1.0` supported the expected behavior, finding a way to revert https://github.com/hotwired/turbo/pull/412 (incorporating changes made since its merge) might be the most backwards-compatible solution.

acetinick:
This amazing, I'll try it out and let you know. 

I guess for large applications like ours it's impossible to switch so easily, it's a slow transition screen by screen. We know we need to pull the pin eventually but for a point release of turbo it's hard for us to refactor soo many existing UJS methods, which are in the hundreds.

I think what always was the goal of turbo was move away from UJS which we agree. Baby steps. For a lot of large apps, it's a few years to rid completely. Like a jquery that sticks around

acetinick:
To add to your comment about expectations, I believe it would be safe to assume that any link with data-remote should be ignored from turbo link intercepts completely. 

If anyone is using data-remote they will not be expecting turbo to do anything with this request.  

I believe same expectation will go for all remote situations of links/buttons and forms.  

It's either one or the other.

acetinick:
@dhh when you get chance could you review and if possible do a release on this. It's a blocker for anyone still using data-remote to upgrade to 7.2

jeremy:
> When the above code is using Turbo 7.2, both the link and the form issue duplicate requests. With this PR, both link & form will use UJS, which would be a change from 7.1.

True, but… seems worth breaking while it's still undefined behavior? Considering the same markup would then behave differently in & out of a frame.

Curious whether `turbo-rails` could manage Rails UJS integration. If Turbo offered a hook or delegate call for users to make this determination, then turbo-rails could use it to check for UJS conflict (and emit deprecation warnings, throw errors, etc.)

seanpdoyle:
@Jeremy I agree that package is a more appropriate home for UJS-specific code. I've previously explored what that might entail in https://github.com/hotwired/turbo-rails/pull/93 and https://github.com/hotwired/turbo-rails/pull/257. 

> If Turbo offered a hook or delegate call for users to make this determination

Applications can call `event.preventDefault()` on `turbo:click` events to opt-out of Turbo integration. Could a similar mechanism on `turbo:submit-start` support opting out of Form Submissions?

Then, Turbo Rails could define event listeners like:

```js
addEventListener("turbo:click", (event) => {
  const remote = event.target.closest('[data-remote="true"]')

  if (remote) event.preventDefault()
})

addEventListener("turbo:submit-start", (event) => {
  const remote = event.target.closest('[data-remote="true"]')

  if (remote) event.preventDefault()
})
```

kevinmcconnell:
> True, but… seems worth breaking while it's still undefined behavior?

The downside to breaking it is that upgrading 7.2 would then require people to update any forms inside `<turbo-frame>`s that had been marked `remote: true`, as they would now be handled differently. And as I say, that sounds like it would be an edge case, but given that many people will have been adding `remote: true` to their forms by default, it's likely to come up quite a bit.

I agree that the inconsistency is not great, but we could choose to preserve existing behaviour while also adding a nicer path for people to resolve the conflict on their own time. As you say, adding things like deprecation warnings would be helpful.

We can prevent the duplicate form submission by adding back the `stopImmediatePropagation` that was in 7.1.

> Applications can call event.preventDefault() on turbo:click events to opt-out of Turbo integration

The ability to do this on `click` events is what changed in #412, I believe, which results in the double submissions from clicks in frames. The event handlers moving to the frame element itself means cancelling the event on `document` no longer works. But if we restored that behaviour, we could stop the double submissions without having UJS & Turbo need to be aware of each other at all. I think that's a nicer way to do it than Turbo having knowledge of `data-remote`. It also means that if there's anyone who has Turbo co-existing with some library other than UJS, it's less likely to break for them.

Personally I think restoring the behaviour of both those things would be the best option, since it'll make the upgrade path easier for folks. But I also agree that `turbo-rails` should provide help here, whether through deprecation warnings or (opt-in) errors where both UJS & Rails are competing, to help people work through updating their forms to be unambiguously one or the other.

dhh:
Yeah, I'd very much like to see a quick 7.2.1 shipped with a solution to this, so we don't create an expectation of the new timing being something you can rely on, when we're changing it back to ensure compatibility 👍

seanpdoyle:
I've pushed up changes that:

* revert #412 
* adjust the restored `LinkInterceptor` to accommodate tangential changes made since it was removed
* Restore behavior for `<form data-remote="true">` elements nested within `<turbo-frame>` from 7.1.0 based on https://github.com/hotwired/turbo/pull/744#issuecomment-1266872816 by adding a call to `event.stopImmediatePropagation()` to mimic the [FormInterceptor][] that existed in `7.1.0`

[FormInterceptor]: https://github.com/hotwired/turbo/blob/v7.1.0/src/core/frames/form_interceptor.ts#L31

Issue #745, opened 2022-09-28
Frames are caching too late


manuelpuyol:
Kinda related to https://github.com/hotwired/turbo/issues/643

When doing a frame navigation, the frame `src` changes _before_ the page is cached, causing some weird interactions when going back/forwards.

### Example

Frames with `data-turbo-action="advance"`

https://user-images.githubusercontent.com/11280312/192871541-55fcfbac-1845-4caa-bb02-8eb054797a56.mov

manuelpuyol:
@seanpdoyle it looks like when the frame is promoted with `advance`, we start a visit **after** the frame is already loaded, so the PageView snapshot cache is running with the new HTML instead of the original one

seanpdoyle:
@manuelpuyol thank you for opening this issue and providing instructions on how to reproduce.

I've pushed up a breaking test to https://github.com/hotwired/turbo/compare/seanpdoyle:fix-issue-745, and am investigating a resolution.

manuelpuyol:
Thanks for looking into this @seanpdoyle !

I opened a possible solution here https://github.com/hotwired/turbo/pull/746, gonna pull in your tests to check if they pass

PR #746, opened 2022-09-28
Cache PageSnapshot with original HTML from the frame's page

manuelpuyol:
Closes https://github.com/hotwired/turbo/issues/745

Before this change, when a promoted frame finished its navigation, a visit was started and the visit would add a new PageSnapshot to the cache. The problem is that this snapshot was looking at a page that was already modified by the frame navigation, so the snapshot cache would be incorrect.

With this proposed change, we are taking a PageSnapshot as soon as the frame navigation starts, so we guarantee that the page hasn't been modified yet. Then, I'm passing this snapshot forward to the Visit and using it as the correct cache instead of calculating a new one.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -358,7 +358,7 @@ export class FrameController
 
   private navigateFrame(element: Element, url: string, submitter?: HTMLElement) {
     const frame = this.findFrameElement(element, submitter)
-    this.pageSnapshot = PageSnapshot.fromElement(element).clone()
+    this.pageSnapshot = PageSnapshot.fromElement(frame).clone()
 
     this.proposeVisitIfNavigatedWithAction(frame, element, submitter)
 
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -358,6 +358,7 @@ export class FrameController
 
   private navigateFrame(element: Element, url: string, submitter?: HTMLElement) {
     const frame = this.findFrameElement(element, submitter)
+    this.pageSnapshot = PageSnapshot.fromElement(element).clone()
 
     this.proposeVisitIfNavigatedWithAction(frame, element, submitter)
 
@@ -569,7 +570,6 @@ export class FrameController
   }
 
   private withCurrentNavigationElement(element: Element, callback: () => void) {
-    this.pageSnapshot = PageSnapshot.fromElement(element).clone()
     this.currentNavigationElement = element
     callback()
     delete this.currentNavigationElement
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -31,6 +31,7 @@ import { isAction, Action } from "../types"
 import { VisitOptions } from "../drive/visit"
 import { TurboBeforeFrameRenderEvent } from "../session"
 import { StreamMessage } from "../streams/stream_message"
+import { PageSnapshot } from "../drive/page_snapshot"
 
 type VisitFallback = (location: Response | Locatable, options: Partial<VisitOptions>) => Promise<void>
 export type TurboFrameMissingEvent = CustomEvent<{ response: Response; visit: VisitFallback }>
@@ -64,6 +65,7 @@ export class FrameController
   readonly restorationIdentifier: string
   private previousFrameElement?: FrameElement
   private currentNavigationElement?: Element
+  pageSnapshot?: PageSnapshot
 
   constructor(element: FrameElement) {
     this.element = element
@@ -382,6 +384,7 @@ export class FrameController
             willRender: false,
             updateHistory: false,
             restorationIdentifier: this.restorationIdentifier,
+            pageSnapshot: this.pageSnapshot,
           }
 
           if (this.action) options.action = this.action
@@ -566,6 +569,7 @@ export class FrameController
   }
 
   private withCurrentNavigationElement(element: Element, callback: () => void) {
+    this.pageSnapshot = PageSnapshot.fromElement(session.view.element).clone()
     this.currentNavigationElement = element
     callback()
     delete this.currentNavigationElement
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -384,7 +384,7 @@ export class FrameController
             willRender: false,
             updateHistory: false,
             restorationIdentifier: this.restorationIdentifier,
-            pageSnapshot: this.pageSnapshot,
+            snapshot: this.pageSnapshot,
           }
 
           if (this.action) options.action = this.action
@@ -569,7 +569,7 @@ export class FrameController
   }
 
   private withCurrentNavigationElement(element: Element, callback: () => void) {
-    this.pageSnapshot = PageSnapshot.fromElement(session.view.element).clone()
+    this.pageSnapshot = PageSnapshot.fromElement(element).clone()
     this.currentNavigationElement = element
     callback()
     delete this.currentNavigationElement
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -767,7 +767,7 @@ test("test navigating frame with button[data-turbo-action=advance] pushes URL st
   assert.ok(await hasSelector(page, "#frame[complete]"), "marks the frame as [complete]")
 })
 
-test.only("test navigating back after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames previous contents", async ({
+test("test navigating back after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames previous contents", async ({
   page,
 }) => {
   await page.click("#add-turbo-action-to-frame")
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -767,7 +767,7 @@ test("test navigating frame with button[data-turbo-action=advance] pushes URL st
   assert.ok(await hasSelector(page, "#frame[complete]"), "marks the frame as [complete]")
 })
 
-test("test navigating back after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames previous contents", async ({
+test.only("test navigating back after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames previous contents", async ({
   page,
 }) => {
   await page.click("#add-turbo-action-to-frame")
@@ -778,12 +778,11 @@ test("test navigating back after pushing URL state from a turbo-frame[data-turbo
 
   const title = await page.textContent("h1")
   const frameTitle = await page.textContent("#frame h2")
-  const src = new URL((await attributeForSelector(page, "#frame", "src")) || "")
 
   assert.equal(title, "Frames")
   assert.equal(frameTitle, "Frames: #frame")
   assert.equal(pathname(page.url()), "/src/tests/fixtures/frames.html")
-  assert.equal(src.pathname, "/src/tests/fixtures/frames/frame.html")
+  assert.equal(await attributeForSelector(page, "#frame", "src"), null)
   assert.equal(await propertyForSelector(page, "#frame", "src"), null)
 })
 
--- a/src/core/drive/page_view.ts
+++ b/src/core/drive/page_view.ts
@@ -39,12 +39,12 @@ export class PageView extends View<HTMLBodyElement, PageSnapshot, PageViewRender
     this.snapshotCache.clear()
   }
 
-  async cacheSnapshot(pageSnapshot?: PageSnapshot) {
-    if (this.shouldCacheSnapshot(pageSnapshot)) {
+  async cacheSnapshot(snapshot: PageSnapshot = this.snapshot) {
+    if (snapshot.isCacheable) {
       this.delegate.viewWillCacheSnapshot()
-      const { snapshot, lastRenderedLocation: location } = this
+      const { lastRenderedLocation: location } = this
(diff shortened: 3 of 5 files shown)


src/tests/functional/frame_tests.ts
```diff
@@ -778,12 +778,11 @@ test("test navigating back after pushing URL state from a turbo-frame[data-turbo
 
   const title = await page.textContent("h1")
   const frameTitle = await page.textContent("#frame h2")
-  const src = new URL((await attributeForSelector(page, "#frame", "src")) || "")
 
   assert.equal(title, "Frames")
   assert.equal(frameTitle, "Frames: #frame")
   assert.equal(pathname(page.url()), "/src/tests/fixtures/frames.html")
-  assert.equal(src.pathname, "/src/tests/fixtures/frames/frame.html")
+  assert.equal(await attributeForSelector(page, "#frame", "src"), null)
```
manuelpuyol, line 785: this was actually testing the _incorrect_ behavior 🤷

src/tests/functional/frame_tests.ts
```diff
@@ -767,7 +767,7 @@ test("test navigating frame with button[data-turbo-action=advance] pushes URL st
   assert.ok(await hasSelector(page, "#frame[complete]"), "marks the frame as [complete]")
 })
 
-test("test navigating back after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames previous contents", async ({
+test.only("test navigating back after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames previous contents", async ({
```
seanpdoyle, line 770: Did the suite pass with this qualifier removed?

seanpdoyle (approved):
Thank you for figuring out this changeset!

I've tested it out locally. How would you feel about some of these variable name and type tweaks?

```diff
diff --git a/src/core/drive/page_view.ts b/src/core/drive/page_view.ts
index c4ee0ee..80635ab 100644
--- a/src/core/drive/page_view.ts
+++ b/src/core/drive/page_view.ts
@@ -39,12 +39,12 @@ export class PageView extends View<HTMLBodyElement, PageSnapshot, PageViewRender
     this.snapshotCache.clear()
   }
 
-  async cacheSnapshot(pageSnapshot?: PageSnapshot) {
-    if (this.shouldCacheSnapshot(pageSnapshot)) {
+  async cacheSnapshot(snapshot: PageSnapshot = this.snapshot) {
+    if (snapshot.isCacheable) {
       this.delegate.viewWillCacheSnapshot()
-      const { snapshot, lastRenderedLocation: location } = this
+      const { lastRenderedLocation: location } = this
       await nextEventLoopTick()
-      const cachedSnapshot = (pageSnapshot || snapshot).clone()
+      const cachedSnapshot = snapshot.clone()
       this.snapshotCache.put(location, cachedSnapshot)
       return cachedSnapshot
     }
@@ -57,8 +57,4 @@ export class PageView extends View<HTMLBodyElement, PageSnapshot, PageViewRender
   get snapshot() {
     return PageSnapshot.fromElement(this.element)
   }
-
-  shouldCacheSnapshot(pageSnapshot?: PageSnapshot) {
-    return (pageSnapshot || this.snapshot).isCacheable
-  }
 }
diff --git a/src/core/drive/visit.ts b/src/core/drive/visit.ts
index 5a73c64..f99f14a 100644
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -42,6 +42,7 @@ export type VisitOptions = {
   action: Action
   historyChanged: boolean
   referrer?: URL
+  snapshot?: PageSnapshot
   snapshotHTML?: string
   response?: VisitResponse
   visitCachedSnapshot(snapshot: Snapshot): void
@@ -51,7 +52,6 @@ export type VisitOptions = {
   shouldCacheSnapshot: boolean
   frame?: string
   acceptsStreamResponse: boolean
-  pageSnapshot?: PageSnapshot
 }
 
 const defaultOptions: VisitOptions = {
@@ -62,7 +62,6 @@ const defaultOptions: VisitOptions = {
   updateHistory: true,
   shouldCacheSnapshot: true,
   acceptsStreamResponse: false,
-  pageSnapshot: undefined,
 }
 
 export type VisitResponse = {
@@ -102,7 +101,7 @@ export class Visit implements FetchRequestDelegate {
   snapshotHTML?: string
   snapshotCached = false
   state = VisitState.initialized
-  pageSnapshot?: PageSnapshot
+  snapshot?: PageSnapshot
 
   constructor(
     delegate: VisitDelegate,
@@ -118,6 +117,7 @@ export class Visit implements FetchRequestDelegate {
       action,
       historyChanged,
       referrer,
+      snapshot,
       snapshotHTML,
       response,
       visitCachedSnapshot,
@@ -125,7 +125,6 @@ export class Visit implements FetchRequestDelegate {
       updateHistory,
       shouldCacheSnapshot,
       acceptsStreamResponse,
-      pageSnapshot,
     } = {
       ...defaultOptions,
       ...options,
@@ -133,6 +132,7 @@ export class Visit implements FetchRequestDelegate {
     this.action = action
     this.historyChanged = historyChanged
     this.referrer = referrer
+    this.snapshot = snapshot
     this.snapshotHTML = snapshotHTML
     this.response = response
     this.isSamePage = this.delegate.locationWithActionIsSamePage(this.location, this.action)
@@ -142,7 +142,6 @@ export class Visit implements FetchRequestDelegate {
     this.scrolled = !willRender
     this.shouldCacheSnapshot = shouldCacheSnapshot
     this.acceptsStreamResponse = acceptsStreamResponse
-    this.pageSnapshot = pageSnapshot
   }
 
   get adapter() {
@@ -456,7 +455,7 @@ export class Visit implements FetchRequestDelegate {
 
   cacheSnapshot() {
     if (!this.snapshotCached) {
-      this.view.cacheSnapshot(this.pageSnapshot).then((snapshot) => snapshot && this.visitCachedSnapshot(snapshot))
+      this.view.cacheSnapshot(this.snapshot).then((snapshot) => snapshot && this.visitCachedSnapshot(snapshot))
       this.snapshotCached = true
     }
   }
diff --git a/src/core/frames/frame_controller.ts b/src/core/frames/frame_controller.ts
index d93d650..0681783 100644
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -384,7 +384,7 @@ export class FrameController
             willRender: false,
             updateHistory: false,
             restorationIdentifier: this.restorationIdentifier,
-            pageSnapshot: this.pageSnapshot,
+            snapshot: this.pageSnapshot,
           }
 
           if (this.action) options.action = this.action
@@ -569,7 +569,7 @@ export class FrameController
   }
 
   private withCurrentNavigationElement(element: Element, callback: () => void) {
-    this.pageSnapshot = PageSnapshot.fromElement(session.view.element).clone()
+    this.pageSnapshot = PageSnapshot.fromElement(element).clone()
     this.currentNavigationElement = element
     callback()
     delete this.currentNavigationElement
diff --git a/src/tests/functional/frame_navigation_tests.ts b/src/tests/functional/frame_navigation_tests.ts
index c01a043..fbdac2a 100644
--- a/src/tests/functional/frame_navigation_tests.ts
+++ b/src/tests/functional/frame_navigation_tests.ts
@@ -1,5 +1,5 @@
 import { test } from "@playwright/test"
-import { getFromLocalStorage, nextBeat, nextEventNamed, nextEventOnTarget, pathname } from "../helpers/page"
+import { getFromLocalStorage, nextEventNamed, nextEventOnTarget, pathname } from "../helpers/page"
 import { assert } from "chai"
 
 test("test frame navigation with descendant link", async ({ page }) => {
@@ -56,28 +56,30 @@ test("test promoted frame navigations are cached", async ({ page }) => {
   await page.goto("/src/tests/fixtures/tabs.html")
 
   await page.click("#tab-2")
-  await nextEventNamed(page, "turbo:frame-render")
+  await nextEventOnTarget(page, "tab-frame", "turbo:frame-load")
+  await nextEventNamed(page, "turbo:load")
 
   assert.equal(await page.textContent("#tab-content"), "Two")
   assert.equal(pathname((await page.getAttribute("#tab-frame", "src")) || ""), "/src/tests/fixtures/tabs/two.html")
   assert.equal(await page.getAttribute("#tab-frame", "complete"), "", "sets [complete]")
 
   await page.click("#tab-3")
-  await nextEventNamed(page, "turbo:frame-render")
+  await nextEventOnTarget(page, "tab-frame", "turbo:frame-load")
+  await nextEventNamed(page, "turbo:load")
 
   assert.equal(await page.textContent("#tab-content"), "Three")
   assert.equal(pathname((await page.getAttribute("#tab-frame", "src")) || ""), "/src/tests/fixtures/tabs/three.html")
   assert.equal(await page.getAttribute("#tab-frame", "complete"), "", "sets [complete]")
 
   await page.goBack()
-  await nextBeat()
+  await nextEventNamed(page, "turbo:load")
 
   assert.equal(await page.textContent("#tab-content"), "Two")
   assert.equal(pathname((await page.getAttribute("#tab-frame", "src")) || ""), "/src/tests/fixtures/tabs/two.html")
   assert.equal(await page.getAttribute("#tab-frame", "complete"), "", "caches two.html with [complete]")
 
   await page.goBack()
-  await nextBeat()
+  await nextEventNamed(page, "turbo:load")
 
   assert.equal(await page.textContent("#tab-content"), "One")
   assert.equal(await page.getAttribute("#tab-frame", "src"), null, "caches one.html without #tab-frame[src]")
diff --git a/src/tests/functional/frame_tests.ts b/src/tests/functional/frame_tests.ts
index 517165e..66c9610 100644
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -767,7 +767,7 @@ test("test navigating frame with button[data-turbo-action=advance] pushes URL st
   assert.ok(await hasSelector(page, "#frame[complete]"), "marks the frame as [complete]")
 })
 
-test.only("test navigating back after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames previous contents", async ({
+test("test navigating back after pushing URL state from a turbo-frame[data-turbo-action=advance] restores the frames previous contents", async ({
   page,
 }) => {
   await page.click("#add-turbo-action-to-frame")
```

src/tests/functional/frame_tests.ts
manuelpuyol: ha, I ha this to test it locally but forgot to remove it

seanpdoyle (approved):

src/core/frames/frame_controller.ts
```diff
@@ -566,6 +569,7 @@ export class FrameController
   }
 
   private withCurrentNavigationElement(element: Element, callback: () => void) {
+    this.pageSnapshot = PageSnapshot.fromElement(session.view.element).clone()
```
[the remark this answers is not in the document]
seanpdoyle: That's surprising to me, the `PageSnapshot.fromElement` implementation reaches through the `Element` argument to access its [ownerDocument](https://github.com/hotwired/turbo/blob/732db00a3a50a449854c00a2132ef27dc8d24e43/src/core/drive/page_snapshot.ts#L12), which should be the same document in both the `element` and `session.view.element` case.

Are they owned by different `Document` instances?

manuelpuyol:
@dhh would you mind taking a look?

src/core/frames/frame_controller.ts
```diff
@@ -566,6 +569,7 @@ export class FrameController
   }
 
   private withCurrentNavigationElement(element: Element, callback: () => void) {
+    this.pageSnapshot = PageSnapshot.fromElement(element).clone()
```
seanpdoyle, line 572: @manuelpuyol should we have a corresponding `delete this.pageSnapshot`?

```diff
diff --git a/src/core/frames/frame_controller.ts b/src/core/frames/frame_controller.ts
index bc92544..26c01db 100644
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -553,6 +553,7 @@ export class FrameController
     this.currentNavigationElement = element
     callback()
     delete this.currentNavigationElement
+    delete this.pageSnapshot
   }
 }
 ```

Comment (seanpdoyle) [reply src/core/frames/frame_controller.ts:572]: Hmm, I tested that idea out locally and tests fail. I can investigate that in the future.

In the meantime, I think cleaning up the assignment would maintain the symmetry in the `withCurrentNavigationElement` implementation.

Comment (manuelpuyol) [reply src/core/frames/frame_controller.ts:572]: from what I investigated, this is the last call before the frame `src` changes, that's why I added it there
In this case `callback()` is going to change the src and we don't want to cache it with the updated attribute

Comment (manuelpuyol) [reply src/core/frames/frame_controller.ts:572]: about the `delete`, the problem would be that the `delete` would run before `proposeVisitIfNavigatedWithAction`, which uses the pageSnapshot to pass it forward to the visit

Comment (manuelpuyol) [reply src/core/frames/frame_controller.ts:572]: we could probably move the snapshot assignment to `navigateFrame`

Comment (manuelpuyol) [reply src/core/frames/frame_controller.ts]: done!

File: src/core/frames/frame_controller.ts [hunk 5]
```diff
@@ -358,6 +358,7 @@ export class FrameController
 
   private navigateFrame(element: Element, url: string, submitter?: HTMLElement) {
     const frame = this.findFrameElement(element, submitter)
+    this.pageSnapshot = PageSnapshot.fromElement(element).clone()
```
seanpdoyle, line 361: It'll end up assigning the same `Document` instance, but how do you feel about passing the `frame` instead of the `element` here?

seanpdoyle (approved):

PR #752, opened 2022-10-03
Merge in remaining header elements (eg: link icons, manifest, meta and title)

### LINKED ISSUE #719: Turbo recreates link elements in head on every page transition
Thanks for the great turbo project!

The HTML head section might contain link elements like [favicons](https://developer.mozilla.org/en-US/docs/Learn/HTML/Introduction_to_HTML/The_head_metadata_in_HTML#adding_custom_icons_to_your_site) and [preload](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload). Turbo [replace them](https://github.com/hotwired/turbo/blob/main/src/core/drive/page_renderer.ts#L65-L66) on every transition and browsers reloading them in that case. Depending on the browser and privacy settings, this results in local cache or additional fetch request. Turbo should handle them like scripts, compare the signature and only remove/create them on changes.

Example HTML:
```html
<html>
  <head>
    <link rel="shortcut icon" href="/some-icon.png" />
    <link rel="preload" as="font" type="font/woff2" href="/some-font.woff2" />
    ....
```
Not sure if there is a way to create a workaround. The `turbo:before-render` event provides only [access to the body](https://github.com/hotwired/turbo/blob/2332d3d0153886ff0c616198e377687a76ac5226/src/core/session.ts#L360) and not the head. Any suggestions how to handle this?

scuml:
Resolves #719 

When turbo loads a page, the current behavior is to remove and replace all provisional head elements.  This would include the title tag, meta tags, and link tags to icons, fonts and manifest.json.  This can lead to additional http requests as demonstrated below:  

https://user-images.githubusercontent.com/1790447/193615680-850b415e-5e99-4cc3-ba37-f9c5da353404.mp4

This PR merges the elements instead providing a cleaner page load experience and preventing these potential additional requests.

https://user-images.githubusercontent.com/1790447/193615692-ce90c1c1-91d8-4d78-8350-6839d16b8aac.mp4

**Unit test note:**
I was not able to get the test suite running and encountered this issue:

```
Tunnel started
```

If this PR has potential, I could use some assistance getting things up and running.

### DIFF
--- a/src/core/drive/page_renderer.ts
+++ b/src/core/drive/page_renderer.ts
@@ -60,10 +60,10 @@ export class PageRenderer extends Renderer<HTMLBodyElement, PageSnapshot> {
   }
 
   async mergeHead() {
+    const mergedHeadElements = this.mergeProvisionalElements()
     const newStylesheetElements = this.copyNewHeadStylesheetElements()
     this.copyNewHeadScriptElements()
-    this.removeCurrentHeadProvisionalElements()
-    this.copyNewHeadProvisionalElements()
+    await mergedHeadElements
     await newStylesheetElements
   }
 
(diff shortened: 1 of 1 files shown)


seanpdoyle:
@dhh could you grant permission to run CI for this change?

scuml:
> @dhh could you grant permission to run CI for this change?
@dhh ^

dhh:
Hmm. It’s not offering the choice to run it on this PR. Not quite sure why. We have it on for other PRs.

scuml:
Warning disappeared for me too.  Pushed a spacing change and it is back.  Could you try once more?

scuml:
Thanks.  I need another manual trigger now that I've fixed the lint warnings.

synth:
Hello. I'm noticing that this work will remove HEAD elements such as CSP meta nonces (and others) when a Turbo Frame request is made with `action: "advance"`. It may occur with other modes as well, not sure, this is just what I'm seeing. 

What I'm seeing is that the Turbo Frame request will render this layout `turbo-rails-2.0.5/app/views/layouts/turbo_rails/frame.html.erb` which ends up having an empty head which combines with the code in this PR which seems to check elements in the new HEAD (which is empty) and remove elements that aren't there.

It took a while to figure this out and find a workaround which is to ensure a full application layout with a proper HEAD section is rendered for these Turbo Frame requests, but that doesn't seem right, unless I'm missing something. 

I can try to put together a reproducible test app but wanted to gauge the temperature first :) Thanks.

PR #755, opened 2022-10-07
Avoid bubbling form submissions that target a new window/tab

### LINKED ISSUE #754: Submitting a form with `[target=_blank]` ignores the target
When submitting the following form

```html
<form id="targeted-form-post" method="post" action="/__turbo/redirect" target="_blank">
  <input type="hidden" name="path" value="/src/tests/fixtures/one.html">
  <input type="submit" id="form-post-submit" value="Submit with target"/>
</form>
```

Turbo treats it as a normal submission and ignores the target. If `data-turbo="false"` is added to the form, it behaves as I would expect, triggering the submission from a new tab/window.

martinzamuner:
Resolves #754 

Avoid bubbling `<form>` element submissions when they target a new window/tab. This occurs when:

* a `<form>` declares a [target](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#attr-target) of `_blank`
* a `<form>` element's submitting `<button>` element declares a [formtarget](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-formtarget) of `_blank`
* a `<form>` element's submitting `<input type="submit">` element declares a [formtarget](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#formtarget) of `_blank`

Based on #389

### DIFF
--- a/src/observers/link_click_observer.ts
+++ b/src/observers/link_click_observer.ts
@@ -38,7 +38,7 @@ export class LinkClickObserver {
     if (event instanceof MouseEvent && this.clickEventIsSignificant(event)) {
       const target = (event.composedPath && event.composedPath()[0]) || event.target
       const link = this.findLinkFromClickTarget(target)
-      if (link && doesNotTargetIFrame(link)) {
+      if (link && doesNotHaveSpecificTarget(link)) {
         const location = this.getLocationForLink(link)
         if (this.delegate.willFollowLinkToLocation(link, location, event)) {
           event.preventDefault()
@@ -71,10 +71,6 @@ export class LinkClickObserver {
   }
 }
 
-function doesNotTargetIFrame(anchor: HTMLAnchorElement): boolean {
-  for (const element of document.getElementsByName(anchor.target)) {
-    if (element instanceof HTMLIFrameElement) return false
-  }
-
-  return true
+function doesNotHaveSpecificTarget(anchor: HTMLAnchorElement): boolean {
+  return anchor.target != "_self"
 }
--- a/src/observers/link_click_observer.ts
+++ b/src/observers/link_click_observer.ts
@@ -72,5 +72,7 @@ export class LinkClickObserver {
 }
 
 function doesNotHaveSpecificTarget(anchor: HTMLAnchorElement): boolean {
-  return !anchor.target || anchor.target == "_self"
+  const target = anchor.getAttribute("target")
+
+  return !target || target == "_self"
 }
--- a/src/observers/link_click_observer.ts
+++ b/src/observers/link_click_observer.ts
@@ -62,7 +62,7 @@ export class LinkClickObserver {
 
   findLinkFromClickTarget(target: EventTarget | null) {
     if (target instanceof Element) {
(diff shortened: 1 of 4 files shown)


seanpdoyle:
@martinzamuner thank you for proposing this change!

When I'd proposed the `submissionDoesNotTargetIFrame` method, I'd forgotten about `_blank`:
 https://github.com/hotwired/turbo/blob/732db00a3a50a449854c00a2132ef27dc8d24e43/src/observers/form_submit_observer.ts#L59-L67

Instead of adding a new method, what do you think about combining `submissionDoesNotTargetIFrame` and `submissionDoesNotTargetBlankWindow` so that it checks for the presence of _any_ `[target]` on the `<form>` or `[formtarget]` on the submitter?

Similarly, could you add coverage for the same scenario for `<a>` elements and make similar changes to the `LinkClickObserver`?

https://github.com/hotwired/turbo/blob/732db00a3a50a449854c00a2132ef27dc8d24e43/src/observers/link_click_observer.ts#L74-L80?

src/observers/link_click_observer.ts
```diff
@@ -71,10 +71,6 @@ export class LinkClickObserver {
...
 }
 
-function doesNotTargetIFrame(anchor: HTMLAnchorElement): boolean {
-  for (const element of document.getElementsByName(anchor.target)) {
-    if (element instanceof HTMLIFrameElement) return false
-  }
-
-  return true
+function doesNotHaveSpecificTarget(anchor: HTMLAnchorElement): boolean {
+  return anchor.target != "_self"
```
seanpdoyle, line 75: If we're going to have a special case handler for `_self`, could we also add test coverage to guarantee that behavior and guard against regressions?

martinzamuner:
@seanpdoyle Sure thing! I've added a special case for `_self`, which I think should behave the same as having no target. Let me know if you think this is unnecessary.

There was already a test for this behavior with anchors: https://github.com/hotwired/turbo/blob/main/src/tests/functional/navigation_tests.ts#L208-L213.

src/observers/link_click_observer.ts
martinzamuner: Done! BTW, I had to remove the check on `target` from `LinkClickObserver#findLinkFromClickTarget` to avoid a full-page load in the case of `[target=_self]`.

seanpdoyle (approved):

martinzamuner:
@dhh Thanks for running the tests! There were a couple failures related to anchors inside SVGs behaving kind of weird. Their `target` is implemented through an [`SVGAnimatedString`](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString), which I didn't know. Using the `getAttribute` function fixed those.

Aside from that, those same tests were now broken because I added some elements before the SVGs in the fixture, and they were not above the fold anymore. That means that `page.click` would have to scroll to the anchors inside, but was failing to do so. I'm not exactly sure why (maybe something to do with [this](https://github.com/microsoft/playwright/issues/15566))? My solution was to scroll to the next element before attempting the click. It's not very elegant, though. Suggestions are welcomed.

martinzamuner:
Fixed by https://github.com/hotwired/turbo/pull/835.

PR #758, opened 2022-10-11
Add better support for shadow DOM

yuki24:
This PR adds better support for the Shadow DOM integration. Right now, Turbo fails to find the closest link when the click target is within the shadow DOM but the link target is not in the same shadow DOM.

turbo-shodow-dom

This PR changes the `findLinkFromClickTarget` function so that it can cross the boundary and continues to reverse-traverse until it gets out of the shadow roots.

### DIFF
--- a/src/observers/link_click_observer.ts
+++ b/src/observers/link_click_observer.ts
@@ -1,4 +1,5 @@
 import { expandURL } from "../core/url"
+import { findClosestRecursively } from "../util"
 
 export interface LinkClickObserverDelegate {
   willFollowLinkToLocation(link: Element, location: URL, event: MouseEvent): boolean
@@ -60,10 +61,8 @@ export class LinkClickObserver {
     )
   }
 
-  findLinkFromClickTarget(target: EventTarget | null) {
-    if (target instanceof Element) {
-      return target.closest<HTMLAnchorElement>("a[href]:not([target^=_]):not([download])")
-    }
+  findLinkFromClickTarget(target: EventTarget | null): HTMLAnchorElement | undefined {
+    return findClosestRecursively<HTMLAnchorElement>(target as Element, "a[href]:not([target^=_]):not([download])")
   }
 
   getLocationForLink(link: Element): URL {
--- a/src/core/session.ts
+++ b/src/core/session.ts
@@ -14,7 +14,7 @@ import { StreamMessage } from "./streams/stream_message"
 import { StreamMessageRenderer } from "./streams/stream_message_renderer"
 import { StreamObserver } from "../observers/stream_observer"
 import { Action, Position, StreamSource } from "./types"
-import { clearBusyState, dispatch, getVisitAction, markAsBusy } from "../util"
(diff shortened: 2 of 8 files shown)


izaguirrejoe:
Hey @yuki, I encountered this problem myself. This bug breaks Turbo Native apps that use Web Components, so definitely an important fix. Here's my implementation, which uses recursion. I'd love to know what people think: 

```typescript
  findLinkFromClickTarget(target: EventTarget | null): HTMLAnchorElement | null {
    if (target instanceof Element) {
      return target.closest<HTMLAnchorElement>("a[href]:not([target^=_]):not([download])") || this.findLinkFromClickTarget((target.getRootNode() as ShadowRoot)?.host)
    }
    else{
      return null
    }
  }
```

Typescript is forcing me to provide the else statement to provide an explicit return type since I'm referencing the function in itself, otherwise my method would simply be adding an OR operator to the original code. 

Maybe the tests can be simplified a bit? Rather than creating an entirely new fixture file, you can create a custom element in [src/tests/fixtures/navigation.html](https://github.com/hotwired/turbo/pull/758#diff-0b53d3ae36fc0091d05b11c0c56c48e0ab150659d0d9ddfe12770052eef9b85c). I took inspiration from #351 in thinking of a way to test custom elements. https://github.com/hotwired/turbo/pull/351/files

yuki24:
Thanks, @izaguirrejoe! All great suggestions. The recursion definitely looks simpler. The test can totally be merged into the navigation test.

src/observers/link_click_observer.ts
```diff
@@ -60,9 +60,10 @@ export class LinkClickObserver {
     )
   }
 
-  findLinkFromClickTarget(target: EventTarget | null) {
+  findLinkFromClickTarget(target: EventTarget | null): HTMLAnchorElement | undefined {
```
yuki24, line 63: @izaguirrejoe Explicitly specifying the return type removed the need of having to declare the `else { ... }` statement. This type used to be `HTMLAnchorElement | null | undefined` and I don't think changing that to `HTMLAnchorElement | undefined` will cause any issues.

izaguirrejoe: I agree, that looks simpler! At first I felt uncomfortable returning undefined, but that's what the original function did anyways, and it make the PR a one-liner.

src/observers/link_click_observer.ts
```diff
@@ -60,9 +60,10 @@ export class LinkClickObserver {
...
     if (target instanceof Element) {
-      return target.closest<HTMLAnchorElement>("a[href]:not([target^=_]):not([download])")
+      return target.closest<HTMLAnchorElement>("a[href]:not([target^=_]):not([download])") ||
+        this.findLinkFromClickTarget(target.assignedSlot || (target.getRootNode() as ShadowRoot)?.host)
```
yuki24, line 66: While I was looking into #801 I noticed another edge case where an element in the `<slot>` may not be able to reverse-traverse, so I had to add `target.assignedSlot` to cover that case as well. There are also two additional tests to cover this case.

PR #759, opened 2022-10-11
Reload page if failed form changes tracked content

kevinmcconnell:
When a form submission fails with a client error, the response from the form replaces the current page content. In Turbo 7.1, whenever this replacement would affect resources marked with `data-turbo-track="reload"`, a reload of the page is triggered instead.

In 7.2.0, we do not perform this reload. Instead, the form's response is discarded. This happens because the content replacement does not occur during a visit, and so we have no new location to reload.

This difference can affect situations like expired sessions, where reloading the page would cause a new session to be established. If we instead allow the response to be discarded, actions can fail silently.

This commit changes the behaviour to reload the current page when `reload` is called outside of a current visit, in order to match the behaviour of 7.1, and to not discard the response from the form submission.

### DIFF
--- a/src/core/native/browser_adapter.ts
+++ b/src/core/native/browser_adapter.ts
@@ -123,9 +123,7 @@ export class BrowserAdapter implements Adapter {
   reload(reason: ReloadReason) {
     dispatch("turbo:reload", { detail: reason })
 
-    if (!this.location) return
-
-    window.location.href = this.location.toString()
+    window.location.href = this.location?.toString() || window.location.href
   }
 
   get navigator() {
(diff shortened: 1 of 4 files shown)


packagethief (approved):
Thanks @kevinmcconnell.

It looks like this changed in https://github.com/hotwired/turbo/pull/601. This updated the browser adapter to depend on the its own `location` (assigned during visit proposal) instead of `window.location` when reloading the page. The thing is, we won't have a visit if we're rendering in response to a failed form submission. Falling back to `window.location` seems reasonable to restore the previous behavior.

Ideally we'd use `window.location` always in the context of reloading and page invalidation. The intent is that it behave exactly like the browser's native "reload". It operates on the _current_ location, not the intended one.

kevinmcconnell:
@packagethief thanks for the review! 🙏

> Ideally we'd use window.location always in the context of reloading and page invalidation.

Yes, I totally agree. I think it's worth shipping this small fix first to match the previous behaviour. But we can also look into removing the `location` from the browser adapter, as it would simplify things a little if it could have a single concept of the current location in all cases.

packagethief:
> I think it's worth shipping this small fix first to match the previous behaviour

Definitely.

PR #764, opened 2022-10-13
Let developers associate lifecycle events

domchristie:
Currently there is no way to know which events are tied to a given user action. For example, it's not possible to associate a `turbo:click` with a `turbo:visit`. One might link a `turbo:click` with the next `turbo:visit`, but this isn't accurate because a navigation can be ended by cancelling `turbo:click` or `turbo:before-visit` (as mentioned by @pfeiffer in https://github.com/hotwired/turbo/issues/99#issuecomment-757535307).

With a lifecycle ID present in each event, it'll be possible to reference the initiator element, providing a way for developers to customise visit behaviour based on HTML attributes. For example:

```js
let initiator = []
addEventListener('turbo:click', ({ detail }) =>
  initiator = [detail.lifecycleIdentifier, detail.originalEvent.target]
)
addEventListener('turbo:visit', ({ detail }) => {
  const [id, element] = initiator
  if (detail.lifecycleIdentifier === id && element.dataset.animate) {
    // do something
  }
}) 
```

It could also be used to determine redirect visits, as `turbo:visit` is dispatched twice:

```js
let lifecycleIdentifier
addEventListener('turbo:visit', ({ detail }) =>
  const isRedirect = detail.lifecycleIdentifier === lifecycleIdentifier
  lifecycleIdentifier = detail.lifecycleIdentifier
)
```

This is a lower-level alternative to https://github.com/hotwired/turbo/pull/750, and whilst I still think that the type annotations in that PR are worthwhile, this might provide some useful information for other cases without having to manipulate `VisitOptions` too much.

If we think this approach is worth pursuing, I can finish it up by adding identifiers to the remaining events.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -8,7 +8,7 @@ import { Visit, VisitDelegate, VisitOptions } from "./visit"
 import { PageSnapshot } from "./page_snapshot"
 
 export type NavigatorDelegate = VisitDelegate & {
-  allowsVisitingLocationWithAction(location: URL, action?: Action): boolean
+  allowsVisitingLocation(location: URL, options: Partial<VisitOptions>): boolean
   visitProposedToLocation(location: URL, options: Partial<VisitOptions>): void
   notifyApplicationAfterVisitingSamePageLocation(oldURL: URL, newURL: URL): void
 }
@@ -24,7 +24,7 @@ export class Navigator {
   }
 
   proposeVisit(location: URL, options: Partial<VisitOptions> = {}) {
-    if (this.delegate.allowsVisitingLocationWithAction(location, options.action)) {
+    if (this.delegate.allowsVisitingLocation(location, options)) {
       if (locationIsVisitable(location, this.view.snapshot.rootLocation)) {
         this.delegate.visitProposedToLocation(location, options)
       } else {
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -322,6 +322,7 @@ export class Visit implements FetchRequestDelegate {
       this.adapter.visitProposedToLocation(this.redirectedToLocation, {
         action: "replace",
         response: this.response,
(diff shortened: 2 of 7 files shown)


seanpdoyle:
Thank you for exploring this option @domchristie!

While I appreciate the simplicity of passing around a uuid string, I wonder how likely it is that applications that adopt this pattern are going to want more and more details about the details from events that precede any given action.

The hypothetical you shared is related to the lifecycle of a `Visit` instance. While I'd really prefer to not expose internal objects within `CustomEvent.detail`, I wonder if there is a middle ground between sharing the `Visit` instance across related events and sharing an abstract uuid string.

Would it be worth pushing each `Event` instance onto an `Array`? Capturing an object that adheres to a `Delegate` interface (maybe like `VisitDelegate`)? Something else entirely?

domchristie:
Thanks for reviewing @seanpdoyle.

Yes, the idea for a just a UUID string stems from the issue of passing around complex objects that don't conform to the structured clone algorithm (discovered by @kevinmcconnell in https://github.com/hotwired/turbo/issues/720), which caused an error when communicating with the iOS adapter.

Related: when augmenting the code, it's really inviting to add arbitrary properties to `VisitOptions`. Many of these make sense in terms of modelling (e.g. `Visit#initiator:Element` feels natural), and can be _really_ handy because `VisitOptions` are passed around everywhere prior to a `Visit`, however, this is risky because of the above.

The other consideration is that there currently isn't a model that encompasses the pre-visit events like `turbo:click`. Perhaps this is a `Lifecycle` or `Navigation`?

domchristie:
> Would it be worth pushing each Event instance onto an Array? Capturing an object that adheres to a Delegate interface (maybe like VisitDelegate) during a turbo:click, then forwarding that along? Something else entirely?

@seanpdoyle just to clarify, are you suggesting that this array of events (or alternative) be available in the `Event` object, or referencable via some other API, e.g `Turbo.currentLifecycle.events`?

PR #767, opened 2022-10-17
Fix navigation when the page contains an iframe with an empty name

lserman:
Links/forms are currently being disabled if any iframe exists on the page with an empty name. This is unfortunately more common than you'd think due to browser extensions commonly injecting iframes carelessly into the page body.

The `target` attribute for both form and link elements returns a blank string when they are not set, so the existing checks since 7.2 were improperly catching any `<iframe name>` elements and disabling the submissions/navigations.


seanpdoyle:
Thank you for opening this pull request!

It seems to overlap with https://github.com/hotwired/turbo/pull/755. Are you able to reproduce your issues with the changes made by https://github.com/hotwired/turbo/pull/755?

PR #770, opened 2022-10-19
Use `devEngines` to relax the minimum required `node` version in external projects

cvkef:
This PR moves the minimum required `node` version under a `devEngines` section at the `package.json`.

The introduction of the [playwright][] test suite forced the minimum required `node` version to v14. As a side effect, using the `engines` section of the `package.json` to set this dependency, introduces an unreasonable constraint to use the `turbo` library in other projects. Considering that `playwright` is a development only dependency, it seems justified to keep it under a `devEngines` section of the `package.json`.

[playwright]: https://github.com/hotwired/turbo/pull/609

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "release": "npm publish && git commit -am \"$npm_package_name v$npm_package_version\" && git push",
     "lint": "eslint . --ext .ts"
   },
-  "engines": {
+  "devEngines": {
     "node": ">= 14"
   }
 }


marcoroth:
Maybe I'm missing something, but it seems like `devEngines` is not (yet) supported in npm/yarn

cvkef:
@marcoroth you're not missing anything. The `devEngines` is not an official object of the `package.json` and it's not supported either by `npm` or `yarn`. It is used, though, to workaround the described issue. An example is the [facebook/react](https://github.com/facebook/react/blob/main/package.json#L103) library.

marcoroth:
@cvkef gotcha! But how are the engines defined in `devEngines` enforced then? I guess the packages/dependencies itself have to manually check for it then?

cvkef:
@marcoroth Unless you include a helper script to utilize the `devEngines` [^1], then yes, you have to check the devDependencies manually. Or, even go with nvm and .nvmrc.

But the point here is that the `engines` directive targets the production dependencies, or as the npm docs [^2] describe it
> **engines**
> You can specify the version of node that your stuff works on:

The way I understand it, `engines` is not about safeguarding the node version for the development dependencies. As I said you can do this with other tools.

And since `playwright` is only a devDependency, it shouldn't lock the required node version for the turbo library.

[^1]: An example is [check-dev-engines.js](https://github.com/facebook/fbjs/blob/main/packages/fbjs-scripts/node/check-dev-engines.js) by [facebook/fbjs](https://github.com/facebook/fbjs)
[^2]: npm Docs - [package-json#engines](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#engines)

cvkef:
Hey @marcoroth,

Just a follow-up to provide a bit more context and try to make clear why this is important to us. Working in a fairly large and mature RoR application, that short a while embraced the Hotwire, it can be challenging to keep up with software major version upgrades. 
That said, our application still relies on node v12 engine. And it was disappointing to realize that we can't update the `turbo-rails` gem beyond v1.1.1 because of the `hotwired/turbo` v7.2.0 node version limitation. Even though it's in our roadmap to upgrade to `node` v14, it will take some time to get there (update libraries, find alternatives where upgrade is not an option, test all libraries, etc).

brunoprietog:
Hey @cvkef, I think that in this case it's better to simply delete that key. Let me know if you prefer to do that and I can merge.

cvkef:
Hi @brunoprietog,

We've managed to upgrade our node production version so now this is a bit irrelevant. I'd have to agree though, that deleting the key is the better solution.

Thank you.

brunoprietog:
Thank you, so closing this one

PR #772, opened 2022-10-20
Check last rendered location for same-page visits

kevinmcconnell:
In [#595][] the logic for determining whether a visit is a "same page" visit was changed. Previously we were comparing the new location with the current location. In #595 we began comparing with the location of the last visit that was started.

This works great when all the visits are web visits. However, it introduced some problems with apps using the native adapters. In native apps, navigations can be a combination of native screens and WebView navigations, which means that we can't rely on the last visit having been sent.

This manifests as failing navigations in certain situations, such as navigating back from a native modal to a WebView. It also caused some problems restoring scroll positions when navigating back from one WebView to another.

Restoring the original logic prevents these problems.

[#595]: https://github.com/hotwired/turbo/pull/595

This reverts commit 2e344b35d983fca636e064bc5ee1801c16908a04.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -17,7 +17,6 @@ export class Navigator {
   readonly delegate: NavigatorDelegate
   formSubmission?: FormSubmission
   currentVisit?: Visit
-  lastVisit?: Visit
 
   constructor(delegate: NavigatorDelegate) {
     this.delegate = delegate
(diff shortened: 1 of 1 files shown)


seanpdoyle:
@domchristie could you please review this change.

jayohms (approved):
I worked with @kevinmcconnell on isolating and testing this change and it fixes the same-page location visits and scroll restoration issues in the mobile apps/adapters.

@domchristie A lot has changed in 7.2 since this initial fix. If this opens up the original edge-case bug, please let me know and we can test another PR. As-is, the behavior breaks critical functionality in the mobile adapters, so the revert is necessary.

domchristie (approved):
The [navigating back test](https://github.com/hotwired/turbo/blob/19ba07f4873b828e56394215519c8e17b2186a2f/src/tests/functional/navigation_tests.ts#L379-L392) is still passing, so this looks good to me.

kevinmcconnell:
Thanks @domchristie. Yes I think that other changes since your original PR have meant that the change was no longer necessary.

PR #782, opened 2022-10-28
Allow users to define an async custom render function

manuelpuyol:
Since https://github.com/hotwired/turbo/pull/431, turbo allows users to customize the render function they want to use, but when an user defines an async function that requires some kind of wait, turbo's timing gets messed up and it fires events before the render actually happened.

With these changes, Turbo will correctly handle async render functions, properly waiting for it to finish before continuing the lifecycle.

### DIFF
--- a/src/core/bardo.ts
+++ b/src/core/bardo.ts
@@ -9,14 +9,14 @@ export class Bardo {
   readonly permanentElementMap: PermanentElementMap
   readonly delegate: BardoDelegate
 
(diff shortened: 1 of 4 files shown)


manuelpuyol:
@seanpdoyle any opinions on this change?

seanpdoyle (approved):

manuelpuyol:
@dhh anything I should do to get this merged?

PR #788, opened 2022-11-04
Add cancelable turbo:frame-click event

manuelpuyol:
Similar to `turbo:click`, `turbo:frame-click` indicates when a frame link was clicked. The event can be canceled to let the browser fallback to the default click.


dhh:
What's the reason `turbo:click` can't work directly for frame clicks?

manuelpuyol:
> What's the reason `turbo:click` can't work directly for frame clicks?

there's some context in https://github.com/hotwired/turbo/issues/610

Issue #554, opened 2022-03-24
Not clearing cache when server responds with text/vnd.turbo-stream.html


tleish:
When submitting a non-GET forms, turbo clears it's cache:

```javascript
class FormSubmission {
  requestSucceededWithResponse(request, response) {
      //...
      this.delegate.formSubmissionSucceededWithResponse(this, response);
    }
  }
}

class Navigator {
  async formSubmissionSucceededWithResponse(formSubmission, fetchResponse) {
    if (formSubmission == this.formSubmission) {
      const responseHTML = await fetchResponse.responseHTML;
      if (responseHTML) {
        if (formSubmission.method != FetchMethod.get) {
          this.view.clearSnapshotCache();
        }
        // ...
      }
    }
  }
}
```

However, if the server responds with header `Content-Type: text/vnd.turbo-stream.html`, then the above logic is skipped and the cache is not cleared. 

```javascript
class StreamObserver {
  constructor(delegate) {
    //...
    this.inspectFetchResponse = event => {
      const response = fetchResponseFromEvent(event);
      if (response && fetchResponseIsStream(response)) {
        event.preventDefault();
        this.receiveMessageResponse(response);
      }
    };
    //...
  }
}

class FetchRequest {
  async receive(response) {
    const fetchResponse = new FetchResponse(response);
    const event = dispatch("turbo:before-fetch-response", {
      cancelable: true,
      detail: {
        fetchResponse: fetchResponse
      },
      target: this.target
    });
    if (event.defaultPrevented) {
      this.delegate.requestPreventedHandlingResponse(this, fetchResponse);
    } else if (fetchResponse.succeeded) {
      this.delegate.requestSucceededWithResponse(this, fetchResponse);
    } else {
      this.delegate.requestFailedWithResponse(this, fetchResponse);
    }
    return fetchResponse;
  }
}
```

When `StreamObserver` detects `fetchResponseIsStream` it calls `event.preventDefault()`.  Then in `FetchRequest#receive` follows the condition `event.defaultPrevented` which calls `FormSubmission#requestPreventedHandlingResponse` instead of `FormSubmission#requestSucceededWithResponse` (mentioned at the beginning of this post).  `FormSubmission#requestPreventedHandlingResponse` does not clear cache.

Is this intentional?  For us it feels like a bug.

Testing a possible solution, the following change clears the cache when the server responds with a turbo-stream:

```diff
- requestPreventedHandlingResponse(request, response) {
[...]

PR #797, opened 2022-11-18
Clear cache when when server responds to non-GET form submit with text/vnd.turbo-stream.html response

tleish:
Fix for #554

Submitting failing test first


src/tests/functional/navigation_tests.ts
```diff
@@ -138,19 +138,6 @@ test("test following a same-origin POST form button[data-turbo-action=replace]",
   assert.equal(await visitAction(page), "replace")
 })
 
-test("test following a POST form clears cache", async ({ page }) => {
```
tleish, line 141: Moving this test to `src/tests/functional/form_submission_tests.ts`

PR #798, opened 2022-11-20
Simplify `FetchRequestDelegate.prepareHeadersForRequest`

seanpdoyle:
With the introduction of `FetchRequest.acceptResponseType()` in [hotwired/turbo#647][], the
`FetchRequestDelegate.prepareHeadersForRequest` callback interacts with more than just the `FetchRequestHeaders` instance passes as its first argument.

Additionally, every implementation of the `FetchRequestDelegate` interface implements the `prepareHeadersForRequest` method, so the fact that it's listed as optional and invoked with a guard clause are unnecessary.

This commit renames the method to `prepareRequest`, reduces the signature to a single `FetchRequest` argument, and no longer declares it as a conditional property on the interface.

[hotwired/turbo#647]: https://github.com/hotwired/turbo/pull/647/files#diff-d4ee4683f7121e24a87566ef6854ee6090ee723a5299430338f5602febea8c1f

### DIFF
--- a/src/core/drive/form_submission.ts
+++ b/src/core/drive/form_submission.ts
@@ -1,4 +1,4 @@
-import { FetchRequest, FetchMethod, fetchMethodFromString, FetchRequestHeaders } from "../../http/fetch_request"
+import { FetchRequest, FetchMethod, fetchMethodFromString } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { expandURL } from "../url"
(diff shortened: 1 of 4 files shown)


seanpdoyle:
Hey @kevinmcconnell, this came out of some of the Turbo Stream work you've done. If you're available, I'd really appreciate some code review!

kevinmcconnell (approved):
Looks great, @seanpdoyle! Thanks for pinging me on this 🙏

PR #799, opened 2022-11-20
Re-use getVisitAction utility function

seanpdoyle:
The logic for reading an `Action` from a list of `Element` instances'
`[data-turbo-action]` attribute is still scattered across several
modules, in spite of the `getVisitAction` declared in the `util` module.

This commit replaces duplicate implementations of the `getVisitAction`
method with the utility function.

With those calls added, the `isAction` type checking utility is only
ever invoked by `getVisitAction`, so this commit moves its
implementation to the `util` module, and removes all other invocations.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -1,9 +1,9 @@
-import { Action, isAction } from "../types"
+import { Action } from "../types"
+import { getVisitAction } from "../../util"
 import { FetchMethod } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { FormSubmission } from "./form_submission"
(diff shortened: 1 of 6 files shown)


seanpdoyle:
@kevinmcconnell if you're available, I'd really appreciate some code review on this extraction.

kevinmcconnell (approved):
Looks good to me! 👍

It looks like we could isolate the uses of `isAction` to only `getVisitAction` as well. After the changes here, the only other place I see it being used is [this conditional in the frame controller](https://github.com/hotwired/turbo/blob/main/src/core/frames/frame_controller.ts#L384). But that has already called `getVisitAction` and so could be simplified to `if (this.action) { ...`. What do you think?

seanpdoyle:
@kevinmcconnell great point! I've made that change.

PR #800, opened 2022-11-21
Reorganize Turbo Events and declare events on global event map

marcoroth:
This Pull Request reorganizes the events Turbo knows about and puts them into a single `src/events.ts` file.

Additionally it declares the Turbo events on `GlobalEventHandlersEventMap` and `ElementEventMap` so that the callbacks-arguments of event-listeners can be typed.

#### Before:
```ts
import { TurboBeforeFetchRequestEvent } from "@hotwired/turbo"

document.addEventListener("turbo:before-fetch-request", (e) => {
  // `e` is typed as `Event`

  const event = e as TurboBeforeFetchRequestEvent
  // `event` is now typed as `TurboBeforeFetchRequestEvent`
})
```

#### After:
```ts
import { TurboBeforeFetchRequestEvent } from "@hotwired/turbo"

document.addEventListener("turbo:before-fetch-request", (event: TurboBeforeFetchRequestEvent) => {
  // `event` is typed as `TurboBeforeFetchRequestEvent`
})
```

Previously you would get a `No overload matches this call` error:
```
error TS2769: No overload matches this call.
  Overload 1 of 2, '(type: keyof WindowEventMap, listener: (this: Window, ev: MessageEvent<any> | Event | UIEvent | AnimationEvent | MouseEvent | ... 22 more ... | StorageEvent) => any, options?: boolean | ... 1 more ... | undefined): void', gave the following error.
```

---
Additional note: this came up as part of https://github.com/hotwired/turbo-rails/pull/392


seanpdoyle (approved):
I love this! Thank you for taking it on.

There are a bunch of event listeners in the test suite, would you be able to remove their type casting and type guards as part of this change?

marcoroth:
Good point, will do 👍🏼

marcoroth:
I'm kinda confused on why and how the tests are failing. I get the same errors locally, trying to figure out what's going on.

seanpdoyle:
@marcoroth I've encountered the CI flakiness as well, and have opened https://github.com/hotwired/turbo/pull/833 to resolve it. Could you try pulling in that change and seeing if you have better luck in CI?

seanpdoyle:
@marcoroth if you rebase the latest, you should be able to pass CI.

marcoroth:
Thanks @seanpdoyle! It seems like the failures I was seeing are gone now. Now it just fails with the same failures as on `main`

seanpdoyle:
@marcoroth 🎉 ! I've opened https://github.com/hotwired/turbo/pull/838 to try and fix `main`.

seanpdoyle:
@marcoroth if you want to get this branch passing while we wait for #838 to be merged, you can apply this diff:

```diff
diff --git a/src/tests/fixtures/navigation.html b/src/tests/fixtures/navigation.html
index 48953b7..f2fd53d 100644
--- a/src/tests/fixtures/navigation.html
+++ b/src/tests/fixtures/navigation.html
@@ -15,7 +15,7 @@
 
     <a href="#ignored-link" id="ignored-link">Skipped Content</a>
 
-    <section id="main" style="height: 200vh">
+    <section id="main">
       <h1>Navigation</h1>
       <p><a id="same-origin-unannotated-link" href="/src/tests/fixtures/one.html">Same-origin unannotated link</a></p>
       <p><a id="same-origin-unannotated-link-search-params" href="/src/tests/fixtures/one.html?key=value">Same-origin unannotated link ?key=value</a></p>
diff --git a/src/tests/functional/rendering_tests.ts b/src/tests/functional/rendering_tests.ts
index ba5555e..2ac873d 100644
--- a/src/tests/functional/rendering_tests.ts
+++ b/src/tests/functional/rendering_tests.ts
@@ -88,6 +88,7 @@ test("test reloads when tracked elements change due to failed form submission",
   })
 
   await page.click("#tracked-asset-change-form button")
+  await nextBeat()
 
   const reason = await page.evaluate(() => localStorage.getItem("reason"))
   const unloaded = await page.evaluate(() => localStorage.getItem("unloaded"))
```

marcoroth:
Thanks @seanpdoyle 🙏🏼

Looking good now!

src/events.ts
```diff
@@ -0,0 +1,86 @@
...
+  "turbo:frame-load": TurboFrameLoadEvent
+  "turbo:frame-render": TurboFrameRenderEvent
+  "turbo:frame-missing": TurboFrameMissingEvent
+  "turbo:submit-start": TurboSubmitStartEvent
+  "turbo:submit-end": TurboSubmitEndEvent
+}
+
+declare global {
+  interface DocumentEventMap extends TurboGlobalEventMap {}
+  interface WindowEventMap extends TurboGlobalEventMap {}
```
seanpdoyle, line 84: @marcoroth does this clobber the existing `WindowEventMap` interface? For example, if I were to add a `"click"` event listener, would it be able to infer the type as `MouseEvent`?

marcoroth: That should still work since this is just extendeding the existing interface

seanpdoyle: It might be due to misconfigurations in my editor (Vim), but when I try to jump to the definition for `WindowEventMap`, it brings me to this one, instead of the TypeScript built-in one with all of the event names.

Is there a way to hint to the code to disambiguate between extending an existing `interface` and declaring a new one?

marcoroth: I'm not sure if there's a different way of doing this. But my gut tells me it's more of an editor-tooling/LSP thing.

src/events.ts
```diff
@@ -0,0 +1,86 @@
...
+  interface ElementEventMap extends TurboElementEventMap {}
```
seanpdoyle, line 85: Should the `TurboElementEventMap` extend from [HTMLElementEventMap](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.htmlelementeventmap.html) instead of [ElementEventMap](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.elementeventmap.html)?

src/events.ts
```diff
@@ -0,0 +1,86 @@
...
+}>
+
+export type TurboFetchRequestErrorEvent = CustomEvent<{
+  request: FetchRequest
+  error: Error
+}>
+
+interface TurboGlobalEventMap {
+  "turbo:before-cache": TurboBeforeCacheEvent
+  "turbo:before-fetch-request": TurboBeforeFetchRequestEvent
```
seanpdoyle, line 51: Is it possible to cut down on the duplication between lists by declaring `TurboGlobalEventMap` and `Turbo{HTML,}ElementEventMap` with `type` declarations, then making `TurboGlobalEventMap` union with `Turbo{HTML,}ElementEventMap`?

```diff
diff --git a/src/events.ts b/src/events.ts
index 0fd77d1..7a8ae2e 100644
--- a/src/events.ts
+++ b/src/events.ts
@@ -46,28 +46,19 @@ export type TurboFetchRequestErrorEvent = CustomEvent<{
   error: Error
 }>
 
-interface TurboGlobalEventMap {
+type TurboGlobalEventMap = {
   "turbo:before-cache": TurboBeforeCacheEvent
-  "turbo:before-fetch-request": TurboBeforeFetchRequestEvent
-  "turbo:before-fetch-response": TurboBeforeFetchResponseEvent
-  "turbo:before-frame-render": TurboBeforeFrameRenderEvent
   "turbo:before-stream-render": TurboBeforeStreamRenderEvent
   "turbo:before-render": TurboBeforeRenderEvent
   "turbo:before-visit": TurboBeforeVisitEvent
   "turbo:click": TurboClickEvent
-  "turbo:fetch-request-error": TurboFetchRequestErrorEvent
-  "turbo:frame-load": TurboFrameLoadEvent
-  "turbo:frame-render": TurboFrameRenderEvent
-  "turbo:frame-missing": TurboFrameMissingEvent
   "turbo:load": TurboLoadEvent
   "turbo:render": TurboRenderEvent
   "turbo:reload": TurboReloadEvent
-  "turbo:submit-start": TurboSubmitStartEvent
-  "turbo:submit-end": TurboSubmitEndEvent
   "turbo:visit": TurboVisitEvent
-}
+} & TurboHTMLElementEventMap
 
-interface TurboElementEventMap {
+type TurboHTMLElementEventMap = {
   "turbo:before-frame-render": TurboBeforeFrameRenderEvent
   "turbo:before-fetch-request": TurboBeforeFetchRequestEvent
   "turbo:before-fetch-response": TurboBeforeFetchResponseEvent
@@ -82,5 +73,5 @@ interface TurboElementEventMap {
 declare global {
   interface DocumentEventMap extends TurboGlobalEventMap {}
   interface WindowEventMap extends TurboGlobalEventMap {}
-  interface ElementEventMap extends TurboElementEventMap {}
+  interface HTMLElementEventMap extends TurboHTMLElementEventMap {}
 }
```

src/events.ts
seanpdoyle: Similarly, could the [DocumentEventMap](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.documenteventmap.html) and [WindowEventMap](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.windoweventmap.html) declarations be combined into a single extension of the [GlobalEventHandlersEventMap](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.globaleventhandlerseventmap.html) interface?

marcoroth: I picked `ElementEventMap` on purpose since we've been using `Element` over `HTMLElement` almost everywhere, but if you think we just just limit it to `HTMLElement` let me know.

OH, I didn't know about `GlobalEventHandlersEventMap`, maybe it would make sense to use that then, thanks for the hint!

seanpdoyle:
> if you think we just just limit it to HTMLElement let me know.

I was thinking it'd make sense since the custom elements that dispatch those events descend from `HTMLElement`. Thinking more on it now, the type of element that _dispatches_ the event matters a lot less than the type of element that _listens_ for the event.

Having said that, I'm not even sure which tags in a Real World application descend from `Element` instead of `HTMLElement`, and `HTMLElement` descends from `Element`, so it might all be a moot point.

stof:
Given that `querySelector` and `querySelectorAll` are typed as returning Element, adding those events only for HTMLElement would force apps to refine the type of the element to be able to register the events with those type benefits. So I would prefer registering them on Element.

jdelStrother:
Anything we can do that would help get this merged?

brunoprietog:
Does this still make sense with the switch to JavaScript?

jdelStrother:
@marcoroth hope it's ok, I reused a lot of your work here to update @types/hotwired__turbo at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69392 .

marcoroth:
@jdelStrother this is great, thank you!

PR #802, opened 2022-11-22
Allow for listening to turbo events raised from within the Shadow DOM

### LINKED ISSUE #801: Turbo data attributes not working when on custom elements
Hi,

I am using Turbo data attributes on an <sl-button> custom element coming from the Shoelace library and they don't appear to be working, even if I pass the data attribute down into the a of the element shadow dom.

Here's an issue I posted in Shoelace about this: https://github.com/shoelace-style/shoelace/discussions/998 @claviska

Just wanted to post here to get some insight from the Turbo side.

I'm wondering if Turbo needs some sort of config option to specify non default elements to watch?

Thanks.

yuki24:
Fixes #801.

This issue seems to be caused by the fact that events fired from inside the Shadow DOM won't propagate across the shadow DOM boundary into the standard DOM.

Normally, the turbo frame listens to the `turbo:click` event in the `LinkInterceptor`:

https://github.com/hotwired/turbo/blob/27384ecf1a22afacccc96cf01864fc6a82fa8cc0/src/core/frames/link_interceptor.ts#L20

And a link click will fire a `turbo:click` event that is eventually handled by the turbo frame link interceptor:

https://github.com/hotwired/turbo/blob/27384ecf1a22afacccc96cf01864fc6a82fa8cc0/src/core/session.ts#L329-L333

However, when the target link is in the shadow DOM, like declared in this PR, the event wouldn't reach the turbo frame link interceptor.

This could be fixed by specifying `composed: true`, so the event will bubble up regardless of where the element is located.

### DIFF
--- a/src/tests/fixtures/navigation.html
+++ b/src/tests/fixtures/navigation.html
@@ -6,22 +6,6 @@
     <title>Turbo</title>
     <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
     <script src="/src/tests/fixtures/test.js"></script>
-    <script>
-      class CustomLinkElement extends HTMLElement {
-        constructor() {
-          super()
-          this.attachShadow({ mode: 'open' })
-        }
-        connectedCallback() {
-          this.shadowRoot.innerHTML = `
-            <a href="${this.getAttribute('link')}">
-              ${this.getAttribute('text')}
-            </a>
-          `
-        }
-      }
-      window.customElements.define('custom-link-element', CustomLinkElement)
-    </script>
     <meta name="referrer" content="origin-when-cross-origin">
   </head>
   <body>
--- a/src/tests/fixtures/test.js
+++ b/src/tests/fixtures/test.js
@@ -66,3 +66,17 @@
   "turbo:frame-missing",
   "turbo:reload"
 ])
+
+window.customElements.define('custom-link-element', class extends HTMLElement {
+  constructor() {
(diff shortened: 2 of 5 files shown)


src/tests/fixtures/navigation.html
```diff
@@ -6,22 +6,6 @@
...
-        connectedCallback() {
-          this.shadowRoot.innerHTML = `
-            <a href="${this.getAttribute('link')}">
-              ${this.getAttribute('text')}
-            </a>
-          `
-        }
-      }
-      window.customElements.define('custom-link-element', CustomLinkElement)
-    </script>
```
yuki24, line 24: Moving this to `src/tests/fixtures/test.js` since now this is used by two different tests.

src/tests/fixtures/test.js
```diff
@@ -66,3 +66,17 @@
...
+
+window.customElements.define('custom-link-element', class extends HTMLElement {
+  constructor() {
+    super()
+    this.attachShadow({ mode: 'open' })
+  }
+  connectedCallback() {
+    this.shadowRoot.innerHTML = `
+      <a href="${this.getAttribute('link')}">
+        <slot></slot>
```
yuki24, line 78: This is minor but I think using `slot` is more common than using a separate attribute.

yuki24: This might actually change the subject that is being tested, so I have changed this back to the original code.

Issue #794, opened 2022-11-11
Body is replaced twice when following a redirect during Drive visit


calleluks:
In a Rails app that I'm working on we have a (Capybara) system test that fails intermittently. The test looks something like this:

```rb
test "the thing" do
  visit some_path
  click_on 'Some link' # The target of this link responds with a 302 redirect

  # Sleeping here makes the test pass

  # On the second page
  fill_in 'Some field' # This input has the "required" attribute set
  click_on 'Submit button'

  # On the third page
  assert_text 'The thing succeeded!'
end
```

Sometimes the test fails because it can't find the text "The thing succeeded!". When inspecting the screenshot, the Capybara driver is stuck on the second page. The form field is empty and Chrome is showing it's built in "Please fill out this field." validation error. It looks as if the field was never filled in but the `fill_in` method never failed with a missing field error.

I first tried adding `sleep`s in the test to see if there was a timing issue somewhere. I quickly found that adding a sleep before trying to fill in the field made the test pass consistently.

After looking into many other things, I eventually tried changing the target of the link to point directly to the second page instead of the redirect, and the test passed consistently again!

This made me suspect that the redirect somehow affected Turbo's rendering so I tried adding a `console.log` in `replaceBody` and found that it was only called once (as expected) when linking directly to the second page but **twice** when following the redirect.

The test was failing because it filled in the field after the first call to `replaceBody` but clicked the submit button after the second call, when the filled in field had been replaced with an empty one.

In an attempt to fix this, I then tried adding back the `willRender: false` parameter to the call to `visitProposedToLocation` that was removed in 256418fee0178ee483d82cd9bb579bd5df5a151f—and the double render was gone! It also made the test that was added in 256418fee0178ee483d82cd9bb579bd5df5a151f fail, though, and I'm not sure how to fix it.

I don't quite understand the change in 256418fee0178ee483d82cd9bb579bd5df5a151f but it seems to have been introduced to fix a somewhat unrelated issue? @seanpdoyle, do you think there is any chance we could achieve the same result without introducing a second call to `replaceBody`?

seanpdoyle:
@calleluks thank you for opening this issue.

I've opened https://github.com/hotwired/turbo/pull/804 to try and resolve this. Could you test that change out locally to confirm if it resolves the issue?

calleluks:
Hey @seanpdoyle, thanks so much for looking into this!

The changes in #804 do resolve the issue! 🎉

PR #804, opened 2022-11-25
Skip Snapshot Caching for redirect visits

seanpdoyle:
Closes [hotwired/turbo#794][]

Reverts the implementation change made in [hotwired/turbo#674][], and in its place passes `shouldCacheSnapshot: false` alongside `willRender: false` for the `action: "replace"` Visit proposed when following a redirect.

In order to test this behavior, this commit introduces the `readBodyMutationLogs` test utility function, along with the `window.bodyMutationLogs` property and the `BodyMutationLog` type.

`BodyMutationLog` instances are pushed onto the log `Array` whenever the `<body>` element is replaced.

[hotwired/turbo#794]: https://github.com/hotwired/turbo/issues/794
[hotwired/turbo#674]: https://github.com/hotwired/turbo/pull/674

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -317,6 +317,8 @@ export class Visit implements FetchRequestDelegate {
       this.adapter.visitProposedToLocation(this.redirectedToLocation, {
         action: "replace",
         response: this.response,
+        shouldCacheSnapshot: false,
+        willRender: false,
       })
       this.followedRedirect = true
     }
--- a/src/tests/fixtures/navigation.html
+++ b/src/tests/fixtures/navigation.html
@@ -15,7 +15,7 @@
 
     <a href="#ignored-link" id="ignored-link">Skipped Content</a>
 
-    <section id="main" style="height: 200vh">
+    <section id="main">
       <h1>Navigation</h1>
       <p><a id="same-origin-unannotated-link" href="/src/tests/fixtures/one.html">Same-origin unannotated link</a></p>
       <p><a id="same-origin-unannotated-link-search-params" href="/src/tests/fixtures/one.html?key=value">Same-origin unannotated link ?key=value</a></p>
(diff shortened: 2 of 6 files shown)


seanpdoyle:
@kevinmcconnell depending on the response to https://github.com/hotwired/turbo/issues/794#issuecomment-1327798645, this is likely ready for code review if you're available.

drjayvee:
Could I politely ask for a release of this fix?

I've now had several issues because of this. They're easy to fix using `data-turbo-method="post"`, but finding instances of GET redirects is really hard, therefore users are running into bugs in our app because of #848 .

jody90:
+1
I'm also looking forward to a new release which will include this fix.

kevinmcconnell:
@drjayvee @jody90 [v7.2.5](https://github.com/hotwired/turbo/releases/tag/v7.2.5) is being released now, which contains this fix.

src/core/drive/visit.ts
```diff
@@ -317,6 +317,8 @@ export class Visit implements FetchRequestDelegate {
       this.adapter.visitProposedToLocation(this.redirectedToLocation, {
         action: "replace",
         response: this.response,
+        shouldCacheSnapshot: false,
+        willRender: false,
```
marcelolx, line 321: @seanpdoyle I understand why we don't want to render again, but I wonder if the flow should be the opposite?

For example, now what happens is the following
-> User clicks a link
-> Link redirects to somewhere else
-> Fetch API follows the redirect
-> Turbo then renders the response
-> Turbo then follows the redirect (updating the history, for example)
-> Turbo then doesn't render the response (because it doesn't want to render it twice)

But shouldn't it be something like
-> User clicks a link
-> Link redirects to somewhere else
-> Fetch API follows the redirect
-> Turbo then SKIPS rendering the response (because it is a redirect)
-> Turbo then follows the redirect (updating the history, for example)
-> Turbo then RENDERS the response

Following this flow, it would be more akin to what browsers do, which is first visiting the redirected location and then rendering the content of the page of that location, wdyt?

*NOTE*: My assumption may be wrong, and I am just misunderstanding how things work, but I thought that it is worth bringing it up anyway

PR #808, opened 2022-12-02
Ensure turbo no longer uses replaceChildren

MaxHatfull:
This method does not exist in older versions of safari (only 14+). This
meand a polyfill would be required in order to use turbo when using the
update action. In order to stop this, we replace the call to this
function with native code.

### DIFF
--- a/src/core/streams/stream_actions.ts
+++ b/src/core/streams/stream_actions.ts
@@ -31,6 +31,9 @@ export const StreamActions: TurboStreamActions = {
   },
 
   update() {
-    this.targetElements.forEach((e) => e.replaceChildren(this.templateContent))
+    this.targetElements.forEach((targetElement) => {
+      targetElement.innerHTML = ""
+      targetElement.append(this.templateContent)
+    })
   },
 }


seanpdoyle:
Since there is [a pre-existing issue](https://github.com/github/browser-support/pull/4/files) that mentions this method, I wonder if it'd be better to encourage applications to depend on something like [@github/browser-support](https://github.com/github/browser-support) or another polyfill?

@keithamus do you have an opinion on this?

keithamus (approved):
This seems fine to me. :+1:

src/core/streams/stream_actions.ts
```diff
@@ -31,6 +31,9 @@ export const StreamActions: TurboStreamActions = {
   },
 
   update() {
-    this.targetElements.forEach((e) => e.replaceChildren(this.templateContent))
+    this.targetElements.forEach((targetElement) => {
+      targetElement.innerHTML = ""
+      targetElement.append(this.templateContent)
```
seanpdoyle, line 36: @MaxHatfull if we're going to go forward with rolling back the `replaceChildren` calls, what do you think about implementing update in terms of the [polyfill](https://github.com/github/browser-support/pull/4/files#diff-08a775c335de1875f6d61c8046ccef7b5f90744fac68d6bc9621b8526b77514eR1-R4):

```typescript
export function replaceChildren(this: Element | Document | DocumentFragment, ...children: Node[]) {
   while (this.firstChild) this.removeChild(this.firstChild)
   this.append(...children)
 }
```

MaxHatfull: I did think about that, the reason I didn't do that in this PR is to minimise the blast radius of this change. If you think this won't affect the projects of the people using hotwire and you think this is clearer, then I think it's a good idea. There's also the question of "is this simpler?" which I don't think it is as unsetting the innerHTML and setting the child to the template is what I'd expect to see when investigating a method called "update".

seanpdoyle:
@kevinmcconnell are you able to execute CI on this branch?

MaxHatfull:
@kevinmcconnell could you run this again please? it looks like it is failing because of issues unrelated to the changes

kevinmcconnell:
@MaxHatfull we had some outdated packages that were causing the CI issues. Can you rebase your changes onto the latest `main` to pick those up? Doing that should solve the failures. Thanks!

MaxHatfull:
@kevinmcconnell that's the rebase done

kevinmcconnell:
@MaxHatfull PR looks great btw. Thanks for taking care of it 🙏

Issue #812, opened 2022-12-08
Does the `Render` type have the parameter order wrong?


TastyPi:
I'm not sure if I'm just misunderstanding something or if they are actually the wrong way round, but it took me a while to figure out what the issue was so this at least might help someone figure it out. The [Render](https://github.com/hotwired/turbo/blob/1ec72ac3236038af50a0010f0821e0664eeac087/src/core/renderer.ts#L10) type is defined as `(newElement: E, currentElement: E) => void`, but it seems like it should be `(currentElement: E, newElement: E) => void`. Let me explain.

For background: We use Front's live chat widget, and their library adds an iframe to the body than can never be reloaded. This means we need to keep the same body element as well. The [original workaround we tried used morphdom](https://github.com/hotwired/turbo/issues/391#issuecomment-1339184391), but then we ran into [this issue of Stimulus controllers not reloading](https://github.com/hotwired/stimulus/issues/209). So we decided to implement a custom renderer that keeps the body and iframe but copies everything else across.

This is the first attempt:

```javascript
  event.detail.render = (newElement, currentElement) => {
    // Remove all the old elements except the Front iframe
    Array.from(currentElement.children).forEach((element) => {
      if (element.id !== FRONT_ELEMENT_ID) {
        element.remove()
      }
    })

    // Move the new elements across
    Array.from(newElement.children).forEach((element) => currentElement.appendChild(element))
  }
```

This doesn't work, it looks like it just removes all the children without copying the new ones across. I assumed the `newElement` was the newly rendered body and the `currentElement` was the body currently in the DOM. Turns out they're the other way round, and swapping the parameter order made it start working - it looked like it was just removing elements because the elements were being moved the wrong way.

The naming of these parameters is very confusing to me, how could the existing DOM be considered the `newElement`? Hence this issue, are the parameters named the wrong way round or am I misunderstanding the meaning of the terms `new` and `current`?

TastyPi:
@seanpdoyle who introduced the `Render` type in #431

TastyPi:
It looks like it should be defined `(currentElement, newElement)` https://github.com/hotwired/turbo/blob/0f3632a2fe1e6118f8c0e8870e59ace91445e282/src/core/drive/page_renderer.ts#L161

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/814 to resolve this.

TastyPi:
[...]

PR #814, opened 2022-12-08
Correct `Render<E>` method signature argument order

seanpdoyle:
Closes [hotwired/turbo#812][]

The order of the arguments in the `Render` signature are incorrect, and [were introduced in that incorrect order][incorrect].

Luckily, the default implementations and the tests that exercise that feature are in the correct order.

The signature's argument order was inspired by
[morphdom's](https://github.com/patrick-steele-idem/morphdom#usage), with the latter argument serving as overrides for the prior argument.

Custom rendering isn't yet covered by Turbo's documentation website, so we have an opportunity to document it thoroughly and correctly, starting with TypeScript's understanding of the interface.

[hotwired/turbo#812]: https://github.com/hotwired/turbo/issues/812
[incorrect]: https://github.com/hotwired/turbo/pull/431/files#diff-d554b20c605a72c68b2d429ff9915cca4afa8e118026f56ede93382d130382a7R11

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "access": "public"
   },
   "devDependencies": {
-    "@playwright/test": "^1.22.2",
+    "@playwright/test": "^1.28.0",
     "@rollup/plugin-node-resolve": "13.1.3",
     "@rollup/plugin-typescript": "^8.5.0",
     "@types/multer": "^1.4.5",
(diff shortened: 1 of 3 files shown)


seanpdoyle:
@kevinmcconnell would you be able to review and merge this change?

PR #822, opened 2022-12-15
Update `@playwright/test` dependency

seanpdoyle:
Our [CI][] is currently failing while executing `yarn run playwright install --with-deps`.

This commit updates the dependency in an effort to resolve whatever underlying dependency issues are causing the failure.

[CI]: https://github.com/hotwired/turbo/actions/runs/3664875382/jobs/6195570624#step:6:1

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "access": "public"
   },
   "devDependencies": {
-    "@playwright/test": "^1.22.2",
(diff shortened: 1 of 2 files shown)


kevinmcconnell (approved):
Looks great 👏 Thanks for sorting this out, @seanpdoyle!

PR #827, opened 2022-12-22
Only snapshot frame visits that have actions

kevinmcconnell:
Frames that use `data-turbo-action` to promote their navigations to page visits require cache snapshots to be taken prior to them loading, in order for those snapshots to be accurate.

However snapshots can be expensive to create for complex pages. In cases where a frame is *not* being promoted, we can avoid taking that snapshot.

This helps with some performance issues observed when using frames on very large pages.

/cc @afcapel

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -64,7 +64,6 @@ export class FrameController
   readonly restorationIdentifier: string
   private previousFrameElement?: FrameElement
   private currentNavigationElement?: Element
-  pageSnapshot?: PageSnapshot
 
   constructor(element: FrameElement) {
     this.element = element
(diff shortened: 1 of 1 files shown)


seanpdoyle (approved):
I'm 👍 for this change!

I wonder how we might add test coverage to guard against regressing with a future change.

Based on this change, I've updated https://github.com/hotwired/turbo/pull/430 to [use the same guard clause](https://github.com/hotwired/turbo/pull/430/files#diff-69797f53a6cb95b1b154ced467a8277baeb57a8a2bcb978c9249bf09000fca48R81-R83).

seanpdoyle:
To pass the test suite in CI, you might want to review and merge https://github.com/hotwired/turbo/pull/822, then rebase.

PR #831, opened 2022-12-25
Add support for defining the matching frame from the caller

p8:
Sometimes you to want define which frame should be matched in the response, other than the one defined in `data-turbo-frame`.

Let's say we have two forms for creating posts and authors:

```html
<!-- GET posts/new -->
<turbo-frame id="new_post">
  <form ...
</turbo-frame>

<!-- GET authors/new -->
<turbo-frame id="new_author">
  <form ...
</turbo-frame>
```
If we want to open them both in a modal we could do the following without having to change the responses:

```html
<turbo-frame id="modal">
</turbo-frame>

<a href="posts/new"   data-turbo-frame="modal" data-turbo-src-frame="new_post"  >New post   modal</a>
<a href="authors/new" data-turbo-frame="modal" data-turbo-src-frame="new_author">New author modal</a>
```

Instead trying to match a frame called "modal" in the response it will try to match what is defined in `data-turbo-src-frame`, in this case "new_post" or "new_author".

I'm not sure about the attribute name. Maybe something like `data-turbo-matching-frame` would be clearer.

Another implementation could be using the anchor part of the URL. But that might be too magical and have bigger implications:

```html
<a href="posts/new#new_post"  data-turbo-frame="modal">New post modal</a>
```

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -370,8 +370,10 @@ export class FrameController
 
     frame.delegate.proposeVisitIfNavigatedWithAction(frame, element, submitter)
 
+    const srcFrame = element.getAttribute("data-turbo-src-frame")
     this.withCurrentNavigationElement(element, () => {
       frame.src = url
+      if (srcFrame) frame.setAttribute("data-turbo-src-frame", srcFrame)
     })
   }
 
@@ -447,7 +449,7 @@ export class FrameController
 
   async extractForeignFrameElement(container: ParentNode): Promise<FrameElement | null> {
     let element
-    const id = CSS.escape(this.id)
+    const id = CSS.escape(this.element.getAttribute('data-turbo-src-frame') || this.id)
 
     try {
       element = activateElement(container.querySelector(`turbo-frame#${id}`), this.sourceURL)
--- a/src/tests/fixtures/frames.html
+++ b/src/tests/fixtures/frames.html
@@ -124,6 +124,10 @@
       <input id="outer-frame-submit" type="submit" value="Outer form submit">
     </form>
 
+    <a id="src-frame-link" href="/src/tests/fixtures/frames/part.html" data-turbo-frame="src-frame" data-turbo-src-frame="part">Src frame link</a>
+    <turbo-frame id="src-frame">
+    </turbo-frame>
+
     <hr class="push-off-screen">
     <input id="below-the-fold-input">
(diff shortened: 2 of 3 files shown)


seanpdoyle:
@p8 thank you for proposing this feature!

In the experimentation leading up to this PR, had you tried adding a server-side condition to render the `<turbo-frame id="<%= ... %>">`?

For example, Turbo will submit the requesting frame ID in the `Turbo-Frame` header (it's what the [turbo-rails gem reads from](https://github.com/hotwired/turbo-rails/blob/main/app/controllers/turbo/frames/frame_request.rb). You could use that to guarantee that the `[id]` value is reflected in the request and response:

```erb
<!-- the request -->
<turbo-frame id="modal">
</turbo-frame>

<a href="posts/new"   data-turbo-frame="modal">New post   modal</a>
<a href="authors/new" data-turbo-frame="modal">New author modal</a>

<!-- GET posts/new -->
<turbo-frame id="<%= request.headers.fetch("Turbo-Frame", "new_post") %>">
  <form ...
</turbo-frame>

<!-- GET authors/new -->
<turbo-frame id="<%= request.headers.fetch("Turbo-Frame", "new_author") %>">
  <form ...
</turbo-frame>
```

If reading the `Turbo-Frame` header feels too low-level or out of your control, you could encode something into the `<a href="...">` as a query parameter that your server could act upon in the same way (for example, `?turbo_src_frame=author` in the request and `params[:turbo_src_frame]` in the response).

p8:
Thanks for the quick feedback @seanpdoyle !

Passing `params[:turbo_src_frame]` is what I was hoping to avoid. 😄 
For the link there isn't much difference:
```html
<a href="posts/new"   data-turbo-frame="modal" data-turbo-src-frame="new_post">New post  modal</a>
<a href="posts/new?turbo_src_frame=new_post" data-turbo-frame="modal">New post  modal</a>
```
But with a param I'll have to sanitize it for the view and I'll have an extra code path in the view in case it is missing.

I didn't know about the `Turbo-Frame` header and find that solution cleaner.
Still I'd rather avoid adding a dynamic part, as it'll make things more difficult to reason about.
For example if the modal `turbo-frame` is included in the layout, it will render two turbo-frames with id "modal" basically breaking Turbo:
```html
<!-- GET authors/new -->
<turbo-frame id="modal">
</turbo-frame>

<!-- this will also render a turbo-frame with id "modal"  -->
<turbo-frame id="<%= request.headers.fetch("Turbo-Frame", "new_author") %>">
  <form ...
</turbo-frame>
```

dhh:
Appreciate what this is trying to solve, but I don't think the complication of the API is worth it. Passing the turbo src frame as Sean suggested strikes me as a fine way to go.

PR #832, opened 2022-12-27
Refactor head_snapshot

scuml:
This modifies head_snapshot to sort elements by type as the snapshot is created resulting in less code, a more compact minimized codebase, and increased performance.

The current head snapshot comparison method compares all items in the header against themselves.  This branch compares scripts to scripts, styles to styles, meta to meta, and so on. This allows for more efficient processing of items during merge.  The primary benefit to this refactor is how it affects the size of the Turbo library - reducing the size of the minified code base by 1.3% (72.78k -> 71.84k).

### DIFF
--- a/src/core/drive/head_snapshot.ts
+++ b/src/core/drive/head_snapshot.ts
@@ -1,71 +1,37 @@
 import { Snapshot } from "../snapshot"
 
-type ElementDetailMap = { [outerHTML: string]: ElementDetails }
-
-type ElementDetails = {
-  type?: ElementType
-  tracked: boolean
-  elements: Element[]
-}
-
-type ElementType = "script" | "stylesheet"
-
 export class HeadSnapshot extends Snapshot<HTMLHeadElement> {
-  readonly detailsByOuterHTML = this.children
-    .filter((element) => !elementIsNoscript(element))
-    .map((element) => elementWithoutNonce(element))
-    .reduce((result, element) => {
-      const { outerHTML } = element
-      const details: ElementDetails =
-        outerHTML in result
-          ? result[outerHTML]
-          : {
-              type: elementType(element),
-              tracked: elementIsTracked(element),
-              elements: [],
-            }
-      return {
-        ...result,
-        [outerHTML]: {
-          ...details,
-          elements: [...details.elements, element],
-        },
-      }
-    }, {} as ElementDetailMap)
+  elements: { [key: string]: Array<Element> } = {}
+  stylesheetElements: Array<Element> = []
+  trackedElemements: Array<Element> = []
+
+  constructor(element: HTMLHeadElement) {
+    super(element)
+    this.parseDetailsByOuterHTML()
+  }
+
+  parseDetailsByOuterHTML() {
+    for (let element of this.children) {
+      element = elementWithoutNonce(element)
+      if (elementIsStylesheet(element)) this.stylesheetElements.push(element)
+      else if (Object.prototype.hasOwnProperty.call(this.elements, element.localName))
+        this.elements[element.localName].push(element)
+      else this.elements[element.localName] = [element]
+
+      if (element.getAttribute("data-turbo-track") == "reload") this.trackedElemements.push(element)
+    }
+  }
 
   get trackedElementSignature(): string {
-    return Object.keys(this.detailsByOuterHTML)
-      .filter((outerHTML) => this.detailsByOuterHTML[outerHTML].tracked)
-      .join("")
+    return this.trackedElemements.map((e) => e.outerHTML).join("")
   }
 
-  getScriptElementsNotInSnapshot(snapshot: HeadSnapshot) {
-    return this.getElementsMatchingTypeNotInSnapshot<HTMLScriptElement>("script", snapshot)
+  get stylesheets(): Array<Element> {
+    return this.stylesheetElements
   }
 
-  getStylesheetElementsNotInSnapshot(snapshot: HeadSnapshot) {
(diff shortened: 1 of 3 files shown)


src/core/drive/head_snapshot.ts
```diff
@@ -74,47 +40,15 @@ export class HeadSnapshot extends Snapshot<HTMLHeadElement> {
   }
 
   findMetaElementByName(name: string) {
-    return Object.keys(this.detailsByOuterHTML).reduce((result, outerHTML) => {
-      const {
-        elements: [element],
-      } = this.detailsByOuterHTML[outerHTML]
-      return elementIsMetaElementWithName(element, name) ? element : result
-    }, undefined as Element | undefined)
+    return this.getElements("meta").filter((e) => (e as HTMLMetaElement).name == name)[0] as Element | undefined
```
seanpdoyle, line 43: Thank you for suggesting these changes!

Since we're grouping tag names, have you considered using class names as keys instead of strings? `HTMLMetaEelement` instead of `"meta"`, `HTMLScriptElement` instead of `"script"`, etc.

That way, when we read the values from the arrays, we don't need to cast them.

seanpdoyle: Since this is grabbing the first occurrence, have you considered using [Array.find](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead of `filter` and index access?

src/core/drive/head_snapshot.ts
```diff
@@ -1,71 +1,37 @@
...
-        ...result,
-        [outerHTML]: {
-          ...details,
-          elements: [...details.elements, element],
-        },
-      }
-    }, {} as ElementDetailMap)
+  elements: { [key: string]: Array<Element> } = {}
+  stylesheetElements: Array<Element> = []
+  trackedElemements: Array<Element> = []
```
seanpdoyle, line 6: Should this property declaration read `trackedElements` instead of `trackedElemements`?

src/core/drive/head_snapshot.ts
```diff
@@ -1,71 +1,37 @@
...
+    this.parseDetailsByOuterHTML()
   }
 
-  getScriptElementsNotInSnapshot(snapshot: HeadSnapshot) {
-    return this.getElementsMatchingTypeNotInSnapshot<HTMLScriptElement>("script", snapshot)
+  parseDetailsByOuterHTML() {
+    for (let element of this.children) {
+      element = elementWithoutNonce(element)
+      if (elementIsStylesheet(element)) this.stylesheetElements.push(element)
+      else if (Object.prototype.hasOwnProperty.call(this.elements, element.localName))
```
seanpdoyle, line 17: A separate "miss" conditional that uses `Object.prototype.hasOwnProperty` combined with `call` and a bound `this` feels more complicated than access-or-assignment. Have you considered combining both `else` branches?

```typescript
if (elementIsStylesheet(element)) {
  this.stylesheetElements.push(element)
} else {
  const collection = this.elements || []
  collection.push(element)
  this.elements[element.localName] = collection
}
```

src/core/drive/head_snapshot.ts
scuml: Yes.  Exactly what I wanted to implement, but I get a typescript error when using `element.constructor` as a key.  Would you know another method to accomplish this?

```
(!) Plugin typescript: @rollup/plugin-typescript TS2538: Type 'Function' cannot be used as an index type.
src/core/drive/head_snapshot.ts: (18:42)

18         const collection = this.elements[element.constructor] || []
```

src/core/drive/head_snapshot.ts
scuml: Great example of how one can look at code for days, comb over it for that perfect PR, and miss the simplest errors.

src/core/drive/head_snapshot.ts
scuml: Ooh - that's nice.

src/core/drive/head_snapshot.ts
scuml: Yes.  Exactly what I wanted to implement.  Thanks for the pointers, however, I get this error using the `extends Element` recommendation:
```
(!) Plugin typescript: @rollup/plugin-typescript TS1005: '(' expected.
src/core/drive/head_snapshot.ts: (4:40)

4   elements: { [key: <E extends Element>]: E[] } = []
```


My original attempt was to use `function` as a key and get the type with `element.constructor` - but that fails with this error. 

```
(!) Plugin typescript: @rollup/plugin-typescript TS2538: Type 'Function' cannot be used as an index type.
src/core/drive/head_snapshot.ts: (18:42)

18         const collection = this.elements[element.constructor] || []
```

scuml:
Sean's recommendations bring the minimized package size down further to 71.79k or 1.36% smaller.

scuml:
@seanpdoyle Tried once more to implement the grouping by class, but couldn't get typescript to accept my various attempts to generic-tize the elements array.  

Would you be able to take a stab at it?  Or can we consider merging this branch to realize the size & speed benefits for now, and attempt to make further optimizations later.

scuml:
Bump for attention.

scuml:
Typescript removed.  Ready to go and a great candidate for Turbo 8.

Looking for review and approval.

shiftyp:
Not planning on helping here too much, but I think your test is still in TS

scuml:
Tests updated.  Ready for review.

scuml:
Verified: works great in Turbo 8.

scuml:
Closing as stale

Issue #371, opened 2021-09-03
Page anchor not working when using CJK characters


DotIN13:
An anchor such as the following one will always bring the page to the top.

```
<h2 id="前世">
  <a href="#前世">#</a>
  前世
</h2>
```

PR #839, opened 2023-01-01
Fix `getElementForAnchor` for non-URLsafe anchors

DotIN13:
Currently, Turbo directly extracts safely encoded anchors from an url object. Thus, if the anchor contains non-URLsafe characters, Turbo will never be able to find the corresponding element.

I am able to fix this issue by adding decoded anchor string to the CSS selectors.

The fix attempts to mimic vanilla browser behaviours, that is, when an non-URLsafe anchor is clicked, the browser would first try to find an element with an id of the safely encoded anchor, and if not found, the browser would then directly look for the element with an id of the unsafe anchor string.

### DIFF
--- a/src/core/snapshot.ts
+++ b/src/core/snapshot.ts
@@ -18,7 +18,10 @@ export class Snapshot<E extends Element = Element> {
   }
 
   getElementForAnchor(anchor: string | undefined) {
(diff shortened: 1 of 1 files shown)


DotIN13:
Ref #371

PR #840, opened 2023-01-02
implement data-turbo-progress-bar on html elements feature

### LINKED ISSUE #540: Progress bar is not shown when using Turbo Frames (even with turbo-action="advance")
Hi,

The documentation specifically mentions that the progress bar is for Turbo Drive but if you use Turbo Frames with URL history support such as when using `<a href="/profile" data-turbo-frame="main" data-turbo-action="advance">`. It's somewhat expected that the progress bar would be shown here because this page transition may happen within 50ms or 5000ms due to external network events that's out of our control.

You can test that it's not being shown by adding a `sleep 1` between Turbo Frame page transitions and not seeing the progress bar where as if you remove the frame so that Turbo Drive picks it up then the progress bar is shown.

Can we make a case here to show the progress bar for all Turbo Frame events regardless of it being done with `turbo-action="advance"`? I'm guessing this was very thoroughly investigated at Hey / Basecamp but that discussion isn't in the issue tracker here. Am I missing something important that would make showing the progress bar for all Turbo Frame events a bad idea?

zhisme:
Closes #540 

PR implements `data-turbo-progress-bar` feature. When added to element, we will always show progress bar on page load, applies for both `a` and `form` html elements. Adding to element `data-turbo-progress-bar="true"` will trigger this behavior.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -1,5 +1,5 @@
 import { Action } from "../types"
-import { getVisitAction } from "../../util"
+import { getVisitAction, getProgressBarValue } from "../../util"
 import { FetchMethod } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { FormSubmission } from "./form_submission"
@@ -92,8 +92,10 @@ export class Navigator {
 
         const { statusCode, redirected } = fetchResponse
         const action = this.getActionForFormSubmission(formSubmission)
+        const withProgressBar = this.getProgressBarForFormSubmission(formSubmission)
         const visitOptions = {
           action,
+          withProgressBar,
           shouldCacheSnapshot,
           response: { statusCode, responseHTML, redirected },
         }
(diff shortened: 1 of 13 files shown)


OpakAlex:
+1

dhh:
I thought the intention from 540 was to get the progress bar for frame loading, and that we wanted it by default on advance? Is this general approach working for those two criteria?

zhisme:
@dhh I will update the PR to match this two criteria, currently it is shown only when data-turbo-progress-bar attribute specified explicitly

zhisme:
ready for review.

Now it is enabled by default for `action="advance"` also can be disabled for that action explicitly specifying `data-turbo-progress-bar="false"`. However, I investigated a bit more and added the same logic to link (as they were anyway triggering `action="advance"` by default, thus triggering progress bar by default). So on links you can also disable it explicitly. And the same behavior applies for forms, so please review and check whether it is needed or not. IMO this all the same logic and we should treat them with the same behavior, however I need some thoughts from maintainers to tell whether it is wrong or right path

brunoprietog:
Will it be a good default to enable it by default for the replace action as well?

zhisme:
@dhh @marcoroth how to proceed now after removing typescript. Should I close this one and reopen with plain js changes?
really appreciate dropping typescript! :tada:

dhh:
You can also just update this PR with the changes. Thanks!

courtsimas:
What's the status of this @zhisme @dhh ? is it just going to make it's way into turbo8 maybe? or some 7.3.x version?

zhisme:
I would like to fix this in turbo8 if you want my opinion. A bit busy latest few months so couldn't find time to handle it, but I will for sure :smiley:

PR #843, opened 2023-01-05
Fix `assert.equal` in frame tests

kevinmcconnell:
The `assert.equal` matcher had been overridden for some tests to allow string comparisons to act on trimmed versions of the strings. Unfortunately the overridden matcher was faulty, in that it didn't fail when the values differed. That is, `assert.equal(1, 2)` was passing rather than failing. As a result, there are a few tests that have failures that were not being reported.

This change addresses this by:

- Restoring the original `assert.equal`
- Adding a new `assert.equalIgnoringWhitespace` so we can be explicit about the cases where we want that behaviour
- Fixing a few existing test failures that had been hidden by the faulty `assert.equal`

**Note:** There are still a couple of test failures here that'll need to be fixed before this can be merged.

### DIFF
--- a/src/tests/functional/frame_tests.ts
+++ b/src/tests/functional/frame_tests.ts
@@ -20,13 +20,16 @@ import {
   searchParams,
 } from "../helpers/page"
 
-assert.equal = function (actual: any, expected: any, message?: string) {
-  actual = typeof actual == "string" ? actual.trim() : actual
-  expected = typeof expected == "string" ? expected.trim() : expected
+declare global {
+  namespace Chai {
+    interface AssertStatic {
+      equalIgnoringWhitespace(actual: string | null | undefined, expected: string, message?: string): void
+    }
+  }
+}
 
-  const assertExpectation = new Assertion(expected)
-
-  assertExpectation.to.equal(expected, message)
+assert.equalIgnoringWhitespace = function (actual: string | null | undefined, expected: string, message?: string) {
+  new Assertion(actual?.trim()).to.equal(expected.trim(), message)
 }
 
 test.beforeEach(async ({ page }) => {
@@ -47,7 +50,7 @@ test("test navigating a frame with Turbo.visit", async ({ page }) => {
   await page.evaluate((pathname) => window.Turbo.visit(pathname, { frame: "frame" }), pathname)
   await nextBeat()
 
-  assert.equal(await page.textContent("#frame h2"), "Frame: loaded", "navigates the target frame")
+  assert.equal(await page.textContent("#frame h2"), "Frame: Loaded", "navigates the target frame")
 })
 
 test("test navigating a frame a second time does not leak event listeners", async ({ page }) => {
@@ -229,7 +232,7 @@ test("test the turbo:frame-missing event following a link to a page without a ma
   await nextEventNamed(page, "turbo:load")
 
   assert.equal(pathname(page.url()), "/src/tests/fixtures/frames.html")
-  assert.equal(await innerHTMLForSelector(page, "#missing"), "")
+  assert.ok(await hasSelector(page, "#missing-frame-link"))
 })
 
 test("test following a link to a page with a matching frame does not dispatch a turbo:frame-missing event", async ({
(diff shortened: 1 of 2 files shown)


kevinmcconnell:
@seanpdoyle I wondered if you have some context for the remaining couple of test failures on this? I can keep digging in if not, but thought some of this might ring a bell.

src/tests/functional/frame_tests.ts
```diff
@@ -229,7 +232,7 @@ test("test the turbo:frame-missing event following a link to a page without a ma
   await nextEventNamed(page, "turbo:load")
 
   assert.equal(pathname(page.url()), "/src/tests/fixtures/frames.html")
-  assert.equal(await innerHTMLForSelector(page, "#missing"), "")
+  assert.ok(await hasSelector(page, "#missing-frame-link"))
```
kevinmcconnell, line 235: I _think_ the original assertion wasn't right here. When navigating back to the previous page, we'd expect the original content of that frame to be there. So I changed this one to check for the presence of the link that was on the page before.

src/tests/functional/frame_tests.ts
```diff
@@ -397,9 +400,11 @@ test("test evaluates frame script elements on each render", async ({ page }) =>
   assert.equal(await frameScriptEvaluationCount(page), undefined)
 
   await page.click("#body-script-link")
+  await nextEventOnTarget(page, "body-script", "turbo:frame-load")
   assert.equal(await frameScriptEvaluationCount(page), 1)
 
   await page.click("#body-script-link")
+  await nextEventOnTarget(page, "body-script", "turbo:frame-load")
   assert.equal(await frameScriptEvaluationCount(page), 2)
 })
```
kevinmcconnell, line 409: This test was correct, but race conditions meant that sometimes it tries to evaluate the count too soon. So changed to wait for the load to finish before each check.

seanpdoyle (approved):
Thank you for making these changes. In the short-term, I think doing whatever is necessary to reduce test flakiness will get the 👍 from me.

In the long-term, I think the test harness's `this.assert` is due for a reckoning, and would benefit from a migration to [Playwright's built-in assertions](https://playwright.dev/docs/test-assertions), since they're written in a way that will synchronize with the document, wait, and even retry.

It's on my to-do list, but I haven't started it in earnest just yet.

kevinmcconnell:
Thanks @seanpdoyle. To be clear, the issue here isn't so much the use of `assert.equal`, it's that we had an [overridden version of `assert.equal`](https://github.com/hotwired/turbo/blob/main/src/tests/functional/frame_tests.ts#L23-L30) that was broken. Hence some of the existing tests that looked like they were passing were actually failing 😬

The remaining two failures currently are:
```
frame_tests.ts:56:1 › test navigating a frame a second time does not leak event listeners 
frame_tests.ts:848:1 › test navigating a eager frame with a link[method=get] that does not fetch eager frame twice
```

So we'll need to figure out if those are just problems with tests, or legit problems with the code that had been going unnoticed. (Hopefully it's the former!).

seanpdoyle:
@kevinmcconnell I can't make `frame_tests.ts:56` fail locally.

I'm able to make `frame_tests.ts:848` pass consistently with the following changes:

```diff
diff --git a/src/tests/fixtures/frames.html b/src/tests/fixtures/frames.html
index c308995..e2e110d 100644
--- a/src/tests/fixtures/frames.html
+++ b/src/tests/fixtures/frames.html
@@ -128,6 +128,6 @@
     <input id="below-the-fold-input">
     <a id="below-the-fold-link-frame-action" data-turbo-action="advance" data-turbo-frame="frame" href="/src/tests/fixtures/frames/frame.html">Navigate #frame</a>
 
-    <a id="link-to-eager-loaded-frame" href="/__turbo/redirect?path=/src/tests/fixtures/page_with_eager_frame.html">Eager-loaded frame after GET redirect</a>
+    <a id="link-to-eager-loaded-frame" href="/src/tests/fixtures/page_with_eager_frame.html">Eager-loaded frame after GET redirect</a>
   </body>
 </html>
```

The fact that it fails during a redirect and passes during a direct visit makes me worried that the second visit that happens behind-the-scenes is clashing with rendering in a way that makes the `<turbo-frame>` connect twice. That seems like a genuine regression or new bug.

seanpdoyle:
@kevinmcconnell is your branch up to date with `main`? It looks like it's missing https://github.com/hotwired/turbo/commit/e9692e75d29383eedaa3d3bb7c164a57798edab0 (https://github.com/hotwired/turbo/compare/assert-equal-fix...main), which might help with passing `:848`.

seanpdoyle:
> I can't make frame_tests.ts:56 fail locally.

I've been able to pass it pretty consistently, but **there are still inconsistent failures**. I think it's a flaky test, and otherwise unrelated to the regressions introduced by the `assert.equals` changes.

kevinmcconnell:
> It looks like it's missing [e9692e7] ... which might help with passing `:848`.

@seanpdoyle you're right - that did fix that failure, thanks! 👍 So other than some existing flakiness, the tests affected by this are all passing now with the change.

Issue #825, opened 2022-12-19
Hash is lost when redirecting after a form submission


stof:
When a form submission redirects to a URL containing a hash, Turbo does not preserve that hash. This breaks the behavior of normal browser interactions where the target element would receive the focus (and be targetted with `:target()` in CSS). This mean that turbo Drive is not actually a drop-in enhancement.

Note: https://github.com/hotwired/turbo/pull/765 is adding tests reproducing the issue, but I haven't found any report for the issue yet.

dhh:
Indeed, dupe of #211

PR #849, opened 2023-01-17
Preserve URL hash when response is redirected

manuelpuyol:
Closes https://github.com/hotwired/turbo/issues/825
Used tests from https://github.com/hotwired/turbo/pull/765

When a response is redirected, the initial request's hash param is lost. This happens because the `response.location` that we use to initiate the visit redirect **does not have a hash**.
I'm fixing this problem by setting the request's `hash` in the response.location object in case there was a redirect.

### DIFF
--- a/src/core/drive/navigator.ts
+++ b/src/core/drive/navigator.ts
@@ -97,7 +97,12 @@ export class Navigator {
           shouldCacheSnapshot,
           response: { statusCode, responseHTML, redirected },
         }
-        this.proposeVisit(fetchResponse.location, visitOptions)
+
+        const location = fetchResponse.location
+        if (redirected) {
+          location.hash = formSubmission.fetchRequest.location.hash
+        }
+        this.proposeVisit(location, visitOptions)
       }
     }
   }
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -359,6 +359,9 @@ export class Visit implements FetchRequestDelegate {
       })
     } else {
       this.redirectedToLocation = response.redirected ? response.location : undefined
+      if (this.redirectedToLocation) {
+        this.redirectedToLocation.hash = request.url.hash
+      }
       this.recordResponse({ statusCode: statusCode, responseHTML, redirected })
     }
   }
--- a/src/tests/fixtures/form.html
+++ b/src/tests/fixtures/form.html
@@ -25,9 +25,8 @@
         <input type="hidden" name="greeting" value="Hello from a redirect">
         <input id="standard-get-form-submit" type="submit" value="form[method=get]">
       </form>
-      <form action="/__turbo/redirect" method="get" class="redirect">
+      <form action="/__turbo/redirect#element-id" method="get" class="redirect">
         <input type="hidden" name="path" value="/src/tests/fixtures/one.html">
-        <input type="hidden" name="hash" value="element-id">
         <button id="standard-get-form-submit-redirect-with-hash">Redirect with hash</button>
       </form>
       <form action="/__turbo/redirect" method="get" data-turbo-stream class="redirect">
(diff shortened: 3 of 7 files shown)


seanpdoyle:
@manuelpuyol thank you for opening this!

When I opened https://github.com/hotwired/turbo/pull/765, I was trying to reproduce the behavior for something like this from a Rails controller:

```rb
class PostsController < ApplicationController
  def create
    @post = Post.create!(post_params)

    redirect_to posts_url(anchor: helpers.dom_id(@post))
  end
end
```

Note the [anchor:](https://edgeapi.rubyonrails.org/classes/ActionDispatch/Routing/PolymorphicRoutes.html#method-i-polymorphic_url) keyword argument.

The idea is that it'd redirect to `/posts#1`.


I might be misunderstanding the implementation changes, but is the code reading the `.hash` value from a URL that's known at request-time, or is it reading it as part of the response-time URL?

manuelpuyol:
@seanpdoyle Interesting, I thought that the test was supposed to showcase #825 since they are really similar.

Here I want to emulate the browser's redirect behavior when the initial URL has a hash. When redirected, the browser will reapply the hash to the new URL:

https://user-images.githubusercontent.com/11280312/213001881-c8e84bc4-fb52-4825-b6b1-beedbac92224.mov

But when Turbo was driving the request, the hash was not preserved:

https://user-images.githubusercontent.com/11280312/213002064-621e5672-e64a-4365-9460-f1493b44c784.mov

So yeah, I'm reading the `hash` from the initial request to pass it forward to redirects, since the `response` object will ignore it for some reason

seanpdoyle:
I'm sorry to confuse the point. Yes, the tests in #765 are meant to exercise the behavior described in #825.

src/core/drive/navigator.ts
```diff
@@ -97,7 +97,12 @@ export class Navigator {
           shouldCacheSnapshot,
           response: { statusCode, responseHTML, redirected },
         }
-        this.proposeVisit(fetchResponse.location, visitOptions)
+
+        const location = fetchResponse.location
+        if (redirected) {
+          location.hash = formSubmission.fetchRequest.location.hash
```
seanpdoyle, line 103: > When a form submission redirects to a URL containing a hash
>
> -- description from https://github.com/hotwired/turbo/issues/825

In my understanding, the "redirects to a URL" part is describing a URL decided on the server-side. This implementation seems like it's reading the `formSubmission.fetchRequest.location.hash` value from a URL decided on the client-side.

To make it concrete, I was hoping to highlight the fact that Turbo cannot access the `anchor:` value from `redirect_to posts_url(anchor: helpers.dom_id(@post))`, and not the `#a-hash` value from the `[action]` attribute from an element like `<form action="/posts#a-hash"></form>`.

seanpdoyle: With that being said, let's fix both!

I think the server-side value will be inaccessible for the same reasons that headers in intermediate redirect responses are inaccessible (see the "Making it work with Turbo Frames and redirects" heading in https://github.com/hotwired/turbo/issues/257#issuecomment-968131438).

manuelpuyol: any idea how we could make Turbo know about hashes coming from the server? It looks like we have two bugs here and this implementation fixes the one I mentioned

manuelpuyol: @seanpdoyle looks like there's no way to get the hash coming from the server since fetch responses are always stripped https://github.com/node-fetch/node-fetch/issues/211#issuecomment-265353508

Would it be ok to move forward with the client fix first?

stof: I just found https://github.com/whatwg/fetch/issues/1167 as a proposal to expose the fragment for the response url. I mentioned the Turbo tickets I found about it as use cases for that proposal.

PR #850, opened 2023-01-24
Restore dev-builds

marcoroth:
For some reason they were disabled in https://github.com/hotwired/turbo/commit/65be4bd76eb0167f077329289ee61283cc82d124, but they were quite helpful if you wanted to test new upcoming features before they get released.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,3 +39,7 @@ jobs:
 
     - name: Firefox Test
       run: yarn test:browser --project=firefox
+
+    - name: Publish dev build
+      run: .github/scripts/publish-dev-build '${{ secrets.DEV_BUILD_GITHUB_TOKEN }}'
+      if: ${{ github.event_name == 'push' }}


.github/workflows/ci.yml
```diff
@@ -39,3 +39,7 @@ jobs:
 
     - name: Firefox Test
       run: yarn test:browser --project=firefox
+
+    - name: Publish dev build
+      run: .github/scripts/publish-dev-build '${{ secrets.DEV_BUILD_GITHUB_TOKEN }}'
+      if: ${{ github.event_name == 'push' }}
```
seanpdoyle, line 45: Is there a way to make this conditional on the pull request author's merge permissions? 

When I open a pull request that passes the rest of CI, I'll receive an email about this portion of the build failing due to lack of permissions.

marcoroth: I guess we could make an own workflow for it which just listens for push events on `main`

marcoroth: I opened https://github.com/hotwired/turbo/pull/851 for this

Issue #852, opened 2023-01-24
Images with relative `src` are loaded incorrectly in frames


manuelpuyol:
Looks like `img` tags are being evaluated before the URL is modified, which can cause unexpected changes in the image's `src`.

https://user-images.githubusercontent.com/11280312/214365315-0ad1dd04-62c0-4dbf-ab78-cc4b2976e888.mov

In this example, the image src is `../images/turbo.png`. In a full page load, the browser requests `http://localhost:9000/src/tests/fixtures/images/turbo.png` and in a frame navigation it requests `http://localhost:9000/src/tests/images/turbo.png`

Note that the second url is the image's path relative to `http://localhost:9000/src/tests/fixtures/tabs.html` which was the original URL when starting the frame navigation.

PR #853, opened 2023-01-24
Relative image

manuelpuyol:
Closes https://github.com/hotwired/turbo/issues/852

Similar to how we activate scripts, I'm activating `img` tags to ensure relative paths are respected according to the updated URL after a frame navigation.


seanpdoyle (approved):
@manuelpuyol if you're able, we'll need to port these changes to JavaScript.

If you're unable, or uninterested, I'm happy to open a new pull request to carry this across the finish line.

PR #856, opened 2023-01-30
Send link as submitter to confirmMethod. Fixes #811

### LINKED ISSUE #811: Pass arbitrary dataset values to `data-turbo-confirm` via `data-turbo-method` links?
Following @excid3's awesome work in #525 enabling custom confirmation dialogs via `data-turbo-confirm`, I've run into a scenario where it doesn't seem like it's currently possible to pass additional information to the `Turbo.setConfirmMethod` function when `data-turbo-confirm` is placed on `data-turbo-method` links.

Consider:
I have a delete link:
```html.erb
<a data-turbo-method="delete" data-turbo-confirm="Are you sure?" data-title="My custom confirm title" data-type="negative" href="/delete/path">
Delete
</a>
```

Clicking that link causes a hidden `form` element to get created in order to trigger the form submission, here: https://github.com/hotwired/turbo/blob/main/src/observers/form_link_click_observer.ts#L33-L60

The problem is, only a subset of the dataset is passed along with the `form` element to the `confirmMethod` call.

In order to work around this, I'm manually querying the DOM for the original link via something like this:
```js
window.Turbo.setConfirmMethod((message, element, submitter) => {
  let originalElement = submitter

  if (!submitter) {
    let matchingElements = document.querySelectorAll(
      `a[data-turbo-confirm="${message}"][data-turbo-method="${element.getAttribute('method')}"]`
    )
    originalElement = [...matchingElements].find((el) => element.action.includes(el.href))
  }

  const {
    type,
    title,
    turboConfirm: description,
    cancel,
    submit,
    turboConfirmDetail: eventDetail,
  } = originalElement.dataset

  ...
}
```

I'd love to be able to just pull those extra data attributes off of the `element` (the `form` element) instead. Would you be open to accepting a PR that allows that to be possible?

excid3:
Since `a[data-turbo-confirm]` dynamically generates a form, there is no way to access the original element. 

This PR stashes the original element on the generated form and passes it as the `submitter` to mimic form submitters.

```html
<a href="#" data-turbo-confirm="Are you sure?">Delete</a>
```

```js
window.Turbo.setConfirmMethod((message, element, submitter) => {
  // submitter is <a href="#" data-turbo-confirm="Are you sure?">Delete</a>
})
```

Fixes #811


src/observers/form_link_click_observer.ts
```diff
@@ -53,7 +53,10 @@ export class FormLinkClickObserver implements LinkClickObserverDelegate {
     if (turboAction) form.setAttribute("data-turbo-action", turboAction)
 
     const turboConfirm = link.getAttribute("data-turbo-confirm")
-    if (turboConfirm) form.setAttribute("data-turbo-confirm", turboConfirm)
+    if (turboConfirm) {
+      form.setAttribute("data-turbo-confirm", turboConfirm)
+      form.originalElement = link
```
seanpdoyle, line 58: I'm unsure about unintended side effects of introducing a property on `Element`. Is there another place we could stash this reference?

seanpdoyle:
Thank you for making this change! Could we add a test to the suite to exercise this behavior and guard against future regressions?

src/observers/form_link_click_observer.ts
excid3: I couldn't think of any other good place to store the reference. Open to suggestions.

Might be better to have a turbo specific name here like `turboConfirmElement` so that it wouldn't clash with any other properties?

src/core/drive/form_submission.ts
```diff
@@ -126,7 +126,11 @@ export class FormSubmission {
     const confirmationMessage = getAttribute("data-turbo-confirm", this.submitter, this.formElement)
 
     if (typeof confirmationMessage === "string") {
-      const answer = await FormSubmission.confirmMethod(confirmationMessage, this.formElement, this.submitter)
+      const answer = await FormSubmission.confirmMethod(
+        confirmationMessage,
+        this.formElement,
+        this.submitter || this.formElement.originalElement
```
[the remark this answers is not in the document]
excid3: This is for`submitter` which is `_submitter: HTMLElement | undefined`, so that shouldn't need changing right?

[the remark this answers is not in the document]
seanpdoyle: I think the `submitter:` type makes sense. How does the compiler understand the `HTMLFormElement.originalElement` property? That assignment is happening in an entirely different file, and there aren't any type annotations in the diff.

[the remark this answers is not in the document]
seanpdoyle: Even if the argument expects `undefined`, I'm surprised you aren't getting compile time errors for accessing the property.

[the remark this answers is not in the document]
excid3: Oh yeah, I'm not sure why that is either. 🤔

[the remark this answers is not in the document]
excid3: I'm not too familiar with TypeScript, any advice on how I should address this?

seanpdoyle:
@excid3 what about incorporating the submitted ID into the confirm message?

```diff
diff --git a/src/tests/functional/form_submission_tests.ts b/src/tests/functional/form_submission_tests.ts
index 56cd951..5d1184a 100644
--- a/src/tests/functional/form_submission_tests.ts
+++ b/src/tests/functional/form_submission_tests.ts
@@ -971,6 +971,24 @@ test("test link method form submission inside frame with confirmation cancelled"
   assert.notOk(await hasSelector(page, "#frame div.message"), "Not confirming form submission does not submit the form")
 })
 
+test("test link method form submission with custom confirmation", async ({ page }) => {
+  page.on("dialog", (dialog) => {
+    assert.equal(dialog.message(), "Submitter: #link-method-inside-frame-with-confirmation")
+    dialog.accept()
+  })
+
+  await page.evaluate(() =>
+    window.Turbo.setConfirmMethod((_message, _form, submitter) =>
+      Promise.resolve(submitter ? confirm(`Submitter: #${submitter.id}`) : false)
+    )
+  )
+
+  await page.click("#link-method-inside-frame-with-confirmation")
+  await nextBeat()
+
+  assert.equal(await page.textContent("#frame div.message"), "Link!")
+})
+
 test("test link method form submission outside frame", async ({ page }) => {
   await page.click("#link-method-outside-frame")
   await nextBody(page)
```

seanpdoyle:
@excid3 with the issue of testing resolved, I'd like to take this opportunity to get up on my soapbox and vent a little bit.

With the re-introduction of the `<a>` based `<form>` submissions, [Turbo is responsible for](https://github.com/hotwired/turbo/blob/f04a21c9418bc384f5cdafaf9d2ed21caafbf438/src/observers/form_link_click_observer.ts#L33):

* creating a `<form>` element
* transforming each query parameter in the `a[href]` value into an `<input type="hidden">` element, and inserting that element into the `<form>`
* transforming the `a[href]` value into a String _without_ and query parameters, and assigning that to the `<form>` element's `[action]` attribute 
* making sure the form declares `[data-turbo="true"]` so that it is handled by Turbo, just in case it's inserted into the document in an area that has otherwise has Turbo disabled
* making sure the `<form>` is visually hidden with `[hidden]` so that it doesn't flash visually during submission
* transforming `a[data-turbo-method]` into `form[method]`
* ensuring the presence or absence of `form[data-turbo-action]` based on `a[data-turbo-action]`
* ensuring the presence or absence of `form[data-turbo-confirm]` based on `a[data-turbo-confirm]`
* ensuring the presence or absence of `form[data-turbo-stream]` based on `a[data-turbo-stream]` **which normally only affects GET requests** in case `a[data-turbo-method="GET"]`, which is an entirely counter productive yet possible use case
* ensuring the presence or absence of the `Turbo-Frame` header in the request based on the `<a>` elements targeted Turbo Frame
* appending the `<form>`, submitting it, listening for its next `turbo:submit-end`, then removing it

Now, with the addition of this change, passing the `<a>` element to another UJS-inspired configuration as a `SubmitEvent.submitter`, despite the fact that they are usually only ever `<button>` or `<input type="submit">` elements.

I've always struggled to understand how Turbo owning the entire (probably incomplete and ever-growing) burden of responsibility outlined above is preferable over client applications using `<form>`+`<button>` elements for requests that are not `GET`. 

What is gained by Turbo providing this support? What is lost by requiring/forcing client applications to use forms?

@excid3 I know you are not the inventor of this feature, so I want to be clear that this criticism isn't being leveled at you!

seanpdoyle (approved):
👍 to the change, 👎 for the feature! 🙃

excid3:
> @excid3 what about incorporating the submitted ID into the confirm message?

Oh smart!

brunoprietog:
I also agree with this. Using links as forms that perform an action other than get should not be done for accessibility reasons, unless the link specifies the button role, which many people realistically ignore. In fact, this is mentioned in the Turbo handbook. Other needs include performing actions using buttons that are inside a form, in which case, instead of using a link, you could use a button with the formaction and formmethod attributes instead.

src/observers/form_link_click_observer.ts
marcoroth: Here's an idea to resolve this: https://github.com/excid3/turbo/pull/2

src/core/drive/form_submission.ts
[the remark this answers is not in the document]
marcoroth: `this.formElement.originalElement` is typed `any` which technically is assignable to anything, even though it might not make sense

src/observers/form_link_click_observer.ts
seanpdoyle: What do you think about actually writing to the `SubmitEvent.submitter` property?

```ts
form.addEventListener("submit", (event: SubmitEvent) => event.submitter = link, { once: true }
```

marcoroth: That might even be cleaner, given you can override `submitter` and it's not being lost in-between.

seanpdoyle: If that doesn't work, you could try passing it to [HTMLFormElement.requestSubmit](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit#parameters) as an argument:

```ts
requestAnimationFrame(() => form.requestSubmit(link))
```

If `requestSubmit` rejects the `link` argument because it's an `HTMLAnchorElement`, and the only reason we're passing it along is to make its content and attributes available to the confirmation callback, what do you think about creating a `<button>` element that mimics the `<a>`?


```ts
const submitter = document.createElement("button")
submitter.textContent = link.textContent

for (const {name, value} of link.attributes) {
  button.setAttribute(name, value)
}

requestAnimationFrame(() => form.requestSubmit(button))
```

excid3: If browser currently allow this, but start enforcing the types in the future, this could break (and possibly only in some browsers).

@marcoroth's idea of adding attributes and a querySelector seems like a safer solution?

seanpdoyle: @excid3 is https://github.com/hotwired/turbo/pull/856#discussion_r1091943137 a future-compatible technique?

Its approach matches the style from the rest of the transformation logic, and it circumvents the need to stash-then-read.

excid3: I think so, but it still lose access to the original element that way. 

It does solve the problem of passing the attributes along, so it's probably good enough™. I can't foresee a reason to need the original element personally.

I'll refactor to use that.

gucki:
@excid3 Thank you for your work on this! May I ask why this hasn't been merged yet? Is there any other workaround (except using buttons, which breaks ex. button group bootstrap styling)?

brunoprietog:
Thanks for the work on this, but we have decided not to continue to fix anything around `data-turbo-method`. Any of those things can be done with traditional forms. While we will not remove it, we will not continue to maintain it. We should reflect this in the documentation as well.

PR #858, opened 2023-01-31
Fix flaky UJS integration test

afcapel:
Copule of changes:

- Ensure that the UJS response is completed before asserting that no Turbo request was issued.

- The form test should assert that a `turbo:frame-load` was indeed fired. It passed some times, even if it was mistakenly asserting that the event didn't happen, because the check is asynchronous so it could happen before the event was captured.


src/tests/functional/ujs_tests.ts
```diff
@@ -1,29 +1,29 @@
...
 import { assert } from "chai"
-import { noNextEventOnTarget } from "../helpers/page"
+import { nextEventOnTarget, noNextEventOnTarget } from "../helpers/page"
 
 test.beforeEach(async ({ page }) => {
   await page.goto("/src/tests/fixtures/ujs.html")
 })
 
-test("test clicking a [data-remote=true] anchor within a turbo-frame", async ({ page }) => {
+test("does NOT handle [data-remote=true] anchors within a turbo-frame", async ({ page }) => {
```
kevinmcconnell, line 9: Maybe this description should be something like "allows UJS to intercept and cancel Turbo requests for anchors inside a turbo-frame"?

Because it's not so much that Turbo doesn't handle those anchors -- normally it would do so. It's that UJS was able to respond first, and cancel the event before Turbo processed it.

kevinmcconnell (approved):
I think we could tweak the test descriptions to make them clearer.

Also what do you think about moving this test into somewhere like `src/tests/integration` so it's clear they are about Turbo's interaction with another library?

afcapel:
@kevinmcconnell makes sense, I've updated the test decription and moved it to ` src/tests/integration`.

kevinmcconnell (approved):
Looks great!

PR #859, opened 2023-01-31
Stub external request in external visit test

kevinmcconnell:
This test was failing locally due to the external request destroying the page execution context. Stubbing a response to the request allows it to complete without relying on an external request.


afcapel (approved):
LGTM 👍

PR #862, opened 2023-01-31
Fix test failure on Firefox for Mac

afcapel:
This test was consistently failing on my Mac machine.

It was because Firefox for Mac doesn't focus on links on TAB presses.

Ref. https://stackoverflow.com/a/11713537


kevinmcconnell (approved):
Looks good to me 👍

PR #863, opened 2023-02-02
Don't break out of frames when frame missing

kevinmcconnell:
This changes the default behaviour when a frame response is missing its expected `<turbo-frame>` element.

Previously, when the response was missing its frame, we would trigger a `turbo:frame-missing` event, and then (provided that event wasn't cancelled) perform a full page visit to the requested URL.

However there are cases where the full reload makes things worse:

- If the frame contents were non-critical, reloading the page can turn a minor bug into a major one.
- It can mask some bugs where frames were intend to explicitly navigate out of the frame (`target="_top"`), by incurring a second request that loads the page that makes it seem as if it's working corrects.
- It leaves the user at a URL that may never be capable of rendering a valid response (since that URL was only intended to serve a particular frame). That means refreshing the page is no help in getting back to a working state.
- It can lose other temporary state on a page, like form values.

With this change, we no longer perform the full page visit. Instead, we handle a missing frame by doing two things:

- Write a short error message into the frame, so that the problem is visible on the page.
- Throw an exception, which should make the problem quite obvious in development, and which allows it to be easily gathered by exception monitoring tools in production.

We keep the `turbo:frame-missing` event exactly as before, so applications can still hook in to perform alternative behaviour if they want.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -427,15 +427,15 @@ export class FrameController
   }
 
   private handleFrameMissingFromResponse(fetchResponse: FetchResponse) {
-    this.showFrameMissingError()
-    this.throwFrameMissingException(fetchResponse)
+    this.renderFrameMissingError()
+    this.throwFrameMissingError(fetchResponse)
   }
 
-  private showFrameMissingError() {
+  private renderFrameMissingError() {
     this.element.innerHTML = `<strong class="turbo-frame-error">Content missing</strong>`
   }
 
-  private throwFrameMissingException(fetchResponse: FetchResponse) {
+  private throwFrameMissingError(fetchResponse: FetchResponse) {
     const message = `The response (${fetchResponse.statusCode}) did not contain the expected <turbo-frame id="${this.element.id}">`
     throw new TurboFrameMissingError(message)
   }
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -32,6 +32,7 @@ import { VisitOptions } from "../drive/visit"
 import { TurboBeforeFrameRenderEvent } from "../session"
 import { StreamMessage } from "../streams/stream_message"
 import { PageSnapshot } from "../drive/page_snapshot"
+import { TurboFrameMissingError } from "../errors"
 
 type VisitFallback = (location: Response | Locatable, options: Partial<VisitOptions>) => Promise<void>
 export type TurboFrameMissingEvent = CustomEvent<{ response: Response; visit: VisitFallback }>
@@ -181,15 +182,9 @@ export class FrameController
           session.frameLoaded(this.element)
           this.fetchResponseLoaded(fetchResponse)
         } else if (this.willHandleFrameMissingFromResponse(fetchResponse)) {
-          console.warn(
-            `A matching frame for #${this.element.id} was missing from the response, transforming into full-page Visit.`
-          )
-          this.visitResponse(fetchResponse.response)
+          this.handleFrameMissingFromResponse(fetchResponse)
         }
       }
-    } catch (error) {
-      console.error(error)
-      this.view.invalidate()
     } finally {
       this.fetchResponseLoaded = () => {}
     }
@@ -264,7 +259,6 @@ export class FrameController
   }
 
   async requestFailedWithResponse(request: FetchRequest, response: FetchResponse) {
-    console.error(response)
     await this.loadResponse(response)
     this.resolveVisitPromise()
   }
@@ -430,6 +424,20 @@ export class FrameController
     })
 
     return !event.defaultPrevented
+  }
+
+  private handleFrameMissingFromResponse(fetchResponse: FetchResponse) {
+    this.showFrameMissingError()
+    this.throwFrameMissingException(fetchResponse)
+  }
+
+  private showFrameMissingError() {
+    this.element.innerHTML = `<strong class="turbo-frame-error">Content missing</strong>`
+  }
+
+  private throwFrameMissingException(fetchResponse: FetchResponse) {
+    const message = `The response (${fetchResponse.statusCode}) did not contain the expected <turbo-frame id="${this.element.id}">`
+    throw new TurboFrameMissingError(message)
   }
 
   private async visitResponse(response: Response): Promise<void> {
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -427,12 +427,8 @@ export class FrameController
   }
 
   private handleFrameMissingFromResponse(fetchResponse: FetchResponse) {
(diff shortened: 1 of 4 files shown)


dhh:
We need the allowlist concept to go along with this, so you can deal with "session expired" cases without having to write a custom turbo:frame-missing handler.

src/core/frames/frame_controller.ts
```diff
@@ -432,6 +426,20 @@ export class FrameController
     return !event.defaultPrevented
   }
 
+  private handleFrameMissingFromResponse(fetchResponse: FetchResponse) {
+    this.showFrameMissingError()
+    this.throwFrameMissingException(fetchResponse)
+  }
+
+  private showFrameMissingError() {
```
packagethief, line 434: Consider renaming to `renderFrameMissingError` to stick with the "render" language, and to suggest that if we're going to do any heavier rendering, we should delegate to the frame's view, [using an ErrorRenderer](https://github.com/hotwired/turbo/blob/b3facd8d486f89f0104baa7c64390f4990c72d79/src/core/drive/page_view.ts#L32-L36) as we do for [page visits](https://github.com/hotwired/turbo/blob/b3facd8d486f89f0104baa7c64390f4990c72d79/src/core/drive/navigator.ts#L113).

I could be convinced we should use a renderer now, it just seemed a bit too heavy. Thoughts?

kevinmcconnell: Agreed on the rename 👍

I'd be inclined not to use a renderer here, as long as we don't have anything more involved to do that set some text. But I also could be convinced otherwise :)

kevinmcconnell: Or, maybe without a renderer it should at least use a view method to set the content? Might be a lightweight middle ground. I'll take a look at this tomorrow.

packagethief: Good call. It would fit well in `FrameView`, which also sets `innerHTML` to invalidate the frame. Could add a `missing()` to go with `invalidate()`.

afcapel (approved):

sandrods:
I've upgraded gem "turbo-rails" to 1.4 and "@hotwired/turbo-rails" to  7.3.0 and still get the old behavior of "second request that loads the page". No exception. 
Am I missing something?

kevinmcconnell:
Hi @jonathanloos, yeah, I totally understand! Thanks for bringing this up.

It's hard to make this change without some disruption because the behaviour we want is a bit stricter than what we had in 7.2. (Prior to 7.2 it was stricter too). In 7.2. we have been emitting console warnings to try to highlight where the requests were missing their frames. But it's very easy to miss those, so it's totally understandable that you could end up with links missing frames/targets without noticing. (This situation was a big part of the motivation for the change.)

One idea that might help would be to use a global handler to catch any `turbo:frame-missing` events, and perform a visit in place of the default handling. That would give you effectively the same behaviour as was in 7.2. If you use an error monitoring service (like Sentry or similar) you could report these occurrences as they happen.

That should keep your app working as it did before, but now you can also see what's coming in to the error monitoring to find any links/frames that need to be updated, and you can make those changes gradually instead of all at once.

Something along these lines should work for that:

```javascript
document.addEventListener("turbo:frame-missing", function(event) {
  // TODO: Report this event somewhere...

  // Visit the response
  event.preventDefault()
  event.detail.visit(event.detail.response)
})
```

Hope that helps!

jonathanloos:
@kevinmcconnell we'll be taking a look into that! We appreciate the quick response as well 🙏

jonathanloos:
Hey @kevinmcconnell !

Once again thank you for the idea behind mitigating the issues. Our team has successfully implemented the patch and are tracking the offending DOM elements using [Ahoy](https://github.com/ankane/ahoy). That way we can gather info on which turbo frames are offending and use it as a todo list in the future.

For reference and anyone also looking to track, here's the basis of our implementation. 
```javascript
import { Controller } from '@hotwired/stimulus'
import ahoy from 'ahoy.js'
import debounce from 'lodash/debounce'

  connect () {
    document.addEventListener('turbo:frame-missing', this.handleMissingFrame)
  }

  handleMissingFrame (event) {
    this.trackAhoy(event)

    // Visit the response
    event.preventDefault()
    event.detail.visit(event.detail.response)
  }

  trackAhoy (event) {
    ahoy.track('turbo_frame_missing', event.target.id)
  }
}
```
Note we had to do some solid debouncing and event listener disconnecting in a stimulus controller to make sure we didn't get duplicates all over the place.

Hope that helps! Thanks again.

kevinmcconnell:
@jonathanloos great! I'm glad to hear this worked out. Thanks for letting me know! 🙏

jon-sully:
Howdy! Just spent a couple hours integrating this (and its [companion](#867)) (and/with the Turbo-Rails companion, https://github.com/hotwired/turbo-rails/pull/428) into an app with a partner of mine and we had one particularly subtle hiccup I wanted to note for others / consideration.

In 7.2 (or at least in RC3, which we were running previously), if a frame requests something and the response didn't contain the matching frame, _that same response_ ended up being painted. It wasn't requested again.

I believe the idea with the new API is, "use `turbo_page_requires_reload` / the raw meta tag equivalent in the `<head>`" but `turbo-visit-control="reload"` is made to force a _full_ page load AFAIK — so in that case it's another request but it's actually a heavier request (all assets and resources in the `head` will be reloaded).

There's [this](https://github.com/hotwired/turbo/pull/863#issuecomment-1470184953) 'escape hatch':

```js
document.addEventListener("turbo:frame-missing", function(event) {
  event.preventDefault()
  event.detail.visit(event.detail.response)
})
```

Which, based on the way `visit` is written under the hood, _should_ actually paint that same `response`'s content (nice API btw, this is clean), but a full page load does end up happening.

As we found after a couple hours of digging and experimenting is that this `.visit` call _does_ do what it was written to: paint that response as-is and not re-request the page... but the problem is that that response was constructed with a minimal `layout` — one with a `<head>` that will not match the `<head>` of the current page and thus will cause Turbo Drive to do its own full-page-reload if you have any `turbo-track="reload"`s in your view. 

In short, the solution we found for a frame 'breaking out' to a full page load, but using the singular response it already received as the content for that new page load (e.g. not requesting again) is to use the escape hatch above _plus_ forcing the layout to be our application layout (which will have a matching `head` block):

```ruby
  # ...
  layout -> { "application" if turbo_frame_request? }
  # ...
end
```

which should override the custom layout introduced and used by default for frame responses via https://github.com/hotwired/turbo-rails/pull/428.

This is sort of a half Turbo and half Turbo-Rails concern so hopefully this is an okay place to post it, just wanted to put forth this use case for consideration and thoughts as it seemed like the `event.detail.visit` API was wired up carefully to re-use the same `Response` object (e.g. not request again) but the Turbo-rails 'minimal-layout' setup is a recipe for having a forced reload no matter what.

PR #864, opened 2023-02-02
Allow opt-in to breaking out of missing frames

kevinmcconnell:
**This builds on #863.**

When a frame response is missing its expected `turbo-frame` element, we consider it an error. Normally this is what we want.
    
In certain specific cases, though, it's better if a response without a matching frame is treated as a full page instead. The classic example of this is a login page being rendered in response to an expired session.

To allow specific paths to be opted in to full reloads when they are returned in response to a missing frame, this adds support for an allowlist specified in a meta tag in the document:

    <meta name="turbo-frame-escape-paths" content="/sessions/new /sign_up">

When a response is missing its frame, we check if its path matches anything in that list. If so, we perform a full page visit.

This configuration is app-wide, and would typically be included in an application layout.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -427,8 +427,8 @@ export class FrameController
   }
 
   private handleFrameMissingFromResponse(fetchResponse: FetchResponse) {
-    if (this.responsePermittedToBreakOutOfFrame(fetchResponse)) {
-      this.warnFrameMissingBreakout(fetchResponse)
+    if (this.responsePermittedToEscapeFrame(fetchResponse)) {
+      this.warnResponseEscapingFrame(fetchResponse)
       this.visitResponse(fetchResponse.response)
     } else {
       this.view.missing()
@@ -436,16 +436,16 @@ export class FrameController
     }
   }
 
-  private responsePermittedToBreakOutOfFrame(fetchResponse: FetchResponse) {
+  private responsePermittedToEscapeFrame(fetchResponse: FetchResponse) {
     const allowedPathsTag = this.element.ownerDocument.querySelector<HTMLMetaElement>(
-      `meta[name=turbo-frame-breakout-paths]`
+      `meta[name=turbo-frame-escape-paths]`
     )
     const allowedPaths = allowedPathsTag?.content?.split(/\s+/) || []
 
     return allowedPaths.includes(fetchResponse.location.pathname)
   }
 
-  private warnFrameMissingBreakout(fetchResponse: FetchResponse) {
+  private warnResponseEscapingFrame(fetchResponse: FetchResponse) {
     console.warn(this.frameMissingMessage(fetchResponse, "Performing a full-page visit."))
   }
 
@@ -453,7 +453,7 @@ export class FrameController
     throw new TurboFrameMissingError(
       this.frameMissingMessage(
         fetchResponse,
-        "To transform the response into a full-page visit, include its path in a turbo-frame-breakout-paths meta tag."
+        "To transform the response into a full-page visit, include its path in a turbo-frame-escape-paths meta tag."
       )
     )
   }
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -427,15 +427,15 @@ export class FrameController
   }
 
   private handleFrameMissingFromResponse(fetchResponse: FetchResponse) {
-    this.showFrameMissingError()
-    this.throwFrameMissingException(fetchResponse)
+    this.renderFrameMissingError()
+    this.throwFrameMissingError(fetchResponse)
   }
 
-  private showFrameMissingError() {
+  private renderFrameMissingError() {
     this.element.innerHTML = `<strong class="turbo-frame-error">Content missing</strong>`
   }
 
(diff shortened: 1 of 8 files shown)


src/core/frames/frame_controller.ts
```diff
@@ -427,17 +427,43 @@ export class FrameController
...
+    return allowedPaths.includes(fetchResponse.location.pathname)
+  }
+
+  private warnFrameMissingBreakout(fetchResponse: FetchResponse) {
+    console.warn(this.frameMissingMessage(fetchResponse, "Performing a full-page visit."))
   }
 
   private showFrameMissingError() {
     this.element.innerHTML = `<strong class="turbo-frame-error">Content missing</strong>`
   }
```
kevinmcconnell, line 454: I realise there are some very tiny methods here, but adding them made the logic of `handleFrameMissingFromResponse` read more clearly.

src/core/frames/frame_controller.ts
```diff
@@ -427,17 +427,43 @@ export class FrameController
...
+      this.visitResponse(fetchResponse.response)
+    } else {
+      this.showFrameMissingError()
+      this.throwFrameMissingException(fetchResponse)
+    }
+  }
+
+  private responsePermittedToBreakOutOfFrame(fetchResponse: FetchResponse) {
+    const allowedPathsTag = this.element.ownerDocument.querySelector<HTMLMetaElement>(
+      `meta[name=turbo-frame-breakout-paths]`
```
kevinmcconnell, line 441: Could do with a better name for this tag, I think.

src/core/frames/frame_controller.ts
packagethief: I agree. They're worth it as explaining methods, but they also allow higher-level functions to operate at the same level of abstraction. +1

src/core/frames/frame_controller.ts
packagethief: I think `turbo-frame-escape-paths` would be appropriate. Escape has a similar meaning to "breakout", but it's more suggestive of this being configuration for an exceptional case: an escape hatch.

kevinmcconnell: Yes! That's much better 👍

src/core/frames/frame_controller.ts
```diff
@@ -427,17 +427,43 @@ export class FrameController
...
+    } else {
+      this.renderFrameMissingError()
+      this.throwFrameMissingError(fetchResponse)
+    }
+  }
+
+  private responsePermittedToEscapeFrame(fetchResponse: FetchResponse) {
+    const allowedPathsTag = this.element.ownerDocument.querySelector<HTMLMetaElement>(
+      `meta[name=turbo-frame-escape-paths]`
+    )
```
packagethief, line 442: Since this is a global setting we're reading from the current page, it seems like we ought to access it through the `session`'s `PageSnapsot`. It already has support for getting the document's meta values, which we can do at a low level like this:

```js
this.session.snapshot.getSetting("turbo-frame-escape-paths")
```

But that's private API. It would be better to provide a parsed version on `PageSnapshot`, like we have for `rootLocation` and others. Then we can depend on an `Array` return value and reduce to:

```js
return this.session.snapshot.frameEscapePaths.includes(pathname)
```

src/core/frames/frame_controller.ts
```diff
@@ -427,17 +427,43 @@ export class FrameController
...
+    if (this.responsePermittedToEscapeFrame(fetchResponse)) {
+      this.warnResponseEscapingFrame(fetchResponse)
+      this.visitResponse(fetchResponse.response)
+    } else {
+      this.renderFrameMissingError()
+      this.throwFrameMissingError(fetchResponse)
+    }
+  }
+
+  private responsePermittedToEscapeFrame(fetchResponse: FetchResponse) {
```
packagethief, line 439: What do you think of `responseShouldEscapeFrame()` to match the "should" convention used elsewhere? I'm thinking of things like `shouldRender`, `shouldCacheSnapshot`, etc.

src/core/frames/frame_controller.ts
kevinmcconnell: Ah, nice. I hadn't noticed that! Fixed in https://github.com/hotwired/turbo/pull/864/commits/d64b279101bcc9ecaed4831ac4a30d3821afb641

src/core/frames/frame_controller.ts
kevinmcconnell: I don't have very strong feelings about this, so am happy to change if you prefer. But at times I find the "should" naming dilutes the intention of methods a bit.

Maybe this one isn't the best example, because "permittedTo" isn't hugely illuminating :) But it does suggest that this is about responses that have been somehow granted the ability to do something, that not every response is allowed to do. I hints at a meaning. Whereas "should" just means "a piece of logic elsewhere will determine this".

I find the "will..." methods are similar sometimes. They are all good conventions to have when crossing API boundaries when it conveys something about the interaction (`shouldX`, `willX`, `didX`) but for internal/private methods I feel like they sometimes get in the way of what could be a clearer name.

It's subtle & subjective for sure, but it can add up I think. What do you think?

kevinmcconnell:
We decided to take a different approach to this: #867 lets us solve the problem without introducing new configuration options.

PR #865, opened 2023-02-02
Replace intern dependency with Jest

afcapel:
### The problem

Before this change following the README steps to install the dev dependencies and run the tests didn't work out of the box. This was because one dependency, [the intern](https://theintern.io/), was vendoring a chromedriver version that is incompatible with recent Chrome.

When you tried run the tests you saw this:

```
$ yarn test
yarn run v1.22.19
$ yarn test:unit && yarn test:browser
Debugger listening on ws://127.0.0.1:9229/4391aecc-bf57-4134-84ea-488a1383896c
Tunnel started
(ノಠ益ಠ)ノ彡┻━┻
SessionNotCreatedException: [POST http://localhost:4444/wd/hub/session / {"desiredCapabilities":{"name":"intern","idle-timeout":60,"browserName":"chrome","goog:chromeOptions":{"args":["headless","disable-gpu","no-sandbox"]},"browser":"chrome"}}] session not created: This version of ChromeDriver only supports Chrome version 105
Current browser version is 109.0.5414.119 with binary path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
System info: host: 'mbp.local', ip: 'fe80:0:0:0:4ad:c2df:e1d8:aa49%en0', os.name: 'Mac OS X', os.arch: 'aarch64', os.version: '13.1', java.version: '17.0.4'
Driver info: driver.version: unknown
remote stacktrace: 0   chromedriver                        0x0000000104bd9788 chromedriver + 4515720
1   chromedriver                        0x0000000104b5d9d3 chromedriver + 4008403
2   chromedriver                        0x00000001047f012a chromedriver + 413994
3   chromedriver                        0x000000010481861c chromedriver + 579100
4   chromedriver                        0x0000000104813f8b chromedriver + 561035
5   chromedriver                        0x000000010480ffd3 chromedriver + 544723
...
```

The way to solve this issue was to run `CHROMEVER="$(chromedriver --version | cut -d' ' -f2)" yarn test`, [which is what CI is doing](https://github.com/hotwired/turbo/blob/0571735a1991d2875d7b0eb1329e4a0e66311718/.github/workflows/ci.yml#L21-L25), but that was not documented in the README.

### Solution

We moved the functional tests to run with playwright in https://github.com/hotwired/turbo/pull/609 and since then we've only used the intern as a test server (to serve fixtures) and to run unit tests.

I think keeping the intern just for that is a bit overkill. Instead we can use a test runner like Jest, which can run inside node, without having to launch a browser. When a browser is needed, we can use playwright and functional tests.

//cc @seanpdoyle


seanpdoyle (approved):
@afcapel thank you for taking this one on!

I'm not familiar enough with Playwright or its underlying tooling, but I _really_ wish there were a way to test the entire suite with a single, unified harness.

I'm 👍 on swapping Jest in for Intern, but it still feels like there's more to explore here.

afcapel:
> I'm not familiar enough with Playwright or its underlying tooling, but I really wish there were a way to test the entire suite with a single, unified harness.

@seanpdoyle that was my first idea too, and I agree it'd be nicer if we only had one test runner. Unfortunately I couldn't make playwright work with the kind of unit tests that we have. They mostly test code internals, and playwright is more tailored to excercise the UI through the browser. But maybe there's something that I'm missing! As I said, I'd be happy if we can unify both.

KonnorRogers:
Just my $0.02, but Jest is terrible because it tests in an environment that doesn't exist.

Nobody browses the web in JSDOM. Jest also wraps all your code and transforms it from ESM to CJS which create aa whole host of problems. 

Things like intersectionObserver, resizeObserver, etc won't work. JSDOM mocks the world.

+1 for playwright, alternatively, you can also use Web-Test-Runner and use playwright as the launcher.

https://modern-web.dev/docs/test-runner/overview/

marcoroth:
> Jest also wraps all your code and transforms it from ESM to CJS which create a whole host of problems.

this is also what have been doing in Stimulus and I've been looking to migrate away from it because it has quite some nasty side-effects.

I was also looking at web-test-runner (+playwright) for the exact same reason, so also a +1 from my side for that combination.

afcapel:
web-test-runner + playwright sounds like a better option, if we can make it work with our tests. I'll give it a try 👍

dhh:
Superseded by https://github.com/hotwired/turbo/pull/866.

PR #866, opened 2023-02-03
Migrate from intern to web-test-runner

KonnorRogers:
intern -> web-test-runner

Builds upon #865 

Co-Authored-By @afcapel


marcoroth:
🚀

web-test-runner.config.mjs
```diff
@@ -0,0 +1,19 @@
...
+
+export default {
+  browsers: [
+    playwrightLauncher({ product: 'chromium' }),
+    playwrightLauncher({ product: 'firefox' }),
+    playwrightLauncher({ product: 'webkit' }),
+  ],
+  nodeResolve: true,
+  files: "./src/tests/unit/**/*_tests.ts",
+
```
afcapel, line 15: Here we can add:

```js
  testFramework: {
    config: {
      ui: "tdd"
    }
  },
```

That configures mocha to use [the TDD interface](https://devhints.io/mocha-tdd) - `test(...)` instead of `it(..)` - whic is more consistent with the functional tests.

src/tests/unit/deprecated_adapter_support_tests.ts
```diff
@@ -3,8 +3,9 @@ import { FormSubmission } from "../../core/drive/form_submission"
 import { Adapter } from "../../core/native/adapter"
 import * as Turbo from "../../index"
 import { DOMTestCase } from "../helpers/dom_test_case"
+import { assert } from "@open-wc/testing"
 
-export class DeprecatedAdapterSupportTest extends DOMTestCase implements Adapter {
+class DeprecatedAdapterSupportTest extends DOMTestCase implements Adapter {
```
afcapel, line 8: I think we can directly call mocha's `setup` and `teardown` functions here and get rid of `DOMTestCase`.

afcapel (commented):
Nice one! Left a couple a couple of suggestions, but it looks great overall. Much neater testing setup.

web-test-runner.config.mjs
KonnorRogers: Perfect! Thanks!

src/tests/unit/deprecated_adapter_support_tests.ts
KonnorRogers: Done!

PR #867, opened 2023-02-06
Respect `turbo-visit-control` for frame requests

kevinmcconnell:
Turbo normally performs a fill page reload whenever a response contains the appropriate `turbo-visit-control` meta tag:

    <meta name="turbo-visit-control" content="reload">

Such responses are considered "not visitable".

For frame requests, we have previously been ignoring any `turbo-visit-control` set in the response, and instead treating all valid frame responses as "visitable".

This commit changes this behaviour so that `turbo-visit-control` will be treated consistently for both frame and non-frame requests.

As well as being more consistent, this provides a useful escape hatch for situations where a frame request redirects to something that should be a full page reload, but which would be prevented due to that content missing the expected frame. The class example of this is when an expired session causes a frame request to be redirected to a login page. By including `turbo-visit-control` on that login page, we can ensure that it is always rendered as a full page, and never hidden by a failed frame request.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -160,29 +160,13 @@ export class FrameController
     try {
       const html = await fetchResponse.responseHTML
       if (html) {
-        const { body } = parseHTMLDocument(html)
-        const newFrameElement = await this.extractForeignFrameElement(body)
+        const document = parseHTMLDocument(html)
+        const pageSnapshot = PageSnapshot.fromDocument(document)
 
-        if (newFrameElement) {
-          const snapshot = new Snapshot(newFrameElement)
-          const renderer = new FrameRenderer(
-            this,
-            this.view.snapshot,
-            snapshot,
-            FrameRenderer.renderElement,
-            false,
-            false
-          )
-          if (this.view.renderPromise) await this.view.renderPromise
-          this.changeHistory()
-
-          await this.view.render(renderer)
-          this.complete = true
-          session.frameRendered(fetchResponse, this.element)
-          session.frameLoaded(this.element)
-          this.fetchResponseLoaded(fetchResponse)
-        } else if (this.willHandleFrameMissingFromResponse(fetchResponse)) {
-          this.handleFrameMissingFromResponse(fetchResponse)
+        if (pageSnapshot.isVisitable) {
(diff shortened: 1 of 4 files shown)


seanpdoyle:
[turbo-rails](https://github.com/hotwired/turbo-rails/blob/main/app/controllers/turbo/frames/frame_request.rb#L16) omits the contents of the document from `<html>` to the `<body>` element, including the `<head>` where this `<meta>` would likely be rendered.

How does this strategy interact with servers that omit portions of the page when handling a request with a `Turbo-Frame:` header value?

kevinmcconnell:
Re Turbo Rails, the idea is to stop stripping out the head for frame responses (https://github.com/hotwired/turbo-rails/pull/428).

For providing a way for certain pages to render outside the frame when needed (the "redirect to login page" problem described above), there's an alternative approach to this we discussed (#864), which avoids that issue by letting you specify a list of paths on the requesting page (most likely in an application layout). Paths listed there would be permitted ahead of time to break out of a frame if returned in response to a frame request.

But, after talking this over with some folks, we felt that `turbo-visit-control` is already a good fit for solving that problem without introducing a new concept. And arguably it's a bit surprising that `turbo-visit-control` didn't already work this way.

afcapel (approved):
LGTM 👍

dhh:
Clever judo! Let's make sure we document this hatch together with the specific use case (how to deal with expiring sessions).

kevinmcconnell:
Will do 👍 I'll work up a doc PR to cover the missing frame behaviour, including this escape hatch.

davidalejandroaguilar:
@kevinmcconnell Can you please help me figure out how to apply this to, for example, break out of a turbo frame if a form is successful, but stay within the frame if not? 

Previous to https://github.com/hotwired/turbo/pull/863, a simple `redirect_to` from the controller was enough.

In this case, I'm having this issue on a page that displays a login form within a turbo frame, because it goes through multiple steps, and when the final step is successful, I redirect to a new page without the frame.

davidalejandroaguilar:
It has come to my attention that [there's already a PR](https://github.com/hotwired/turbo-rails/pull/367) for my previous suggestion! 

I think that would also fix the example scenario I provided.

PR #871, opened 2023-02-14
Deprecate [data-turbo-cache=false] in favor of [data-turbo-temporary]

packagethief:
Renames the `[data-turbo-cache=false]` attribute (used to denote temporary elements that should be removed before caching) to `[data-turbo-temporary]` for better similarity with `[data-turbo-permanent]`, its conceptual opposite.

This is a superficial change, but worth it in terms of cohesion, I think. The pairing of "temporary" with "permanent" is just too good to ignore. Also, given the existence of `turbo-cache-*` as the namespace for _page-level_ cache control, a unique name is less likely to confuse.

References:
- https://github.com/hotwired/turbo/pull/238

### DIFF
--- a/src/observers/cache_observer.ts
+++ b/src/observers/cache_observer.ts
@@ -1,27 +1,44 @@
 import { TurboBeforeCacheEvent } from "../core/session"
 
 export class CacheObserver {
+  readonly selector: string = "[data-turbo-temporary]"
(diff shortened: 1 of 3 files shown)


dhh:
Love it!

seanpdoyle (approved):

packagethief:
Thanks! I updated the documentation in https://github.com/hotwired/turbo-site/pull/130

PR #874, opened 2023-02-16
Fix `data-turbo-confirm` on `<a>` without `data-turbo-method`

marcoroth:
The documentation in the ["Requiring Confirmation for a Visit" section](https://turbo.hotwired.dev/handbook/drive#requiring-confirmation-for-a-visit) mentions the use of the `[data-turbo-confirm]` attribute on an `<a>` element to prompt a confirmation before executing a visit. An example is given as follows:

```html
<a href="/articles" data-turbo-confirm="Do you want to leave this page?">Back to articles</a>
```

Prior to version `v7.2.5`, this functionality did not work unless the `[data-turbo-method]` attribute was also present on the link. 

This pull request modifies the behavior so that `[data-turbo-confirm]` can now be used on links without the `[data-turbo-method]` attribute.


seanpdoyle (changes requested):
Would an `<a>` without `[data-turbo-method]` will be treated as a typical Visit navigation?

In the absence of that attribute, what is directing the code to flow through this observer? Could confirmation be accommodated elsewhere, closer to a typical `<a>`-driven Visit?

marcoroth:
Thanks @seanpdoyle, this is good feedback! You are right, this logic doesn't belong to the `FormLinkClickObserver`.

I pushed up a version where we check for the presence of `data-turbo-confirm` attribute in the regular visit flow for links in `Session`.

Since `confirmMethod` was tied to `FormSubmission` I extracted it into a new `Confirm` class which both the form flow and link flow now make use of. I like this a lot better!

searls:
Seems weird that the [documentation](https://turbo.hotwired.dev/handbook/drive#requiring-confirmation-for-a-visit) still says this is possible when the example cited doesn't actually work:

```
<a href="/articles" data-turbo-confirm="Do you want to leave this page?">Back to articles</a>
```

Just spent 15 minutes verifying it wasn't a problem in my turbo version or import maps setup

marcoroth:
Let me get this rebased so we get a chance to get this this merged in

marcoroth:
I just rebased this PR again, but I feel like this is not quite right anymore after the merged PR #1217 from @seanpdoyle. The new `Turbo.config.forms.confirm` config option doesn't quite fit the naming now, since the confirm method is also used on non-form elements here.

I'm wondering if we should either rename the config option or also use the `Turbo.config.forms.confirm` config option for the `<a>` tags.

PR #882, opened 2023-02-27
Form submissions from frames should clear cache

kevinmcconnell:
Usually when submitting a form with a method other than `GET`, we clear the snapshot cache. This helps to avoid cases where we might show a stale view of some data that we just mutated.

Form submissions from inside frames were not doing this. But they ought to, for the same reason: they may have mutated some data that causes other recent pages to become stale.

To avoid this, we can clear the snapshot cache after processing a non-`GET` form, or after any failed form submission, to mirror what we do when the form is outside of any frames.

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -266,12 +266,16 @@ export class FrameController
     const frame = this.findFrameElement(formSubmission.formElement, formSubmission.submitter)
 
     frame.delegate.proposeVisitIfNavigatedWithAction(frame, formSubmission.formElement, formSubmission.submitter)
-
     frame.delegate.loadResponse(response)
+
+    if (formSubmission.method !== FetchMethod.get) {
+      session.clearCache()
+    }
   }
 
   formSubmissionFailedWithResponse(formSubmission: FormSubmission, fetchResponse: FetchResponse) {
     this.element.delegate.loadResponse(fetchResponse)
+    session.clearCache()
   }
 
   formSubmissionErrored(formSubmission: FormSubmission, error: Error) {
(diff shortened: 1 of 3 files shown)


src/tests/functional/frame_tests.ts
```diff
@@ -877,6 +877,22 @@ test("test navigating a eager frame with a link[method=get] that does not fetch
...
+
+  await expect(page.locator("h1")).toHaveText("Changed")
+
+  await page.click("#navigate-form-redirect-as-new")
+  await expect(page.locator("h1")).toHaveText("Page One Form")
+  await page.click("#submit-form")
+  await expect(page.locator("h2")).toHaveText("Form Redirected")
+  await page.goBack()
+
+  await expect(page.locator("h1")).toHaveText("Frames")
```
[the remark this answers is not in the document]
kevinmcconnell: Good idea, that's a bit clearer.

src/core/frames/frame_controller.ts
```diff
@@ -266,12 +266,16 @@ export class FrameController
     const frame = this.findFrameElement(formSubmission.formElement, formSubmission.submitter)
 
     frame.delegate.proposeVisitIfNavigatedWithAction(frame, formSubmission.formElement, formSubmission.submitter)
-
     frame.delegate.loadResponse(response)
+
+    if (formSubmission.method !== FetchMethod.get) {
```
seanpdoyle, line 271: For what it's worth, there's a [FormSubmission.isIdempotent](https://github.com/hotwired/turbo/blob/39affe528cd01be9b8cb0cb0e8091997c254f3ed/src/core/drive/form_submission.ts#L113-L115) property with this implementation.

kevinmcconnell: Thanks 👍 I saw that, but used the same check as the `Navigator` since the cache logic isn't really about idempotency. But I think actually a better option is to rename that method, as it was a bit misleading. Then we can use it. I've done that in #883.

PR #883, opened 2023-02-28
Rename `isIdempotent` to `isSafe`

kevinmcconnell:
The method name was a bit misleading, as what we're actually checking for here is whether the method is [safe][0].

We can also now make use of this method in a couple of places where we clear the snapshot cache after form submissions. The logic there is to clear the cache when performing an operation that is unsafe.

[0]: https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP

### DIFF
--- a/src/http/fetch_request.ts
+++ b/src/http/fetch_request.ts
@@ -144,7 +144,7 @@ export class FetchRequest {
       credentials: "same-origin",
       headers: this.headers,
       redirect: "follow",
-      body: this.isIdempotent ? null : this.body,
+      body: this.isSafe ? null : this.body,
       signal: this.abortSignal,
       referrer: this.delegate.referrer?.href,
     }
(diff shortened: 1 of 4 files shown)


src/http/fetch_request.ts
```diff
@@ -156,8 +156,8 @@ export class FetchRequest {
     }
   }
 
-  get isIdempotent() {
-    return this.method == FetchMethod.get
+  get isSafe() {
+    return this.method === FetchMethod.get
```
afcapel, line 160: As pointed by MDN, `HEAD` and `OPTIONS` are safe too. Do we want to include them here? Don't think we're using them, but they are part of the fetch API.

afcapel (approved):
Looks good! Precision in language is always nice 🏅

src/http/fetch_request.ts
kevinmcconnell: I thought about that, but opted to leave it as-is since we [don't define them](https://github.com/hotwired/turbo/blob/060dd7afde467d718dd1afb4ed450026402388d0/src/http/fetch_request.ts#L30-L36). Adding them to the enum suggests (perhaps) that we are using them somewhere,

afcapel: Fair enough 👍

PR #885, opened 2023-03-02
Extract `HTMLFormSubmission` tuple

seanpdoyle:
The `HTMLFormElement, HTMLElement?` [data clump][] occurs in various places across the internals of the codebase.

This commit extracts that pairing into an `HTMLFormSubmission` class that knows how to properly extract various pieces of HTML Form Submission context from the pairing, like the `method`, `action`, or `target` values (which can be overridden by `formmethod`, `formaction`, and `target`, respectively).

With the new extraction various call sites can be simplified to delegate to the object, with the most impactful changes being made in the `FormSubmission` class (mostly changes to the `constructor` and the removal of various dynamic properties).

[data clump]: https://refactoring.guru/smells/data-clumps


seanpdoyle:
I don't love the `HTMLFormSubmission` name, but it felt better than `SubmissionTuple`, `FormSubmissionTuple`, `FormSubmissionRequest`, or `SubmitRequest` (derived from `HTMLFormElement.requestSubmit`). I'm open to any and all naming suggestions!

@kevinmcconnell Additionally, changing the `FormSubmission` properties to be pre-determined and readonly will help enable (and shrink!) the changes proposed in https://github.com/hotwired/turbo/pull/691.

src/core/drive/form_submission.ts
```diff
@@ -65,61 +56,27 @@ export class FormSubmission {
...
-    this.location = expandURL(this.action)
-    if (this.method == FetchMethod.get) {
-      mergeFormDataEntries(this.location, [...this.body.entries()])
-    }
+    this.submission = submission
+    this.formElement = submission.form
+    this.submitter = submission.submitter
+    this.formData = submission.formData
+    this.location = submission.location
+    this.method = submission.fetchMethod
```
seanpdoyle, line 66: The majority of these `readonly` properties were once defined as `get`-powered properties, which are part of the `FormSubmission` public API. Since `turbo:submit-start` and `turbo:submit-end` event listeners are able to access `FormSubmission` instances, it feels like we should go through a deprecation cycle before moving them to be `private` or replaced by `this.submission[...]` lookups.

seanpdoyle:
@afcapel @kevinmcconnell are either of you able to review these changes (and https://github.com/hotwired/turbo/pull/691, the PR that these changes are in service of).

seanpdoyle:
@afcapel @kevinmcconnell I've rebased these changes to account for the migration from TypeScript. It's ready for review.

src/core/drive/form_submission.ts
seanpdoyle: If the next version of Turbo is a major version bump to `8.0`, that might have implications for deprecation.

afcapel (commented):
@seanpdoyle I think the change makes sense, but I'm not convinced that replacing `element, submitter` argument tuples with an `elementOrSubmission` argument helps readability. The method then have to figure out what the `elementOrSubmission` really is, and extract the the HTML elements that it needs. All while those elements were readily available without any ambiguity in the calling place.

PR #887, opened 2023-03-02
Create Frame Snapshot from Fetch Response HTML

seanpdoyle:
After the changes made in [@hotwired/turbo#867][] and changes made in [@hotwired/turbo-rails#428][] (the canonical server-side implementation), Turbo expects full HTML documents in response to requests with `Turbo-Frame:` headers.

Prior to this commit, the `FrameController` compensated for missing pieces of an HTML document by taking an HTML "snapshot" of the current page through the `<html>` element's [outerHTML][].

This commit changes the `fetchResponseLoaded` callback to read the `responseHTML` directly from the `FetchResponse`, since that will be a fully formed HTML document in Turbo v7.3.0 and later.

To support that change, this commit also updates various `src/test/fixtures` files to render fully-formed HTML documents.

[@hotwired/turbo#867]: https://github.com/hotwired/turbo/pull/867
[@hotwired/turbo-rails#428]: https://github.com/hotwired/turbo-rails/pull/428
[outerHTML]: https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -26,7 +26,7 @@ import { PageSnapshot } from "../drive/page_snapshot"
 import { TurboFrameMissingError } from "../errors"
 
 export class FrameController {
-  fetchResponseLoaded = (_fetchResponse) => {}
+  fetchResponseLoaded = (_fetchResponse) => Promise.resolve()
   #currentFetchRequest = null
   #resolveVisitPromise = () => {}
   #connected = false
@@ -140,7 +140,7 @@ export class FrameController {
         }
       }
     } finally {
-      this.fetchResponseLoaded = () => {}
+      this.fetchResponseLoaded = () => Promise.resolve()
     }
   }
 
@@ -315,7 +315,7 @@ export class FrameController {
       this.complete = true
       session.frameRendered(fetchResponse, this.element)
       session.frameLoaded(this.element)
-      this.fetchResponseLoaded(fetchResponse)
+      await this.fetchResponseLoaded(fetchResponse)
     } else if (this.#willHandleFrameMissingFromResponse(fetchResponse)) {
       this.#handleFrameMissingFromResponse(fetchResponse)
     }
@@ -354,10 +354,10 @@ export class FrameController {
       const pageSnapshot = PageSnapshot.fromElement(frame).clone()
       const { visitCachedSnapshot } = frame.delegate
 
-      frame.delegate.fetchResponseLoaded = (fetchResponse) => {
+      frame.delegate.fetchResponseLoaded = async (fetchResponse) => {
         if (frame.src) {
           const { statusCode, redirected } = fetchResponse
-          const responseHTML = frame.ownerDocument.documentElement.outerHTML
+          const responseHTML = await fetchResponse.responseHTML
           const response = { statusCode, redirected, responseHTML }
           const options = {
             response,
--- a/src/tests/fixtures/frame_preloading.html
+++ b/src/tests/fixtures/frame_preloading.html
@@ -1,14 +1,12 @@
 <!DOCTYPE html>
 <html>
-
-<head>
-  <meta charset="utf-8">
-  <title>Page With Preloading Frame</title>
-  <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
-</head>
-
-<body>
-  <turbo-frame id="menu" src="/src/tests/fixtures/frames/preloading.html"></turbo-frame>
-</body>
-
+  <head>
+    <meta charset="utf-8">
+    <title>Frame Preloading</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+  </head>
+  <body>
+    <turbo-frame id="menu" src="/src/tests/fixtures/frames/preloading.html"></turbo-frame>
+  </body>
 </html>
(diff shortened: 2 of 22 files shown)


packagethief (approved):
This looks good to me, @seanpdoyle. Thanks for tidying up!

hebertcisco (approved):

BALOTIAS:
I honestly didn't know if I should create an issue or write it here. Since v7.3.0 the turbo frame missing event doesn't break out of the frame anymore, this has a big impact for a project that I'm working on. We had to downgrade to v.7.2.5 again, and would like to have re-add this behavior back. Will this PR also support this?

```js
document.addEventListener('turbo:frame-missing', event => {
  event.preventDefault() // Cancel event to not replace turbo-frame content with "Content missing"

  // With 7.3.0: Break out of the turbo frame and visit the response url -> this sucks since we have to request a second time
  event.detail.visit(event.detail.response.url) // Currently we have to do this 

  // Break out of the turbo frame and replace current view with its response -> doesn't currently work as the responseHTML is missing in the response
  // TODO: Wait for https://github.com/hotwired/turbo/pull/887
  // event.detail.visit(event.detail.response.url, { response: event.detail.response })
})
```

Maybe I'm unaware of something?

seanpdoyle:
I've rebased this to account for the migration away from TypeScript.

packagethief (approved):
Thanks @seanpdoyle. This looks great.

@afcapel, is this something we can merge as part of the work you've been doing? Not sure if you've got a released planned or not.

afcapel:
Yes, it looks good, thanks @seanpdoyle!

src/core/frames/frame_controller.js
```diff
@@ -354,10 +354,10 @@ export class FrameController {
       const pageSnapshot = PageSnapshot.fromElement(frame).clone()
       const { visitCachedSnapshot } = frame.delegate
 
-      frame.delegate.fetchResponseLoaded = (fetchResponse) => {
+      frame.delegate.fetchResponseLoaded = async (fetchResponse) => {
         if (frame.src) {
           const { statusCode, redirected } = fetchResponse
-          const responseHTML = frame.ownerDocument.documentElement.outerHTML
+          const responseHTML = await fetchResponse.responseHTML
```
domchristie, line 360: This will likely cause a full page reload because the `responseHTML` will only have a minimal `<head>` and therefore the tracked elements won't be identical.

seanpdoyle: Thank you for catching this @domchristie. I'm not sure how to best resolve the issue mentioned in #1047. Do you have an idea on what to change?

domchristie: @seanpdoyle Can this line be reverted or will that break the intention of https://github.com/hotwired/turbo/pull/867 / https://github.com/hotwired/turbo-rails/pull/428?

Otherwise, I'm beginning to feel that the different visit usages warrant a different modelling.

> For example:
>
> 1. A standard Visit that navigates from one location to the next
> 2. A same-page anchor Visit
> 3. A redirected Visit
> 4. A Frame Visit (promoted to a Visit using data-turbo-action
>
> Each has different expectations and outcomes, which hints that they could be different classes that share the same API
> https://github.com/hotwired/turbo/pull/1044#issuecomment-1780692991

This would be a fairly big refactor, but might be worth it?

seanpdoyle: https://github.com/hotwired/turbo/pull/430 explores a `FrameVisit`, but I'm unsure of what changes would be necessary to incorporate a fix for this issue.

> Can this line be reverted or will that break the intention of #867 / [hotwired/turbo-rails#4](https://github.com/hotwired/turbo-rails/pull/428)

I'm unsure of the implications of reverting it. Is there a conditional that could be introduced to support responses without a `<head>` **and** responses from newer `turbo-rails` versions that send a minimal head?

domchristie: > I'm unsure of the implications of reverting it.

It seems like v7.3.0 includes https://github.com/hotwired/turbo/pull/867 so reverting _should_ work (although I'm not sure how!?) I'm not 

Off the top of my head (pun intended ;), I'm not sure how `HeadSnapshot` could distinguish between a "minimal layout" head from a frame visit and standard visit, where the page's tracked assets have actually changed

domchristie: > but I'm unsure of what changes would be necessary to incorporate a fix for this issue.

It doesn't feel right to me that we call `session.proposeVisitIfNavigatedWithAction` after rendering a frame response. (In the same way it doesn't feel right to propose a second visit after rendering a redirected response.) In these cases, we're not "visiting" a new location, we're just using the side-effects of the visit lifecycle.

In terms of what I had in mind for implementing different visit types: I was thinking that each Visit type could share an interface, and each implementation could respond accordingly (e.g. a `RedirectVisit` wouldn't dispatch any events because the Visit lifecycle events have already been broadcast by the initial Visit.)

qinmingyuan: Any progress?

tleish: This change broke Restoration Visits with `turbo-frame data-turbo-action="advance"`.

see: 
- [Issue: Turbo frame history not navigating properly w/ back button after upgrading from 7 to 8](https://github.com/hotwired/turbo/issues/1241)
- [PR: Support back button (restore) even if page loaded in frame does not contain matching HEAD](https://github.com/hotwired/turbo/pull/1281#issuecomment-2185427909)

Is this something we can revert or is there a different approach we can take?

fonji: also see #1300

Issue #780, opened 2022-10-27
Flipping content on Turbo Drive navigation


AmShaegar13:
When clicking a link using Turbo Drive, which you do by default when creating a new Rails app, the content is flipping between the current version, the old version and the new version when the content was changed for example by another user.

Imagine a counter that is incremented on every request.

1. Counter shows `1`
2. Click link to the same page
3. Counter shows `2`
4. Click link again
5. Counter immediately changes back to `1`
6. After the response arrives the counter changes to `3`

This can be observed especially with slow connections. I created a fresh rails application with the following code to demonstrate the issue:

Controller
```
def index
  session['counter'] ||= 0
end
```
View
```
<span><%= session['counter'] += 1 %></span>
<br>
<%= link_to 'Click me!', root_path %>
```
I used Firefox' Development Tools to simulate a slow connection. This is what happens:

![Screencast from 27 10 2022](https://user-images.githubusercontent.com/3840390/198225683-e9f6f998-6263-4302-b6e4-de28fd962e5c.gif)

Basically, what I try to achieve is a shared shopping cart with an indicator for how many items are in it which is present on every page. But when navigating the site and someone changes the shopping cart I get this nasty behavior.

It looks like this is an issue with caching. Because when I disable it using
```
<meta name="turbo-cache-control" content="no-cache">
```
the issue is gone. However, I think that this behavior is not intended. Why would I want and old version from the cache while waiting for the new version when I already have the current version displayed?

tleish:
See: [Persisting Elements Across Page Loads](https://turbo.hotwired.dev/handbook/building#persisting-elements-across-page-loads)

AmShaegar13:
Thanks for your reply. I tried that but it doesn't work and doesn't make sense to me. I don't want to persist the cart. It must be updated on navigation. In the counter example the counter would stay at 1 until you refresh the whole page. This is not the goal.

I know that I could change the the cart with action cable but I don't need it to be super up-to-date. Refresh on page visits enough. I thought I'd save the overhead.

tleish:
The behavior you are seeing is from Turbo [Restoration Visits](https://turbo.hotwired.dev/handbook/drive#restoration-visits).  

> It must be updated on navigation.

This is how `data-turbo-permanent`, but...
- it won't work unless you also add a unique id to the same element
- if you are navigating to the same page over and over (which is not typical behavior), then it will appear as if this feature is not working.

### GET REQUESTS

Example: without `data-turbo-permanent`
[...]

PR #895, opened 2023-03-09
Stop showing stale cache snapshot on same-page link click

compumike:
Ran into a similar problem as #780 with a stale cache entry showing when using Turbo Drive and clicking a link to the current page. I've identified and fixed the stale cache bug, and added a new test to prevent regression. This new test fails on `main` and passes on my PR branch.

Here's a simple example with a page that shows the current timestamp, and has a link to itself:

**Before:**
[before.webm](https://user-images.githubusercontent.com/1057385/224182485-5217424a-7160-4592-9644-30482a475ca7.webm)
Notice how on the 2nd-and-beyond click (red flash), Turbo briefly previews an old cache entry, and the timestamp rolls backwards briefly! :open_mouth: 

Expectation: the timestamp should never roll backwards. 

**After:**
[after.webm](https://user-images.githubusercontent.com/1057385/224182498-dea20321-4b17-4086-b47f-999de39ab821.webm)
The timestamp increases monotonically without rolling back.

**Origin of caching bug:**

`Visit#loadCachedSnapshot` had a logic bug because it would `snapshot = getCachedSnapshot()` **before** it called `this.cacheSnapshot()` to preserve the old page. This is fine if you're clicking a link from one page to different page because they operate on different cache keys. However, if you're clicking a link to the current page itself, then the old and new pages have the same cache key (equivalent to the same request URL). The incorrect get-before-set results in getting an older version of the page than is currently being shown!

To fix the bug, there are a few options. One would be to reverse the order of cache operations, and synchronously save a new snapshot to the cache before loading it right back and showing it as a preview. Alternatively, one can simply skip the rendering of the preview in this case. The `if (this.isSamePage)` was attempting to do something like that, but I'm guessing the behavior of `isSamePage` conditions changed in a way that no longer makes sense to use it to see if there's this stale cache preview issue. So I've added an `isSameCacheKey` and use that to determine whether to skip rendering the cached snapshot.

I've also added a page http://localhost:9000/__turbo/timestamp so you can easily test it out.

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -2,7 +2,7 @@ import { Adapter } from "../native/adapter"
 import { FetchMethod, FetchRequest, FetchRequestDelegate } from "../../http/fetch_request"
 import { FetchResponse } from "../../http/fetch_response"
 import { History } from "./history"
-import { getAnchor } from "../url"
+import { getAnchor, toCacheKey } from "../url"
 import { Snapshot } from "../snapshot"
 import { PageSnapshot } from "./page_snapshot"
 import { Action } from "../types"
@@ -293,11 +293,12 @@ export class Visit implements FetchRequestDelegate {
 
   loadCachedSnapshot() {
     const snapshot = this.getCachedSnapshot()
+    const isSameCacheKey = toCacheKey(this.location) === toCacheKey(this.view.lastRenderedLocation)
     if (snapshot) {
       const isPreview = this.shouldIssueRequest()
       this.render(async () => {
         this.cacheSnapshot()
-        if (this.isSamePage) {
+        if (isSameCacheKey) {
           this.adapter.visitRendered(this)
         } else {
           if (this.view.renderPromise) await this.view.renderPromise
--- a/src/tests/fixtures/timestamp.html
+++ b/src/tests/fixtures/timestamp.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html id="html">
+  <head>
+    <meta charset="utf-8">
+    <title>Turbo</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+  </head>
+  <body>
+    <p>Hello $TIMESTAMP</p>
+    <p><a href="/__turbo/timestamp">Refresh</a></p>
+  </body>
+</html>
--- a/src/tests/fixtures/timestamp.html
+++ b/src/tests/fixtures/timestamp.html
@@ -6,7 +6,7 @@
     <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
   </head>
   <body>
-    <p>Hello $TIMESTAMP</p>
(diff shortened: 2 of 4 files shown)


domchristie:
> The if (this.isSamePage) was attempting to do something like that

`isSamePage` checks if the visit is a same-page anchor visit. This is a special case so parts of the visit lifecycle are skipped (e.g. `this.view.renderPage(…)` and `this.complete()`). I don't _think_ we want to skip these operations when revisiting the same page.

> To fix the bug, there are a few options. One would be to reverse the order of cache operations, and synchronously save a new snapshot to the cache before loading it right back and showing it as a preview.

This might be worth exploring although I think there might be some history re. the order of caching/rendering so it might be fragile.

What if `getCachedSnapshot` didn't return a snapshot if it's a revisit to the same location?

compumike:
Hi @domchristie I really appreciate the feedback as I've been using Turbo for a while but am fairly new to the internals!

If `getCachedSnapshot` returned `null` when it's a click to the same location, that seems fine to me, but it seems that it would similarly skip `this.view.renderPage` and `this.complete` which you expressed some concerned about skipping.

Could you help me understand an example or use case where the current solution in this PR produces the incorrect result?

domchristie:
> Could you help me understand an example or use case where the current solution in this PR produces the incorrect result?

Hey @compumike, it may work fine :) But I suppose it's a question of which method should be responsible for performing the check, and due to the [complexity of checking of same-page anchor visits](https://github.com/hotwired/turbo/blob/4593d06ce58d17af5b17495ad8524eaa9bc2f5d2/src/core/drive/navigator.ts#L140-L150), I'm not 100% confident that changing the condition here will work in all cases

compumike:
Yes I understand, I came across those `locationWithActionIsSamePage` while I was investigating this issue. 

That's why I simplified this check it in terms of being just a pure cache logic bug if the cache key is the same, where it is will obviously show stale data if you read from a cache key just before you write to the same cache key.

domchristie:
I could be wrong, and it might not matter, but I don't _think_ that check was there to fix a caching issue; I think it's purpose was just to bypass the rendering process specifically for same-page anchor visits, as they are special cases.

> where it is will obviously show stale data if you read from a cache key just before you write to the same cache key.

In this case, does it make sense to have a snapshot if it's not valid? and is it necessary to go through the `Visit#render` flow before chhecking? If not, then it might be preferable to move this check to `getCachedSnapshot`

brunoprietog:
With the new page refreshes, there is a similar problem with cache when morphing. In #1098 I'm disabling preview rendering if it's a page refresh and if the render method is morph. Seeing this PR, I think we could take advantage of the same and disable the preview rendering on any page refresh, independent of the method. Could you test if that solves this problem too please?

brunoprietog:
The caveat of #1098 is that it is only for page refreshes, which means that it only skips the preview rendering if the page is the same and if the action is replace.

brunoprietog:
Thanks for your contribution, but the plan is simplify same page anchor visits in #1285.

PR #898, opened 2023-03-23
Allow to override turbo fetch response (or export FetchRequest class)

omohokcoj:
I'm attempting to implement http://instantclick.io functionality (prefetch on link hover) in my turbo app - one of the ways to implement it is via turbo:before-fetch-request event handler with the ability to override turbo fetch response:
https://github.com/hotwired/turbo/commit/8de3c39a0b51d27fb3c820f1d6d48ca9d5dfb477

https://gist.github.com/omohokcoj/781a18b3615e4e721ae5fa7f5d5c44ec (prefetch on hover implementation on top of the `turbo:before-fetch-request` event)

Another way to implement it is via the `FetchRequest` class monkey patch (not exported in turbo currently).

Would appreciate it a lot if any of those two commits (or both) could be merged and released in the next version.

Related issue: https://github.com/hotwired/turbo/issues/174

### DIFF
--- a/src/http/fetch_request.ts
+++ b/src/http/fetch_request.ts
@@ -6,6 +6,7 @@ export type TurboBeforeFetchRequestEvent = CustomEvent<{
   fetchOptions: RequestInit
   url: URL
   resume: (value?: any) => void
+  response?: Promise<any>
 }>
 export type TurboBeforeFetchResponseEvent = CustomEvent<{
   fetchResponse: FetchResponse
@@ -104,10 +105,10 @@ export class FetchRequest {
   async perform(): Promise<FetchResponse | void> {
     const { fetchOptions } = this
     this.delegate.prepareRequest(this)
-    await this.allowRequestToBeIntercepted(fetchOptions)
+    const event = await this.allowRequestToBeIntercepted(fetchOptions)
     try {
       this.delegate.requestStarted(this)
-      const response = await fetch(this.url.href, fetchOptions)
+      const response = await (event.detail.response || fetch(this.url.href, fetchOptions))
       return await this.receive(response)
     } catch (error) {
       if ((error as Error).name !== "AbortError") {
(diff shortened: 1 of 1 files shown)


omohokcoj:
Another solution could be via `turbo:before-fetch-request` event handler with ability to override turbo fetch response:
https://github.com/hotwired/turbo/pull/898/commits/8de3c39a0b51d27fb3c820f1d6d48ca9d5dfb477

davidalejandroaguilar:
@omohokcoj This PR might be of interest https://github.com/hotwired/turbo/pull/1000, since it touches the same files as this PR.

davidalejandroaguilar:
@seanpdoyle @dhh Have you guys looked at this? 

Oleksandr and Pete from [Docuseal](https://github.com/docusealco/docuseal) (Open Source Document Signing built with Rails and Turbo) [added](https://github.com/docusealco/docuseal/commit/787df43ac025889372ad9411291bef3076107682) a Turbo version of [Instantclick](http://instantclick.io/), and the perceived speed benefits are simply _amazing_ (they can be experienced on [their page](https://docuseal.co/).

dhh:
I'd like to see us add instaclick functionality straight into Turbo, if anyone feels like investigating that.

shiftyp:
Maybe a good place to “start”

https://github.com/hotwired/turbo/blob/3d30fe15d9c4231a6acec42567355aed98186336/src/core/frames/link_interceptor.js#L7-L17

https://github.com/hotwired/turbo/blob/3d30fe15d9c4231a6acec42567355aed98186336/src/observers/link_click_observer.js#L12-L24

Stunkymonkey:
hey @omohokcoj.

tried packaging docuseal for nixos, with success, but weird workarounds. that is why i came across this PR. 

I think this PR is superseeded by #1101 and is therefore already implemented.

maybe you can look at it again. and update to the latest turbo version in docuseal?

brunoprietog:
We already have Turbo prefetch and in the worst case scenario you can listen to the `turbo:before-fetch-request` event. Closing this.

PR #899, opened 2023-03-24
Allow remote content to have different ID

### LINKED ISSUE #151: multiple turbo-frames on page with same id
Hi,

currently we are using a custom element which does some similiar tricks as turbo-frame does. Though, when switching to turbo-frame, in one case we'd have two turbo-frames with the same id. There is a long list and we'd like to provide two entry points for a sequence of forms.

It basically works, but one would rather like to avoid having duplicate ids on a page.

Do you think this is something worth considering for an update?

kind regards
Steffen

jbennett:
This lets you specify the id of the foreign turbo_frame tag. By default this will match the id of this turbo_frame.

This is useful if you are loading multiple turbo_frames from the same endpoint on a single page, maybe with different parameters without having duplicate IDs on the page

This would resolve #151 .

```html
<turbo-frame id="pinned_items" foreign-id="items" src="/items?pinned=1"></turbo-frame>
<turbo-frame id="items" foreign-id="items" src="/items"></turbo-frame>
<turbo-frame id="items" src="/items"></turbo-frame> <!-- equivalent to previous turbo-frame -->
```

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -466,8 +466,8 @@ export class FrameController
 
   async extractForeignFrameElement(container: ParentNode): Promise<FrameElement | null> {
     let element
-    const id = CSS.escape(this.id)
-
+    const id = CSS.escape(this.foreignId)
+    
     try {
       element = activateElement(container.querySelector(`turbo-frame#${id}`), this.sourceURL)
       if (element) {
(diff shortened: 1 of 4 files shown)


brunoprietog:
Hi @jbennett, thanks for your contribution! We don't plan to expand the Turbo frames API to accept more than one frame. If you need more complex operations, I suggest you explore Turbo streams to achieve similar things

PR #908, opened 2023-04-18
Remove `requestSubmit` polyfill

seanpdoyle:
A polyfill for [HTMLFormElement.requestSubmit][] was introduced and inlined as part of the [7.1.0][] release.

As of April 2023, that method has total support and 100% [Compatibility][] with evergreen browsers.

This commit removes that polyfill.

[HTMLFormElement.requestSubmit]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit
[7.1.0]: https://github.com/hotwired/turbo/commit/d4805a4daeaf17d44e5a24c6da07c55e00099169
[Compatibility]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit#browser_compatibility


rik:
I believe it is too early to remove it. Many people are still on devices that can't upgrade to Safari 16+. See https://caniuse.com/mdn-api_htmlformelement_requestsubmit with "Usage relative" selected.

As a side note, is there a browser support policy for the project?

seanpdoyle:
Would it be an appropriate stance to consider inlined, built-in polyfills outside the bounds of `turbo@7.4.x`, while encouraging those that need them to depend on third-party packages like [event-submitter-polyfill](https://github.com/idea2app/event-submitter-polyfill) and [form-request-submit-polyfill](https://github.com/javan/form-request-submit-polyfill)?

afcapel:
As a data point, [we still support Safari 14.1+ at Basecamp](https://3.basecamp-help.com/article/102-system-requirements). Safari is particularly sticky because versions are tied to the OS and the OS is tied to the hardware so when we deprecate one Safari version we may be forcing customers to updated their hardware.

jorgemanrubia (approved):
Thanks @seanpdoyle

starzonmyarmz:
Late comment for anyone landing here from a broken build: this PR shipped in **8.0.21** alongside #909, which is a patch bump from 8.0.20. We hit this in production at Planning Center — form submissions broke for our iOS 15 users and we had to declare an incident.

If you support browsers that still need these polyfills, you'll want to pin to 8.0.20 or pull in `form-request-submit-polyfill` and `event-submitter-polyfill` directly (per @seanpdoyle's suggestion above).

I've opened #1537 to discuss the broader semver and browser-support-policy questions this raised — happy to take that conversation there rather than reopening this thread.

PR #909, opened 2023-04-18
Remove `SubmitEvent` polyfill

seanpdoyle:
The codebase's [SubmitEvent][] polyfill code involves three parts:

* a general purpose polyfill for browsers that do not provide a `window.SubmitEvent` prototype that includes a `submitter` property for their [submit][] events
* a Safari-specific bugfix
* a TypeScript `type` declaration.

As of April 2023, all facets of the `SubmitEvent` have fully support and [Compatibility][] with evergreen browsers.

As of 2021-08-30, the [Safari bug][] mentioned in the code comment has been resolved, and a fix has been released as part of Safari 15.4.

The TypeScript compiler supports the `SubmitEvent` type declaration.

This commit removes the polyfill from the codebase, and relies on the built-in support instead.

[submit]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event
[SubmitEvent]: https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent
[Compatibility]: https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent#browser_compatibility
[Safari bug]: https://bugs.webkit.org/show_bug.cgi?id=229660


src/polyfills/submit-event.ts
```diff
@@ -1,44 +0,0 @@
...
-  // Certain versions of Safari 15 have a bug where they won't
-  // populate the submitter. This hurts TurboDrive's enable/disable detection.
-  // See https://bugs.webkit.org/show_bug.cgi?id=229660
-  if ("SubmitEvent" in window && /Apple Computer/.test(navigator.vendor)) {
-    prototype = window.SubmitEvent.prototype
-  } else if ("SubmitEvent" in window) {
-    return // polyfill not needed
-  }
-
-  addEventListener("click", clickCaptured, true)
```
seanpdoyle, line 32: On a related note, capturing a `click` event for a submission **does not** polyfill the built-in behavior for calls to `HTMLFormElement.requestSubmit()` with a `submitter` argument. For example:

```html
<form>
  <button id="the-submitter">Submit</button>
  <button type="reset" id="not-the-submitter">Reset</button>
</form>

<script>
  const form = document.querySelector("form")
  const submitter = document.getElementById("the-submitter")

  form.requestSubmit(submitter)
</script>
```

In this case, the `submitter` argument is ignored, since it's never clicked and will not dispatch a `click` event.

rik:
I believe it is too early to remove it. Many people are still on devices that can't upgrade to Safari 15.4+. See https://caniuse.com/mdn-api_submitevent_submitter with "Usage relative" selected.

seanpdoyle:
Would it be an appropriate stance to consider inlined, built-in polyfills outside the bounds of `turbo@7.4.x`, while encouraging those that need them to depend on third-party packages like [event-submitter-polyfill](https://github.com/idea2app/event-submitter-polyfill) and [form-request-submit-polyfill](https://github.com/javan/form-request-submit-polyfill)?

PR #910, opened 2023-04-20
Export Preloader to use preloadURL dynamically

julianrubisch:
# Why?
I am currently testing the preloader's capabilities and found that it could be beneficial to call `preloadURL` dynamically, for example

- using an `IntersectionObserver`, like https://github.com/GoogleChromeLabs/quicklink, for example.
- using `mouseover` events, like https://instant.page/
- using `MutationObservers`, (e.g. in Stimulus controllers) that listen for a certain data attribute to be set,
- or even in other custom web components


julianrubisch:
heads up, I've put up an alternative here: #911

PR #913, opened 2023-04-22
Fix navigation to the turbo-root not using Turbo Drive

### LINKED ISSUE #912: Navigating to the `turbo-root` doesn't use Turbo Drive
If you configure a Turbo root as follows:

```html
<meta name="turbo-root" content="/app">
```

Turbo Drive will handle navigations to URLs within that root (like `/app/foo`), but not navigations to the root itself (`/app`).

We can see why this happens by following the logic from the [`locationIsVisitable` function](https://github.com/hotwired/turbo/blob/4593d06ce58d17af5b17495ad8524eaa9bc2f5d2/src/core/url.ts#L36):

1. `locationIsVisitable` is called with `location = http://example.com/app` and `rootLocation = http://example.com/app`.
2. These are passed into `isPrefixedBy`, which via `getPrefix`, adds a trailing slash to the root URL:

```ts
// location -> baseURL, rootLocation -> url
export function isPrefixedBy(baseURL: URL, url: URL) {
  const prefix = getPrefix(url) // = http://example.com/app/
  return baseURL.href === expandURL(prefix).href || baseURL.href.startsWith(prefix)
}
```

3. Now the comparison is `'http://example.com/app' === 'http://example.com/app/' || 'http://example.com/app'.startsWith('http://example.com/app/')`, which is false. No Turbo Drive.

tobyzerner:
Fixes #912 by simply not adding a trailing slash to the root URL prefix.

I'm not sure if this is the best way to fix this issue, but it makes the test case pass and doesn't break any other tests.

### DIFF
--- a/src/core/url.ts
+++ b/src/core/url.ts
@@ -59,9 +59,5 @@ function getLastPathComponent(url: URL) {
 }
 
 function getPrefix(url: URL) {
-  return addTrailingSlash(url.origin + url.pathname)
-}
-
-function addTrailingSlash(value: string) {
-  return value.endsWith("/") ? value : value + "/"
+  return url.origin + url.pathname;
 }
--- a/src/tests/fixtures/root/index.html
+++ b/src/tests/fixtures/root/index.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html id="html">
+  <head>
+    <meta charset="utf-8">
+    <title>Turbo</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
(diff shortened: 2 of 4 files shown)


seanpdoyle:
@afcapel @kevinmcconnell could we grant permission to execute CI on this branch?

afcapel:
We need to fix the merge conflict before we can run CI.

fritzmg (approved):
I can confirm the [original issue](https://github.com/hotwired/turbo/issues/912) and that this PR fixes it.

Any chance to get this merged and released?

PR #923, opened 2023-05-10
Make Turbo Frames' IntersectionObserver configurable

julianrubisch:
This PR adds configuration options akin to those controlling the autoscroll behavior to `AppearanceObserver`.

Note that I allowed only one threshold value (which could be many, see https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#thresholds, but that doesn't make sense here).

Note further that all those options are optional, and just fall back to the default if omitted.

### DIFF
--- a/src/observers/appearance_observer.js
+++ b/src/observers/appearance_observer.js
@@ -4,7 +4,7 @@ export class AppearanceObserver {
   constructor(delegate, element) {
     this.delegate = delegate
     this.element = element
-    this.intersectionObserver = new IntersectionObserver(this.intersect)
+    this.intersectionObserver = new IntersectionObserver(this.intersect, readIntersectionObserverOptions(element))
   }
 
   start() {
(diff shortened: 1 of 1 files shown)


adrienpoly:
I love it this is really useful to start lazy loading before the element appears.

Little naming comments:
- most of the data attributes in Turbo are prefixed by `turbo`. Shouldn't we have here `data-turbo-intersection-root-selector`, `data-turbo-intersection-root-margin` etc

julianrubisch:
Thanks! FWIW I took inspiration from the autoscroll attributes: https://turbo.hotwired.dev/reference/frames#html-attributes

It's `data-autoscroll-...` there.

julianrubisch:
heads up: I've rewritten this in pure JS, will add a test if you think it's worth it ✌

adrienpoly:
Hello @afcapel 

Thanks for the beta of Turbo 8. I was wondering if this PR could be part of the final V8

As per my comment https://github.com/hotwired/turbo/pull/923#issuecomment-1542776338 we could add a `turbo` prefix. Julian went that way to be consistent with `data-autoscroll` that was introduced before. 

Given that we have a major release this could be a good time to rename `data-autoscroll` in `data-turbo-autoscroll` and introduce those new settings for the intersectionObserver

- data-turbo-intersection-root-selector
- data-turbo-intersection-root-margin
- data-turbo-intersection-threshold

I have seen quite a few apps where a custom stimulus controller was created just for that. I think it would be a good DX improvement

rik:
I think those options are essential to mimic the native lazy loading from browsers. They load images and iframes before they intersect with the viewport so that it's likely to be ready when lazy loaded elements appear on screen.

rik:
Ping @dhh since you seem to be focused on Hotwire at the moment.

adrienpoly:
Hello @brunoprietog,

I was wondering if you would consider this PR if we were to reopen it.
I personally have three apps where I use a custom Stimulus controller for this functionality.

Here is a typical example: https://github.com/adrienpoly/rubyvideo/blob/4d98815a36a557d36fb99a911cde82efaef1f777/app/javascript/controllers/lazy_loading_controller.js

Defining IntersectionObserver options really helps to make lazy-loaded Turbo Frames almost imperceptible.

Issue #924, opened 2023-05-24
Proposal: Rename `Vnd.Prefetch` header set when preloading links


pfeiffer:
Introduced in #552 by @hey-leon is the ability to preload links tagged with `[data-turbo-preload]`. This works great for prepopulating the cache. The links are fetched with the header `Vnd.Prefetch` set to `true`, which the server can use to determine that the request was made as a preloaded request. To be honest it's not clear where the `Vnd.Prefetch` name for the header originated - it was introduced in the PR and I see no mention of it elsewhere than in the Turbo PR.

One gotcha with this is that [CloudFlare seems to strips all request headers containing dots](https://community.cloudflare.com/t/how-to-allow-dot-in-headers/142183). It seems like this also happens for other vendors as the header name is invalid(?). For instance [Oracle](https://support.oracle.com/knowledge/Oracle%20Cloud/2748855_1.html), [nginx](https://serverfault.com/questions/994595/nginx-filtering-headers-with-a-dot-in-the-header-name-even-after-enabling-ignor) etc. An easy fix would be to rename the header to something else that doesn't contain a dot.

At the same time it could be worth considering if the header should be renamed to something that conforms with what other vendors use. Safari and Mozilla uses a `(X-)Purpose: prefetch` for prefetching `<link prefetch ..>` 

Chrome, Safari and Facebook in-app browser use a `X-Purpose: preview` header to indicate prefetching a preview of a page before the user navigates to it. I think this would be very similar to how the Turbo preloading happens, as Turbo per default triggers a re-fetch of a preloaded page when tapping the preloaded link.

TLDR; It seems like `vnd.prefetch` header is invalid and stripped by many vendors. Setting `X-Purpose: preview` would probably be a better option and conform with the behavior of other vendors for fetching a light preview of a page before fetching it fully.

Happy to provide a PR that changes this!

rik:
The current standard for this is [`Sec-Purpose: prefetch`](https://fetch.spec.whatwg.org/#sec-purpose-header). [Browsers might not yet be up-to-date with the spec](https://chromestatus.com/feature/6247959677108224) but if possible, I think using that header would be more forward looking.

hey-leon:
@pfeiffer good call out. 

I think I was mistaken when adding this header. Some samples of code I found before implementing this feature included it, and I believe I included it under a false assumption it was used by turbo. If you search the repo on github for the header it does not come up (other than the case I added). I think we are probably good to simply remove the header or opt for a standardised version.

[...]

PR #925, opened 2023-05-24
Rename header for preload requests

pfeiffer:
This PR renames header and value for preloaded requests to `x-purpose: preview`. See discussion in #924.

### DIFF
--- a/src/core/drive/preloader.ts
+++ b/src/core/drive/preloader.ts
@@ -42,7 +42,7 @@ export class Preloader {
     }
 
     try {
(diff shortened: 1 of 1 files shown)


pfeiffer:
Thanks @afcapel! I've updated this PR with the header name and value matching the standard spec.

This is a breaking change and would require version bump. It would be great to have this included in Turbo as the current preload header is stripped by eg. Cloudflare and thus pretty unusable.

afcapel:
Thanks @pfeiffer, we'll bump the version before releasing this.

PR #926, opened 2023-05-24
Proposal: pass isPreview value into the event detail in render events

Matt-Yorkley:
Passes the renderer's `isPreview` attribute into the event detail on `'turbo:before-render'` and `'turbo:render'` events.

`isPreview` signifies whether the content currently being rendered is from the turbo cache or not, and this bit of information is very useful for any listeners that might want to act on those two events differently for cached or non-cached content.

Eg, the following listeners:
```ruby
  console.log("turbo:before-render")
  console.log(event.detail)
})

document.addEventListener("turbo:render", (event) => {
  console.log("turbo:render")
  console.log(event.detail)
})
```
now produce these event details (when navigating to a page that has a cache entry, both events fire twice):
Screenshot from 2023-05-24 13-35-12

So it's now possible to do things like...
```ruby
  if (event.detail.isPreview) {
    resetTheThings()
    modifyTurboCachedContent(event.detail.newBody)
  } else {
    modifyFinalContent(event.detail.body)
  }
})
```

### DIFF
--- a/src/core/frames/frame_controller.ts
+++ b/src/core/frames/frame_controller.ts
@@ -288,7 +288,11 @@ export class FrameController
 
   // View delegate
 
-  allowsImmediateRender({ element: newFrame }: Snapshot<FrameElement>, options: ViewRenderOptions<FrameElement>) {
+  allowsImmediateRender(
+    { element: newFrame }: Snapshot<FrameElement>,
+    _isPreview: boolean,
+    options: ViewRenderOptions<FrameElement>
+  ) {
     const event = dispatch<TurboBeforeFrameRenderEvent>("turbo:before-frame-render", {
       target: this.element,
       detail: { newFrame, ...options },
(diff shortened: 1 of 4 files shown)


afcapel:
Looks good, thanks @Matt-Yorkley 👍

domchristie:
Could this be inferred from `html[data-turbo-preview]` attribute?:

```js
document.addEventListener("turbo:before-render", () => {
  console.log("turbo:before-render")
  console.log(document.documentElement.hasAttribute('data-turbo-preview'))
})

document.addEventListener("turbo:render", () => {
  console.log("turbo:render")
  console.log(document.documentElement.hasAttribute('data-turbo-preview'))
})
```

afcapel:
@Matt-Yorkley @domchristie is making a good point. Can you use `data-turbo-preview` instead?

Matt-Yorkley:
> @Matt-Yorkley @domchristie is making a good point. Can you use `data-turbo-preview` instead?

Ah, I must have missed `data-turbo-preview` in the documentation. Yeah, @domchristie's comment about preferring to store state in the DOM sounds good https://github.com/hotwired/turbo/pull/1007#issuecomment-1816888966

PR #933, opened 2023-06-14
Fix for SubmitEvent polyfill breaking requestSubmit() for newer Safari versions

mrtnin:
This PR addresses an issue where the current polyfill for SubmitEvent is not functioning as expected on the newer versions of Safari. It appears that the original webkit issue the polyfill was meant to address has been resolved, and the polyfill is now introducing undesired behavior.

Specifically, this issue is observed when invoking the requestSubmit() method with a submitter as an argument in Safari 16.5. According to the current behavior, the submitter argument seems to be ignored (ref https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit).

In an attempt to correct this, I've made adjustments to the polyfill logic and added a browser test to confirm the fix. The test fails when executed in the webkit environment with the old polyfill but passes successfully with the revised version in this PR. The command used to run the test is as follows:

```bash
yarn test:browser src/tests/functional/form_submission_tests.ts:237 --project=webkit
```

Please note that for this test to work, I had to add the webkit browser to the playwright.config.ts. I've omitted those changes in this PR to keep it focused on the issue at hand.

### DIFF
--- a/src/polyfills/submit-event.ts
+++ b/src/polyfills/submit-event.ts
@@ -23,10 +23,14 @@ function clickCaptured(event: Event) {
   // Certain versions of Safari 15 have a bug where they won't
   // populate the submitter. This hurts TurboDrive's enable/disable detection.
   // See https://bugs.webkit.org/show_bug.cgi?id=229660
-  if ("SubmitEvent" in window && /Apple Computer/.test(navigator.vendor)) {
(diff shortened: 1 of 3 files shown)


dhh:
Thanks! Please do open a PR with the playwright changes as well so we have tests.

PR #948, opened 2023-07-20
Add ! to the copyright banner comment

afcapel:
So that build tools like esbuild or rollup-plugin-license can detect the banner and keep the license in the minified output.

The banner is only a few bytes, so bundle size is not a concern.

Ref. https://github.com/evanw/esbuild/issues/2745


kevinmcconnell (approved):
Looks great! And well spotted!

PR #949, opened 2023-07-20
Add disk cache store

afcapel:
This commit extends the Turbo cache API to allow for the use of different cache stores. The default store is still the in-memory store, but a new persistent cache store is now available.

The disk store uses the [Cache API](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage) to store and retrieve snapshots from the browser's cache. This allows for the snapshots to be persisted across different tabs, page reloads and even browser restarts.

The disk store is not enabled by default. To enable it, you need to set the `Turbo.cache.store` property to `"disk"`.

```js
Turbo.cache.store = "disk"
```

This is also a stepping stone to implement offline support with Service Workers. With a Service Worker in place, and the disk cache store enabled, we can still serve cached snapshots even when the browser is offline.

### DIFF
--- a/src/core/cache.js
+++ b/src/core/cache.js
@@ -1,4 +1,5 @@
 import { setMetaContent } from "../util"
+import { SnapshotCache } from "./drive/snapshot_cache"
 
 export class Cache {
   constructor(session) {
@@ -6,7 +7,7 @@ export class Cache {
   }
 
   clear() {
-    this.session.clearCache()
+    this.store.clear()
   }
 
   resetCacheControl() {
@@ -19,6 +20,18 @@ export class Cache {
 
   exemptPageFromPreview() {
     this.#setCacheControl("no-preview")
+  }
+
+  set store(store) {
+    if (typeof store === "string") {
+      SnapshotCache.setStore(store)
+    } else {
+      SnapshotCache.currentStore = store
+    }
+  }
+
+  get store() {
+    return SnapshotCache.currentStore
   }
 
   #setCacheControl(value) {
--- a/src/core/drive/cache_stores/disk_store.js
+++ b/src/core/drive/cache_stores/disk_store.js
@@ -1,0 +1,64 @@
+import { PageSnapshot } from "../page_snapshot"
+
+export class DiskStore {
+  _version = "v1"
+
+  constructor() {
+    if (typeof caches === "undefined") {
+      throw new Error("windows.caches is undefined. CacheStore requires a secure context.")
+    }
+
+    this.storage = this.openStorage()
+  }
+
+  async has(location) {
+    const storage = await this.openStorage()
+    return (await storage.match(location)) !== undefined
+  }
+
+  async get(location) {
+    const storage = await this.openStorage()
+    const response = await storage.match(location)
+
+    if (response && response.ok) {
+      const html = await response.text()
+      return PageSnapshot.fromHTMLString(html)
+    }
+  }
+
+  async put(location, snapshot) {
+    const storage = await this.openStorage()
+
+    const response = new Response(snapshot.html, {
+      status: 200,
+      statusText: "OK",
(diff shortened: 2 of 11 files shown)


src/core/drive/cache_stores/disk_store.ts
```diff
@@ -0,0 +1,54 @@
...
+        "Content-Type": "text/html",
+      },
+    })
+    await this.storage.put(location, response)
+    return snapshot
+  }
+
+  async clear() {
+    const keys = await this.storage.keys()
+    await Promise.all(keys.map((key) => this.storage.delete(key)))
```
kevinmcconnell, line 43: I wonder if there's a point where it's quicker to delete the cache itself, rather than the individual items. Perhaps it's very unlikely to come up in any case. But I wondered if that's something you looked at?

src/tests/functional/disk_cache_tests.ts
```diff
@@ -0,0 +1,58 @@
+import { Page, test, expect } from "@playwright/test"
+import { nextBody } from "../helpers/page"
+
+const path = "/src/tests/fixtures/disk_cache.html"
+
+test.beforeEach(async ({ page }) => {
+  await page.goto(path)
+})
+
+test("test stores pages in the disk cache", async ({ page }) => {
```
kevinmcconnell, line 10: I think this does verify that the cache gets populated, but is it also worth testing that those cached entries definitely get used? Maybe the test could listen for the `turbo:render` event for the cached version when following a link?

src/core/cache.ts
```diff
@@ -24,6 +26,18 @@ export class Cache {
     this.setCacheControl("no-preview")
   }
 
+  set store(store: string | CacheStore) {
```
kevinmcconnell, line 29: The cache API is only available in a secure context. It's an edge case for people to be on plain http these days, but just in case, would it be worth checking that the API is available when setting the cache, and raising an error with an explanation if it's not?

kevinmcconnell (approved):
This looks great! 🎉 It's a very good idea, and the implementation is simple and clear 👏

I left a couple of questions but it's all just minor curiosity stuff.

Also, as we discussed elsewhere, it'll be worth us looking at ways to keep more of the cache for longer, rather than clear it after every mutation. I reckon there's lot of potential in having a longer-lived cache once this is in place. But that seems better explored separately.

src/core/drive/cache_stores/disk_store.ts
afcapel: The cache API only has a method to delete individual keys https://developer.mozilla.org/en-US/docs/Web/API/Cache

src/tests/functional/disk_cache_tests.ts
afcapel: As discussed offline, it'd be nice to have a test checking the cache (we don't have any at the moment), but for some reason this doesn't work in the test env. Will have to punt this for now.

src/core/cache.ts
afcapel: 👍 added in 2d0985c7bc131bfb3a72c40bf0cb7941baf50bd4

src/core/drive/cache_stores/disk_store.ts
kevinmcconnell: You can delete the `Cache` instance though, instead of deleting the individual items that it contains: https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage/delete

afcapel: Ah, I missed that 👍

I think deleting the keys is ok, though. Don't think there will be a performance penalty, and because the operations are async there's a small chance of race condition if we delete the cache and reopen it again when there are other operations inflight.

dyaskur (changes requested):
there is a conflict. Please fix it

PR #953, opened 2023-07-26
Perform view transitions on cached snapshots

kevinmcconnell:
Extends the recently-added support for view transitions to better support cached snapshot previews.

Whenever Turbo navigates to a page with a cached snapshot, we actually render the page twice: first with the preview, then with the freshly loaded version. To prevent this extra rendering from interfering with a view transition, we previously disabled snapshot caching for pages that had opted in to using a view transition. This keeps the transitions uninterrupted, but has the disadvantage of preventing caching in cases where we might still want it.

We can provide a better experience by allowing caching as normal, but performing the view transition on whichever render happens first. In the case of a cached snapshot, this allows us to fetch the real content while we're transitioning to the cached view; when the new content is ready and the transition has complete, we swap the new content into place.

This also means that history navigations can still be served entirely from cache, as they would be when not using view transitions.

transition

In cases where a particular transition looks better without the snapshot, we can use Turbo's existing mechanisms to opt out of caching on those specific pages.

Thanks to @domchristie for suggesting this change 🙏


src/core/drive/view_transitioner.ts
```diff
@@ -0,0 +1,31 @@
...
+        await document.startViewTransition!(render).finished
+      })
+    } else {
+      this.lastOperation = this.lastOperation.then(render)
+    }
+
+    return this.lastOperation
+  }
+
+  private get canTransition() {
```
afcapel, line 28: Maybe we can call this property `viewTransitionAvailable`? To be a bit more specific, there're a few different reasons why you can or can't use transitions, like meta tags.

src/core/drive/visit.ts
```diff
@@ -472,6 +477,13 @@ export class Visit implements FetchRequestDelegate {
     delete this.frame
   }
 
+  async renderPageSnapshot(snapshot: PageSnapshot, isPreview: boolean) {
+    await this.viewTransitioner.renderChange(this.view.shouldTransitionTo(snapshot), async () => {
+      await this.view.renderPage(snapshot, isPreview, this.willRender, this)
+      this.performScroll()
```
afcapel, line 483: I'd move the scrolling call outside `renderPageSnapshot`. We always call render and scroll together, but conceptually scrolling feels like a different concern than rendering.

afcapel (approved):
LGTM 👍

src/core/drive/view_transitioner.ts
kevinmcconnell: Good idea 👍

src/core/drive/visit.ts
kevinmcconnell: It's here because correcting the scroll position needs to happen within the transition animation. Otherwise it could transition to the page at the wrong scroll position, and then jump back to the correct scroll position after the animation is over.

So in that sense, I think scrolling _is_ part of rendering the snapshot, in as much as it's all about getting the right content into the right place on the screen. I could look at combining them into another concern to see if it makes that clearer. Might do that separately though.

Issue #954, opened 2023-07-28
aria-busy attribute is not being toggled after a Form Submit


JuannFerrari:
**Turbo version: 7.3.0**

**Description:**

According to the Handbook:

> In tandem with the progress bar, Turbo Drive will also toggle the [[aria-busy] attribute](https://www.w3.org/TR/wai-aria/#aria-busy) on the page’s <html> element during page navigations started from Visits or Form Submissions. Turbo Drive will set [aria-busy="true"] when the navigation begins and will remove the [aria-busy] attribute when the navigation completes.

After a Form Submit, the progress bar is shown but the aria-busy attribute is not being toggled.

**Steps to reproduce:**
Make a GET form submission (like for filtering records) with Turbo Drive enabled.

**Expected Behavior:**
`.turbo-progress-bar` element should be appended to the DOM and the `aria-busy` attribute on the page's <html> element should be set to `true`

**Actual Behavior:**
`.turbo-progress-bar` element is being appended but the `aria-busy` attribute is not being toggled

**Proposed Solution:**
 I checked this PR https://github.com/hotwired/turbo/pull/317 about how the progress bar is being handled for FormSubmissions and is different  from the form submission method at `frame_controller` (because the aria-busy attribute is working for turbo frames)

```
  // Form submission delegate

  formSubmissionStarted({ formElement }: FormSubmission) {
    markAsBusy(formElement, this.findFrameElement(formElement))
  }
```

Should we call `markAsBusy` at `browser_adapter#formSubmissionStarted`  to toggle the aria-busy attribute or I'm missing something?

seanpdoyle:
@JuannFerrari thank you for opening this issue.

In an attempt to to reproduce the behavior you're describing here, I've added some additional coverage to the Form Submission test suite in https://github.com/hotwired/turbo/pull/955.

Unfortunately, the new assertions pass without any implementation-side changes. Are there any other details you could provide that might contribute to the unexpected behavior?

JuannFerrari:
Hi @seanpdoyle thanks for your answer. I did the same and the assertions pass too... I don't know what's really happening :/.
The stack that I'm using is Ruby on Rails 7 + turbo_rails gem and the form for filtering my records is provided by ransack + simpleform.

I have even tried writing a pure HTML form and it is not toggling the attribute either.

[...]

PR #955, opened 2023-07-29
Add test coverage toggling `[aria-busy]` for `GET` Form Submissions

seanpdoyle:
[Coverage for 954][]

Assert that `[aria-busy="true"]` at the start of a `GET` request
initiated by a `<form>` element and handled as a page-wide navigation by
Turbo Drive.

[Coverage for 954]: https://github.com/hotwired/turbo/issues/954


afcapel:
Thanks @seanpdoyle 👍

Issue #932, opened 2023-06-10
Proposal: Turbo Stream apply to Template Tags


faqndo97:
## TL;DR

This proposal involves enhancing the functionality of Turbo streams to manipulate the content within a template tag. At present, Turbo streams are unable to access the content within a template tag due to the inability of `getElementById` and `querySelectorAll` to reach the `DocumentFragment`.

## Proposal

Consider the following scenario where we have a form that contains a template tag:

```erb
<%= form_tag url: some_path, model: ResourceA.new, data: { controller: "nested-form"} do |f| %>
  <%# TEMPLATES %>
    <template data-nested-form-target="nestedFields" data-fields-type="resourceB" >
      <%= f.fields_for :resource_b, ResourceB.new, child_index: 'TEMPLATE_RECORD' do |resource_b_form| %> 
        <%= resource_b_form.select :options, ... %>
      <% end %>
    </template>
  <%# END TEMPLATES %>

  <%= f.select :values, ... %>

  <%= f.fields_for :resource_b do |resource_b_form| %> 
    <%= resource_b_form.select :options, ... %>
  <% end %>

  <span data-action="click->nested-form#add_relationship" data-nested-form-field-type-param="resourceB">Add more resource B</span>
<% end %>
```

In this form, when a user clicks on the span, the `nested-form#add_relationship` is activated. This mechanism scans the template, takes the `innerHTML`, replaces `TEMPLATE_RECORD` with the current timestamp (`new Date().getTime()`) and appends it to the end of the form.

This functionality worked flawlessly until the requirement changed to "Update the options on `resource_b_form.select :options` when `f.select :values, changes`". Initially, I assumed this would be straightforward. By assigning a stimulus controller to `f.select :values`, I expected that any changes would prompt a server request using `request.js`. Subsequently, the Turbo stream would use a CSS selector to update all the options on the page. So far this continue working perfectly.

However, when a new `resourceB` was added, it did not contain the updated values from the Turbo stream. This issue arises because Turbo cannot reach the content inside a template tag, as it is a `DocumentFragment` which is not accessible by `getElementById` and `querySelectorAll`.

To address this, I propose that we should have the option to target these elements with Turbo streams. Template tags have numerous advantages: they are hidden by default, they don't initiate stimulus controllers, and they don't trigger anything else. The only disadvantage we face is the limited functionality of Turbo streams in this context.

[...]

PR #957, opened 2023-08-03
Streams search for element(s) inside template tag

viniciusoyama:
Up until now `StreamElement` searches for a target (by id) or targets (by generic css selector) by traversing the DOM using `document.getElementById` or `document.querySelectorAll`. These methods doesn't search for tags inside templates. 

This PR changes `StreamElement` to also search for target/targets inside templates. 

Ref: https://github.com/hotwired/turbo/issues/932

### DIFF
--- a/src/elements/stream_element.js
+++ b/src/elements/stream_element.js
@@ -169,22 +169,31 @@ export class StreamElement extends HTMLElement {
   }
 
   get targetElementsById() {
+    const elements = []
     const element = this.ownerDocument?.getElementById(this.target)
 
     if (element !== null) {
-      return [element]
-    } else {
-      return []
+      elements.push(element)
     }
+
+    for (const template of (this.ownerDocument?.querySelectorAll("template") || [])) {
+      const elementInTemplate = template.content.getElementById(this.target)
+
+      if (elementInTemplate !== null) {
+        elements.push(elementInTemplate)
+      }
+    }
+
+    return elements
   }
 
   get targetElementsByQuery() {
-    const elements = this.ownerDocument?.querySelectorAll(this.targets)
+    const elements = [...(this.ownerDocument?.querySelectorAll(this.targets) || [])]
 
-    if (elements.length !== 0) {
-      return Array.prototype.slice.call(elements)
-    } else {
-      return []
+    for (const template of (this.ownerDocument?.querySelectorAll("template") || [])) {
+      elements.push(...template.content.querySelectorAll(this.targets))
     }
+
+    return elements
   }
 }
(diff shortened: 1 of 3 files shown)


src/elements/stream_element.ts
```diff
@@ -166,22 +166,34 @@ export class StreamElement extends HTMLElement {
...
-      return []
+    const elementInDom = this.ownerDocument?.getElementById(this.target!)
+    
+    let foundElements : HTMLElement[] = [];
+    
+    if (elementInDom !== null) {
+      foundElements = [elementInDom]
     }
+
+    this.ownerDocument?.querySelectorAll("template").forEach((template) => {
```
seanpdoyle, line 177: Since [querySelectorAll](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll) returns a `NodeList`, what do you think about using a `for` loop and `Array.push` here:

```js
const elements = []
const element = this.ownerDocument?.getElementById(this.target!)

if (element !== null ) {
  elements.push(element)
}

for (const template of this.ownerDocument?.querySelectorAll("template")) {
  const elementInTemplate = template.content.getElementById(this.target!)

  if (elementInTemplate) {
    elements.push(elementInTemplate)
  }
}

return elements
```

src/elements/stream_element.ts
```diff
@@ -166,22 +166,34 @@ export class StreamElement extends HTMLElement {
...
+        foundElements.push(elementInTemplate)
+      }
+    });
+
+    return foundElements;
   }
 
   private get targetElementsByQuery() {
-    const elements = this.ownerDocument?.querySelectorAll(this.targets!)
+    let elements = Array.prototype.slice.call(this.ownerDocument?.querySelectorAll(this.targets!))
```
seanpdoyle, line 189: Similar to https://github.com/hotwired/turbo/pull/957/files#r1283567754, we can avoid the `Array.prototype.slice.call` with a `for` loop, and the `Array.prototype.slice.call` with `Array.push` and the `...` spread operator:

```js
const elements = this.ownerDocument?.querySelectorAll(this.targets!)

for (const template of this.ownerDocument?.querySelectorAll("template")) {
  elements.push(...template.content.querySelectorAll(this.targets!))
}

return elements
```

src/elements/stream_element.ts
viniciusoyama: Makes total sense. Thanks 👍 .

src/elements/stream_element.ts
viniciusoyama: Will do that 👍 .

seanpdoyle:
@jorgemanrubia could you enable CI for this PR?

Issue #958, opened 2023-08-14
`turbo:load` is not dispatched when response status is 404/Not Found


domchristie:
`turbo:before-render` and `turbo:render` are dispatched, but `turbo:load` is not.

PR #960, opened 2023-08-14
Dispatch turbo:load after loading 404 response

domchristie:
I'm not 100% sure if this is the correct approach, but as far as I can tell, this just dispatches the `turbo:load` event after rendering a 404.

Fixes https://github.com/hotwired/turbo/issues/958

### DIFF
--- a/src/core/drive/visit.ts
+++ b/src/core/drive/visit.ts
@@ -202,6 +202,7 @@ export class Visit implements FetchRequestDelegate {
     if (this.state == VisitState.started) {
       this.state = VisitState.failed
       this.adapter.visitFailed(this)
+      this.delegate.visitCompleted(this)
     }
   }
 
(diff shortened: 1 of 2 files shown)


afcapel:
I think this is correct. `visitCompleted` is basically [clearing the busy state and firing the load event](https://github.com/hotwired/turbo/blob/6915c54d78eff8650c5ac2ca5306390bbb57d10e/src/core/session.ts#L223-L224), both of which we want on load.

@domchristie happy to merge if you fix the merge conflict.

domchristie:
@afcapel I suppose the only consideration is if we want to maintain symmetry with `complete` i.e. call `this.delegate.visitFailed(this)` (which would then just call `this.visitCompleted`

afcapel:
@domchristie I think it's better if we avoid the indirection and keep the call to `this.visitCompleted` inside `Visit` itself.

PR #976, opened 2023-09-07
perf: Replace all occurrences of `0` with `undefined | undefined`

tomhicks:
## What?

Replaces all occurrences of the literal `0` with `undefined | undefined`

## Why?

In a #971, a new pattern was introduced of using the construct `undefined | undefined` which uses bitwise operations to obtain the literal value `0`.

As bitwise operations are always faster than anything else, this will be a performance enhancement.

### DIFF
--- a/src/core/drive/progress_bar.js
+++ b/src/core/drive/progress_bar.js
@@ -8,28 +8,28 @@ export class ProgressBar {
       .turbo-progress-bar {
         position: fixed;
         display: block;
-        top: 0;
-        left: 0;
+        top: undefined | undefined;
+        left: undefined | undefined;
         height: 3px;
         background: #0076ff;
         z-index: 2147483647;
         transition:
           width ${ProgressBar.animationDuration}ms ease-out,
           opacity ${ProgressBar.animationDuration / 2}ms ${ProgressBar.animationDuration / 2}ms ease-in;
-        transform: translate3d(0, 0, 0);
+        transform: translate3d(undefined | undefined, undefined | undefined, undefined | undefined);
       }
     `
   }
 
   hiding = false
-  value = 0
+  value = undefined | undefined
   visible = false
 
   constructor() {
     this.stylesheetElement = this.createStylesheetElement()
     this.progressElement = this.createProgressElement()
     this.installStylesheetElement()
-    this.setValue(0)
+    this.setValue(undefined | undefined)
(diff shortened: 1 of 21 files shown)


Decencies (approved):

nstojanovikj (approved):

d-bucur:
This looks very much like idiomatic JS, I approve

image

src/core/drive/progress_bar.js
```diff
@@ -64,14 +64,14 @@ export class ProgressBar {
   }
 
   installProgressElement() {
-    this.progressElement.style.width = "0"
+    this.progressElement.style.width = "0"|"0"
```
maurer2, line 67: Hello, this seems to change the type from string to number. Perhaps the previous version should be kept here.

src/core/drive/progress_bar.js
```diff
@@ -64,14 +64,14 @@ export class ProgressBar {
...
     this.progressElement.style.opacity = "1"
     document.documentElement.insertBefore(this.progressElement, document.body)
     this.refresh()
   }
 
   fadeProgressElement(callback) {
-    this.progressElement.style.opacity = "0"
+    this.progressElement.style.opacity = "0"|"0"
```
maurer2, line 74: Same as above.

tomhicks: This one I think we should change. Let's make sure we enjoy JavaScript in the glorious spirit it was originally designed and keep the interpreter on its toes by sometimes passing a string and sometimes a number.

KaulSe (approved):
undefined

tomhicks:
> undefined

0?

KnlnKS (approved):

src/core/index.js
```diff
@@ -80,7 +80,7 @@ export function renderStreamMessage(message) {
  * Removes all entries from the Turbo Drive page cache.
  * Call this when state has changed on the server that may affect cached pages.
  *
- * @deprecated since version 7.2.0 in favor of `Turbo.cache.clear()`
+ * @deprecated since version 7.2.undefined | undefined in favor of `Turbo.cache.clear()`
```
muescha, line 83:
```suggestion
 * @deprecated since version 7.2.(undefined | undefined) in favor of `Turbo.cache.clear()`
```

PR #984, opened 2023-09-07
Started adding JSDoc

shashkovdanil:
Even though I love TypeScript, I respect the decision made by the owners. I can admit that sometimes TypeScript causes a headache and you have to use `as/any/unknown`. It's also hard to read and understand what is happening in code like this.

```ts
export type GetKeysByValue<T extends Values> = {
  [K in keyof ItemMap]: ItemMap[K] extends { value: infer U } ? (U extends T ? K : never) : never;
}[keyof ItemMap];
```

However, it must be acknowledged that the code's readability has decreased. It will be harder for users to understand how to use turbo, and for other contributors to understand what functions and methods are doing. And instead of arguing, I prefer to find compromises. Therefore, I decided to combine the best of both worlds and started adding JSDoc. You can see what this looks like in this PR. I don't want to waste my time in vain, so I'm putting this up for public judgment. If the maintainers approve this PR, I will cover the remaining code with JSDoc.

### DIFF
--- a/src/core/session.js
+++ b/src/core/session.js
@@ -17,6 +17,10 @@ import { PageView } from "./drive/page_view"
 import { FrameElement } from "../elements/frame_element"
 import { Preloader } from "./drive/preloader"
 
+/**
+ * The Session class represents a user session, managing the various observers, navigators,
+ * and controllers to provide a seamless user experience in a Turbo Drive enabled application.
+ */
 export class Session {
   navigator = new Navigator(this)
   history = new History(this)
@@ -40,6 +44,9 @@ export class Session {
   started = false
   formMode = "on"
 
+  /**
+   * Starts all observers and services if not already started.
+   */
   start() {
     if (!this.started) {
       this.pageObserver.start()
@@ -57,10 +64,16 @@ export class Session {
     }
   }
 
+  /**
+   * Disables the session.
+   */
   disable() {
     this.enabled = false
   }
 
+  /**
+   * Stops all observers and services if they are running.
+   */
   stop() {
     if (this.started) {
       this.pageObserver.stop()
@@ -76,10 +89,22 @@ export class Session {
     }
   }
 
+  /**
+   * Registers a new browser adapter for the session.
+   *
+   * @param {BrowserAdapter} adapter - The new browser adapter to register.
+   */
   registerAdapter(adapter) {
     this.adapter = adapter
   }
 
+  /**
+   * Initiates a visit to a specified location, with optional parameters.
+   *
+   * @param {string} location - The location to visit.
+   * @param {Object} options - Optional parameters for the visit.
+   * @param {string} [options.frame] - The ID of the frame element to use for the visit.
+   */
   visit(location, options = {}) {
     const frameElement = options.frame ? document.getElementById(options.frame) : null
 
@@ -91,26 +116,54 @@ export class Session {
     }
   }
 
+  /**
+   * Connects a new stream source.
+   *
+   * @param {any} source - The new stream source to connect.
+   */
   connectStreamSource(source) {
     this.streamObserver.connectStreamSource(source)
   }
 
+  /**
+   * Disconnects an existing stream source.
+   *
+   * @param {any} source - The stream source to disconnect.
+   */
   disconnectStreamSource(source) {
     this.streamObserver.disconnectStreamSource(source)
   }
 
+  /**
+   * Renders a stream message.
+   *
+   * @param {Object} message - The stream message to render.
+   */
   renderStreamMessage(message) {
     this.streamMessageRenderer.render(StreamMessage.wrap(message))
   }
 
+  /**
+   * Clears the cache.
+   */
   clearCache() {
     this.view.clearSnapshotCache()
   }
 
+  /**
+   * Sets the delay for the progress bar.
+   *
+   * @param {number} delay - The delay in milliseconds.
+   */
   setProgressBarDelay(delay) {
     this.progressBarDelay = delay
   }
 
+  /**
+   * Sets the form mode.
+   *
+   * @param {string} mode - The form mode to set, either "on" or "off".
+   */
   setFormMode(mode) {
     this.formMode = mode
   }
@@ -361,6 +414,14 @@ export class Session {
 
   // Helpers
 
+  /**
+   * Helper function to check if a submission is navigatable based on the current form mode and element attributes.
+   *
+   * @private
+   * @param {HTMLFormElement} form - The form element being submitted.
+   * @param {HTMLElement} submitter - The element that triggered the form submission.
+   * @return {boolean} - True if the submission is navigatable, false otherwise.
+   */
   submissionIsNavigatable(form, submitter) {
     if (this.formMode == "off") {
       return false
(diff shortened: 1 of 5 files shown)


src/core/session.js
```diff
@@ -91,26 +116,54 @@ export class Session {
     }
   }
 
+  /**
+   * Connects a new stream source.
+   *
+   * @param {any} source - The new stream source to connect.
```
seanpdoyle, line 122: This is either [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) | [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket)

src/core/session.js
```diff
@@ -91,26 +116,54 @@ export class Session {
...
+   */
   connectStreamSource(source) {
     this.streamObserver.connectStreamSource(source)
   }
 
+  /**
+   * Disconnects an existing stream source.
+   *
+   * @param {any} source - The stream source to disconnect.
```
seanpdoyle, line 131: This is either [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) | [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket)

seanpdoyle:
Thank you for opening this. I think this is an improvement over what's currently available. 

I'm in favor of documentation of any kind, especially documentation that brings typed specificity.

I worry that the core team's disinterest in TypeScript might extend to a disinterest in typed code comments. Even the most correctly typed JSDoc comments risk becoming useless without the necessary tooling to enforce them.

Could this pull request also introduce tooling for CI and local development?

Does the core team have interest in incorporating that tooling into the project?

WagnerMoreira:
I'm highly interested in making contributions here and adding JSDoc over the whole codebase if this move is accepted. 

@seanpdoyle 
> Could this pull request also introduce tooling for CI and local development?

you mean things like using `eslint-plugin-jsdoc` and checking for the presence and correctness of JSDoc comments?

FroggyPanda:
I would also like to help out with reading types for this project.  However, I would like one of the main contributors, like @.dhh, to say if they're willing to add this PR so our time doesn't get wasted reading something that was removed.

src/core/session.js
```diff
@@ -76,10 +89,22 @@ export class Session {
...
+   */
   registerAdapter(adapter) {
     this.adapter = adapter
   }
 
+  /**
+   * Initiates a visit to a specified location, with optional parameters.
+   *
+   * @param {string} location - The location to visit.
+   * @param {Object} options - Optional parameters for the visit.
```
DASPRiD, line 105: Would it make sense to define the options properties? This can be done via [`@typedef`](https://jsdoc.app/tags-typedef.html).

src/core/session.js
```diff
@@ -91,26 +116,54 @@ export class Session {
...
+   * @param {number} delay - The delay in milliseconds.
+   */
   setProgressBarDelay(delay) {
     this.progressBarDelay = delay
   }
 
+  /**
+   * Sets the form mode.
+   *
+   * @param {string} mode - The form mode to set, either "on" or "off".
```
DASPRiD, line 165: This could be typed as `@param {('on'|'off')} mode`.

src/core/session.js
```diff
@@ -361,6 +414,14 @@ export class Session {
 
   // Helpers
 
+  /**
+   * Helper function to check if a submission is navigatable based on the current form mode and element attributes.
+   *
+   * @private
+   * @param {HTMLFormElement} form - The form element being submitted.
+   * @param {HTMLElement} submitter - The element that triggered the form submission.
```
DASPRiD, line 422: `submitter` is optional, it should therefore be typed as `@param {HTMLElement} [submitter]`.

DASPRiD:
This is a good idea. I did check out your branch and ran tsc to generate type definitions from this, which produces quite usable results for the files you touched:

```bash
npx -p typescript tsc src/**/*.js --declaration --allowJs --emitDeclarationOnly --outDir types
```

shashkovdanil:
Hey guys! So I'm waiting for answer from maintainers, if they're interested in it, I'll add infrastructure like eslint and continue writing documentation

WebReflection:
imho, "*this is the way*" (cit)

the benefits of JSDoc TS are many:

  * the `tsc` build step is not strictly necessary anymore, but it can always be used to `--check` from time to time via `npx` without having it as dependency
  * JSDoc itself encourage better documentation, not just types, and there are tools able to use that documentation to generate out of the box proper static pages anyone can read or crawl whenever it's needed
  * if focused only on the public API, the rest of the code will be still enjoyable (for the team) and maintenance would be kept minimal as public API signatures are all dependent projects need to have TS work, as [TS works already very well with JSDoc](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html)

last, but not least, I love the fact somebody proposed something instead of insulting people involved in this project (not everyone, but what I've read these days show how TS community can be way more toxic and hostile than anything happened in here).

I hope maintainers will give this PR a chance.

w3bdesign (approved):

afcapel:
@shashkovdanil thanks for the civil and constructive approach to the PR, I really appreciate it.

What we found that works for us to document Turbo is having a [documentation project](https://github.com/hotwired/turbo-site) that you can [access online](https://turbo.hotwired.dev/reference/drive).

Most of the public Turbo APIs are HTML annotations (data- attributes, mostly), custom elements and events. Having a documentation website allows us to document all of those together in a consistent way.

There are a few cases where you might want to call a Turbo public API directly from your code, like `Turbo.visit` or `Turbo.cache.clear`. For those cases we were thinking to ship some type annotations to [DefinitelyTyped](https://github.com/DefinitelyTyped), so people that like TS can enjoy their types.

shashkovdanil:
@afcapel Hi, JSDoc is not only good for users who will use the public API, but it is also good for contributors. If the documentation covers all this and JSDoc is not useful, I think you can close the PR

ArnavK-09 (approved):

FroggyPanda (approved):

Gravy59:
Hi, you might consider using [`ts-to-jsdoc`](https://github.com/futurGH/ts-to-jsdoc) or similar to expedite the process.

marcoroth:
Here's an attempt to migrate the whole codebase to JSDoc using the above mentioned `ts-to-jsdoc` package: https://github.com/marcoroth/turbo/commit/1ea3bfc3c061e693dffd9a597929479be254519d

The generated code is not 100% perfect in all cases, but I would be more than happy to polish this by hand if we decide to continue moving forward with this.

I think having the comments/documentation is valuable for contributors to see and understand how the internals work. This is also why it's not really a solution to just push up the types to DefinitelyTyped. It's not too valuable for users using Turbo and consuming the exported types the packages ships with, it's about the context and documentation we lost about the internals.

TechStudent10 (approved):
This is a great step forward. If the devs (looking at you DHH) really want to ditch TypeScript, I think JSDoc is a great in-between

shashkovdanil:
@afcapel hey, should I continue this?

ChristianECG:
image

afcapel:
@shashkovdanil I'm afraid we're not interested in bringing back the type system, even if it's in the form of comments.

But thanks again for the constructive approach to the discussion.

PR #988, opened 2023-09-08
Refactor ErrorRenderer class for better modularity and maintainability

AfolabiOlaoluwa:
This refactors the 'ErrorRenderer' Class based of few principles or opinion. They are as listed below:

1. **Global Variables**: 

    The refactor addressed the issue of global variable dependency on `document` by passing it as a parameter to the class constructor. This change makes the code more modular and testable. Additionally, `newSnapshot` was passed as a parameter for better encapsulation.

2. **Magic Numbers**: 

    The refactor also addresses problem of magic strings for import paths was tackled by defining them as constants. This change improves code maintainability as it provides descriptive names for these paths.

3. **Code Duplication**: 

    Consolidated code duplication in the `replaceHeadAndBody` method by combining two calls to `documentElement.replaceChild` into a single call, enhancing code readability and reducing redundancy.

4. **Static Methods**: 
 
    Removed the static `renderElement` method as it didn't seem to have a clear reason to be static, making the code more consistent.

5. **Method Size**: 

    Split the `render` method into two separate methods, `replaceHeadAndBody` and `activateScriptElements`, to improve code organization and readability.

6. **Hardcoded Queries**: 

    The hardcoded query for script elements has been abstracted into the `getScriptElements` method, making it more maintainable.


afcapel:
@AfolabiOlaoluwa thanks for the effort to improve the code base, but I'm not convinced these changes are in the right direction. Changing `ErrorRenderer` class signature will require changing all the client code. Also `ErrorRenderer` follows the same structure than `PageRenderer` and `FrameRenderer`, which will need updates too for consistency.

Depending on the global variable `document` is okay and widespread around the code base. Using globals, in moderation, is okay when the simplify the code. You can see [this video](https://www.youtube.com/watch?v=D7zUOtlpUPw) or [this article](https://dev.37signals.com/globals-callbacks-and-other-sacrileges/) for a longer discussion of the subject.

In this case the alternative is to initially fetch the `document` from the global, (because document is created by the browser and that's how client side code can access it) and then pass it around in every method call, which I don't think will improve readability.

And I don't think extracting the "script" string is worth it either. It adds indirection, forcing you to scroll back to to the top of the file to check what the constant means, when the original `querySelector` was already perfectly clear.

PR #993, opened 2023-09-09
fix: Enhance test stability in drive_tests.js

AfolabiOlaoluwa:
In the `drive_tests.js` file, I improved test stability by introducing steps in two test cases: "test drive to external link" and "test drive enabled by default; click link inside data-turbo='false'."

The added  calls ensure that the respective elements (e.g., `#drive_enabled_external` and `#drive_disabled`) are present and visible before interacting with them, reducing the chances of element unavailability issues during test execution.

This resolves test failures where elements were not immediately accessible, leading to test timeouts.

### DIFF
--- a/src/tests/functional/drive_tests.js
+++ b/src/tests/functional/drive_tests.js
@@ -9,7 +9,6 @@ test.beforeEach(async ({ page }) => {
 })
 
 test("test drive enabled by default; click normal link", async ({ page }) => {
-  await page.waitForSelector("#drive_enabled")
   await page.click("#drive_enabled")
   await nextBody(page)
   assert.equal(pathname(page.url()), path)
@@ -28,7 +27,7 @@ test("test drive to external link", async ({ page }) => {
 })
 
 test("test drive enabled by default; click link inside data-turbo='false'", async ({ page }) => {
-  await page.waitForSelector("#drive_disabled")
+  await page.click("#drive_disabled")
   await nextBody(page)
 
   assert.equal(pathname(page.url()), path)
--- a/src/tests/functional/drive_tests.js
+++ b/src/tests/functional/drive_tests.js
@@ -29,7 +29,7 @@ test("test drive to external link", async ({ page }) => {
 
 test("test drive enabled by default; click link inside data-turbo='false'", async ({ page }) => {
   await page.waitForSelector("#drive_disabled")
-  await page.click("#drive_disabled")
+  await nextBody(page)
 
   assert.equal(pathname(page.url()), path)
   assert.equal(await visitAction(page), "load")
--- a/src/tests/functional/drive_tests.js
+++ b/src/tests/functional/drive_tests.js
@@ -9,7 +9,8 @@ test.beforeEach(async ({ page }) => {
 })
 
 test("test drive enabled by default; click normal link", async ({ page }) => {
-  page.click("#drive_enabled")
+  await page.waitForSelector("#drive_enabled")
+  await page.click("#drive_enabled")
   await nextBody(page)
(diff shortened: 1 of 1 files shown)


israelalagbe (approved):

src/tests/functional/drive_tests.js
```diff
@@ -19,16 +20,17 @@ test("test drive to external link", async ({ page }) => {
...
+  await page.click("#drive_enabled_external")
   await nextBody(page)
 
   assert.equal(await page.evaluate(() => window.location.href), "https://example.com/")
   assert.equal(await page.textContent("body"), "Hello from the outside world")
 })
 
 test("test drive enabled by default; click link inside data-turbo='false'", async ({ page }) => {
-  page.click("#drive_disabled")
-  await nextBody(page)
```
seanpdoyle, line 31: What is the significance of removing the call to wait for the `nextBody`?

AfolabiOlaoluwa: The significance of the change is to simplify the code by leveraging Playwright's built-in waiting capabilities, making it more concise and still ensuring that the test waits for the expected state transitions before performing assertions. To buttress further:

1. `nextBody` waits for navigation: 
The `nextBody` function waits for the page to navigate or load a new content body. This is useful when you want to ensure that the page has fully transitioned to a new state before making assertions.

2. Replacing with await page.click: 
By using `await page.click("#drive_disabled")` instead, you are now explicitly waiting for the page to click the element with the id `drive_disabled`. Playwright's `page.click` function includes implicit waiting behavior. It will wait for the element to become clickable and for any resulting navigation. DOM changes to complete. In other words, it combines the click action with the necessary waiting for the page to update.

Having said all that, I am open to revert that line back to `await nextBody(page)` given unformnity with line 24.

AfolabiOlaoluwa: I have reverted it back to `nextBody` in another commit solely for uniformity in explicitness sake.

seanpdoyle: Thank you for the explanation. Unfortunately, Playwrights built-in synchronization when waiting for navigations isn't effective with Turbo, since Turbo intervenes in the page navigation process and issues fetch requests for content. There are no full page navigations for Playwright to listen for. That's why manual timing synchronization with helpers like nextBody are necessary.

afcapel:
@AfolabiOlaoluwa I don't think these changes are necessary. Playwright already automatically waits for an element to be available before performing an action like `page.click`. 

For more details, see https://playwright.dev/docs/actionability

AfolabiOlaoluwa:
> @AfolabiOlaoluwa I don't think these changes are necessary. Playwright already automatically waits for an element to be available before performing an action like `page.click`. 
> 
> 
> 
> For more details, see https://playwright.dev/docs/actionability 

But test cases aren't passing despite. Should we rather leave it hanging and not fixed?

afcapel:
@AfolabiOlaoluwa adding `await` to `page.click` is correct, what we don't need is the `page.waitForSelector` before `page.click`. If you can update the branch with those changes I'll merge 👍

AfolabiOlaoluwa:
> adding `await` to `page.click` is correct, what we don't need is the `page.waitForSelector` before `page.click`. If you can update the branch with those changes I'll merge

I think should have been the comments as a review of the PR rather than the PR being closed. I do think we need to be patient about closing PRs. 

I have made amends as I agree with your opinion clearly listed out on https://github.com/hotwired/turbo/pull/993#issuecomment-1715355438.

seanpdoyle (approved):

afcapel:
Thanks @AfolabiOlaoluwa 👍

PR #1000, opened 2023-09-12
[Deprecation]: Remove internal objects from public API

seanpdoyle:
The `FetchRequest`, `FetchResponse`, and `FormSubmission` objects are internal abstractions that facilitate Turbo's management of HTTP and Form Submission life cycles.

Currently, references to instances of those classes are available through `event.detail` for the following events:

* `turbo:frame-render`
* `turbo:before-fetch-request`
* `turbo:before-fetch-response`
* `turbo:submit-start`
* `turbo:submit-end`

Similarly, the `turbo:before-fetch-request` exposes a `fetchOptions` object that is a separate instance from the one used to submit the request. This means that **any** modifications to **any** value made from within an event listener are ineffective and **do not change** the ensuing request.

This commit deprecates those properties in favor of their built-in foundations, namely:

* [Request][]
* [Response][]

The properties that expose those instances will remain as deprecations, but will be inaccessible after an `8.0` release.

[Request]: https://developer.mozilla.org/en-US/docs/Web/API/Request
[Response]: https://developer.mozilla.org/en-US/docs/Web/API/Response

### DIFF
--- a/src/core/drive/form_submission.js
+++ b/src/core/drive/form_submission.js
@@ -116,64 +116,106 @@ export class FormSubmission {
 
   // Fetch request delegate
 
-  prepareRequest(request) {
-    if (!request.isSafe) {
+  prepareRequest(fetchRequest) {
+    if (!fetchRequest.isSafe) {
       const token = getCookieValue(getMetaContent("csrf-param")) || getMetaContent("csrf-token")
       if (token) {
-        request.headers["X-CSRF-Token"] = token
+        fetchRequest.headers.set("X-CSRF-Token", token)
       }
     }
 
-    if (this.requestAcceptsTurboStreamResponse(request)) {
-      request.acceptResponseType(StreamMessage.contentType)
+    if (this.requestAcceptsTurboStreamResponse(fetchRequest)) {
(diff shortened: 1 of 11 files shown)


src/core/drive/form_submission.js
```diff
@@ -116,64 +116,106 @@ export class FormSubmission {
...
     this.setSubmitsWith()
     dispatch("turbo:submit-start", {
       target: this.formElement,
-      detail: { formSubmission: this }
+      detail: {
+        request: fetchRequest.request,
+        submitter: this.submitter,
+
+        get formSubmission() {
+          console.warn("`event.detail.formSubmission` is deprecated. Use `event.target`, `event.detail.submitter`, and `event.detail.request` instead")
```
seanpdoyle, line 145: @afcapel @kevinmcconnell I realize now that the removal of TypeScript in anticipation of an 8.0 release means that a deprecation cycle as part of a minor release is no longer possible.

I'm curious of your thoughts here. Ultimately, we can remove the properties entirely and skip the `console.warn` calls.

shiftyp:
Congrats, you're the 1000th PR! You win a prize 💯 * 🔟 🥳

Issue #1005, opened 2023-09-14
Turbo Drive form submission doesn't render exceptional 4xx page.


gnattali:
I found that on Turbo Drive form submission, raised exception like RecordInvalid which results in unprocessable_entity status won't be rendered and only to be displayed in browser console.

This is caused by these two underlying situations.

1. On Turbo Drive form submission, there's no way to distinguish usual validation error from raised exception like RecordInvalid, since both result in unprocessable_entity.

2. Raised exception screen has different head tag contents from normal screen, and this prevents page renderer to render.

https://github.com/hotwired/turbo/blob/32cadc0abf9ea623cc5f1e5aa752091579ea8ea3/src/core/drive/page_renderer.js#L13-L15

I think this can be resolved by stop using PageView#renderPage in Navigator#formSubmissionFailedWithResponse and change it to something like PageView#renderFailedFormSubmission which detects raised exception from some clue like changed head contents.

https://github.com/hotwired/turbo/blob/32cadc0abf9ea623cc5f1e5aa752091579ea8ea3/src/core/drive/navigator.js#L92-L105

I'll create a PR with a failing test.

yokonao:
I would be very happy if this issue is resolved!

Unhandled errors that occur during Rails development are not necessarily Server Errors (5xx). In addition to RecordInvalid, as mentioned by the issue submitter, there are also ActiveRecord::RecordNotFuond and ActionController::ParameterMissing.

The development experience would be improved if an error page were displayed instead of a page reload in these cases.

It appears that the diff in https://github.com/hotwired/turbo/pull/1006 resolves the issue and takes compatibility into consideration.
I hope this issue is resolved!

PR #1006, opened 2023-09-14
add failing test about form submission responded with exceptional 422

gnattali:
https://github.com/hotwired/turbo/issues/1005


gnattali:
@afcapel 
Thank you for your comment and commit!
I think your solution forces a full reload of assets on normal validation error response which is not necessary.
How about avoiding it this way?
https://github.com/hotwired/turbo/pull/1006/commits/9d3d6bec7f8510386e637596eeede9b59cca1b6f
Any refactoring, correction, or even reverting is welcome!

PR #1007, opened 2023-09-14
Add history navigation direction to turbo:visit events

domchristie:
This pull request adds a `direction` detail in `turbo:visit` events. This enables differentiation between back and forward history navigations and is useful for customising animations.

A summary of the changes:

When a history entry is updated (pushState or replaceState):
  - increment the current index if it's a pushState
  - store the current index history's state (as `restorationIndex`)

When the history is popped (back or forward):
  - determine the direction from and current index and the popstate event state
  - create a Visit with that direction (and dispatch the `turbo:visit` event with the direction detail)
  - update the current index to the restored index

`direction` can be:
  - `forward` for advance or forward restoration visits
  - `back` for back restoration visits
  - `none` for back replace visits

### DIFF
--- a/src/core/drive/history.js
+++ b/src/core/drive/history.js
@@ -6,6 +6,7 @@ export class History {
   restorationData = {}
   started = false
   pageLoaded = false
+  currentIndex = 0
 
   constructor(delegate) {
     this.delegate = delegate
@@ -15,6 +16,7 @@ export class History {
     if (!this.started) {
       addEventListener("popstate", this.onPopState, false)
       addEventListener("load", this.onPageLoad, false)
+      this.currentIndex = history.state?.turbo?.restorationIndex || 0
       this.started = true
       this.replace(new URL(window.location.href))
     }
@@ -37,7 +39,9 @@ export class History {
   }
 
   update(method, location, restorationIdentifier = uuid()) {
-    const state = { turbo: { restorationIdentifier } }
+    if (method === history.pushState) ++this.currentIndex
+
+    const state = { turbo: { restorationIdentifier, restorationIndex: this.currentIndex } }
     method.call(history, state, "", location.href)
     this.location = location
     this.restorationIdentifier = restorationIdentifier
@@ -81,9 +85,11 @@ export class History {
       const { turbo } = event.state || {}
       if (turbo) {
         this.location = new URL(window.location.href)
-        const { restorationIdentifier } = turbo
+        const { restorationIdentifier, restorationIndex } = turbo
         this.restorationIdentifier = restorationIdentifier
-        this.delegate.historyPoppedToLocationWithRestorationIdentifier(this.location, restorationIdentifier)
+        const direction = restorationIndex > this.currentIndex ? "forward" : "back"
+        this.delegate.historyPoppedToLocationWithRestorationIdentifierAndDirection(this.location, restorationIdentifier, direction)
+        this.currentIndex = restorationIndex
       }
     }
   }
--- a/src/tests/functional/visit_tests.js
+++ b/src/tests/functional/visit_tests.js
@@ -220,6 +220,29 @@ test("test Visit with network error", async ({ page }) => {
   await nextEventNamed(page, "turbo:fetch-request-error")
 })
 
+test("test turbo:visit direction details", async ({ page }) => {
+  await page.click("#same-origin-link")
+  let details = await nextEventNamed(page, "turbo:visit")
+  assert.equal(details.direction, "forward")
+
+  await nextEventNamed(page, "turbo:load")
+  await page.goBack()
+  details = await nextEventNamed(page, "turbo:visit")
+  assert.equal(details.direction, "back")
+
+  await nextEventNamed(page, "turbo:load")
+  await page.goForward()
+  details = await nextEventNamed(page, "turbo:visit")
+  assert.equal(details.direction, "forward")
+
+  await nextEventNamed(page, "turbo:load")
+  await page.goBack()
+  await nextEventNamed(page, "turbo:load")
+  await page.click("#same-origin-replace-link")
+  details = await nextEventNamed(page, "turbo:visit")
+  assert.equal(details.direction, "none")
+})
+
 async function visitLocation(page, location) {
   return page.evaluate((location) => window.Turbo.visit(location), location)
 }
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -36,6 +36,12 @@ export const SystemStatusCode = {
   contentTypeMismatch: -2
 }
 
+export const Direction = {
+  advance: "forward",
+  restore: "back",
+  replace: "none"
+}
+
 export class Visit {
   identifier = uuid() // Required by turbo-ios
   timingMetrics = {}
@@ -65,7 +71,8 @@ export class Visit {
       willRender,
       updateHistory,
       shouldCacheSnapshot,
-      acceptsStreamResponse
+      acceptsStreamResponse,
+      direction
     } = {
       ...defaultOptions,
       ...options
@@ -83,6 +90,7 @@ export class Visit {
     this.scrolled = !willRender
     this.shouldCacheSnapshot = shouldCacheSnapshot
     this.acceptsStreamResponse = acceptsStreamResponse
+    this.direction = direction || Direction[action]
   }
 
   get adapter() {
--- a/src/core/session.js
+++ b/src/core/session.js
@@ -125,11 +125,12 @@ export class Session {
 
   // History delegate
 
-  historyPoppedToLocationWithRestorationIdentifier(location, restorationIdentifier) {
+  historyPoppedToLocationWithRestorationIdentifierAndDirection(location, restorationIdentifier, direction) {
     if (this.enabled) {
       this.navigator.startVisit(location, restorationIdentifier, {
         action: "restore",
-        historyChanged: true
+        historyChanged: true,
+        direction
       })
     } else {
       this.adapter.pageInvalidated({
@@ -188,7 +189,7 @@ export class Session {
     }
     extendURLWithDeprecatedProperties(visit.location)
     if (!visit.silent) {
-      this.notifyApplicationAfterVisitingLocation(visit.location, visit.action)
+      this.notifyApplicationAfterVisitingLocation(visit.location, visit.action, visit.direction)
     }
   }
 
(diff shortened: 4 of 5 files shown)


src/core/drive/history.js
```diff
@@ -15,6 +16,7 @@ export class History {
     if (!this.started) {
       addEventListener("popstate", this.onPopState, false)
       addEventListener("load", this.onPageLoad, false)
+      this.currentIndex = history.state?.turbo?.restorationIndex || 0
```
domchristie, line 19: I wasn't aware of this until today, but browsers persist the state between refreshes (🤯). This means this feature should still work after a reload (e.g. after assets are invalidated and a user taps Back)

domchristie:
Also /ht @matthewp for the [inspiration](https://github.com/withastro/astro/blob/fca6892f8d6a30ceb1e04213be2414dd4cb4d389/packages/astro/components/ViewTransitions.astro)

afcapel:
@domchristie nice change 🙌

I'm wondering if we could also add an attribute to the body, similar to data-turbo-preview, with the navigation direction. That way we could customize animations with just css, no js or event handling required.

domchristie:
> I'm wondering if we could also add an attribute to the body

@afcapel good idea. Added in https://github.com/hotwired/turbo/pull/1007/commits/c39c2236d186cf7fc7f90f55b70155137975ca54. It's difficult to test thoroughly because the attributes get added and removed quicker than it's possible to check them (unless we pause rendering :/)

domchristie:
With this in place, part of me wonders if the event detail is necessary, and whether we just use the data attribute?

domchristie:
> With this in place, part of me wonders if the event detail is necessary, and whether we just use the data attribute?

For consistency, I'd also suggest removing the isPreview detail too, as I think this also can be inferred from a data attribute

afcapel:
> With this in place, part of me wonders if the event detail is necessary, and whether we just use the data attribute? This might feel more Hotwire-y: where state lives at attributes in the DOM

Yes, I think that makes sense. If you can make the change, that'd be great.

I still need to spin the branch for a test, but the changes look 👌

domchristie:
> Yes, I think that makes sense. If you can make the change, that'd be great.

👍 @afcapel Done in https://github.com/hotwired/turbo/pull/1007/commits/badfd694c19614ea47a108503b155ed85d2c0685

> I still need to spin the branch for a test, but the changes look 👌

Demo app with build from this PR and directional View Transitions: https://github.com/domchristie/turbo_view_transition_test

https://github.com/hotwired/turbo/assets/111734/cad63e8f-e7b8-4353-9236-8954d5f85fba

domchristie:
(I should add, the test are flaky, but seem to pass, at least after a couple of automated retries)

domchristie:
@afcapel friendly ping, I think this is ready

src/tests/functional/visit_tests.js
```diff
@@ -220,6 +221,55 @@ test("test Visit with network error", async ({ page }) => {
   await nextEventNamed(page, "turbo:fetch-request-error")
 })
 
+test("test Visit direction data attribute when clicking a link", async ({ page }) => {
```
afcapel, line 224: The `test("test"` prefix was an leftover from having moved the tests to Playwright, but we've removed it from main in 0613595a0c3d42afc3f5fecca8eec413850f464b. We should remove them here too. 

We can use something with an active verb, like `test("sets direction data attribute when clicking a link")`.

src/tests/functional/visit_tests.js
```diff
@@ -220,6 +221,55 @@ test("test Visit with network error", async ({ page }) => {
...
+  await Promise.all([
+    waitUntilSelector(page, "[data-turbo-visit-direction='forward']")
+      .then(() => waitUntilSelector(page, "html:not([data-turbo-visit-direction])")),
+    page.click("#same-origin-link")
+  ])
```
afcapel, line 229: I think we can write these tests without relying on `Promise.all`:

```js
  page.click("#same-origin-link")

  await waitUntilSelector(page, "[data-turbo-visit-direction='forward']")
  await waitUntilNoSelector(page, "[data-turbo-visit-direction]")
```

src/tests/functional/visit_tests.js
```diff
@@ -220,6 +221,55 @@ test("test Visit with network error", async ({ page }) => {
...
+})
+
+test("test Visit direction detdata attribute when navigating back", async ({ page }) => {
```
afcapel, line 232: Typo: `detdata`.

src/core/drive/history.js
afcapel: TIL, fascinating.

afcapel (commented):
@domchristie looks great, just a few suggestions about the tests.

src/tests/functional/visit_tests.js
domchristie:
```suggestion
test("Visit direction data attribute when clicking a link", async ({ page }) => {
```

src/tests/functional/visit_tests.js
domchristie:
```suggestion
test("Visit direction data attribute when navigating back", async ({ page }) => {
```

src/tests/functional/visit_tests.js
```diff
@@ -220,6 +221,55 @@ test("test Visit with network error", async ({ page }) => {
...
+  await page.goBack()
+  await nextEventNamed(page, "turbo:load")
+  await Promise.all([
+    waitUntilSelector(page, "[data-turbo-visit-direction='forward']")
+      .then(() => waitUntilSelector(page, "html:not([data-turbo-visit-direction])")),
+    page.goForward()
+  ])
+})
+
+test("test Visit direction attribute on a replace visit", async ({ page }) => {
```
domchristie, line 254:
```suggestion
test("Visit direction attribute on a replace visit", async ({ page }) => {
```

src/tests/functional/visit_tests.js
```diff
@@ -220,6 +221,55 @@ test("test Visit with network error", async ({ page }) => {
...
+  await Promise.all([
+    waitUntilSelector(page, "[data-turbo-visit-direction='none']")
+      .then(() => waitUntilSelector(page, "html:not([data-turbo-visit-direction])")),
+    page.click("#same-origin-replace-link")
+  ])
+})
+
+test("test Turbo history state after a reload", async ({ page }) => {
```
domchristie, line 262:
```suggestion
test("Turbo history state after a reload", async ({ page }) => {
```

src/tests/functional/visit_tests.js
domchristie: I think I tried this and I don't this it worked  because the attribute is added and removed quicker than the assertions can track.

If I recall correctly, in the time it takes for `page.click` to resolve, the attribute it already added and removed, so `waitUntilSelector(page, "[data-turbo-visit-direction='forward']")` never passes

domchristie:
@afcapel "test" prefixes and typo fixed, and left a comment about the reason for `Promise.all`

src/tests/functional/visit_tests.js
domchristie: also line 226-227 ensures the attribute is added and removed in the correct order

afcapel: @domchristie what do you think of a660c7bf644daf2f29abfdc146eec2367698cf10?

The trick is that you don't `await` for the `page.click`, so it runs in parallel with the first `waitUntilSelector`. Then we `await` for the two assertions to ensure they happen one after the other.

afcapel: Huh, that works locally but seems very flaky on CI.

afcapel: I think our best option is to use the mutation logs so we are not so dependent on the exact assertion timing. With the mutation log even if the attribute is no longer present in the page, we can still assert that it was present at some point.

See https://github.com/hotwired/turbo/compare/2bc7c13c66b2a31c4c46c63819d4f7f052538e6f..5d4758969663e4633cf5c61eb9b2bb23cf0a77a9

brunoprietog:
> With this in place, part of me wonders if the event detail is necessary, and whether we just use the data attribute?

> For consistency, I'd also suggest removing the isPreview detail too, as I think this also can be inferred from a data attribute (#926 (comment))

Why remove them? Can't both coexist? Even HEY have a helper to know if it is turboPreview. If it was necessary to make the helper, wouldn't it be better to read the attribute directly from the event as well?

<https://[app.hey.com/assets/helpers/turbo_helpers-5496333b991567941c18f8dd0485c0d0703e9c67.js>

src/tests/functional/visit_tests.js
domchristie: > I think our best option is to use the mutation logs

This looks good to me 👍

domchristie:
> Why remove them? Can't both coexist?

@brunoprietog for me, it's about what feels right for the API. Turbo's event's details are pretty minimal, and as such, `isPreview` feels out-of-place (particularly since Hotwire encourages state to be stored in the DOM). If event details were to include comprehensive Visit information, then I could see a place for it.

I think there's a good argument for more information in event details, but until that shape is decided upon, it feels preferable to stick with the data attribute approach only. Otherwise, if we decide on a different event API, we'd be introducing a breaking change.

src/tests/functional/visit_tests.js
afcapel: Cool, @domchristie if you can push those changes to your branch I'll merge.

domchristie: @afcapel I will do (I’ve just moved house, and will have a reliable internet connection next week)

domchristie: > if you can push those changes to your branch I'll merge

@afcapel Done

afcapel:
Thanks @domchristie!

brunoprietog:
With this argument, it should #926 be reverted?

I don't think there is much consistency. Some things are reported via HTML attributes and other things in the event detail. The most recent example is the renderMethod introduced in #1019.

https://github.com/hotwired/turbo/blob/ac0035982e2f8a6a72055acc954d813330afa771/src/core/session.js#L348-L350

Other events that are emitted also provide quite a bit of information in the detail.

As developers, isn't it more ergonomic to receive the information in the detail? Rather than having to make custom helpers to read this information from the HTML.

I feel that exposing the information in the HTML is perfect for the CSS but not ergonomic for the event listeners. What feels wrong with exposing the information both ways?

If something is still undecided, I think it would be good to document the approach and make the API consistent. What euristics will be used to define whether to expose via HTML attributes or via the event detail?

PR #1012, opened 2023-09-22
Added frozen lockfile option on ci and dev workflows

shiftyp:
Added frozen lockfile option on ci and dev yarn install to fail if the lockfile is out of date


shiftyp:
Prevents review work like https://github.com/hotwired/turbo/pull/1010#issuecomment-1727675122

afcapel:
Nice, thanks @shiftyp 🙏

PR #1014, opened 2023-09-27
Fix back navigation after POST form submission

afcapel:
This commit fixes a bug where navigating back after submitting a POST form would not show the previous page. The bug was introduced in https://github.com/hotwired/turbo/pull/949. The new cache API is async since it needs to access `CacheStorage` which is async. In that PR, `getCachedSnapshot` was changed to be async, but `hasCachedSnapshot` was not changed. This meant that `getCachedSnapshot` would always return a promise, which is truthy, and so `hasCachedSnapshot` which just checked that the return value was not null would always return true.

The bug would show up when you tried to navigate back to a page after a post form submission. The form submission would clear the cache, but the restoration visit would think it had a cached snapshot and so would not issue a request. This meant that the page would not be restored and nothing would happen.

There's a new test in navigation_tests.js that reproduces this bug.

The fix is to make `hasCachedSnapshot` async and await it in the places where it is used.

### DIFF
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -154,10 +154,10 @@ export class Visit {
     }
   }
 
-  issueRequest() {
+  async issueRequest() {
     if (this.hasPreloadedResponse()) {
       this.simulateRequest()
-    } else if (this.shouldIssueRequest() && !this.request) {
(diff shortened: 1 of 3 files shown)


kevinmcconnell (approved):
Looks great! Nice catch 👏

Issue #951, opened 2023-07-21
Custom rendering is incompatible with caching


palkan:
**Turbo version**: 7.3.0

I tried using Turbo Drive with morphdom and encountered an issue with restoration visits ("Back" button).

Given the basic morphdom setup as per [handbook](https://turbo.hotwired.dev/handbook/drive#custom-rendering):

```js
import morphdom from "morphdom"

addEventListener("turbo:before-render", (event) => {
  event.detail.render = (currentElement, newElement) => {
    morphdom(currentElement, newElement)
  }
})
```

We see that going back always shows an incorrect page (it's lagging one page behind). I did some debugging and nailed down the problem—we call `snapshot.clone()` on the _next tick_, and it happens that during the next tick the `document.body` has been already morphed but still refers to the same DOM element as the one that is connected to the `snapshot` object; hence, calling `clone()` actually clones the new, updated DOM, not the previous version (as it should):

https://github.com/hotwired/turbo/blob/da647a63950c6f834fb0fbcf9315bee2be408cbe/src/core/drive/page_view.ts#L50

This makes using morphdom (or any other custom rendering mechanism that preserves `document.body`) incompatible with caching.

The workaround I found is to add the same _next tick_ hack to the `render` function:

```js
addEventListener("turbo:before-render", (event) => {
  event.detail.render = async (currentElement, newElement) => {
    await new Promise((resolve) => setTimeout(() => resolve(), 0));

    morphdom(currentElement, newElement)
  }
})
```

That seems to work but doesn't look like a robust solution.

s3ththompson:
I ran into the same issue and added the same (temporary) fix. If this fix is deterministic, it would be nice if it was handled internally. If it's not deterministic, then it seems like the `turbo:before-render` hook itself needs to be rethought since this behavior is definitely expected...

I also ran into a related issue: usually replacing DOM elements in the render function disconnects and reconnects Stimulus controllers (since the intersection observer registers the old and new DOM nodes changing). I had some Stimulus logic that relied on controllers reconnecting after every render. Once I added `morphdom`, I had to manually manage this lifecycle since the DOM nodes themselves don't change. This isn't necessarily something that can be handled internally by Turbo, but at the very least, it seems like it might be worth calling out in the documentation.

This is what my `render` function ended up looking like:

```js
const application = Application.start();

async function nextEventLoopTick() {
  return new Promise((resolve) => setTimeout(() => resolve(), 0));
}

[...]

Issue #37, opened 2020-12-25
Option to maintain current scroll position?


jaredcwhite:
On a regular Drive page update, it's scrolling up to the top of the page. Normally that would be desired, but on some Drive links I want to maintain the current scroll position. I tried looking for a `data-turbo-preserve-scroll` option or something like that but couldn't find anything. I could use Frames in this scenario instead, but then I'd lose the URL history/back button/etc.

seanpdoyle:
Would a pair of `document`- or `window`-level suit your needs?

```js
let scrollTop = 0

addEventListener("turbo:click", ({ target }) => {
  if (target.hasAttribute("data-turbo-preserve-scroll")) {
    scrollTop = document.scrollingElement.scrollTop
  }
})

addEventListener("turbo:load", () => {
  if (scrollTop) {
    document.scrollingElement.scrollTo(0, scrollTop)
  }

  scrollTop = 0
})
```

jaredcwhite:
@seanpdoyle That works in a pinch! 👏

jmarsh24:
Could you elaborate on how to implement this? I'm also looking for a similar solution, but I'm not exactly sure how to make it work. 🤔

bvsatyaram:
We have a slightly different requirement. Instead of one HTML element, we have several elements for which we wanted to retain the scroll position. It is a Kanban board with multiple columns. The following worked for us:

```
let containerScrollTops = {};

addEventListener("turbo:click", () => {
  document
    .querySelectorAll("[data-turbo-preserve-scroll-container]")
    .forEach((ele) => {
      containerScrollTops[ele.dataset.turboPreserveScrollContainer] =
        ele.scrollTop;
    });
});

addEventListener("turbo:load", () => {
  document
    .querySelectorAll("[data-turbo-preserve-scroll-container]")
    .forEach((ele) => {
      const containerScrollTop =
        containerScrollTops[ele.dataset.turboPreserveScrollContainer];
      if (containerScrollTop) ele.scrollTo(0, containerScrollTop);
    });

  containerScrollTops = {};
});
```

We are essentially setting a unique value for `data-turbo-preserve-scroll-container` for each such columns for which we would like to retain the scroll position.

michalsznajder:
Is there a solution to keep scroll position during form submission?

tommybickerdike:
@michalsznajder this worked for me:

```
let scrollTop = 0

addEventListener("turbo:submit-end", (event) => {
[...]

Issue #920, opened 2023-05-04
Replace call to requestAnimationFrame in render function to update page content even if tab has lost focus


jadetr:
Here is the case we are facing:
- we use Turbo.visit method to fetch a page's content that hold a TurboStream socket connection
- If we switch the browser tab (Tab with Rails app loses focus) before the whole page is loaded. The JS is blocked in Turbo's render method at line: this.frame = requestAnimationFrame(() => resolve())
- The requestAnimationFrame doesn't fire callback when the tab has lost focus
- The result is that the new page'S content is never rendered and the TurboStream socket is not connected.
- Few minutes later, we go back into our app browser Tab, the requestAnimationFrame fire, the page is rendered and the TurboStream socket is connected.

Problem: 
- When we go back into the tab, The page show outdated HTML content that should have been updated by the TurboStream socket.

afcapel:
@jadetr is this happening in a particular browser, or is it a common implementation? The spec [says](https://www.w3.org/TR/animation-timing/#processingmodel) that browser should only clear the callback list if the document is visible and the callback has been processed.

michelson:
@afcapel we are facing this issue on a chat application please see the video examples at https://github.com/hotwired/turbo-rails/issues/338

michelson:
Indeed we have replaced ALL nextAnimationFrame and it works, not sure which one did the trick tho. I've added a screen-recording with the working solution at https://github.com/hotwired/turbo-rails/issues/338

jadetr:
@afcapel sorry for now replying before. I didn't test it in other browser and I had to implement a workaround when I've hit this bug few months ago. In other terms, I cannot easily test it on other browser now. @michelson seems to have a working solution, great!

brunoprietog:
I guess we can close this since #1042 was merged

afcapel:
Fixed in https://github.com/hotwired/turbo/pull/1042

PR #1019, opened 2023-10-05
Page refreshes

afcapel:
This PR introduces the concept of *page refresh*. A page refresh happens when Turbo renders the current page again. We will offer two new options to control behavior when a page refresh happens:

- The method used to update the page: with a new option to use morphing (Turbo currently replaces the body).

- The scroll strategy: with a new option to keep it (Turbo currently resets scroll to the top-left).

The combination of morphing and scroll-keeping results in smoother updates that keep the screen state. For example, this will keep both horizontal and vertical scroll, the focus, the text selection, CSS transition states, etc.

We will also introduce a new turbo stream action that, when broadcasted, will request a page refresh. This will offer a simplified alternative to fine-grained broadcasted turbo-stream actions.

https://github.com/basecamp/turbo-private/assets/150107/a2ded86c-e68e-44ab-b07a-78fd7d20c6b2

By @afcapel and @jorgemanrubia.

### Page refresh configuration

Turbo will detect page refreshes automatically when it renders the current location again. The user configures how those page refreshes are handled.

#### Configurable refresh method

```html
<meta name="turbo-refresh-method" content="replace|morph">
```

* `replace`: Current behavior: it replaces the body.
* `morph`: It morphs the existing body to become the new body.

The default is `replace`. We may switch to `morph` eventually.

#### Configurable scroll behavior:

```html
<meta name="turbo-refresh-scroll" content="reset|preserve">
```

* `reset`: Current behavior: it resets the scroll to the top-left.
* `preserve`: It keeps the scroll in place.

The default is `reset`. We may switch to `preserve` eventually.

#### Support via helpers

==See companion PR in `turbo-rails`==.

```ruby
turbo_refreshes_with scroll method: :morph, scroll: :preserve
```

#### Exclude sections from morphing

There are scenarios where you might want to define sections that you want to ignore while morphing. The most common scenario is a popover or similar UI elements, which state you want to preserve when the page refreshes. We reuse the existing `data-turbo-permanent` attribute to flag such elements.

```html
<div data-turbo-permanent>...</div>
```

#### Reload turbo frames on page refreshes

Refreshing the current page presents a challenge if the page has loaded additional content dynamically. For example, if you have loaded new pages of data. You don’t want to lose that data when a page refresh happens. If the user can act on that data, you want to see it refreshed too.

To deal with this, it will reload turbo-frames with a `src` attribute that are flagged with `[refresh=reload]`. And it will use the configured page refresh method to reload those. 

```html
<turbo-frame refresh="reload">...</turbo-frame>
```

#### Turbo stream action to signal a refresh

This introduces a new turbo stream action called `refresh` that will trigger a page refresh:

```html
<turbo-stream action="refresh"></turbo-stream>
```

#### Broadcast changes that require a page refresh

There is a new method in active record's models to broadcast a *page refresh signal* when the record changes. This is implemented internally by broadcasting the turbo stream action referred to above.

```ruby
module Recording::Broadcasting
  extend ActiveSupport::Concern

  included do
    broadcasts_refreshes_to :bucket
  end
end
```

In view templates, you would subscribe to these updates using the regular turbo_stream_from helper:

```ruby
turbo_stream_from bucket
```

The system will debounce automatically sequences of updates since it doesn't make sense to trigger multiple signals for multiple updates in a row. 

⚠️ ⚠️ ==You can learn more about the Rails helpers in the companion PR in `turbo-rails`==.

### Idiomorph

We started with [`morphdom`](https://github.com/patrick-steele-idem/morphdom) as the DOM-tree morphing library but we eventually changed to [`idiomorph`](https://github.com/bigskysoftware/idiomorph). The reason is that we found `morphdom` to be very picky about `ids` when matching nodes, when `idiomorph` is way more relaxed. In practice, we found that with `morphdom` you had to keep adding `ids` everywhere if you wanted morphing to work, while `idiomorph` worked out of the box with pretty complex HTML structures.

### Mechanism against bounced signals

The system includes a mechanism to ignore refresh actions that originated in a web request that originated in the same browser session. To enable this, these refresh actions can carry a `request-id` attribute to track the request that originated them. Turbo will keep a list of recent requests and, when there is a match, the stream action is discarded.

This is implemented by patching `fetch` to inject a random request id in a custom header `X-Turbo-Request-Id`. We originally intended to use the Rails' standard `X-Request-Id` in the response, but we found a blocking problem: it's not possible with Javascript to read the response headers in a redirect response, you can just read the headers in the target destination response.

The server side where the broadcasts originate is responsible for using that header to flag broadcasted page refreshes with it (==see `turbo-rails` PR with the reference implementation == ).

## Pending

- [ ] Document in guides.
- [ ] Complete code documentation.

### DIFF
--- a/src/core/drive/morph_renderer.js
+++ b/src/core/drive/morph_renderer.js
@@ -1,0 +1,88 @@
+import Idiomorph from "idiomorph"
+import { nextAnimationFrame } from "../../util"
+import { Renderer } from "../renderer"
+
+export class MorphRenderer extends Renderer {
+  async render() {
+    if (this.willRender) await this.#morphBody()
+  }
+
+  get renderMethod() {
+    return "morph"
+  }
+
+  // Private
+
+  async #morphBody() {
+    this.#morphElements(this.currentElement, this.newElement)
+    this.#reloadRemoteFrames()
+
+    this.#dispatchEvent("turbo:morph", { currentElement: this.currentElement, newElement: this.newElement })
+  }
+
+  #morphElements(currentElement, newElement, morphStyle = "outerHTML") {
+    Idiomorph.morph(currentElement, newElement, {
+      morphStyle: morphStyle,
+      callbacks: {
+        beforeNodeMorphed: this.#shouldMorphElement,
+        beforeNodeRemoved: this.#shouldRemoveElement,
+        afterNodeMorphed: this.#reloadStimulusControllers
+      }
+    })
+  }
+
+  #reloadRemoteFrames() {
+    this.#remoteFrames().forEach((frame) => {
+      if (this.#isFrameReloadedWithMorph(frame)) {
+        this.#renderFrameWithMorph(frame)
+      }
+      frame.reload()
+    })
+  }
+
+  #renderFrameWithMorph(frame) {
+    frame.addEventListener("turbo:before-frame-render", (event) => {
+      event.detail.render = this.#morphFrameUpdate
+    }, { once: true })
+  }
+
+  #morphFrameUpdate = (currentElement, newElement) => {
+    this.#dispatchEvent("turbo:before-frame-morph", { currentElement, newElement }, currentElement)
+    this.#morphElements(currentElement, newElement, "innerHTML")
+  }
+
+  #shouldRemoveElement = (node) => {
+    return this.#shouldMorphElement(node)
+  }
+
+  #shouldMorphElement = (node) => {
+    if (node instanceof HTMLElement) {
+      return !node.hasAttribute("data-turbo-permanent")
+    } else {
+      return true
+    }
+  }
+
+  #reloadStimulusControllers = async (node) => {
+    if (node instanceof HTMLElement && node.hasAttribute("data-controller")) {
+      const originalAttribute = node.getAttribute("data-controller")
+      node.removeAttribute("data-controller")
+      await nextAnimationFrame()
+      node.setAttribute("data-controller", originalAttribute)
+    }
+  }
+
+  #isFrameReloadedWithMorph(element) {
+    return element.getAttribute("src") && element.getAttribute("refresh") === "morph"
+  }
+
+  #remoteFrames() {
+    return document.querySelectorAll("turbo-frame[src]")
+  }
+
+  #dispatchEvent(name, detail, target = document.documentElement) {
+    const event = new CustomEvent(name, { bubbles: true, cancelable: true, detail })
+    target.dispatchEvent(event)
+    return event
+  }
+}
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -1,6 +1,7 @@
 import { nextEventLoopTick } from "../../util"
 import { View } from "../view"
 import { ErrorRenderer } from "./error_renderer"
+import { MorphRenderer } from "./morph_renderer"
 import { PageRenderer } from "./page_renderer"
 import { PageSnapshot } from "./page_snapshot"
 import { SnapshotCache } from "./snapshot_cache"
@@ -15,7 +16,10 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const renderer = new PageRenderer(this.snapshot, snapshot, PageRenderer.renderElement, isPreview, willRender)
+    const shouldMorphPage = this.isPageRefresh(visit) && this.snapshot.shouldMorphPage
+    const rendererClass = shouldMorphPage ? MorphRenderer : PageRenderer
+
+    const renderer = new rendererClass(this.snapshot, snapshot, PageRenderer.renderElement, isPreview, willRender)
 
     if (!renderer.shouldRender) {
       this.forceReloaded = true
@@ -53,6 +57,10 @@ export class PageView extends View {
 
   getCachedSnapshotForLocation(location) {
     return this.snapshotCache.get(location)
+  }
+
+  isPageRefresh(visit) {
+    return visit && this.lastRenderedLocation.href === visit.location.href
   }
 
   get snapshot() {
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -154,10 +154,10 @@ export class Visit {
     }
   }
 
-  issueRequest() {
+  async issueRequest() {
     if (this.hasPreloadedResponse()) {
       this.simulateRequest()
-    } else if (this.shouldIssueRequest() && !this.request) {
+    } else if (!this.request && await this.shouldIssueRequest()) {
       this.request = new FetchRequest(this, FetchMethod.get, this.location)
       this.request.perform()
     }
@@ -231,14 +231,14 @@ export class Visit {
     }
   }
 
-  hasCachedSnapshot() {
-    return this.getCachedSnapshot() != null
+  async hasCachedSnapshot() {
+    return (await this.getCachedSnapshot()) != null
   }
 
   async loadCachedSnapshot() {
     const snapshot = await this.getCachedSnapshot()
     if (snapshot) {
-      const isPreview = this.shouldIssueRequest()
+      const isPreview = await this.shouldIssueRequest()
       this.render(async () => {
         this.cacheSnapshot()
         if (this.isSamePage) {
@@ -335,7 +335,7 @@ export class Visit {
   // Scrolling
 
   performScroll() {
-    if (!this.scrolled && !this.view.forceReloaded) {
+    if (!this.scrolled && !this.view.forceReloaded && !this.view.snapshot.shouldPreserveScrollPosition) {
       if (this.action == "restore") {
         this.scrollToRestoredPosition() || this.scrollToAnchor() || this.view.scrollToTop()
       } else {
@@ -391,11 +391,11 @@ export class Visit {
     return typeof this.response == "object"
   }
 
-  shouldIssueRequest() {
+  async shouldIssueRequest() {
     if (this.isSamePage) {
       return false
-    } else if (this.action == "restore") {
-      return !this.hasCachedSnapshot()
+    } else if (this.action === "restore") {
+      return !(await this.hasCachedSnapshot())
     } else {
       return this.willRender
     }
--- a/src/core/streams/stream_actions.js
+++ b/src/core/streams/stream_actions.js
@@ -30,5 +30,14 @@ export const StreamActions = {
       targetElement.innerHTML = ""
       targetElement.append(this.templateContent)
     })
+  },
+
+  refresh() {
+    const requestId = this.getAttribute("request-id")
+    const isRecentRequest = requestId && window.Turbo.recentRequests.has(requestId)
+    if (!isRecentRequest) {
+      window.Turbo.cache.exemptPageFromPreview()
+      window.Turbo.visit(window.location.href, { action: "replace" })
+    }
   }
 }
--- a/src/http/recent_fetch_requests.js
+++ b/src/http/recent_fetch_requests.js
@@ -1,0 +1,15 @@
+import { uuid } from "../util"
+
+const originalFetch = window.fetch
+
+window.fetch = async function(url, options = {}) {
+  const modifiedHeaders = new Headers(options.headers || {})
+  const requestUID = uuid()
+  window.Turbo.recentRequests.add(requestUID)
+  modifiedHeaders.append("X-Turbo-Request-Id", requestUID)
+
+  return originalFetch(url, {
+    ...options,
+    headers: modifiedHeaders
+  })
+}
--- a/package.json
+++ b/package.json
@@ -32,6 +32,9 @@
   "publishConfig": {
     "access": "public"
   },
+  "dependencies": {
+    "idiomorph": "https://github.com/basecamp/idiomorph#rollout-build"
+  },
   "devDependencies": {
     "@open-wc/testing": "^3.1.7",
     "@playwright/test": "^1.28.0",
--- a/src/core/drive/limited_set.js
+++ b/src/core/drive/limited_set.js
@@ -1,0 +1,15 @@
+export class LimitedSet extends Set {
+  constructor(maxSize) {
+    super()
+    this.maxSize = maxSize
+  }
+
+  add(value) {
+    if (this.size >= this.maxSize) {
+      const iterator = this.values()
+      const oldestValue = iterator.next().value
+      this.delete(oldestValue)
+    }
+    super.add(value)
+  }
+}
--- a/src/core/drive/page_snapshot.js
+++ b/src/core/drive/page_snapshot.js
@@ -73,6 +73,14 @@ export class PageSnapshot extends Snapshot {
     return this.headSnapshot.getMetaValue("view-transition") === "same-origin"
   }
 
+  get shouldMorphPage() {
+    return this.getSetting("refresh-method") === "morph"
+  }
+
+  get shouldPreserveScrollPosition() {
+    return this.getSetting("refresh-scroll") === "preserve"
+  }
+
   // Private
 
   getSetting(name) {
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -276,7 +276,7 @@ export class FrameController {
     return !defaultPrevented
   }
 
-  viewRenderedSnapshot(_snapshot, _isPreview) {}
+  viewRenderedSnapshot(_snapshot, _isPreview, _renderMethod) {}
 
   preloadOnLoadLinksForView(element) {
     session.preloadOnLoadLinksForView(element)
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -4,11 +4,13 @@ import { PageRenderer } from "./drive/page_renderer"
 import { PageSnapshot } from "./drive/page_snapshot"
 import { FrameRenderer } from "./frames/frame_renderer"
 import { FormSubmission } from "./drive/form_submission"
+import { LimitedSet } from "./drive/limited_set"
 
 const session = new Session()
 const cache = new Cache(session)
+const recentRequests = new LimitedSet(20)
 const { navigator } = session
-export { navigator, session, cache, PageRenderer, PageSnapshot, FrameRenderer }
+export { navigator, session, cache, recentRequests, PageRenderer, PageSnapshot, FrameRenderer }
 
 export { StreamActions } from "./streams/stream_actions"
 
--- a/src/core/native/browser_adapter.js
+++ b/src/core/native/browser_adapter.js
@@ -22,11 +22,7 @@ export class BrowserAdapter {
 
   visitRequestStarted(visit) {
     this.progressBar.setValue(0)
-    if (visit.hasCachedSnapshot() || visit.action != "restore") {
-      this.showVisitProgressBarAfterDelay()
-    } else {
-      this.showProgressBar()
-    }
+    this.showVisitProgressBarAfterDelay()
   }
 
   visitRequestCompleted(visit) {
--- a/src/core/renderer.js
+++ b/src/core/renderer.js
@@ -79,4 +79,8 @@ export class Renderer {
   get permanentElementMap() {
     return this.currentSnapshot.getPermanentElementMapForSnapshot(this.newSnapshot)
   }
+
+  get renderMethod() {
+    return "replace"
+  }
 }
--- a/src/core/session.js
+++ b/src/core/session.js
@@ -263,9 +263,9 @@ export class Session {
     return !defaultPrevented
   }
 
-  viewRenderedSnapshot(_snapshot, isPreview) {
+  viewRenderedSnapshot(_snapshot, isPreview, renderMethod) {
     this.view.lastRenderedLocation = this.history.location
-    this.notifyApplicationAfterRender(isPreview)
+    this.notifyApplicationAfterRender(isPreview, renderMethod)
   }
 
   preloadOnLoadLinksForView(element) {
@@ -328,8 +328,8 @@ export class Session {
     })
   }
 
-  notifyApplicationAfterRender(isPreview) {
-    return dispatch("turbo:render", { detail: { isPreview } })
+  notifyApplicationAfterRender(isPreview, renderMethod) {
+    return dispatch("turbo:render", { detail: { isPreview, renderMethod } })
   }
 
   notifyApplicationAfterPageLoad(timing = {}) {
--- a/src/core/view.js
+++ b/src/core/view.js
@@ -69,7 +69,7 @@ export class View {
         if (!immediateRender) await renderInterception
 
         await this.renderSnapshot(renderer)
-        this.delegate.viewRenderedSnapshot(snapshot, isPreview)
+        this.delegate.viewRenderedSnapshot(snapshot, isPreview, this.renderer.renderMethod)
         this.delegate.preloadOnLoadLinksForView(this.element)
         this.finishRenderingSnapshot(renderer)
       } finally {
--- a/src/index.js
+++ b/src/index.js
@@ -1,6 +1,7 @@
 import "./polyfills"
 import "./elements"
 import "./script_warning"
+import "./http/recent_fetch_requests"
 
 import * as Turbo from "./core"
 
--- a/src/tests/fixtures/frame_refresh_morph.html
+++ b/src/tests/fixtures/frame_refresh_morph.html
@@ -1,0 +1,3 @@
+<turbo-frame id="refresh-morph">
+  <h2>Loaded morphed frame</h2>
+</turbo-frame>
--- a/src/tests/fixtures/frame_refresh_reload.html
+++ b/src/tests/fixtures/frame_refresh_reload.html
@@ -1,0 +1,3 @@
+<turbo-frame id="refresh-reload">
+  <h2>Loaded reloadable frame</h2>
+</turbo-frame>
--- a/src/tests/fixtures/page_refresh.html
+++ b/src/tests/fixtures/page_refresh.html
@@ -1,0 +1,52 @@
+<!DOCTYPE html>
+<html id="html" data-skip-event-details="turbo:submit-start turbo:submit-end turbo:fetch-request-error">
+  <head>
+    <meta charset="utf-8">
+    <meta name="turbo-refresh-method" content="morph">
+    <meta name="turbo-refresh-scroll" content="preserve">
+
+    <title>Turbo</title>
+    <link rel="icon" href="data:;base64,iVBORw0KGgo=">
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+
+    <style>
+        body {
+            margin: 0;
+            padding: 0;
+
+            /* Ensure the page is large enough to scroll */
+            width: 150vw;
+            height: 150vh;
+        }
+    </style>
+  </head>
+  <body>
+    <h1>Page to be refreshed</h1>
+
+    <turbo-frame id="refresh-morph" src="/src/tests/fixtures/frame_refresh_morph.html" refresh="morph">
+      <h2>Frame to be morphed</h2>
+    </turbo-frame>
+
+    <turbo-frame id="refresh-reload" src="/src/tests/fixtures/frame_refresh_reload.html" refresh="reload">
+      <h2>Frame to be reloaded</h2>
+    </turbo-frame>
+
+    <div id="preserve-me" data-turbo-permanent>
+      Preserve me!
+    </div>
+
+    <div id="stimulus-controller" data-controller="test">
+      <h3>Element with Stimulus controller</h3>
+    </div>
+
+    <p><a id="link" href="/src/tests/fixtures/one.html">Link to another page</a></p>
+
+    <form id="form" action="/__turbo/refresh" method="post" class="redirect">
+      <input type="text" name="text" value="">
+      <input type="hidden" name="path" value="/src/tests/fixtures/page_refresh.html">
+      <input type="hidden" name="sleep" value="50">
+      <input id="form-submit" type="submit" value="form[method=post]">
+    </form>
+  </body>
+</html>
--- a/src/tests/fixtures/page_refresh_replace.html
+++ b/src/tests/fixtures/page_refresh_replace.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html id="html" data-skip-event-details="turbo:submit-start turbo:submit-end turbo:fetch-request-error">
+  <head>
+    <meta charset="utf-8">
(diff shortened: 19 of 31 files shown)


seanpdoyle:
This has the potential to close https://github.com/hotwired/turbo/issues/37.

seanpdoyle:
> Exclude sections from morphing
>
> There are scenarios where you might want to define sections that you want to ignore while morphing. The most common scenario is a popover or similar UI elements, which state you want to preserve when the page refreshes. We reuse the existing `data-turbo-permanent` attribute to flag such elements.

Prior to this behavior, the presence of `[data-turbo-permanent]` required the presence of an `[id]` attribute. Does this change mean that `[data-turbo-permanent]` on its own has new morph-only semantics, or will elements still require `[id]` attributes to be skipped during morphing?

src/core/drive/morph_renderer.js
```diff
@@ -0,0 +1,88 @@
...
+    } else {
+      return true
+    }
+  }
+
+  #reloadStimulusControllers = async (node) => {
+    if (node instanceof HTMLElement && node.hasAttribute("data-controller")) {
+      const originalAttribute = node.getAttribute("data-controller")
+      node.removeAttribute("data-controller")
+      await nextAnimationFrame()
```
seanpdoyle, line 70: There's an open issue (https://github.com/hotwired/turbo/issues/920) about the effects of `requestAnimationFrame()` being our timing mechanism. Animation frame callbacks might not fire when the user is browsing another tab.

I wonder if there's another timing mechanism that could be used, since morphs are able to be broadcast over WebSockets.

Maybe `nextEventLoopTick()` or `nextMicrotask()` are decent candidates?

https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/util.js#L50-L56

src/core/drive/morph_renderer.js
```diff
@@ -0,0 +1,88 @@
...
+      node.setAttribute("data-controller", originalAttribute)
+    }
+  }
+
+  #isFrameReloadedWithMorph(element) {
+    return element.getAttribute("src") && element.getAttribute("refresh") === "morph"
```
seanpdoyle, line 76: The `FrameElement` class defines a `.src` property:

https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/elements/frame_element.js#L60-L65

Could we use that here? Does it make sense to introduce a `.refresh` property in the same style?

src/http/recent_fetch_requests.js
seanpdoyle, line 1: Overriding the global `fetch` presents some risky side-effects. Are there ways we could scope this behavior to be Turbo-only? Maybe some changes to `FetchRequest` delegate classes' [prepareRequest](https://github.com/search?q=repo%3Ahotwired%2Fturbo%20prepareRequest&type=code) callbacks:

https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/drive/visit.js#L285-L289

https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/drive/form_submission.js#L101-L114

https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/frames/frame_controller.js#L192-L200

src/core/drive/morph_renderer.js
```diff
@@ -0,0 +1,88 @@
...
+
+  #remoteFrames() {
+    return document.querySelectorAll("turbo-frame[src]")
+  }
+
+  #dispatchEvent(name, detail, target = document.documentElement) {
+    const event = new CustomEvent(name, { bubbles: true, cancelable: true, detail })
+    target.dispatchEvent(event)
+    return event
+  }
```
seanpdoyle, line 87: Is this implementation different from the `dispatch` defined in the `util` module?

https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/util.js#L29-L44

src/core/drive/morph_renderer.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  #isFrameReloadedWithMorph(element) {
+    return element.getAttribute("src") && element.getAttribute("refresh") === "morph"
+  }
+
+  #remoteFrames() {
+    return document.querySelectorAll("turbo-frame[src]")
+  }
```
seanpdoyle, line 81: Since this is a `Renderer` subclass, it has `currentSnapshot` and `newSnapshot` properties:

https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/renderer.js#L6-L13

They're likely [PageSnapshot](https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/drive/page_snapshot.js) instances, which are subclasses of [Snapshot](https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/snapshot.js).

The `Snapshot` class already defines some querying properties and methods (like [firstAutofocusableElement](https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/snapshot.js)). Have you considered implementing `get remoteFrames() { ... }` on the `PageSnapshot` class, then accessing that in this class from `this.newSnapshot.remoteFrames`?

swanson:
> @swanson does that seem related to #1004?

Yes, that is exactly the situation we have.

We have a frame used as a navigation context but it is not lazy-loaded. We have a settings panel that slides in and changes there could have a big impact on the layout of the frame so we just want to reload the whole thing (while being able to keep the settings panel open, etc)

swanson:
This is going to be so nice for things like auto-saving forms where, previously, you would lose your cursor focus if you did a full reload!

src/http/recent_fetch_requests.js
KonnorRogers: I mean you could just take the existing version of fetch and export it / pass it around.

```js
export async function fetch(url, options = {}) {
  const originalFetch = window.fetch
  const modifiedHeaders = new Headers(options.headers || {})
  const requestUID = uuid()
  window.Turbo.recentRequests.add(requestUID)
  modifiedHeaders.append("X-Turbo-Request-Id", requestUID)

  return originalFetch(url, {
    ...options,
    headers: modifiedHeaders
  })
}

// Somewhere else
import { fetch } from "./fetch.js"
```

If React taught us anything, it's dont patch global fetch.

morki:
How detection of refresh works? Does it consider different query params? I think it should me configurable like:

```html
<meta name="turbo-refresh-strategy" content="same-path|same-path-and-query|only-turbo-stream">
```

The naming is only for discussion.

Am i missing something? Thank you for this awesome enhancement!

src/core/drive/visit.js
```diff
@@ -335,7 +335,7 @@ export class Visit {
   // Scrolling
 
   performScroll() {
-    if (!this.scrolled && !this.view.forceReloaded) {
+    if (!this.scrolled && !this.view.forceReloaded && !this.view.snapshot.shouldPreserveScrollPosition) {
```
Intrepidd, line 338: Since `shouldPreserveScrollPosition` doesn't check if the page is a refresh, this restores scroll even when doing a classic visit instead of a refresh. It was unexpected to me when I tried it out, we should only restore scroll for a refresh, shouldn't we ?

Intrepidd:
> How detection of refresh works? Does it consider different query params? I think it should me configurable like:
> 
> ```
> <meta name="turbo-refresh-strategy" content="same-path|same-path-and-query|only-turbo-stream">
> ```
> 
> The naming is only for discussion.
> 
> Am i missing something? Thank you for this awesome enhancement!

AFAICS different query params won't trigger a refresh but a full visit : 

https://github.com/hotwired/turbo/pull/1019/files#diff-f1972083be28125eb9c224f3769b8224ed588745e49847fd10213bfbb9ccc5fcR62-R64

I agree it would be nice to be able to perform a refresh when query params differ though. An example use case would be to redirect to the same page with a `?success=true` query param that would trigger a success animation after a form submission.

src/core/drive/morph_renderer.js
```diff
@@ -0,0 +1,88 @@
...
+    return "morph"
+  }
+
+  // Private
+
+  async #morphBody() {
+    this.#morphElements(this.currentElement, this.newElement)
+    this.#reloadRemoteFrames()
+
+    this.#dispatchEvent("turbo:morph", { currentElement: this.currentElement, newElement: this.newElement })
```
seanpdoyle, line 20: What is the significance of introducing new `turbo:morph` and `turbo:before-frame-morph` events as an alternative to extending existing events like `turbo:render`, `turbo:before-frame-render`, and `turbo:frame-render`? Could we change the shape of those events' details to incorporate information about the `refresh-method` setting being used?

seanpdoyle:
What changes are necessary to incorporate idiomorph into the [FrameRenderer.loadFrameElement](https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/frames/frame_renderer.js#L39-L42) method?

Unless I'm missing something, I don't see changes to support a `<turbo-frame refresh="morph">` element navigated by a `<a>` or `<form>` being modified by idiomorph.

afcapel:
> Prior to this behavior, the presence of [data-turbo-permanent] required the presence of an [id] attribute. Does this change mean that [data-turbo-permanent] on its own has new morph-only semantics, or will elements still require [id] attributes to be skipped during morphing?

That's correct. Initially we had another name for the data attribute  - `data-turbo-refresh-preserve` -, but we decided to reuse `data-turbo-permanent`. Semantically it does the same in both use cases.

src/core/drive/visit.js
afcapel: This is intentional. We have two different settings `<meta name="turbo-refresh-method">` and `<meta name="turbo-refresh-scroll">` because we have identified some use cases in our apps where we want to do one without the other necessarily.

src/core/drive/morph_renderer.js
afcapel: Good point, implemented in 3f452549ad3f22bd7dc40da197e61962ad631ba0 👍

src/core/streams/stream_actions.js
```diff
@@ -30,5 +30,14 @@ export const StreamActions = {
...
       targetElement.append(this.templateContent)
     })
+  },
+
+  refresh() {
+    const requestId = this.getAttribute("request-id")
+    const isRecentRequest = requestId && window.Turbo.recentRequests.has(requestId)
+    if (!isRecentRequest) {
+      window.Turbo.cache.exemptPageFromPreview()
+      window.Turbo.visit(window.location.href, { action: "replace" })
```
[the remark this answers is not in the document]
afcapel: Importing would be nicer, but unfortunately it creates a circular dependency:

```
(!) Circular dependency
src/core/index.js -> src/core/streams/stream_actions.js -> src/core/index.js
created dist/turbo.es2017-umd.js, dist/turbo.es2017-esm.js in 225ms
```

src/core/drive/morph_renderer.js
afcapel: Not really! Thanks for the pointer, I've changed the code to use the `dispatch` util function in a653d1238da17f4d7813b06b8ffa03a2e91dc300.

src/core/streams/stream_actions.js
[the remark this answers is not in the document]
seanpdoyle: Ah, thanks for exploring that @afcapel.

Since `Turbo` isn't really an instance of a class or namespace, but rather a collection of other exports from the `export * as Turbo` statement, what do you think about importing the individual components themselves?

I've had some success with this diff locally:

```diff
diff --git a/src/core/index.js b/src/core/index.js
index 743380d..34269c5 100644
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -12,8 +12,6 @@ const recentRequests = new LimitedSet(20)
 const { navigator } = session
 export { navigator, session, cache, recentRequests, PageRenderer, PageSnapshot, FrameRenderer }
 
-export { StreamActions } from "./streams/stream_actions"
-
 /**
  * Starts the main session.
  * This initialises any necessary observers such as those to monitor
diff --git a/src/core/streams/stream_actions.js b/src/core/streams/stream_actions.js
index 62801fd..e8e20bd 100644
--- a/src/core/streams/stream_actions.js
+++ b/src/core/streams/stream_actions.js
@@ -1,3 +1,5 @@
+import { cache, recentRequests, visit } from "../"
+
 export const StreamActions = {
   after() {
     this.targetElements.forEach((e) => e.parentElement?.insertBefore(this.templateContent, e.nextSibling))
@@ -34,10 +36,10 @@ export const StreamActions = {
 
   refresh() {
     const requestId = this.getAttribute("request-id")
-    const isRecentRequest = requestId && window.Turbo.recentRequests.has(requestId)
+    const isRecentRequest = requestId && recentRequests.has(requestId)
     if (!isRecentRequest) {
-      window.Turbo.cache.exemptPageFromPreview()
-      window.Turbo.visit(window.location.href, { action: "replace" })
+      cache.exemptPageFromPreview()
+      visit(window.location.href, { action: "replace" })
     }
   }
 }
diff --git a/src/index.js b/src/index.js
index b76e238..336101a 100644
--- a/src/index.js
+++ b/src/index.js
@@ -8,6 +8,7 @@ import * as Turbo from "./core"
 window.Turbo = Turbo
 Turbo.start()
 
+export { StreamActions } from "./core/streams/stream_actions"
 export * from "./core"
 export * from "./elements"
 export * from "./http"
```

According to the [official documentation](https://turbo.hotwired.dev/handbook/streams#custom-actions), applications that need to extend `StreamActions` should be importing it directly from the module instead of accessing it through `window.Turbo.StreamActions`.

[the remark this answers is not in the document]
seanpdoyle: As an alternative to (or in addition to!) resolving the circular dependency and exports, have you considered extending the shape of the `visit` function's optional `VisitOption` interface to accept a `requestId`? Passing that option along could help to encapsulate these details and limit the amount of modules that are dependent on the `cache` and `recentRequests` variables. That might look something like:

```js
import { visit } from "../"

// ...
refresh() {
  visit(this.baseURI, {
    action: "replace",
    requestId: this.getAttribute("request-id")
  })
}
```

Then, the `visit(url, options)` implementation could check for the presence of `options.requestId` to cross reference the `recentRequests`, then call `cache.exemptPageFromPreview()` if necessary.

src/http/recent_fetch_requests.js
seanpdoyle, line 1: I've opened https://github.com/hotwired/turbo/pull/1025 (cut off this branch) to explore an alternative to re-writing the `window.fetch`

src/core/streams/stream_actions.js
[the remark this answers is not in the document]
seanpdoyle: I've opened https://github.com/hotwired/turbo/pull/1026 to explore an alternative to keeping this logic in `StreamActions`. It leans more heavily on encapsulating the logic within the `Session` class. It avoids the [circular dependency trap](https://github.com/hotwired/turbo/pull/1019/files#r1349736505) described above.

palkan:
Hey @afcapel @seanpdoyle!

I'd like to pay your attention to this issue: https://github.com/hotwired/turbo/issues/951

I haven't tried this branch yet but from what I see in the changes, it seems that the current implementation may be prone to the problem mentioned in the issue (tl;dr snapshot referencing new morphed state of the document).

palkan:
@afcapel Could you please share an example of the problem that `<turbo-frame refresh="reload">` solves? From the PR description:

> For example, if you have loaded new pages of data. You don’t want to lose that data when a page refresh happens. If the user can act on that data, you want to see it refreshed too.

So, we want to reload frames after the page has been morphed. But isn't this already happening automatically for remote frames? We receive the `<turbo-frame>` HTML without the `complete` attribute from the server, so morphing would remove it from the current frame node thus triggering the reload (see https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/elements/frame_element.js#L52).

A one exclusion I see is frames within permanent elements. With the proposed solution, they would be reloaded disregarding being within permanent containers—which seems confusing to me. And this is not how the current replace strategy works.

/cc @seanpdoyle since it's related to #1029

jorgemanrubia:
@palkan 

> I haven't tried this branch yet but from what I see in the changes, it seems that the current implementation may be prone to the problem mentioned in the issue (tl;dr snapshot referencing new morphed state of the document).

Yes, I found this problem too during our explorations. The reason is that [Turbo caches snapshots in a deferred way](https://github.com/turbolinks/turbolinks/pull/390), both as a performance optimization and to let stimulus controllers an opportunity to disconnect and cache whatever they do there. Essentially, Turbo is relying on a different `<body>` for the snapshot system to work.

The PR doesn't present that problem because we don't use morphing for regular navigations across different pages, just for cases where you re-render the current page you are in, typically after a redirect. Because of the nature of these page refreshes, caching doesn't really make sense (just like you don't want to cache a stream action response that makes a partial update).

I wonder if the fix by @afcapel from https://github.com/hotwired/turbo/pull/1014 would fix https://github.com/hotwired/turbo/issues/951 too

jorgemanrubia:
@palkan 

> So, we want to reload frames after the page has been morphed. But isn't this already happening automatically for remote frames? We receive the <turbo-frame> HTML without the complete attribute from the server, so morphing would remove it from the current frame node thus triggering the reload (see

We don't want those frames to reload normally because that would trigger a replacement of the content so they would lose screen state. We want them to keep screen state and, to all effects, remain untouched if the remote content didn't change.

An example is pagination:

* You paginate new content in a turbo frame.
* You delete one of the paginated elements, that triggers a page refresh via a "redirect back" response.
* You just want to see the deleted element disappear smoothly, with the scroll kept and such. 

Checking the code I'm noticing that during the upstreaming we lost something we had in the private gem we created for this, which was excluding `turbo-frames` flagged with `refresh=reload` from morphing itself. I think we should do that.

> A one exclusion I see is frames within permanent elements. With the proposed solution, they would be reloaded disregarding being within permanent containers—which seems confusing to me. And this is not how the current replace strategy works.

This is a good point. We should exclude those.

Our availability is a bit choppy this week as we are in a company meetup. Thanks for all the feedback everyone.

palkan:
@jorgemanrubia Thanks for the quick response!

> The PR doesn't present that problem because we don't use morphing for regular navigations across different pages

I see, this is what the `isPageRefresh` guard for. My first impression was that `turbo-refresh-method` meta affects everything (even though the "refresh" is used in the name). That's just a matter of documentation.

> excluding turbo-frames flagged with refresh=reload from morphing itself

Oh, that explains everything 🙂

src/http/recent_fetch_requests.js
natematykiewicz: Also, other libraries such as Sentry patch `fetch` to add instrumentation. If you are overriding `window.fetch`, because libraries are async, I'm pretty sure you run the risk of one library removing another library's `fetch` patches (Sentry removing Turbo's or Turbo removing Sentry's). This seems like a [read-modify-write](https://en.wikipedia.org/wiki/Read–modify–write) race condition of a non-atomic operation.

natematykiewicz: If I'm reading this right, I believe this change means Turbo will break Sentry's instrumentation, as `fetch` is no longer native code.

https://github.com/getsentry/sentry-javascript/blob/a173fa22ae89380114ac4cc8766d9834704aab86/packages/integrations/src/httpclient.ts#L298-L313

src/core/drive/morph_renderer.js
jvillarejo: Is it cohesively correct that Turbo knows about Stimulus controllers and how to trigger their reload? 
Or is it better that Stimulus knows about Turbo morph and configures itself when to reload?

jorgemanrubia:
@seanpdoyle 

> Prior to this behavior, the presence of [data-turbo-permanent] required the presence of an [id] attribute. Does this change mean that [data-turbo-permanent] on its own has new morph-only semantics, or will elements still require [id] attributes to be skipped during morphing?

When using morphing, you can use `[data-turbo-permanent]` **without** a companion `id`.

src/core/drive/morph_renderer.js
jorgemanrubia: @jvillarejo the proposed implementation doesn't reload stimulus controllers automatically in general, only when a node is morphed (e.g: an attribute or its text content changed). We believe this is a good heuristic for stimulus controllers that modify the underlying `element` when connected.

As a general rule, I agree that that morphing should leave stimulus controllers alone.

src/http/recent_fetch_requests.js
jorgemanrubia: The reason for patching `fetch` globally instead of keeping the modification isolated in the library is that we want the new behavior to kick in when users invoke `fetch` manually, to prevent bounced page refreshes in those cases.

Said that, I agree that patching it globally sucks. We could certainly expose a `turbo/fetch` version you can import in your apps if you want to invoke `fetch`manually, and use the patched version internally. Still, I'm concerned about places where you don't direct control over how `fetch` is called (e.g: if using [@rails/request.js](https://github.com/rails/requestjs-rails)).

@natematykiewicz that's a great point and something I hadn't considered at all. 

I think that going with the importable patched `fetch` that @KonnorRogers suggests is the way to go. For invocations you don't have control over, you can still override `window.fetch` in your app if you want that, but sounds like a safer default. I'll take a stab at that.

On a final note, I wouldn't mind reworking the whole thing if you have a better idea about how to detect/prevent page refresh bounces automatically.

Thanks for the feedback here!

src/core/drive/morph_renderer.js
seanpdoyle: > is it better that Stimulus knows about Turbo morph and configures itself when to reload?

Could Stimulus detect the presence of Turbo (either the package, or the new `<meta>` elements) and attach a document-wide `turbo:render` listener to respond to renders with `renderMethod == "morph"`? I might be misunderstand the file, but Stimulus defines a [defaultSchema](https://github.com/hotwired/stimulus/blob/5fd12c7b6af88620741e3aa969b6ab440559c5b7/src/core/schema.ts#L11-L13) export with `[data-controller]` and friends as the default values in what appears to be an overridable Object.

```ts
export const defaultSchema: Schema = {
  controllerAttribute: "data-controller",
  actionAttribute: "data-action",
  targetAttribute: "data-target",
  // ...
}
```

src/http/recent_fetch_requests.js
seanpdoyle: > when users invoke `fetch` manually

If a user is invoking their own `fetch` to an endpoint that responds in a Turbo-enabled way, would it be reasonable to encourage them to use a package like `@rails/request.js`, which could either be extended to be Turbo aware?

I'm imagining using some heuristic in [FetchRequest.headers](https://github.com/rails/request.js/blob/03d484c4ebf00e4f7ef3086dd67a0eeb75a060ef/src/fetch_request.js#L66-L80)). Either that, or could Turbo export some kind of `prepareRequest` or `withTurboHeaders` helper method?

Maybe I'm misunderstanding the problem, but I'm having trouble imagining the code that would be making `fetch` requests.

If a `fetch` were made without the `X-Turbo-Request-Id` header outside of a Turbo-handled `<form>` element submission, how would it communicate the native Response back into the Turbo lifecycle? How would it run the risk of circumventing the refresh debouncing mechanism?

The only way that I can imagine is passing the HTML directly to something like `Turbo.visit(path, { responseHTML: await response.text() })`, which breaches public API and encapsulation in a way that feels dangerous to begin with.

src/core/drive/visit.js
seanpdoyle: @afcapel is scroll preservation something that's specific to Refresh Morph?

Could scroll preservation be implemented separately, with a more generic `<meta>` name?

Maybe something like `<meta name="turbo-scroll" value="preserve">`? The scroll-related portion of the diff feels small by comparison. I'd be happy to open a PR to add that behavior separately.

src/http/recent_fetch_requests.js
marcelolx: > We could certainly expose a `turbo/fetch` version you can import in your apps if you want to invoke `fetch` manually, and use the patched version internally. Still, I'm concerned about places where you don't direct control over how fetch is called (e.g: if using [@rails/request.js](https://github.com/rails/requestjs-rails))

I think depending on how it is implemented in Turbo; we can always support it in [@rails/request.js](https://github.com/rails/requestjs-rails)... @rails/request.js is already aware of Turbo, so making it use a patched version of Fetch from Turbo is also possible. 

I am not closely following this PR, but I just wanted to share that, if possible, we can add support to that in  [@rails/request.js](https://github.com/rails/requestjs-rails). Obviously, other libraries may have the same problem, but I guess nothing will prevent them from supporting it, either.

src/core/drive/visit.js
seanpdoyle: I've opened https://github.com/hotwired/turbo/pull/1040 to add `<meta name="turbo-scroll" content="preserve">` support. That PR is against `main`.

I think there's a lot of value in managing scroll preservation outside of a morphing context, so generalizing to `turbo-scroll` feels like a win.

jorgemanrubia: @seanpdoyle I'd like to keep the scroll configuration under the same "page refresh" umbrella. Our vision here is not adding more flexibility but aiming for seamless smoother page updates (as in re-render the current page again). We believe it's in that scenario where keeping the scroll makes sense, to the point where "preserve" should be the future default for such cases being "reset" the exception. In fact, the only reason we added the option is that we found a case of "page refresh" where you wanted to opt out of scroll preservation due to design considerations.

Just like morphing, scroll preservation for regular navigation could make sense in other scenarios, but we believe those will be quite rare and are out of the scope of what we're addressing here. I'd like to keep both "scroll" and "morph" tied to "page refreshes" for now, as I think it makes for a more cohesive development story.

brunoprietog:
So, `<turbo-frame src="..." refresh="morph"></turbo-frame>` will be used instead of `refresh="reload"`?

src/core/drive/morph_renderer.js
```diff
@@ -0,0 +1,92 @@
...
+    }, { once: true })
+  }
+
+  #morphFrameUpdate = (currentElement, newElement) => {
+    dispatch("turbo:before-frame-morph", {
+      target: currentElement,
+      detail: { currentElement, newElement }
+    })
+    this.#morphElements(currentElement, newElement)
+  }
```
seanpdoyle, line 62: Would it be possible to achieve this branch's goals without introducing new types of events?

Could a morphing-supported `<turbo-frame>` dispatch `turbo:before-frame-render` and `turbo:frame-render` events with `renderMethod: "morph"` in the same way we're dispatching `turbo:render` events?

When handling `turbo:before-frame-render` events with `renderMethod: "morph"`, listeners can access the `currentElement` analog with `event.target` and the `newElement` analog with `event.detail.newFrame`.

I also wonder if would-be `turbo:morph` listeners could instead listen for `turbo:render` events with `renderMethod: "morph"`, then access the `currentElement` from `event.target` and the `newElement` from `event.detail.newBody`.

jorgemanrubia:
> So, <turbo-frame src="..." refresh="morph"></turbo-frame> will be used instead of refresh="reload"?

@brunoprietog we were talking about this yesterday. We are going to remove the attribute and make it always refresh with morphing remote `turbo-frames` when a page refresh with morphing happens. We still need to test in our apps but we believe this is a great default behavior, and removes a confusing element from the API.

northeastprince:
Will this PR make Turbo Frames mostly redundant?

mindaslab:
Thank you so much, bring it into Rails fast. I am getting killed by ReactJs.

jorgemanrubia:
We have removed the logic to reconnect stimulus controllers during morphing.

The implementation was flawed in any case: we only wanted to reload controllers when a certain node's attributes changed, not in all cases. But even if we got that right, we now think that wasn't the right call.

The idea of reconnecting when some attribute changed was not something we needed either for the HEY calendar or for the BC4's card table, which are the two main apps where we've been testing this. In fact, doing that for all the stimulus controllers would break some calendar features, that's how we detected the implementation was flawed 😅.

The heuristic was motivated by some mobile stimulus controllers tracking some state in the DOM. We now think we should not intervene with controller's reconnection at all. If we introduced some API here, it will be with the opposite semantics: flag controllers to reconnect during morphing, but we'll really need to see the need first. 

@palkan's [article](https://evilmartians.com/chronicles/the-future-of-full-stack-rails-turbo-morph-drive) shows a great example of how to leverage a DOM change with morphing.

src/core/drive/page_view.js
```diff
@@ -55,6 +59,10 @@ export class PageView extends View {
     return this.snapshotCache.get(location)
   }
 
+  isPageRefresh(visit) {
+    return visit && this.lastRenderedLocation.href === visit.location.href
```
seanpdoyle, line 63: @jorgemanrubia @afcapel does this conditional incorporate re-rendering failed form submissions where the response is `4xx` (like a `400 Bad Request` or `422 Unprocessable entity`)?

When a Form Submission succeeds, [navigator.formSubmissionSucceededWithResponse()](https://github.com/hotwired/turbo/blob/3d48c865938a745b280feb567ff124ebc144dfb1/src/core/drive/navigator.js#L71-L90) gets called, which is ultimately the place where the Visit gets proposed.

Conversely, when the Form Submission fails with a Client Error (status `400..499`), then [navigator.formSubmissionFailedWithResponse](https://github.com/hotwired/turbo/blob/3d48c865938a745b280feb567ff124ebc144dfb1/src/core/drive/navigator.js#L92-L105) gets called. 

`navigator.formSubmissionFailedWithResponse` doesn't have any concept of a `Visit`, and interacts with the View directly.

jorgemanrubia: Thanks for the heads up @seanpdoyle. That's a great scenario for morphing indeed. See https://github.com/hotwired/turbo/pull/1061.

src/core/drive/morph_renderer.js
seanpdoyle: @jorgemanrubia this PR is really shaping up!

I think idea of introducing new events instead of extending existing events is the only remaining concept I'm curious to learn more about.

afcapel: @seanpdoyle I think having different event types make it easier to target only the actions you're interested on.

jorgemanrubia:
A heads up that we still need to fix an issue with mobile apps before releasing a beta here. Right now, when you redirect back in a native app, you always get a blank flash because it always uses a new page view (whether you use morphing or not). We'll fix that in their corresponding adapters.

jon-sully:
> The system includes a mechanism to ignore refresh actions that originated [from a] web request that originated in the same browser session.

This is _very_ clever. So well done! 🙌 "Don't trigger a Page Refresh (broadcasted) when you _just_ did a Page Refresh (since the controller redirected you to the same page)". So nice ✨

PR #1023, opened 2023-10-08
Add test coverage for `StreamActions` export

seanpdoyle:
According to the [official documentation][], applications that need to extend `StreamActions` should be importing it directly from the module instead of accessing it through `window.Turbo.StreamActions`.

This commit adds test coverage to exercise the corresponding `import`. It also re-arranges the `export` to explicitly occur in the `@hotwired/turbo` module itself, instead of being transitively exported as part of the `@hotwired/turbo/core` module.

[official documentation]: https://turbo.hotwired.dev/handbook/streams#custom-actions

### DIFF
--- a/src/tests/unit/export_tests.js
+++ b/src/tests/unit/export_tests.js
@@ -1,7 +1,9 @@
 import { assert } from "@open-wc/testing"
-import * as Turbo from "../../index"
+import * as Turbo from "../../"
+import { StreamActions } from "../../"
 
 test("test Turbo interface", () => {
+  assert.equal(typeof Turbo.StreamActions, "object")
   assert.equal(typeof Turbo.start, "function")
   assert.equal(typeof Turbo.registerAdapter, "function")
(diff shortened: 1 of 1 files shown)


src/core/index.js
```diff
@@ -10,8 +10,6 @@ const cache = new Cache(session)
 const { navigator } = session
 export { navigator, session, cache, PageRenderer, PageSnapshot, FrameRenderer }
 
-export { StreamActions } from "./streams/stream_actions"
```
afcapel, line 13: Can we keep the export here too? It's not strictly necessary, but it will avoid breaking libraries and client code that depend on this import.

seanpdoyle: I've pushed up a change to continue to export as `Turbo.StreamActions`.

src/index.js
```diff
@@ -10,3 +10,4 @@ Turbo.start()
 export * from "./core"
```
afcapel, line 10: Isn't this export already exporting `StreamActions`?

PR #1024, opened 2023-10-08
Move `Cache` instantiation to `Session`

seanpdoyle:
When the `Cache` class was [introduced in a54ac17][], it depended on a `Session` instance. When it was [modified in f86a376][], that dependency was no longer necessary.

This commit preserves the export of the `Turbo.cache` property while re-arranging the location it's declared. Instead of instantiating the `Cache` instance in `@hotwired/turbo/core/`, this commit defines it as a property on the `Session` instance.

[introduced in a54ac17]: https://github.com/hotwired/turbo/commit/a54ac176a707487feee0424b8d719927305c8b7a
[modified in f86a376]: https://github.com/hotwired/turbo/commit/f86a376b0d088c8ff83ef8180b95d492527902e5

### DIFF
--- a/src/core/cache.js
+++ b/src/core/cache.js
@@ -2,10 +2,6 @@ import { setMetaContent } from "../util"
 import { SnapshotCache } from "./drive/snapshot_cache"
 
 export class Cache {
-  constructor(session) {
-    this.session = session
-  }
-
   clear() {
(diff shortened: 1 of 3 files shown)


afcapel:
Good catch 👍

PR #1025, opened 2023-10-08
Replace global fetch patch with Turbo-specific behavior 

seanpdoyle:
As an alternative to patching `fetch` globally, this commit introduces a
new `FetchRequestObserver` class, and makes the `Session` instance a
delegate.

As prep-work, this commit also moves the `recentRequests` out of the
global `@hotwired/turbo` module an into an instance property on
`Session`.

The `FetchRequestObserver` instance attaches an event listener for
`turbo:before-fetch-request` (in this case, on the `<html>` element).
Instead of generating the request ID in the delegate, this commit also
extends the `FetchRequest` class to generate the `uuid()` and set it to
the new `FetchRequest.id` property.

The `Session.willFetch` delegate method will be invoked before each
request, and will read the `X-Turbo-Request-Id` header and write it to
the instance's `recentRequests` set.


jorgemanrubia (commented):
Hey @seanpdoyle,

Thanks for working on this. What I had in mind here was two things:

1. Define an internal `fetch` that decorates `window.fetch` with the new behavior, without overriding it. Turbo will internally use this internal `fetch`. And applications will be able to import it too:

```javascript
import { fetch } from "@hotwired/turbo-rails"
```

2. Expose it also via `Turbo.window`, so that libraries that doesn't build against Turbo – such as [requestjs-rails](https://github.com/rails/requestjs-rails) – can still do something like:

```javascript
const fetch = window.Turbo?.fetch || window.fetch
```

(1) will be the preferred form of using the new fetch from apps. (2) is intended a fallback mechanism.

I think your refactor is a nice improvement over the original PR, but it's kind of misaligned with (1). I think we need a custom `fetch` to use internally and export to clients, instead of making the current requests observable.

What do you think?

seanpdoyle:
@jorgemanrubia I think [I am misunderstanding](https://github.com/hotwired/turbo/pull/1019#discussion_r1361681176) the use case for applications to utilize a Turbo-specific `fetch` (1. above).

Why would an application need to make a request with an `X-Turbo-Request-Id` header? Even if they were able to import and utilize the Turbo-extended `fetch`, how would they application pass the HTTP response back to Turbo in a way that is compatible with Turbo Drive?

If they're submitting a request outside of `Turbo.visit` or `HTMLFormElement.requestSubmit()`, what risks would an internal `fetch` guard against?

jorgemanrubia:
@seanpdoyle because that way the application makes sure that the request id is registered by `Turbo` internally. So that, if a "refresh" stream action is broadcasted, and if it carries information about the "request id that originated it", Turbo can determine if it was self-originated and, in that case, ignore it.

Practical case: you have a Drag & Drop system that relies on `window.fetch` from a stimulus controller to update stuff. You have configured the server the involved models to broadcast "page refreshes" on changes. If you drop an element, you don't want to get a broadcasted "page refresh" from the same request that you originated.

In the general case, you are interested in all the broadcasted "page refreshes" except the ones that you originate. The vast majority of your page refreshes will come via Turbo actions, but not necessarily all of them, since using `window.fetch` is still relatively common in certain circumstances. That's the problem we're trying to solve here.

jorgemanrubia:
@seanpdoyle not in outcomes but in ergonomics from the caller side, imo. Also, the first form's encapsulation is higher, hiding away the fact that internally it's manipulating the request headers.

adrienpoly:
Hello 

Thanks for all the work going on, it looks really promising.

I wanted to share a few thoughts:

```js
import { fetch } from "@hotwired/turbo"
```

I personally think having a custom fetch option coming from Turbo will add lots of confusion at the end. 

I have been really enjoying the [@rails/request.js](https://github.com/rails/requestjs-rails) library so far. The API is nice with lots of small goodies.  Out of all the fetch libraries that I have been using, I think it has the best API.

in https://github.com/hotwired/turbo/pull/1019#discussion_r1362053564 @marcelolx proposed some help to make Rails/request.js Turbo compatible. My favorite solution would be to be able to use this package for all my fetch-related actions.

Therefore project NOT using  [@rails/request.js](https://github.com/rails/requestjs-rails), I would second @seanpdoyle suggestion for something like that.

```js
import { withTurboHeaders } from "@hotwired/turbo"

window.fetch("/my-path", withTurboHeaders({ ... }))
```

This makes it clear that it is not the preferred option. The preferred option would be documented by using request.js.

And for Rails/request.js let's make sure this works out of the box from the standard fetch library provided by Rails.

jorgemanrubia:
Hey @adrienpoly, +1 to make `@rails/request.js` the preferred option for manual `fetch` operations. We can make sure we document that clearly in the docs also, and we can improve the compatibility with Turbo in the process.

I don't agree with making the not-preferred option less ergonomic though.

adrienpoly:
The issue I see with exposing this new custom fetch from Turbo as a public API is that it will make it harder down the road to unify Turbo and Request.js. To me, it should remain a private API inside Turbo or come from request.js

When I saw Request.js for the first time, I saw it as an extraction of Turbo internal fetch and thought the plan was to ultimately replace Turbo internal fetch with request.js. I am sure there are quite a few differences in both implementations but last time I looked there was also a lot of duplication in both code bases. 

IMHO, the most comprehensive API would be 

```js
import { get, put, post, delete  } from "@hotwired/turbo"
```

and those get, put, post, and delete would be coming from request.js

seanpdoyle:
Closing due to comments above: https://github.com/hotwired/turbo/pull/1025#pullrequestreview-1692732716.

PR #1027, opened 2023-10-09
Make `Renderer` instance available to `View` delegates

seanpdoyle:
The changes proposed in [#1019][] require dispatching the new `Renderer.renderMethod` property as part of the `turbo:render` event. Similarly, there's an opportunity to include that information in the `turbo:before-render`, `turbo:before-frame-render`, and `turbo:frame-render` events.

To simplify those chains of object access, this commit changes the `View` class delegate's `allowsImmediateRender` and `viewRenderedSnapshot` methods to accept the `Renderer` instance, instead of individual properties.

With access to the instance, the delegate's can read properties like `isPreview` along with the `element` (transitively through the `newSnapshot` property).

In order to dispatch the `turbo:frame-render` event closer to the moment in time that the view renders, this commit removes the `Session.frameRendered` callback, and replaces it with a `dispatch("turbo:frame-render")` call inside the
`FrameController.viewRenderedSnapshot(renderer)` delegate method.

In order to do so, this commit must work around the fact that `turbo:frame-render` events include the `FetchResponse` instance involved in the render. Since that instance isn't available at render time, this commit wraps the `await this.view.render(renderer)` in a utility function that injects the `FetchResponse` instance into the Custom event's `event.detail` object immediately after it's initially dispatched.

Ideally, this work around will only be temporary, since the `turbo:frame-load` event also includes `event.detail.fetchResponse`. There's an opportunity to deprecate that property in `turbo:frame-render` events in the future.

[#1019]: https://github.com/hotwired/turbo/pull/1019


src/core/frames/frame_controller.js
```diff
@@ -276,7 +276,13 @@ export class FrameController {
     return !defaultPrevented
   }
 
-  viewRenderedSnapshot(_snapshot, _isPreview) {}
+  viewRenderedSnapshot({ currentSnapshot: { element: frame } }) {
+    return dispatch("turbo:frame-render", {
+      detail: {},
```
seanpdoyle, line 281: This is where we'll be able to assign the `renderMethod` key based on the availability of the `renderer` instance

seanpdoyle:
@afcapel this PR re-arranges the internal View delegate to make it easier to provide the `renderMethod` as part of `turbo:render` and `turbo:before-render` events.

PR #1028, opened 2023-10-09
Infer `renderElement` during `Renderer` construction

seanpdoyle:
Follow-up to [#1192][]
Closes [#1303][]

Remove the argument from all `Renderer` subclass constructor call sites.
In its place, define the default value for the `Renderer.renderElement`
property based on the `static renderElement(currentElement, newElement)`
defined by the inheriting class.

By automatically determining which `renderElement` implementation based
on class, this commit enables a simpler Frame morphing strategy. Instead
of attaching one-time event listeners, this commit maintains a private
`#shouldMorphFrame` property that is set, then un-set during the
rendering lifecycle of a frame.

Similarly, this commit implements the
`MorphingFrameRenderer.preservingPermanentElements` method in the same
style as the `MorphingPageRenderer.preservingPermanentElements`. Since
_instances_ of `MorphingFrameRenderer` are now being used instead of the
static `MorphingFrameRenderer.renderElement`, we're able to utilize an
instance method instead of passing around an additional anonymous
function (which is the implementation that [#1303][] proposed).

[#1192]: https://github.com/hotwired/turbo/pull/1192
[#1303]: https://github.com/hotwired/turbo/pull/1303

Co-authored-by: Micah Geisel <micah@botandrose.com>

### DIFF
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -19,7 +19,7 @@ export class PageView extends View {
     const shouldMorphPage = this.isPageRefresh(visit) && this.snapshot.shouldMorphPage
     const rendererClass = shouldMorphPage ? MorphingPageRenderer : PageRenderer
 
-    const renderer = new rendererClass(this.snapshot, snapshot, rendererClass.renderElement, isPreview, willRender)
+    const renderer = new rendererClass(this.snapshot, snapshot, isPreview, willRender)
 
     if (!renderer.shouldRender) {
       this.forceReloaded = true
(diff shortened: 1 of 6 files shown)


seanpdoyle:
@afcapel this aims to simplify the Renderer constructor and interface.

seanpdoyle:
@jorgemanrubia @brunoprietog while this is an older PR, I've re-purposed the changes to follow up #1192 and to resolve a bug documented by #1303. It's ready for review.

botandrose (approved):
@seanpdoyle Nice! Much better. +1 from me.

jorgemanrubia (approved):
Nice one @seanpdoyle, thanks.

PR #1029, opened 2023-10-09
Extract `Morph` class, then use it in `FrameRenderer`

seanpdoyle:
Declaring all `Idiomorph`-related logic in the `MorphRenderer` limits
its accessibility to other parts of the system. For example,
`<turbo-frame refresh="morph">` elements are unable to morph their
renders when driven by `<a>` navigations or `<form>` submissions.

This commit extracts the bulk of the morphing logic into a new `Morph`
class. The `Morph` encapsulates the call to `Idiomorph`, along with the
remote `<turbo-frame>` reloading and `[data-turbo-permanent]` checking.

With this extraction, the `MorphRenderer` is implemented in terms of
delegating to the static `Morph.render` method.

With that extraction in place, the `FrameRenderer.renderElement` method
can incorporate the `element.src && element.refresh === "morph"` check,
calling `FrameRenderer.morph` when true, then falling back to the
default `FrameRenderer.replace` when false.

For the sake of consistency, declare all `Renderer` subclasses'
`renderElement` methods in terms of `static replace` and `static morph`
methods to be explicit about which styles they support.

This commit includes test coverage for morphing `<turbo-frame
refresh="morph">` elements driven by typical navigation.

The potential for `<turbo-stream action="morph">`
---

With the `Morph.render` function existing separately from
`MorphRenderer`, there's the potential to add a new
`StreamAction.morph`.

The implementation would look something like:

```js
morph() {
  this.targetElements.forEach((targetElement) => {
    Morph.render(targetElement, this.templateContent)
  })
}
```

I've omitted that from this commit because I'm not sure if that's an
interface we're interested in introducing, but I did want to highlight
the possibility here. It'd be an Idiomorph-powered version of the
[turbo-morph][] package.

[turbo-morph]: https://github.com/marcoroth/turbo-morph

### DIFF
--- a/src/core/frames/frame_renderer.js
+++ b/src/core/frames/frame_renderer.js
@@ -1,8 +1,17 @@
-import { activateScriptElement, nextAnimationFrame } from "../../util"
+import { activateScriptElement, nextAnimationFrame, dispatch } from "../../util"
 import { Renderer } from "../renderer"
+import { Morph } from "../morph"
 
 export class FrameRenderer extends Renderer {
   static renderElement(currentElement, newElement) {
+    if (currentElement.src && currentElement.refresh === "morph") {
+      FrameRenderer.morph(currentElement, newElement)
+    } else {
+      FrameRenderer.replace(currentElement, newElement)
+    }
+  }
+
+  static replace(currentElement, newElement) {
     const destinationRange = document.createRange()
     destinationRange.selectNodeContents(currentElement)
     destinationRange.deleteContents()
@@ -15,6 +24,15 @@ export class FrameRenderer extends Renderer {
     }
   }
 
+  static morph(currentElement, newElement) {
+    dispatch("turbo:before-frame-morph", {
+      target: currentElement,
+      detail: { currentElement, newElement }
+    })
+
+    Morph.render(currentElement, newElement.children, "innerHTML")
+  }
+
   constructor(delegate, currentSnapshot, newSnapshot, renderElement, isPreview, willRender = true) {
     super(currentSnapshot, newSnapshot, renderElement, isPreview, willRender)
     this.delegate = delegate
@@ -22,6 +40,14 @@ export class FrameRenderer extends Renderer {
 
   get shouldRender() {
     return true
+  }
+
+  get renderMethod() {
+    if (this.currentElement.refresh === "morph") {
+      return "morph"
+    } else {
+      return super.renderMethod
+    }
   }
 
   async render() {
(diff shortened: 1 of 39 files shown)


src/tests/functional/rendering_tests.js
```diff
@@ -440,6 +442,20 @@ test("test restores focus during turbo-frame rendering when transposing the acti
...
+  const frame = await page.locator("turbo-frame#hello")
+
+  await frame.evaluate((frame) => frame.setAttribute("refresh", "morph"))
+  await input.press("Enter")
+  await nextEventOnTarget(page, "hello", "turbo:frame-render")
+
+  assert.ok(await locatorHasFocus(input), "restores focus after page loads")
+  assert.equal(await input.count(), 1)
+  assert.equal(await frame.count("#permanent-form-in-frame"), 1)
+  assert.equal(await frame.count("#permanent-descendant-input-in-frame"), 1)
```
seanpdoyle, line 456: Something is off, I can't figure out a way to pass these. When the morph occurs, it duplicates the `[data-turbo-permanent]` elements.

src/core/frames/frame_renderer.js
```diff
@@ -15,6 +24,15 @@ export class FrameRenderer extends Renderer {
     }
   }
 
+  static morph(currentElement, newElement) {
+    dispatch("turbo:before-frame-morph", {
+      target: currentElement,
+      detail: { currentElement, newElement }
+    })
+
+    Morph.render(currentElement, newElement, "innerHTML")
```
palkan, line 33: I've just found that Idiomorph's `innerHTML` mode doesn't work as we expect: it only consider this option for the target node, not the new element; so, the correct code would be:

```suggestion
    Morph.render(currentElement, newElement.children, "innerHTML")
```

src/core/morph.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  constructor(currentElement, newElement) {
+    this.currentElement = currentElement
+    this.newElement = newElement
+  }
+
+  render(morphStyle = "outerHTML") {
+    window.Idiomorph.morph(this.currentElement, this.newElement, {
+      ignoreActiveValue: true,
```
seanpdoyle, line 18: The [ignoreActiveValue:](https://github.com/bigskysoftware/idiomorph/tree/4231990153895d058848b3153efc7e1d2ed2727a#options) option was introduced in `0.0.9`, and feels important enough to include as a default for the sake of focus preservation and to minimize disruptions to the element with focus.

Progressing this branch forward hinges on https://github.com/bigskysoftware/idiomorph/pull/12 being merged. Without it, the package needs to depend on `window.Idiomorph`.

PR #1031, opened 2023-10-11
Integrate `[data-turbo-temporary]` with `pagehide` event

### LINKED ISSUE #1030: bfcache and data-turbo-temporary should work together
## Scenario:
Sample html:
```html
<div class="flash" data-turbo-temporary>
  Your cart was updated!
</div>
```
1. load page in a browser that supports bfcache (eg chrome, ff)
2. navigate away using url
3. navigate back using browser back button

**Expectation:** `data-turbo-temporary` elements are remove
**Actual:** `data-turbo-temporary` elements are restored from [bfcache](https://web.dev/articles/bfcache)

Additional information:
- There does not seem to be any way to use the turbo events (eg `turbo:before-cache`) to remove these elements
- I handled this issue for my code base with something similar to:
```js
window.addEventListener('beforeunload', () => {
  document.querySelectorAll('[data-turbo-temporary]:not([data-turbo-temporary="false"])')
    .forEach((elem) => { elem.remove() })
});
```
Questions:
- Is there a nice & sensible way to handle hiding flash messages so they won't be cached by bfcache? 
- Would it make sense for turbo to hook into `beforeunload` and remove the `data-turbo-temporary` elements?

Thanks for any thoughts on this and thank you for Turbo.

seanpdoyle:
Closes #1030

Add a [pagehide][] listener to the `CacheObserver` class to listen for when the page has potential to enter the browser's [Back-Forward Cache][bcfache].

When the event is dispatched with the [persisted][] set to `true`, remove all temporary elements.

[pagehide]: https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event
[bfcache]: https://web.dev/articles/bfcache#observe_when_a_page_is_entering_bfcache
[persisted]: https://developer.mozilla.org/en-US/docs/Web/API/PageTransitionEvent/persisted


seanpdoyle:
The suite is passing for Chrome, but failing for Firefox. I'm not sure how to proceed.

PR #1032, opened 2023-10-12
Expand "Content Missing" Message

ncphillips:
## Change
Provide more context when the turbo-frame from the loaded page is missing.

## Reason

One of the things Rails does well is provide feedback that directs developers toward fixing their issues. 

Currently, Turbo is lacking in this kind of helpful error messages. 

I myself, and others I've known who've tried to learn Turbo, have gotten stuck and confused by the accurate but unhelpful "Content Missing" message that appears.

### DIFF
--- a/src/core/frames/frame_view.js
+++ b/src/core/frames/frame_view.js
@@ -3,7 +3,7 @@ import { View } from "../view"
 
 export class FrameView extends View {
   missing() {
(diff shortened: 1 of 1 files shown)


chrisdmacrae:
Food for thought: Turbo should be dev environment aware somehow (e.g., add a TURBO_ENV into the meta tags) that if present, and `development`, you provide helpful error messages in the console and in this case, an actual HTML error "page" similar to the Rails error page.

Issue #857, opened 2023-01-30
Turbo-frame caching doesn't seem to work when preloading frame links.


bennadel:
I [raised this issue on the dev forum](https://discuss.hotwired.dev/t/caching-of-turbo-frame-not-possible/4754), and it was suggested that I open an Issue here - apologies if this is not the correct course of action.

I am attempting to use a `<turbo-frame id="..." data-turbo-action="advance">` in conjunction with some embedded links that preload the content, `<a href="..." data-turbo-preload>`.

Here's what I **do see**:

1. The turbo-frame works in so much as the embedded links make a request (when clicked) and re-render the frame content only (leaving the rest of the page untouched).
2. Turbo drive pre-fetches the `<a>` tag content immediately when the page loads.

What I **don't see**:

1. A Preview of the cached link rendered immediately upon clicking (as fetch is made in the background).
2. Two `turbo:render` events (documentation says one should fire for the preview and then another for the fresh content).

Here's a truncated version of the code that I am running - it's intended to be a "slide show" kind of thing for quotes:

```html
<turbo-frame id="quote-frame" data-turbo-action="advance">
	<figure>
		#encodeForHtml( quote.excerpt )#
	</figure>
	<div>
		<a
			href="index.htm?quoteIndex=#encodeForUrl( quote.prevIndex )#"
			data-turbo-preload>
			Prev quote
		</a>
		<a
			href="index.htm?quoteIndex=#encodeForUrl( quote.nextIndex )#"
			data-turbo-preload>
			Next quote
		</a>
	</div>
</turbo-frame>
```

Over on the dev forum, someone indicated that caching for turbo-frame links _works_ ... but, **not** in conjunction with `data-turbo-preload`. Thanks for any help you can offer!

seanpdoyle:
Thank you for opening this issue!

At the moment, pre-loading `<a>` elements doesn't incorporate any contextual Turbo Frame information:

https://github.com/hotwired/turbo/blob/f04a21c9418bc384f5cdafaf9d2ed21caafbf438/src/core/drive/preloader.ts#L37-L53

What's crucially missing here is the `Turbo-Frame` header.

If we were to support that, the `Preloader` would need to evaluate _which `<turbo-frame>`_ (if any) the `<a>` element would drive:

* does it have a `<turbo-frame>` ancestor
* does it reference a `<turbo-frame>` through its `[data-turbo-frame]` attribute
[...]

PR #1033, opened 2023-10-12
Guard `[data-turbo-preload]` with conditionals

seanpdoyle:
Closes [#857](https://github.com/hotwired/turbo/issues/857)

Prior to this change, _any_ `<a>` element with `[data-turbo-preload]` would be preloaded. With that behavior comes some risk:

* if an element is nested within a `[data-turbo="false"]` element, the preloading would occur unnecessarily
* if an element has `[data-turbo-stream]`, the preloaded request won't be the same as the ensuing request due to differences in the `Accept:` header
* if an element has `[data-turbo-method]`, the preloaded request won't be the same as the ensuing request due to differences in the method
* if an element is within a `<turbo-frame>` or driving a frame via `[data-turbo-frame]`, the preloaded request won't be the same as the ensuing request due to differences in the `Turbo-Frame:` header

This commit extends the `Preloader` delegate interface to include a `shouldPreloadLink(link: HTMLAnchorElement)` predicate method to give delegates an opportunity to opt-out of preloading. The `Session` implementation of the delegate class adds rejects `<a>` elements that match any of those cases.

### DIFF
--- a/src/core/drive/preloader.js
+++ b/src/core/drive/preloader.js
@@ -4,27 +4,28 @@ import { fetch } from "../../http/fetch"
 export class Preloader {
   selector = "a[data-turbo-preload]"
 
-  constructor(delegate) {
+  constructor(delegate, snapshotCache) {
     this.delegate = delegate
-  }
-
-  get snapshotCache() {
-    return this.delegate.navigator.view.snapshotCache
+    this.snapshotCache = snapshotCache
   }
 
   start() {
     if (document.readyState === "loading") {
-      return document.addEventListener("DOMContentLoaded", () => {
-        this.preloadOnLoadLinksForView(document.body)
-      })
+      document.addEventListener("DOMContentLoaded", this.#preloadAll)
     } else {
       this.preloadOnLoadLinksForView(document.body)
     }
   }
 
+  stop() {
+    document.removeEventListener("DOMContentLoaded", this.#preloadAll)
+  }
+
   preloadOnLoadLinksForView(element) {
     for (const link of element.querySelectorAll(this.selector)) {
-      this.preloadURL(link)
+      if (this.delegate.shouldPreloadLink(link)) {
+        this.preloadURL(link)
+      }
     }
   }
 
@@ -44,5 +45,9 @@ export class Preloader {
     } catch (_) {
       // If we cannot preload that is ok!
     }
+  }
+
+  #preloadAll = () => {
+    this.preloadOnLoadLinksForView(document.body)
   }
 }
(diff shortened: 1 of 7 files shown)


seanpdoyle:
I've opened https://github.com/hotwired/turbo-site/pull/150 to make these edge cases more obvious.

afcapel:
Thanks @seanpdoyle, nice one!

adambutler:
@seanpdoyle I'm trying to achieve some preloading behaviour that appears to be impossible with the changes made here, specifically:

```js
const frame = frameTarget == "_top" ?
  null :
  document.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame:not([disabled])");
``` 

My understanding is:

1. If the `frameTarget` is `"_top:"` then frame is set to an explicit `null`
2. Otherwise; Attempt to find the element by the `data-turbo-frame` attribute.
3. Otherwise; Get the closest `<turbo-frame>`

Then if it's present and a kind of `FrameElement` a prefetch will be prevented.

In my example, I have tabbed content within a page containing other content. I wish for users to be able to click the tabs and for the content to be shown instantly since it has been prefetched. See my diagram for more information:

CleanShot 2024-01-08 at 14 16 04

If I set my `data-turbo-frame="_top"` this causes it to work as expected but does mean that I'm loading and replacing the whole document rather than just the frame I'd like.

**In Turbo `v7.3.0`:**

- The `fetch` for the prefetch pages performs as expected
- However, upon clicking the link the cache is not used.
- Next time the same link is used within in the same page visit the cache is used.

However in Turbo `v8.0.0-beta.2` I believe the issue of the prefrech populating the cache to be used upon click has been solved by some other PR all that remains is fixing an issue that this PR has seemed to introduce where the prefetch is not invoked due to the guard introduced.

My suggestion might be nieve as I'm not completely clear on the why this condition was added. I was thinking though that:

1. If the `frameTarget` is `"_top:"` then it continues to work
1. If the `frameTarget` is not present it allows it to perform the prefetch (a change in behaviour)
2. If the `frameTarget` is present
  a. If it matches the closest turbo frame it allows it to perform the prefetch (a change in behaviour)
  b. Otherwise the prefetch is prevented

The

```diff
    const frame = frameTarget == "_top" ?
      null :
-      document.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame:not([disabled])");
+      document.getElementById(frameTarget);
+    
+    if (frame && frame !== findClosestRecursively(element, "turbo-frame:not([disabled])")) {
+      return false;
+    }
```

I'd appreciate if you could let me know what you think and I'd he happy to submit a PR if you think this change makes sense or if there is a different way of solving my problem.

Issue #963, opened 2023-08-17
Preloader does not use FetchRequest and therefore does not trigger events


sascha-egerer:
When using [`data-turbo-preload`](https://turbo.hotwired.dev/handbook/drive#preload-links-into-the-cache) the FetchRequest implementation is not used but a native fetch(). See https://github.com/hotwired/turbo/blob/c0b6bb6b321e918fc062ba7c5df935fe859a6899/src/core/drive/preloader.ts#L45

So the preload function will not trigger any events and is therefore useless in some cases e.g. if fetchOptions are set in `turbo:before-fetch-request`

Expected behavior: Preload function should trigger fetch events like `turbo:before-fetch-request`

seanpdoyle:
Thank you for opening this issue! I've opened https://github.com/hotwired/turbo/pull/1034 to implement this behavior, and I've opened https://github.com/hotwired/turbo-site/pull/151 to document it.

PR #1034, opened 2023-10-12
Dispatch `turbo:before-fetch-{request,response}` during preloading

seanpdoyle:
Closes [#963][]

Replace the raw call to `fetch` with a new `FetchRequest` instance that treats the `Preloaded` instances as its delegate. During that request's lifecycle, dispatch the `turbo:before-fetch-request` and `turbo:before-fetch-response` events with the `<a>` element as its target.

Prepare the request with the [Sec-Purpose][] header in the `prepareRequest` delegate callback.

Write to the snapshot cache from within the
`requestSucceededWithResponse` delegate callback.

[#963]: https://github.com/hotwired/turbo/issues/963
[Sec-Purpose]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Sec-Purpose#prefetch

### DIFF
--- a/src/core/drive/preloader.js
+++ b/src/core/drive/preloader.js
@@ -1,5 +1,5 @@
 import { PageSnapshot } from "./page_snapshot"
-import { fetch } from "../../http/fetch"
+import { FetchMethod, FetchRequest } from "../../http/fetch_request"
 
 export class Preloader {
   selector = "a[data-turbo-preload]"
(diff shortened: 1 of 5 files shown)


seanpdoyle:
@afcapel this is rebased off #1033 and ready for review.

afcapel:
Makes sense, thanks @seanpdoyle 👍

Issue #967, opened 2023-09-01
Update the html lang attribute


jmsche:
Hi,

In an application, I have the `lang` attribute set on the  `<html>` tag.
When the user switches the locale, the body is changed with Turbo.
However, the `lang` attribute is not updated by Turbo, and some of my JavaScript relies on this attribute to use the correct locale.

I know that I could either add an event listener listening to a Turbo event to update the attribute, or trigger a full page refresh instead.

Would it make sense for Turbo to also update the `lang` attribute by default instead of adding custom logic?

seanpdoyle:
Turbo should probably monitor changes to the `<html>` element's `[lang]` attribute.

Until that support is baked in, applications that drive to pages with different `html[lang]` attributes should disable Turbo Drive for those mechanisms by annotating those `<a>` and `<form>` elements with `[data-turbo="false"]`, or by [requesting a full-page reload](https://turbo.hotwired.dev/handbook/drive#ensuring-specific-pages-trigger-a-full-reload) by including `<meta name="turbo-visit-control" content="reload">` in the response.

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/1035 to implement this behavior.

sedubois:
I came here after encountering this exact issue: the document's `lang` attribute stays stuck in the previous language when switching language. So for example this kind of CSS doesn't work until the page is reloaded: `html[lang="fr"] q { quotes: "« " " »"; }`.

Thank you so much @seanpdoyle for the PR.

PR #1035, opened 2023-10-12
Set `html[lang]` during navigation

seanpdoyle:
Closes [#967][]

Change the `<html>` element's `[lang]` attribute during navigation. Currently, this isn't possible from applications (without a lot of trouble), since the [turbo:before-render][] event is scoped to the `newBody`.

In order to support this behavior, change the `PageSnapshot` constructor to accept an instance of [HTMLHtmlElement][] (the `<html>` element) so that that element can be accessible from the `PageRenderer`.

[#967]: https://github.com/hotwired/turbo/issues/967
[turbo:before-render]: https://turbo.hotwired.dev/reference/events
[HTMLHtmlElement]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLHtmlElement

### DIFF
--- a/src/core/drive/page_renderer.js
+++ b/src/core/drive/page_renderer.js
@@ -29,6 +29,7 @@ export class PageRenderer extends Renderer {
   }
 
   async prepareToRender() {
+    this.#setLanguage()
     await this.mergeHead()
   }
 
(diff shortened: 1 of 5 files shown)


seanpdoyle:
@afcapel thank you for merging this! I wonder if it's worth merging other (or all?) attributes as well.

Issue #962, opened 2023-08-16
Make progress bar finish on `turbo:load`


fritzmg:
We have one page that takes not very long to transmit (low TTFB) but very long to render in the browser. This provides confusing UX when using Turbo as the progress bar indicates that the page has finished loading, but the content isn't actually shown for a long time after that.

Is there any way to make the progress bar actually finish when the `turbo:load` event fires? Rather than when `turbo:before-render` fires (which is roughly what happens now).

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/1036 to change this behavior.

PR #1036, opened 2023-10-12
Hide Progress bar on `turbo:load`

seanpdoyle:
Closes [#962][]

Prior to this commit, the progress bar is hidden between the Turbo Drive visits' `turbo:before-fetch-response` event and the `turbo:render` event.

This commit changes that behavior to hide the bar when the Visit is complete (some time prior to the `turbo:load` event).

[#962]: https://github.com/hotwired/turbo/issues/962


afcapel:
Thanks Sean!

Issue #929, opened 2023-06-02
`broadcast_remove_to` conflicts with input field named "remove"


Tim-Blokdijk:
**Description**:

In my Ruby on Rails application utilizing Hotwire, I'm encountering a conflict when using the `broadcast_remove_to` function on a form element that contains an input field with the name attribute set to "remove".

It seems that, when `broadcast_remove_to` is called, instead of executing the native `remove()` JavaScript method on the target form element, JavaScript instead returns the input field with the name attribute "remove". This behavior seems to occur because JavaScript treats the name attributes of form elements as properties of the form element. 

The error message encountered in this case is `TypeError: e.remove is not a function`, which is not sufficiently informative to understand the problem.

**Steps to Reproduce**:

1. Create an HTML form element with an identifiable ID in a Turbo-enabled Rails view.

```ruby
form_with(model: [a, b, c], method: :delete, html: {class: 'profile', id: dom_id(profile) + '_show'}) do |form|
  form.submit 'X', name: 'remove', class: 'delete button', data: { turbo_confirm: 'Are you sure that you want to remove this?' }
end
```

2. In the corresponding Rails controller, call `broadcast_remove_to` on the form element after an action such as a `destroy` action.

**Expected Behavior**:

The `broadcast_remove_to` function should call the native `remove()` method on the targeted form element, removing it from the DOM.

**Actual Behavior**:

The `broadcast_remove_to` function, when processed by the JavaScript Turbo library, instead returns the input field with the `name="remove"`, and fails to remove the targeted form element. An error "TypeError: e.remove is not a function" is thrown which doesn't clearly indicate the source of the problem.

**Proposed Solution**:

1. Consider modifying how Turbo handles named form elements to prevent this naming conflict. 
2. Provide a clear warning or error message when such a conflict occurs, particularly in development, staging, and test environments.

**Additional Information**:

This behavior was observed in both Chrome and Firefox on up-to-date versions as of May 2023.

seanpdoyle:
Thank you for opening this issue and sharing all of these details.

I've opened https://github.com/hotwired/turbo/pull/1037. It adds test coverage, but doesn't yet resolve the issue.

I'm not sure if there's a predictable way to work around the underlying problem. One thing we could do is detect a collision (with something like `typeof e.remove === "function"`), then log a warning and no-op.

[...]

PR #880, opened 2023-02-23
Use getAttribute to find duplicate ids, not the id property

ethancrawford:
Since form controls can be accessed in javascript directly by name through the form object, (e.g. `my_form.my_input`), `c.id` may end up referring to an _element_ whose name is `id`, instead of referring to the global id _attribute_. Explicitly retrieving the id with `getAttribute` avoids this.

Fixes #879

PR #1037, opened 2023-10-12
Handle `<turbo-stream>` name collisions with `<form>` controls

seanpdoyle:
Closes [#929][]

At the very least, we should provide guidance based on MDN's [Issues with Naming Elements][]:

> Some names will interfere with JavaScript access to the form's
> properties and elements.
>
> For example:
>

> `<input name="id">` will take precedence over `<form id="…">`. This
> means that `form.id` will not refer to the form's `id`, but to the
> element whose `name` is `"id"`. This will be the case with any other
> form properties, such as `<input name="action">` or `<input
> name="post">`. `<input name="elements">` will render the form's
> `elements` collection inaccessible. The reference `form.elements` will
> now refer to the individual element.
>
> To avoid such problems with element names:
>

> * Always use the `elements` collection to avoid ambiguity between an
>   element `name` and a form property.
> * Never use `"elements"` as an element name.

To guard against future regressions, this commit adds coverage for the `stream.html` fixture page by rendering a `<form>` element with `<input type="hidden">` elements whose names match each `StreamActions` function.

[#929]: https://github.com/hotwired/turbo/issues/929
[Issues with Naming Elements]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#issues_with_naming_elements

### DIFF
--- a/src/tests/functional/stream_tests.js
+++ b/src/tests/functional/stream_tests.js
@@ -182,6 +182,92 @@ test("test receiving a remove stream message preserves focus blurs the activeEle
   assert.notOk(await hasSelector(page, ":focus"))
 })
 
+test("test receiving an [action=after] targeting a form", async ({ page }) => {
+  await page.evaluate(() =>
+    window.Turbo.renderStreamMessage(`
+      <turbo-stream action="after" target="form_with_action_names">
+        <template><p id="after_form_with_action_names">After</p></template>
+      </turbo-stream>
+    `)
+  )
+
+  assert.equal(await page.textContent("#after_form_with_action_names"), "After")
+})
+
+test("test receiving an [action=append] targeting a form", async ({ page }) => {
+  await page.evaluate(() =>
+    window.Turbo.renderStreamMessage(`
+      <turbo-stream action="append" target="form_with_action_names">
+        <template>Append</template>
+      </turbo-stream>
+    `)
+  )
+  await nextBeat()
+
+  assert.include(await page.textContent("#form_with_action_names"), "Append")
+})
+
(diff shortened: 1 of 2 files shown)


seanpdoyle:
To re-iterate https://github.com/hotwired/turbo/issues/929#issuecomment-1760395982:

I'm not sure if there's a predictable way to work around the underlying problem. One thing we could do is detect a collision (with something like `typeof e.remove === "function"`), then log a warning and no-op.

It seems that any other counter-measure is kicking the can down the road. For example, `e.remove()` could be replaced by `e.parentElement.removeChild(e)`, but if there's a form field named `parentElement`, we're back to square one. While it's unlikely, there's nothing preventing a re-ordering/sorting `<form>` element from having `<input name="parentElement">`.

src/tests/functional/stream_tests.js
```diff
@@ -182,6 +182,92 @@ test("test receiving a remove stream message preserves focus blurs the activeEle
...
+      <turbo-stream action="update" target="form_with_action_names">
+        <template><input name="updated_field"></template>
+      </turbo-stream>
+    `)
+  )
+  await nextBeat()
+
+  assert.ok(await hasSelector(page, "#form_with_action_names [name=updated_field]"))
+})
+
```
seanpdoyle, line 270: These tests should probably be covered by unit tests, like in https://github.com/hotwired/turbo/pull/880/files#diff-f9df76d4df4de3858e7298e7f36d9e3acc8e7f07af6759c4f66c80bcc309b465.

brunoprietog:
#880 is trying to solve this too

Issue #901, opened 2023-03-29
Turbo Frame on post submit not updating with _top


jonsgreen:
I am not sure if this is a bug or expected behavior but I have found that if I submit a form with `data-turbo-frame="_top"` that the page does not seem to update while if I set it to a target with an id then it will update the matching frame.  

It could easily be that I am doing something incorrect but I thought I would first ask whether this is an expected limitation with form submits.

seanpdoyle:
Thank you for opening this issue.

I've opened https://github.com/hotwired/turbo/pull/1038. I did my best to faithfully reproduce the behavior you're describing here. 

Unfortunately, I wasn't able to reproduce it. Submitting `<form>` elements with `[data-turbo-frame="_top"]` from inside and outside `<turbo-frame>` elements appeared to behave as expected.

I'd really appreciate feedback on https://github.com/hotwired/turbo/pull/1038, particularly the [structural changes to the `form.html` fixture](https://github.com/hotwired/turbo/pull/1038/files#diff-38acdd502c97ee1a1140d761a08c7e07e46c9f266b4efd0c55d0d5b986517158). Do they recreate HTML similar to yours?

fluxsaas:
@jonsgreen @jonsgreen 

i know this is long been closed, but i stumbled into the same situation, i could track this down to following:

* having a rails controller for Comment
* after POST to create a Comment, redirect to another resource e.g. comment.commentable

i change the controller code to include `status: :see_other`

change the controller from:

`format.html { redirect_to comment.commentable, notice: 'Comment have been successfully created.' }`

to:

`format.html { redirect_to comment.commentable, notice: 'Comment have been successfully created.', status: :see_other }`

which solved my problem.

fluxsaas:
@jonsgreen @seanpdoyle 

i know this is long been closed, but i stumbled into the same situation, i could track this down to following:

* having a rails controller for Comment
* after POST to create a Comment, redirect to another resource e.g. comment.commentable

i change the controller code to include `status: :see_other`

this works:

`format.html { redirect_to comment, notice: 'Comment have been successfully created.' }`

this does not work:

`format.html { redirect_to comment.commentable, notice: 'Comment have been successfully created.' }`

PR #1038, opened 2023-10-12
Add test coverage for submitting `form[data-turbo-frame="_top"]`

seanpdoyle:
Closes [#901][]

Add test coverage for submitting `<form>` elements with `[data-turbo-frame="_top"]`. Coverage includes both `<form>` elements outside of `<turbo-frame>` *and* `<form>` elements that are descendants of `<turbo-frame>` elements.

[#901]: https://github.com/hotwired/turbo/issues/901


afcapel:
Thanks again, @seanpdoyle

Issue #884, opened 2023-02-28
Progress Bar finishes prematurely


tropperstyle:
I have backend server that streams responses to the client as soon as possible, so the browser can start painting while a data is still pending.

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8">
    <title>HTML Streaming</title>
    <link rel="stylesheet" href="style.css">
    <script src="script.js"></script>
  </head>
  <body>
    <!-- Paused here, waiting for data -->
```

So even if `await fetch()` returns, we still only have the response headers, but are still waiting for the body to complete. Turbo's progress bar is really jarring in this case. It quickly jumps to 100%, then there is a long delay before the page actually navigates.

I added `await` [here](https://github.com/hotwired/turbo/blob/main/src/http/fetch_request.ts#L134) which fixes my use-case, but it introduced a failing test which I'm not sure what to do with:

```
  1) [chrome] › functional/frame_tests.ts:594:1 › test navigating pushing URL state from a frame navigation fires events 

    AssertionError: sets aria-busy on the <html>: expected null to equal 'true'

      606 |   assert.notOk(await nextAttributeMutationNamed(page, "frame", "aria-busy"), "removes aria-busy from the <turbo-frame>")
      607 |
    > 608 |   assert.equal(await nextAttributeMutationNamed(page, "html", "aria-busy"), "true", "sets aria-busy on the <html>")
          |          ^
      609 |   await nextEventOnTarget(page, "html", "turbo:before-visit")
      610 |   await nextEventOnTarget(page, "html", "turbo:visit")
      611 |   await nextEventOnTarget(page, "html", "turbo:before-cache")
```

seanpdoyle:
@tropperstyle thank you for opening this issue.

In the time since you've opened it, this project has undergone significant structural changes, and files have been renamed.

The "here" that you link to in your description references a file that no longer exists. The code referenced by `src/http/fetch_request.ts` now resides in [src/http/fetch_request.js](https://github.com/hotwired/turbo/blob/main/src/http/fetch_request.js).

Could you re-share the line number that you've prefixed with `await`?

tropperstyle:
Sure, I also updated my previous comment: https://github.com/hotwired/turbo/blob/v7.3.0/src/http/fetch_request.ts#L134

seanpdoyle:
Thank you! I've opened https://github.com/hotwired/turbo/pull/1039. I'd appreciate review. Could we move conversation to that PR?

[...]

PR #1039, opened 2023-10-13
`await` for `FetchRequest` delegate to handle response

seanpdoyle:
Closes [#884]

Despite the fact that `Visit` and `FrameController` are `FetchRequest` delegate classes that implement their delegate methods asynchronously on success ([Visit.requestSucceededWithResponse][] and [FrameController.requestSucceededWithResponse][]) and on failure ([Visit.requestFailedWithResponse][]) and
[FrameController.requestFailedWithResponse][]), the `async FetchRequest.receive` method doesn't use the `await` keyword to wait for those delegates to finish handling their callbacks.

This commit adds those `await` clauses.

[Visit.requestSucceededWithResponse]: https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/drive/visit.js#L297
[FrameController.requestSucceededWithResponse]: https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/frames/frame_controller.js#L210
[Visit.requestFailedWithResponse]: https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/drive/visit.js#L311
[FrameController.requestFailedWithResponse]: https://github.com/hotwired/turbo/blob/c207f5b25758e4a084e8ae42e49712b91cf37114/src/core/frames/frame_controller.js#L215

[#884]: https://github.com/hotwired/turbo/issues/884


seanpdoyle:
I'm hopeful that this resolves the underlying problems mentioned by #884. I'm not sure how to add test coverage for this. With TypeScript, we'd be able to enforce the `FetchRequest` delegate's interface, and have type hints to know that these methods are asynchronous and return `Promise` instances.

In the absence of that type checking, I'd really appreciate some guidance on how to add coverage to the test suite to guard against future regressions.

src/http/fetch_request.js
```diff
@@ -124,7 +124,7 @@ export class FetchRequest {
     try {
       this.delegate.requestStarted(this)
       const response = await fetch(this.url.href, fetchOptions)
-      return await this.receive(response)
```
armstrjare, line 127: Should this `await` remain so that the `finally` block calling `this.delegate.requestFinished(this)` is deferred until the response has been processed/rendered?

As it stands on main branch, and with your changes too, it gets called prior to the response being downloaded/processed.

armstrjare:
I've added a comment on a suggested tweak that seems to resolve a related but different issue I've been having where the turbo-frame busy attribute is removed prior to the frame being rendered, and there can be significant delay between this when there is a high latency connection and/or a large response body on the request.

I'm not quite sure on the best approach to test this in this project because my initial thought was to assert the timing/ordering of related events and DOM mutations to be as expected - but it doesn't seem trivial with the helper methods currently provided since these behaviours are tracked separately.

src/http/fetch_request.js
seanpdoyle: Great catch! I've reverted that.

src/core/drive/visit.js
```diff
@@ -159,7 +159,7 @@ export class Visit {
       this.simulateRequest()
     } else if (this.shouldIssueRequest() && !this.request) {
       this.request = new FetchRequest(this, FetchMethod.get, this.location)
-      this.request.perform()
+      await this.request.perform()
```
brunoprietog, line 162: Wouldn't this require that `issueRequest` be asynchronous? And then the entire preceding calls as well?

ggmueller:
I am seeing a similar problem as @armstrjare.

I have tested your update but it does not solve the issue for me. The busy attribute still is reset too early.

I am using a form to update a turbo-frame like this:

```html
<form name="filter" method="get" data-turbo-frame="skill-results" class="mb-6">
                                <input type="text" id="search" x-on:input.debounce="$event.target.form.requestSubmit()"  placeholder="Suchen" name="SearchText" value="">
</form>
<turbo-frame id="skill-results" src="https://localhost:5001/Skills?SearchText=" complete="">
</turbo-frame>
```

The busy and aria-busy attributes are reset, when the response returns (before it is downloaded). It can be followed when throttling the network connection.

I suspect that, additionally to your fixes,  also in frame_controller.js `formSubmissionSucceededWithResponse` has to be made `async`, and the call to `frame.delegate.loadResponse()` should be awaited.

seanpdoyle:
@ggmueller I've pushed a commit that aims to address the `<form>`-based issues you've outlined in https://github.com/hotwired/turbo/pull/1039#issuecomment-1877341515.

Are you able to try and test this branch again?

PR #1040, opened 2023-10-17
Add control to preserve scroll during Drive visits

seanpdoyle:
Extracted from [#1019][]

Add support for preserving scroll changes from page to page. Controlled by the presence of the `<meta name="turbo-scroll">` element in the `<head>`.

To preserve:

```html
<meta name="turbo-scroll" content="preserve">
```

To reset:

```html
<meta name="turbo-scroll" content="reset">
```

[#1019]: https://github.com/hotwired/turbo/pull/1019

Co-authored-by: Alberto Fernández-Capel <afcapel@gmail.com>
Co-Authored-By: Jorge Manrubia <jorge@hey.com>

### DIFF
--- a/src/core/drive/page_snapshot.js
+++ b/src/core/drive/page_snapshot.js
@@ -73,6 +73,10 @@ export class PageSnapshot extends Snapshot {
     return this.headSnapshot.getMetaValue("view-transition") === "same-origin"
   }
 
+  get shouldPreserveScrollPosition() {
+    return this.getSetting("scroll") === "preserve"
+  }
+
   // Private
 
   getSetting(name) {
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -335,7 +335,7 @@ export class Visit {
   // Scrolling
 
   performScroll() {
-    if (!this.scrolled && !this.view.forceReloaded) {
+    if (!this.scrolled && !this.view.forceReloaded && !this.view.snapshot.shouldPreserveScrollPosition) {
       if (this.action == "restore") {
         this.scrollToRestoredPosition() || this.scrollToAnchor() || this.view.scrollToTop()
       } else {
--- a/src/tests/fixtures/scroll_restoration/preserve.html
+++ b/src/tests/fixtures/scroll_restoration/preserve.html
@@ -2,6 +2,7 @@
 <html>
   <head>
     <meta charset="utf-8">
+    <meta name="turbo-scroll" content="preserve">
     <title>Scroll Restoration</title>
     <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
     <script src="/src/tests/fixtures/test.js"></script>
--- a/src/tests/fixtures/scroll_restoration/reset.html
+++ b/src/tests/fixtures/scroll_restoration/reset.html
@@ -2,6 +2,7 @@
 <html>
   <head>
     <meta charset="utf-8">
+    <meta name="turbo-scroll" content="reset">
     <title>Scroll Restoration</title>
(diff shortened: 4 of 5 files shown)


seanpdoyle:
@jorgemanrubia in https://github.com/hotwired/turbo/pull/1019#discussion_r1363197762, you mentioned skepticism related to making scroll preservation configurable:

> Just like morphing, scroll preservation for regular navigation could make sense in other scenarios, but we believe those will be quite rare and are out of the scope of what we're addressing here. I'd like to keep both "scroll" and "morph" tied to "page refreshes" for now, as I think it makes for a more cohesive development story.

This PR has the potential to provide a solution for https://github.com/hotwired/turbo/issues/37 that draws inspiration from other `<meta>` element-enabled configurations. Has your perspective changed since this was last proposed?

src/core/drive/page_snapshot.js
```diff
@@ -83,6 +83,10 @@ export class PageSnapshot extends Snapshot {
     return this.getSetting("refresh-scroll")
   }
 
+  get visitScroll() {
+    return this.getSetting("visit-scroll")
+  }
```
seanpdoyle: The current `<meta name="turbo-scroll">` is meant to parallel the concept of a "refresh". 

Originally, this PR proposed `<meta name="turbo-scroll">`. An alternative couldbe `<meta name="turbo-render-scroll">`. Are either of those preferable?

src/core/drive/page_view.js
```diff
@@ -60,7 +60,11 @@ export class PageView extends View {
   }
 
   shouldPreserveScrollPosition(visit) {
-    return this.isPageRefresh(visit) && (visit?.refresh?.scroll || this.snapshot.refreshScroll) === "preserve"
+    if (this.isPageRefresh(visit)) {
+      return (visit?.refresh?.scroll || this.snapshot.refreshScroll) === "preserve"
+    } else {
+      return this.snapshot.visitScroll === "preserve"
```
tvararu: This introduces a subtle bug when using the back button.

Without this change, the back button restores scroll position; with this change it resets scroll to the top.

It's tied to `performScroll`:

```js
performScroll() {
  if (!this.scrolled && !this.view.forceReloaded && !this.view.shouldPreserveScrollPosition(this)) {
    if (this.action == "restore") {
      this.scrollToRestoredPosition() || this.scrollToAnchor() || this.view.scrollToTop()
    } else {
      this.scrollToAnchor() || this.view.scrollToTop()
    }

    this.scrolled = true
  }
}
```

Before:

0. On Page A
1. Go to Page B
2. Hit Back button
3. Trigger `performScroll`:
  `this.scrolled` false
  `this.view.forceReloaded` false
  `this.view.shouldPreserveScrollPosition(this)` false
4. All false -> Triggers `this.action === "restore"`
5. Scroll position restored

After:

0. On Page A: Has `visitScroll === "preserve"`
1. Go to Page B
2. Hit Back button
3. Trigger `performScroll`:
  `this.scrolled` false
  `this.view.forceReloaded` false
  `this.view.shouldPreserveScrollPosition(this)` **true** ⚠️ because `this.snapshot.visitScroll === "preserve"`
4. Not all false -> Doesn't trigger `this.action === "restore"`
5. Scroll position not restored

## Possible solution

Not sure if it's the cleanest, but I've added a guard to the top of `shouldPreserveScrollPosition`:

```js
shouldPreserveScrollPosition(visit) {
  if (visit?.action === "restore") return false
  if (this.isPageRefresh(visit)) {
    return (visit?.refresh?.scroll || this.snapshot.refreshScroll) === "preserve"
  } else {
    return this.snapshot.visitScroll === "preserve"
  }
}
```

PR #1042, opened 2023-10-20
replace nextAnimationFrame by nextMicroTask

michelson:
DOM operations on turbostream are queued when the user is not in the browser tab or the tab is not active. This fixes that.
the problem resides when we call `await nextAnimationFrame()` because browsers pause the execution to save resources on the inactive browsers tab. The problem is that for chat applications or other kinds of notifications having a way to update things in the foreground is key.
**more context at:**
ref https://github.com/hotwired/turbo-rails/issues/338

### DIFF
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -1,13 +1,11 @@
 import { Session } from "./session"
-import { Cache } from "./cache"
 import { PageRenderer } from "./drive/page_renderer"
 import { PageSnapshot } from "./drive/page_snapshot"
 import { FrameRenderer } from "./frames/frame_renderer"
 import { FormSubmission } from "./drive/form_submission"
 
 const session = new Session()
-const cache = new Cache(session)
-const { navigator } = session
+const { cache, navigator } = session
 export { navigator, session, cache, PageRenderer, PageSnapshot, FrameRenderer }
 
 export { StreamActions } from "./streams/stream_actions"
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -1,11 +1,13 @@
 import { Session } from "./session"
+import { Cache } from "./cache"
 import { PageRenderer } from "./drive/page_renderer"
 import { PageSnapshot } from "./drive/page_snapshot"
 import { FrameRenderer } from "./frames/frame_renderer"
 import { FormSubmission } from "./drive/form_submission"
 
 const session = new Session()
-const { cache, navigator } = session
+const cache = new Cache(session)
+const { navigator } = session
 export { navigator, session, cache, PageRenderer, PageSnapshot, FrameRenderer }
 
 export { StreamActions } from "./streams/stream_actions"
--- a/src/elements/stream_element.js
+++ b/src/elements/stream_element.js
@@ -1,5 +1,5 @@
 import { StreamActions } from "../core/streams/stream_actions"
-import { nextEventLoopTick, nextAnimationFrame } from "../util"
+import { nextAnimationFrame } from "../util"
 
 // <turbo-stream action=replace target=id><template>...
 
@@ -43,7 +43,7 @@ export class StreamElement extends HTMLElement {
       const event = this.beforeRenderEvent
 
       if (this.dispatchEvent(event)) {
-        await window.Turbo.session.useNextAnimationFrame ? nextAnimationFrame() : nextEventLoopTick()
+        await nextAnimationFrame()
         await event.detail.render(this)
       }
     })())
--- a/src/elements/stream_element.js
+++ b/src/elements/stream_element.js
@@ -1,5 +1,5 @@
 import { StreamActions } from "../core/streams/stream_actions"
-import { nextMicrotask } from "../util"
+import { nextEventLoopTick } from "../util"
 
 // <turbo-stream action=replace target=id><template>...
 
@@ -43,7 +43,7 @@ export class StreamElement extends HTMLElement {
       const event = this.beforeRenderEvent
 
       if (this.dispatchEvent(event)) {
-        await nextMicrotask()
+        await nextEventLoopTick()
         await event.detail.render(this)
       }
     })())
--- a/src/elements/stream_element.js
+++ b/src/elements/stream_element.js
@@ -1,5 +1,5 @@
 import { StreamActions } from "../core/streams/stream_actions"
-import { nextEventLoopTick } from "../util"
+import { nextEventLoopTick, nextAnimationFrame } from "../util"
 
 // <turbo-stream action=replace target=id><template>...
 
@@ -43,7 +43,7 @@ export class StreamElement extends HTMLElement {
       const event = this.beforeRenderEvent
 
       if (this.dispatchEvent(event)) {
-        await nextEventLoopTick()
+        await window.Turbo.session.useNextAnimationFrame ? nextAnimationFrame() : nextEventLoopTick()
         await event.detail.render(this)
       }
     })())
--- a/src/elements/stream_element.js
+++ b/src/elements/stream_element.js
@@ -43,7 +43,7 @@ export class StreamElement extends HTMLElement {
       const event = this.beforeRenderEvent
 
       if (this.dispatchEvent(event)) {
-        nextRepaint()
+        await nextRepaint()
         await event.detail.render(this)
       }
     })())
--- a/src/elements/stream_element.js
+++ b/src/elements/stream_element.js
@@ -1,5 +1,5 @@
 import { StreamActions } from "../core/streams/stream_actions"
-import { nextAnimationFrame } from "../util"
+import { nextEventLoopTick, nextAnimationFrame } from "../util"
 
 // <turbo-stream action=replace target=id><template>...
 
@@ -43,7 +43,7 @@ export class StreamElement extends HTMLElement {
       const event = this.beforeRenderEvent
 
       if (this.dispatchEvent(event)) {
-        await nextAnimationFrame()
+        await (window.Turbo.session.useNextAnimationFrame ? nextAnimationFrame() : nextEventLoopTick())
         await event.detail.render(this)
       }
     })())
--- a/src/elements/stream_element.js
+++ b/src/elements/stream_element.js
@@ -129,13 +129,6 @@ export class StreamElement extends HTMLElement {
   }
 
   /**
-   * Gets the current action.
-   */
-  get usesNextAnimationFrame() {
-    return this.getAttribute("waits_for_next_frame") !== "false"
-  }
-
-  /**
    * Gets the current target (an element ID) to which the result will
    * be rendered.
    */
--- a/src/elements/stream_element.js
+++ b/src/elements/stream_element.js
@@ -1,5 +1,5 @@
 import { StreamActions } from "../core/streams/stream_actions"
-import { nextEventLoopTick } from "../util"
+import { nextRepaint } from "../util"
 
 // <turbo-stream action=replace target=id><template>...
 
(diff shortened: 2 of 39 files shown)


brunoprietog:
This fixes #920.

Hopefully this will be taken into account in #1019, as mentioned by @seanpdoyle in [this comment](https://github.com/hotwired/turbo/pull/1019#discussion_r1347293771).

afcapel:
This is subtly changing behavior because `nextAnimationFrame` and `nextMicrotask` have some other differences besides how the browser chooses to schedule them when the tab is not active. When you run something after `requestAnimationFrame` you are guaranteed that the event loop has run its UI rendering part, while putting a callback in the microtask queue executes it within the same event loop tick, before any UI changes have rendered.

To illustrate this point see this code:

```html
<!DOCTYPE html>
<html lang="en">
<head>
  <link rel="icon" href="data:,">
  <style>
    #box {
      width: 100px;
      height: 100px;
      background-color: blue;
      transition: all 1s;
    }
  </style>
  <script type="text/javascript">
    function nextAnimationFrame() {
      return new Promise((resolve) => requestAnimationFrame(() => resolve()))
    }

    function nextEventLoopTick() {
      return new Promise((resolve) => setTimeout(() => resolve(), 0))
    }

    function nextMicrotask() {
      return Promise.resolve()
    }

    document.addEventListener("DOMContentLoaded", async () => {
      const box = document.getElementById("box")

      box.style.transform = "translateX(500px)"
      await nextAnimationFrame()
      box.style.transform = "translateX(0px)"
    })
  </script>
</head>
<body>
  <div id="box"></div>
</body>
</html>
```

This animates a box moving from right to left.

https://github.com/hotwired/turbo/assets/150107/3cf230cf-e012-48ad-939f-a56e7cd2f88b

However replacing `nextAnimationFrame` with  `nextMicrotask` breaks the animation. The JS code sets `translateX(500px)` _and_ `translateX(0px)` _before_ the UI updates render, and so the property doesn't change between rendering steps and there's no animation.

The current uses of `nextAnimationFrame` are intended to let the browser render UI updates before carrying on with further DOM transformations and I'm wary that changing them can introduce subtle bugs.

However, I'd be interested to see an easy way to reproduce `nextAnimationFrame` not triggering to see if we can solve the issue in some other way.

michelson:
hi @afcapel , Thanks for your answer. I've tried your example and put it in codesandbox.
https://codesandbox.io/s/jolly-kare-yq5mmd?file=/index.html , I've also replaced the nextMicroask to nextEventLoopTick in your example, and the animation is executed correctly.

I've implemented this in this PR by replacing the nextMicroTask with nextEventLoopTick.

Unit tests are passing. But note that I had to keep the nextMicroTask on the stream_element.js; not sure why it failed when used with the nextEventLoopTick. Also, I don't know how to properly run the browser specs on localhost as they keep retrying. Do you have any suggestions on how to run it properly?
 
Screen Shot 2023-10-24 at 12 31 49 PM

afcapel:
@michelson sorry, I'm a wary of changing this behaviour because I find the possible repercussions very difficult to foresee.

Turbo already works well simplifying most simple use cases, but this one is particular enough that you may have to reach to lower level APIs. I think you're best chance is to reach to ActionCable to deliver the new message notifications. 

You can reuse the same WebSocket connection that Turbo opens with the server, and do anything you want when a new message arrives, even if your tab is in the background.

This is how we do it in in one of our apps:

```js
// consumer.js

import { cable } from "@hotwired/turbo-rails"
export default await cable.getConsumer() // this reuses the same websocket that Turbo uses for updates

// typing_notifications_channel.js

import consumer from "channels/consumer"

// Subscribe to a custom channel
export function subscribeToTypingNotificationsChannel(roomId, callbacks) {
  const channel = { channel: "TypingNotificationsChannel", room_id: roomId }
  return consumer.subscriptions.create(channel, callbacks)
}

// typing_notifications_controller.js

import { subscribeToTypingNotificationsChannel } from "channels/typing_notifications_channel"
import { pageIsTurboPreview } from "helpers/turbo_helpers"

// Reacts to new messages on the `TypingNotificationsChannel`
  connect() {
    if (!pageIsTurboPreview()) {
      this.channel = subscribeToTypingNotificationsChannel(this.roomIdValue, { received: this.#received.bind(this) })
    }
  }

  ...

  #received(data) {
    ... // handle received data
  }
}
```

Hope that helps and you can migrate away from React.

michelson:
Hey @afcapel, We've successfully used the fork in our near-complete app port without issues, Our Turbo fork have just three failing browser specs. I'm addressing these shortly as I'm trying to understand how this is implemented. 
However, closing this impacts us significantly. Could you revisit the decision? I believe many apps, especially with chat functionalities, will be constrained without this feature. Isn't such reactivity expected by default? Is this a design choice or an oversight?

```
  **3 failed**
    [chrome] › functional/loading_tests.js:100:1 › test reloading a frame reloads the content ======
    [firefox] › functional/frame_tests.js:385:1 › test submitting a form that redirects to a page with a <turbo-frame recurse> which lazily loads a matching frame 
    [firefox] › functional/loading_tests.js:100:1 › test reloading a frame reloads the content =====

  **1 flaky**
    [firefox] › functional/frame_tests.js:371:1 › test following a link to a page with a <turbo-frame recurse> which lazily loads a matching frame 
 
```

michelson:
This are certain use cases might be affected or limited by this approach:

1. **Immediate Feedback**: In scenarios where immediate feedback is crucial, even a slight delay introduced by `nextAnimationFrame` can be noticeable. For instance, real-time collaborative tools or games might need instantaneous reactions.

2. **Rapid Successive Updates**: If there are rapid successive updates to the DOM, `nextAnimationFrame` might cause them to be bundled together. While this is often an optimization, there are scenarios where you might want updates to render as they come in, rather than waiting.

Chat applications have unique challenges and requirements that might be influenced by the use of `nextAnimationFrame` for rendering updates. Here are some points specific to chat applications:

1. **Instant Message Rendering**: In real-time chat applications, messages from other users need to appear immediately once they're received. Any perceptible delay, even if it's minor, can detract from the "real-time" feel of the chat. `nextAnimationFrame` could potentially introduce such a delay.

2. **Message Order**: If multiple messages are received in quick succession, it's crucial that they render in the correct order. The delay introduced by `nextAnimationFrame` could potentially cause issues if messages are processed in rapid succession but render out of order.

3. **Read Receipts**: Many chat applications show when a message has been read by the other user. If the rendering of this status is delayed, it might confuse users or make the application feel less responsive.

4. **Typing Indicators**: "User is typing..." indicators need to be real-time to be effective. If there's a delay in showing or hiding these indicators due to `nextAnimationFrame`, it can lead to miscommunication or confusion.

5. **Notifications and Alerts**: For unread messages or mentions in a chat, instant notifications are crucial. Any delay in rendering these notifications could mean users might miss out on timely information or updates.

Considering these points, while `nextAnimationFrame` can optimize the rendering process for many web applications, chat applications have specific real-time needs that might be affected by such delays.

While these are potential limitations, it's essential to understand that for many applications, especially typical web applications, the use of `nextAnimationFrame` in Turbo-Streams offers a balanced and optimized user experience. However, like all design decisions, it's crucial to consider the specific requirements and constraints of the project at hand.


what if this could be an opt in feature? would you consider a new PR that implements that?

This is an idea to provide an optional configuration setting for Hotwire's Turbo to control whether it uses `nextAnimationFrame` for rendering updates. If the setting is enabled, Turbo would bypass `nextAnimationFrame` for more immediate rendering, especially useful for applications like real-time chats.

Here's a high-level overview of how we could implement this:

1. **Configuration Setting**:
    Add a configuration option in Turbo to control the use of `nextAnimationFrame`.

```javascript
Turbo.config = {
    ...,
    useNextAnimationFrame: true // default is true to maintain current behavior
};
```

2. **Update Turbo's Rendering Logic**:
    In the part of the Turbo code that handles rendering, conditionally bypass `nextAnimationFrame` based on the configuration.

```javascript
if (Turbo.config.useNextAnimationFrame) {
    requestAnimationFrame(() => {
        // current rendering logic
    });
} else {
    // immediate rendering logic
}
```

3. **Expose API to Change Configuration**:
    Developers should be able to change this configuration at runtime based on specific conditions or use-cases.

```javascript
function enableImmediateRendering() {
    Turbo.config.useNextAnimationFrame = false;
}

function enableDefaultRendering() {
    Turbo.config.useNextAnimationFrame = true;
}
```

michelson:
Doing more tests, I believe that the only change needed is in the render function of `StreamElement,` so I'm guessing if the `<turbo-stream action=replace target=id><template>` element could have some indicator to change how it handles its render, like an attribute like: `<turbo-stream render-timing="event-loop">` is that something that you would accept as a PR, @afcapel ?

afcapel:
Not interested in configuration options, but I'm open to a more surgical change. We can try limiting the change to the `StreamElement` render function and using a new timing function that only skips `nextAnimationFrame` if the tab is hidden:

```js
// util.js

export function nextRepaint() {
  if (document.visibilityState === "hidden") {
    return nextEventLoopTick()
  } else {
    return nextAnimationFrame()
  }
}
```

michelson:
Great solution, @afcapel . On it.

src/elements/stream_element.js
```diff
@@ -43,7 +43,7 @@ export class StreamElement extends HTMLElement {
       const event = this.beforeRenderEvent
 
       if (this.dispatchEvent(event)) {
-        await nextEventLoopTick()
+        nextRepaint()
```
seanpdoyle, line 46: Do we need to `await` this? Should `nextRepaint()` return the `next`-prefixed Promise utilities instead of `await`-ing them?

src/tests/unit/stream_element_tests.js
```diff
@@ -34,13 +34,13 @@ test("action=append", async () => {
   assert.equal(subject.find("#hello")?.textContent, "Hello Turbo")
 
   subject.append(element)
-  await nextAnimationFrame()
+  await nextRepaint()
```
seanpdoyle, line 37: Since these are unit tests, will `nextRepaint()` ever invoke `nextEventLoopTick()`, since the `document.visibilityState` won't ever equal `"hidden"`?

src/core/index.js
```diff
@@ -1,11 +1,13 @@
...
+import { Cache } from "./cache"
 import { PageRenderer } from "./drive/page_renderer"
 import { PageSnapshot } from "./drive/page_snapshot"
 import { FrameRenderer } from "./frames/frame_renderer"
 import { FormSubmission } from "./drive/form_submission"
 
 const session = new Session()
-const { cache, navigator } = session
+const cache = new Cache(session)
+const { navigator } = session
```
seanpdoyle, line 10: This appears to be reverting some recent changes made to `main`:

```suggestion
const { cache, navigator } = session
```

src/util.js
```diff
@@ -43,6 +43,14 @@ export function dispatch(eventName, { target, cancelable, detail } = {}) {
   return event
 }
 
+export async function nextRepaint() {
+  if (document.visibilityState === "hidden") {
+    await nextEventLoopTick()
+  } else {
+    await nextAnimationFrame()
+  }
+}
```
seanpdoyle, line 52: Should this utility return the calls made to the other utilities?

```suggestion
export function nextRepaint() {
  if (document.visibilityState === "hidden") {
    return nextEventLoopTick()
  } else {
    return nextAnimationFrame()
  }
}
```

seanpdoyle (approved):

michelson:
Hi @seanpdoyle, thanks for the suggestions. I've applied those already. Also, CI passing at https://github.com/michelson/turbo/actions/runs/6721616946/job/18267725557

afcapel:
Thanks @michelson & @seanpdoyle!

PR #1044, opened 2023-10-24
Do not dispatch multiple visit events on redirect visit

### LINKED ISSUE #877: `turbo:visit` is dispatched twice on redirect
With the following flow:

1. User clicks link to `/one`
2. `turbo:visit` is fired
3. `/one` redirects to `/two`
4. `turbo:visit` is fired again

There's currently no way to accurately determine whether a Visit was triggered by a click, a redirect, or programmatically via `Turbo.visit`. This is useful for animating page transitions: an animation library could animate on the first `turbo:visit`, and ignore the redirect visit.

Perhaps it'd be worth detailing the `reason` in `event.details`, similar to the `reload` event?

domchristie:
https://github.com/hotwired/turbo/pull/328 fixed an issue with the window's location in native adapters (https://github.com/hotwired/turbo-ios/issues/30 & https://github.com/hotwired/turbo-android/issues/128).

However, proposing a new visit re-runs the visit lifecycle again, duplicating events, and introduced issues such as https://github.com/hotwired/turbo/issues/428, https://github.com/hotwired/turbo/issues/526, https://github.com/hotwired/turbo/issues/537 & https://github.com/hotwired/turbo/issues/877.

This pull request reverts the visit proposal, and instead updates the location as part of `changeHistory()`. (This may not have been doable before https://github.com/hotwired/turbo/pull/601.) It leads to a simpler `followRedirect` method that now just updates the visit's location. As such, we no longer need the fix from https://github.com/hotwired/turbo/pull/563.

Fixes #877

### DIFF
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -128,12 +128,9 @@ export class Visit {
     if (this.state == VisitState.started) {
       this.recordTimingMetric(TimingMetric.visitEnd)
       this.state = VisitState.completed
-      this.followRedirect()
 
-      if (!this.followedRedirect) {
-        this.adapter.visitCompleted(this)
-        this.delegate.visitCompleted(this)
-      }
+      this.adapter.visitCompleted(this)
+      this.delegate.visitCompleted(this)
     }
   }
 
@@ -147,6 +144,7 @@ export class Visit {
 
   changeHistory() {
     if (!this.historyChanged && this.updateHistory) {
+      this.followRedirect()
       const actionForHistory = this.location.href === this.referrer?.href ? "replace" : this.action
       const method = getHistoryMethodForAction(actionForHistory)
       this.history.update(method, this.location, this.restorationIdentifier)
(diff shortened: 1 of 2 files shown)


jayohms:
As mentioned here https://github.com/hotwired/turbo/pull/328#issuecomment-1779997759 this approach is not compatible with with the iOS/Android Turbo adapters and the way that the libraries work. Avoiding a double-visit on a redirect won't be trivial, so maybe there's another way to approach the issue you're experiencing with the double-visit + animation. Most importantly, though, the mobile adapters need the opportunity to see the redirect url and start a new visit.

domchristie:
Thanks @jayohms, that definitely clears things up.

I wonder whether there might be another way to model Visit proposals? The Visit class seems to be used as a way to trigger a Visit lifecycle, but this can mean different things in different contexts. For example:

1. A standard Visit that navigates from one location to the next
2. A same-page anchor Visit
3. A redirected Visit
4. A Frame Visit (promoted to a Visit using `data-turbo-action`

Each has different expectations and outcomes, which hints that they could be different classes that share the same API

domchristie:
Alternatively: https://github.com/hotwired/turbo/compare/main...domchristie:silence_non_rendering_visits?expand=1

PR #1026, opened 2023-10-08
Delegate StreamActions.refresh to Session 

seanpdoyle:
The bulk of the `StreamActions.refresh` implementation was reaching
through the global `window.Turbo` property, which itself was reaching
through either global variables or the `Session`.

This commit moves the implementation out of the `StreamActions` and into
a new `Session.refresh(url, requestId)` method. With that change, all
property access is encapsulated within the `Session`.

To support that change, this commit also introduces the
`StreamElement.requestId` property to read the `[request-id]` attribute.

PR #1049, opened 2023-10-27
Remove Circular Build Dependency

seanpdoyle:
Resolves https://github.com/hotwired/turbo/pull/1026#issuecomment-1782605125


afcapel:
Thanks!

afcapel:
@seanpdoyle I think removing the `StreamActions` export broke the turbo-rails system tests, as they are using it [here](https://github.com/hotwired/turbo-rails/blob/c3ca7009c0759563ec65ecf1bc60a9af1ff33728/test/system/broadcasts_test.rb#L137).

See https://github.com/hotwired/turbo-rails/actions/runs/6971461623/job/18971576977

PR #1062, opened 2023-11-13
Revert disk cache store

afcapel:
This reverts https://github.com/hotwired/turbo/pull/949 and the follow up PRs https://github.com/hotwired/turbo/pull/1014 and https://github.com/hotwired/turbo/pull/1056.

The idea is still appealing but we've found several hard-to-debug issues while testing this in our applications. Most notably we found some meta tags propagating outside the original page on navigation, and issues with the timing of `turbo:load` events.

These issues are solvable, but our priority right now is shipping  [Page Refreshes](https://github.com/hotwired/turbo/pull/1019) ahead of Turbo v8.


jorgemanrubia (approved):

src/tests/unit/native_adapter_support_tests.js
```diff
@@ -62,11 +62,11 @@ setup(() => {
   Turbo.registerAdapter(adapter)
 })
 
-test("navigator adapter is native adapter", async () => {
+test("test navigator adapter is native adapter", async () => {
```
kevinmcconnell, line 65: I think the extraneous `test` in the descriptions were removed after the original changes were added, and so reverting is bringing some of them back. Might we worth removing them again as an additional commit, for consistency?

kevinmcconnell (approved):
Looks good @afcapel.

It's a shame if the disk cache doesn't end up making it into v8, but I agree with you about the priorities so it's a good call to take it back out for now.

src/tests/unit/native_adapter_support_tests.js
afcapel: Good point, I've added the commits to the branch.

The branch was gone, so I had to recreate the commits. Definitely one case in which squashing the branch was not very helpful.

PR #1064, opened 2023-11-14
Add coverage for focus and value preservation during morph

seanpdoyle:
Idiomorph will preserve focus, but discard content. Version [0.0.9][] introduces the `ignoredActiveValue:` option to preserve the value for the element with focus. Since the currently packaged build isn't caught up to that version, the `[data-turbo-permanent]` treatment enables a similar kind of behavior.

This commit adds test coverage to ensure that focus preservation is global, and value preservation requires opt-in with `[data-turbo-permanent]`.

[0.0.9]: https://github.com/bigskysoftware/idiomorph/tree/v0.0.9#options


afcapel (approved):
Nice, thanks @seanpdoyle

PR #1068, opened 2023-11-17
Restore attribute "refresh=morph" to flag turbo frames to reload during a page refresh

jorgemanrubia:
This adds a new turbo-frame attribute: `refresh`. When its value is `morph`:

* It will reload the turbo frame with morphing during a page refresh.
* It won't update the turbo frame with the server response.
* It won't remove it if it's missing in the server response.

An example where these frames are useful are for pagination, or for other scenarios where you load new state in the screen that you want to preserve during page refreshes, but that you also want to reload because the user action might have affected them.

This attribute was part of [the original proposal we presented in Rails World](https://www.youtube.com/watch?v=m97UsXa6HFg), then [we removed it](https://github.com/hotwired/turbo/pull/1019/commits/0c6a95d07fd7661e734ed1fbfe19daa9a00b3fd4), because we thought it wasn't needed. But after testing the library in different scenarios, we've found that assuming certain logic for all the remote frames was problematic. We had to be too clever about what you wanted to do with the frame. The new attribute makes for a simpler story:

* The default behavior will be the expected one: turbo frames will be morphed as any other element. E.g: if they get a new URL they will be reloaded; if the get deleted in the response, they will disappear, etc.
* You can use the new attribute to flag the frames for which you want the special behavior.

Follow up to #1019.

### DIFF
--- a/src/core/drive/morph_renderer.js
+++ b/src/core/drive/morph_renderer.js
@@ -1,6 +1,5 @@
 import Idiomorph from "idiomorph"
 import { dispatch } from "../../util"
-import { urlsAreEqual } from "../url"
 import { Renderer } from "../renderer"
 
 export class MorphRenderer extends Renderer {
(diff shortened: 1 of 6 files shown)


afcapel (approved):
LGTM 👍

krschacht:
@jorgemanrubia I noticed that when you put `refresh="morph"` on a turbo frame and trigger it's refresh with `element.reload()` it does not morph. In this case, it falls back to a full reload of the frame. I'm thinking this may have been an oversight of the implementation. 

In my testing it only morphs when the full page is doing a morphing refresh.

jorgemanrubia:
@krschacht yes I agree that that would be more consistent. I'd appreciate if you can create an issue 🙏 .

krschacht:
Done. Issue #1161 is tee'd up with a good example.

PR #1072, opened 2023-11-20
Change default action for form submissions that redirect to the current location

afcapel:
This changes the default action for form submission that redirect to the current location from "advance" to "replace". This makes more sense since we're re-rendering the current page, there should be no new entry in the history.

We are also changing the morphing behaviour to only kick in when the visit action is "replace". This is to avoid morphing the page when, for example, we are following a link, unless we opt in to morphing by setting the data-turbo-action attribute to replace.

The main use case for morphing remains the same: you submit a form that changes some state and redirect to the same location. Turbo handles the redirect and morphs the page to reflect the new state.

### DIFF
--- a/src/tests/functional/navigation_tests.js
+++ b/src/tests/functional/navigation_tests.js
@@ -307,6 +307,18 @@ test("clicking the forward button", async ({ page }) => {
   assert.equal(await visitAction(page), "restore")
 })
 
+test("form submissions that redirect to a different location have a default advance action", async ({ page }) => {
+  await page.click("#redirect-submit")
+  await nextBody(page)
+  assert.equal(await visitAction(page), "advance")
+})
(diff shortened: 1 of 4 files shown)


jorgemanrubia (approved):
👍 👍

src/tests/functional/navigation_tests.js
```diff
@@ -307,6 +307,18 @@ test("clicking the forward button", async ({ page }) => {
   assert.equal(await visitAction(page), "restore")
 })
 
+test("form submissions that redirect to a different location have a default advance action", async ({ page }) => {
```
jorgemanrubia, line 310: I would add a page refresh test to validate that morphing doesn't trigger if you click on a link pointing to the current page.

PR #1073, opened 2023-11-23
Export StreamActions as property of window.Turbo 

afcapel:
This was removed in https://github.com/hotwired/turbo/pull/1049 but it's still used in the turbo-rails test suite.

https://github.com/hotwired/turbo-rails/blob/c3ca7009c0759563ec65ecf1bc60a9af1ff33728/test/system/broadcasts_test.rb#L137

@seanpdoyle can you have a look, please?


brunoprietog:
Thanks! I've upgraded to beta and can confirm that custom actions are not working. Turbo.StreamActions is undefined

seanpdoyle:
The exports should be equivalent between the reverted change and this one. I'm not sure why this is necessary.

There's existing test coverage introduced prior to this change that guarantees [Turbo.StreamActions](https://github.com/hotwired/turbo/blob/1aa7a876359480a64ff48ed1b842fa6b22de867a/src/tests/unit/export_tests.js#L6) is exported.

afcapel:
@seanpdoyle I think what is happening is that `StreamActions` is being exported as part of the module, but not set up inside `window.Turbo`. It's a bit confusing, but what we set as window.Turbo [here](https://github.com/hotwired/turbo/blob/58a6bb45a72468125b3e06231adfb921d2d30d44/src/index.js#L5-L7) is just the exports of `core/index.js`, not the whole Turbo. 

The unit tests pass because the asserting the exports from the module, not the global `window.Turbo`. I've added some browser tests to check `window.Turbo` as well.

PR #1077, opened 2023-11-27
Avoid infinite recursion from `window.fetch` name collision

seanpdoyle:
Closes [@hotwired/turbo-rails#526][]

Prevent naming collisions between the new `Turbo.fetch` function and the built-in `window.fetch` function.

To do so, create a module-local reference to `window.fetch`, and define the exported function as `fetchWithTurboHeaders`. At export-time, rename to `fetch` so that consumers can continue to import it as `Turbo.fetch`.

[@hotwired/turbo-rails#526]: https://github.com/hotwired/turbo-rails/issues/526

### DIFF
--- a/src/http/fetch.js
+++ b/src/http/fetch.js
@@ -1,13 +1,17 @@
 import { uuid } from "../util"
 
-export function fetch(url, options = {}) {
+const nativeFetch = window.fetch
+
+function fetchWithTurboHeaders(url, options = {}) {
   const modifiedHeaders = new Headers(options.headers || {})
   const requestUID = uuid()
   window.Turbo.session.recentRequests.add(requestUID)
   modifiedHeaders.append("X-Turbo-Request-Id", requestUID)
 
-  return window.fetch(url, {
+  return nativeFetch(url, {
     ...options,
     headers: modifiedHeaders
   })
 }
+
+export { fetchWithTurboHeaders as fetch }
--- a/src/tests/unit/export_tests.js
+++ b/src/tests/unit/export_tests.js
@@ -18,6 +18,7 @@ test("Turbo interface", () => {
   assert.equal(typeof Turbo.cache.clear, "function")
   assert.equal(typeof Turbo.navigator, "object")
   assert.equal(typeof Turbo.session, "object")
+  assert.equal(typeof Turbo.fetch, "function")
 })
 
 test("StreamActions interface", () => {


seanpdoyle:
@jorgemanrubia @afcapel I'm not entirely sure why this resolves the issue.

During my investigation, I found that the build output was using `var` in a watch that was defining any exported `fetch` as a property on the `window.`. First with the Turbo-specific `window.fetch` override, then with the [fetch](https://github.com/hotwired/stimulus/blob/349dc165611238f8ba5b316d86ec6536c58b2a2c/src/multimap/set_operations.ts#L10) method that Stimulus' `MultiMap` implementation ships.

I think something is either wrong built the build tooling configuration, or something else at the browser layer.

jorgemanrubia (approved):
Thanks for fixing this one @seanpdoyle

seanpdoyle:
What can we do to fix the underlying issue?

How can we prevent future changes from reintroducing the regression?

The test suite did not catch it.

afcapel:
I think the problem here is that a new `rails new demo -a propshaft -T -j bun -c tailwind` is adding `<%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>` to `application.html.erb`. The default type for `javascript_include_tag` is `"text/javascript"`, which is incorrect in this case. `javascript_include_tag` is pointing to ES6 code that imports other modules, the correct type should be `"module"`. Otherwise all the variables inside the ESM module are incorrectly assigned in the global scope.

I checked that adding `type: "module"` to the `javascript_include_tag` call fixes the issue.

This doesn't happen using importmaps because `importmap-rails` adds `<script type="module">import "application"</script>` to the application layout instead.

afcapel:
I can reproduce this with Turbo 7.3 and bun, same pollution of the global `window` scope. I think we only noticed now because there was a collision between `window.fetch` and `Turbo.fetch`.

afcapel:
Fix in js-bundling-rails here https://github.com/rails/jsbundling-rails/pull/178

jorgemanrubia:
Amazing work tracking this one down @afcapel!

seanpdoyle:
Would it be worth reverting this commit, since the fix is better suited elsewhere?

jorgemanrubia:
I think the change is ok @seanpdoyle. It uses a clearer naming approach IMO.

KonnorRogers:
> I can reproduce this with Turbo 7.3 and bun, same pollution of the global `window` scope. I think we only noticed now because there was a collision between `window.fetch` and `Turbo.fetch`.

+1. This behavior still exists on @hotwired/turbo@8.beta-1 using Bun which comes with Rails new. Not sure if a new release has been made.

PR #1078, opened 2023-11-27
Restructure to avoid internal `window.Turbo` access

seanpdoyle:
Replace accessing the `recentRequests` property through the `window.Turbo.session` object.

In its place, restructure the object's dependency graph so that the `http/fetch` module constructs the `LimitedSet` instance, and the `Session` accepts that instance as a property.


jorgemanrubia (approved):
👍 thanks @seanpdoyle

jorgemanrubia:
ah @seanpdoyle could you please rebase please?

PR #1079, opened 2023-11-27
Page refreshes: Use pathname instead of href to detect if locations are equal

brunoprietog:
Before this commit, URLs had to be exactly the same for the possibility of a refresh with morphing.

Now, it only checks that the paths are equal, regardless of whether they have different search params.

From my perspective, search params don't determine whether it is a refresh or not. The most representative example that comes to mind is the same case as a Rails controller. A show action could have different search params, but in essence, it remains the same action and the same page.

Additionally, sometimes when clicking a link on social media, for example, tracking parameters are injected. If these parameters exist and a form is submitted, Turbo would not detect that it is a refresh.

These are just small examples of use cases, but obviously there are many more.

### DIFF
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -56,7 +56,7 @@ export class PageView extends View {
   }
 
   isPageRefresh(visit) {
-    return !visit || (this.lastRenderedLocation.href === visit.location.href && visit.action === "replace")
+    return !visit || (this.lastRenderedLocation.pathname === visit.location.pathname && visit.action === "replace")
   }
 
   get snapshot() {
--- a/src/tests/fixtures/page_refresh.html
+++ b/src/tests/fixtures/page_refresh.html
@@ -44,6 +44,7 @@
       <h3>Element with Stimulus controller</h3>
     </div>
 
+    <p><a id="replace-link" data-turbo-action="replace" href="/src/tests/fixtures/page_refresh.html?param=something">Link with params to refresh the page</a></p>
     <p><a id="link" href="/src/tests/fixtures/one.html">Link to another page</a></p>
 
     <form id="form" action="/__turbo/refresh" method="post" class="redirect">
--- a/src/tests/functional/page_refresh_tests.js
+++ b/src/tests/functional/page_refresh_tests.js
@@ -17,6 +17,13 @@ test("renders a page refresh with morphing", async ({ page }) => {
   await nextEventNamed(page, "turbo:render", { renderMethod: "morph" })
 })
 
+test("renders a page refresh with morphing when the paths are the same but search params are diferent", async ({ page }) => {
+  await page.goto("/src/tests/fixtures/page_refresh.html")
+
+  await page.click("#replace-link")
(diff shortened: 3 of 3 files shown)


src/core/drive/page_view.js
```diff
@@ -56,7 +56,7 @@ export class PageView extends View {
   }
 
   isPageRefresh(visit) {
-    return !visit || (this.lastRenderedLocation.href === visit.location.href && visit.action === "replace")
+    return !visit || (this.lastRenderedLocation.pathname === visit.location.pathname && visit.action === "replace")
```
jorgemanrubia, line 59: I'd check that the `host` matches too. It would be a weird scenario that it didn't but I think. You could extract a method `isReloadingCurrentPath` that checks both conditions: `pathname` and `host`.

jorgemanrubia (commented):
I think this makes a lot of sense @brunoprietog 👏 . I left a minor suggestion there. I'll merge it with that. Thanks!

src/core/drive/page_view.js
jorgemanrubia: Great point @brunoprietog!

jorgemanrubia (approved):
Thanks @brunoprietog 👏

krschacht:
@jorgemanrubia or @brunoprietog In my application I have multiple cases where I want the Page Refresh morph action to be used, but the routes differ in a more significant way than simply search params. But currently, the page will only morph when it redirects back to the same route.

I think that I have good reason for my routes to be different, but then it falls back to Turbo Drive's full body replacement. I find myself contorting my code in order to redirect back to the same route just so I can get the Page Refresh morph to work.

Should we also add a way for morph to be explicitly triggered by including `data-turbo-action="morph"` in a form or in a link_to? I think the automatic default is very reasonable, but it doesn't feel right that the automatic default is the only way.

(Maybe it's not right to have this discussion on this pull request but this seems to be the most recent place that the question of route matching / page refresh was being considered.)

brunoprietog:
Yes, I've also seen the need to do so. My most representative case is a page with several tabs, in which to maintain order, I have each tab in its own controller. I had to unify them in one and differentiate them using the tab parameter, but it's certainly not the most comfortable.

The argument for not doing it is that since morphing is an expensive operation, it shouldn't need to be done when the current page doesn't look like the new page. But as we have seen, this is not always the case. Just because the path isn't the same, doesn't mean that the pages don't look similar enough to want to morph. Over time, I feel this will incentivize the use of fat controllers with parameters that differentiate the different use cases, just to be able to take advantage of morphing.

I also understand that it's a cost to introduce a new element to the Turbo API, and that is also a valid point.

I would also love to be able to make Turbo frames navigation with morphing for accessibility reasons. I'm currently doing it but using Idiomorph separately.

krschacht:
@brunoprietog Thanks for the quick reply. I'm glad I'm not crazy. :) I too have started to create fat controllers with parameters that differentiate the use-cases in order to solve for this.

@jorgemanrubia I tee'd up a PR that adds this functionality. It's my first time contributing to rails/hotwired/stimulus so I hope I followed the guidelines correctly: #1145 .  I welcome any guidance on how I implemented it.

PR #1086, opened 2023-11-29
Include idiomorph as a dev dependency

afcapel:
The library is bundled within Turbo so there's no need for apps using Turbo to install it.


jorgemanrubia (approved):
👍

PR #1088, opened 2023-11-29
Remove old head assets

### LINKED ISSUE #1082: [proposal] Removing unused stylesheets and scripts in head
So I understand that any new stylesheets and scripts in the head of the new page will be appended to the current head. Which is great, but what's not so great, is that these appended assets stick around even when not needed. So if I have some page specific CSS or JS, they will persist way beyond the page they are used.

I also understand that the recommended solution to this is to include these page specific assets in the body, instead of the head.

This is however, not ideal. Particularly when it comes to styles, as styles are render blocking assets, resulting in a flash of unstyled content when loading a new page that includes CSS within the body.

I've been playing around and have found what I believe to be a better solution that allows me to include assets in the head, but cleans up any unused or page specific assets.

After Turbo merges the head, the stylesheets and scripts within the new and current snapshots are compared. Any current asset not found in the new snapshot is removed from the head. Something like this...

```js
removeOldStylesheets() {
  const currentSheets = this.currentHeadSnapshot.getElementsMatchingType("stylesheet")
  const newSheets = this.newHeadSnapshot.getElementsMatchingType("stylesheet")

  currentSheets.forEach((element) => {
    if (!this.isCurrentElementInElementList(element, newSheets)) {
      document.head.removeChild(element)
    }
  })
}
```

This works nicely in my [limited] testing, but not so great if caching is enabled - I need to dig deeper into that. But before I do so, I wanted to get some feedback on this idea, and ensure that I've not overlooked anything.

Thanks, and any and all feedback is appreciated.

joelmoss:
fixes #1082


brunoprietog:
Replaced by #1128

Issue #1083, opened 2023-11-28
Morphing: Changes from JS libs that mutate the DOM are lost


weaverryan:
Hi!

Thank you for adding morphing! ❤️. I maintain the Symfony's LiveComponents package where we also morph. We frequently hit a problem with morphing + Stimulus that Turbo morphing also hits.

Reproducer: https://github.com/weaverryan/turbo-morph-tom-select-reproducer

tl;dr

1) Initialize a package like [TomSelect](https://tom-select.js.org/) inside a Stimulus controller. I'm using TomSelect as an example, but it's not unique to this package:

```js
export default class extends Controller {
     connect() {
        new TomSelect(this.element);
    }
}
```

```html
<select data-controller="tom-select">...</select>
```

2) Trigger a page refresh with morphing.

The result is that the rich TomSelect widget is lost. This is because TomSelect works by modifying the `select` (e.g. adding classes) and adding an entirely new `<div>` element after the `select`:

<img width="906" alt="Screenshot 2023-11-28 at 9 22 44 AM" src="https://github.com/hotwired/turbo/assets/121003/52f0ce18-580a-41cb-a24d-14d8c8dfbe52">

So, naturally, when the morph happens, the `select` element is reverted and the new `<div>` is lost entirely. However, because the `<select data-controller="tom-select">` element was not removed and re-added to the page, `disconnect()` and `connect()` are not called again. Also, no Stimulus values were changed and no targets (if we added a target to the `select`) are added/removed.

The result is that we lose the TomSelect widget and there's no hook to reinitialize it.

## Possible Solutions

Is there a solution for this? Or any thoughts? In Symfony LiveComponents, we have a complex MutationObserver system to track changes. But the correct solution would be something much simpler. Ideas:

A) Do we need to switch to a paradigm where we avoid JavaScript packages that mutate the DOM?
B) I know it #1019, there was temporarily a way for a Stimulus controller to "reconnect" when morphed. Is that needed?
C) Or, do we need some way to be able to mark an element to NOT be morphed (permanent)? This is actually not a great solution, as, in this case, you need to be aware of and find the new elements added by TomSelect and manually add some attribute (e.g. `data-turbo-permanent`). And if a Stimulus value *did* change and we *did* want to reinitialize TomSelect after a morph, we might need to manually destroy the `<div>` adde by TomSelect.

Note: TomSelect is an especially annoying/messy library. But from experience, this "JavaScript mutated the DOM and that was lost after morphing" will be a common issue.

[...]

PR #1097, opened 2023-12-03
Introduce `turbo:{before-,}morph-{element,attribute}` events

seanpdoyle:
Follow-up to [9944490][]
Related to [#1083]
Related to [@hotwired/turbo-rails#533][]

The problem
---

Some client-side plugins are losing their state when elements are
morphed.

Without resorting to `MutationObserver` instances to determine when a
node is morphed, uses of those plugins don't have the ability to prevent
(without `[data-turbo-permanent]`) or respond to the morphing.

The proposal
---

This commit introduces a `turbo:before-morph-element` event that'll
dispatch as part of the Idiomorph `beforeNodeMorphed` callback. It'll
give interested parties access to the nodes before and after a morph. If
that event is cancelled via `event.preventDefault()`, it'll skip the
morph as if the element were marked with `[data-turbo-permanent]`.

Along with `turbo:before-morph-element`, this commit also introduces a
`turbo:before-morph-attribute` to correspond to the
`beforeAttributeUpdated` callback that Idiomorph provides. When
listeners (like an `HTMLDetailsElement`, an `HTMLDialogElement`, or a
Stimulus controller) want to preserve the state of an attribute, they
can cancel the `turbo:before-morph-attribute` event that corresponds
with the attribute name (through `event.detail.attributeName`).

Similarly, this commit adds a new `turbo:morph-element` event to be
dispatched for every morphed node (via Idiomorph's `afterNodeMorphed`
callback). The original implementation dispatched the event for the
`<body>` element as part of `MorphRenderer`'s lifecycle. That event will
still be dispatched, since `<body>` is the first element the callback
will fire for. In addition to that event, each individual morphed node
will dispatch one.

This commit re-introduced test coverage for a Stimulus controller to
demonstrate how an interested party might respond. It isn't immediately
clear with that code should live, but once we iron out the details, it
could be part of a `@hotwired/turbo/stimulus` package, or a
`@hotwired/stimulus/turbo` package that users (or
`@hotwired/turbo-rails`) could opt-into.

[9944490]: https://github.com/hotwired/turbo/pull/1019/commits/9944490a3c8aec0c5060401125cc8932e93a32df
[#1083]: https://github.com/hotwired/turbo/issues/1083
[@hotwired/turbo-rails#533]: https://github.com/hotwired/turbo-rails/issues/533

### DIFF
--- a/src/core/drive/morph_renderer.js
+++ b/src/core/drive/morph_renderer.js
@@ -33,7 +33,9 @@ export class MorphRenderer extends Renderer {
       callbacks: {
         beforeNodeAdded: this.#shouldAddElement,
         beforeNodeMorphed: this.#shouldMorphElement,
-        beforeNodeRemoved: this.#shouldRemoveElement
+        beforeAttributeUpdated: this.#shouldUpdateAttribute,
+        beforeNodeRemoved: this.#shouldRemoveElement,
+        afterNodeMorphed: this.#didMorphElement
       }
     })
   }
@@ -44,9 +46,36 @@ export class MorphRenderer extends Renderer {
 
   #shouldMorphElement = (oldNode, newNode) => {
     if (oldNode instanceof HTMLElement) {
-      return !oldNode.hasAttribute("data-turbo-permanent") && (this.isMorphingTurboFrame || !this.#isFrameReloadedWithMorph(oldNode))
-    } else {
-      return true
+      if (!oldNode.hasAttribute("data-turbo-permanent") && (this.isMorphingTurboFrame || !this.#isFrameReloadedWithMorph(oldNode))) {
+        const event = dispatch("turbo:before-morph-element", {
+          cancelable: true,
+          target: oldNode,
+          detail: {
+            newElement: newNode
+          }
+        })
+
+        return !event.defaultPrevented
+      } else {
+        return false
+      }
+    }
+  }
+
+  #shouldUpdateAttribute = (attributeName, target, mutationType) => {
+    const event = dispatch("turbo:before-morph-attribute", { cancelable: true, target, detail: { attributeName, mutationType } })
+
+    return !event.defaultPrevented
+  }
+
+  #didMorphElement = (oldNode, newNode) => {
+    if (newNode instanceof HTMLElement) {
+      dispatch("turbo:morph-element", {
+        target: oldNode,
+        detail: {
+          newElement: newNode
+        }
+      })
     }
   }
 
--- a/src/core/view.js
+++ b/src/core/view.js
@@ -64,7 +64,7 @@ export class View {
         await this.prepareToRenderSnapshot(renderer)
 
         const renderInterception = new Promise((resolve) => (this.#resolveInterceptionPromise = resolve))
-        const options = { resume: this.#resolveInterceptionPromise, render: this.renderer.renderElement }
+        const options = { resume: this.#resolveInterceptionPromise, render: this.renderer.renderElement, renderMethod: this.renderer.renderMethod }
         const immediateRender = this.delegate.allowsImmediateRender(snapshot, options)
         if (!immediateRender) await renderInterception
 
--- a/src/tests/fixtures/page_refresh.html
+++ b/src/tests/fixtures/page_refresh.html
@@ -9,6 +9,45 @@
     <link rel="icon" href="data:;base64,iVBORw0KGgo=">
     <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
     <script src="/src/tests/fixtures/test.js"></script>
+    <script type="module">
+      import { Application, Controller } from "https://unpkg.com/@hotwired/stimulus/dist/stimulus.js"
+
+      const application = Application.start()
+
+      addEventListener("turbo:morph-element", ({ target }) => {
+        for (const { element, context } of application.controllers) {
+          if (element === target) {
+            context.disconnect()
+            context.connect()
+          }
+        }
+      })
+
+      addEventListener("turbo:before-morph-attribute", (event) => {
+        const { target, detail: { attributeName, mutationType } } = event
+
+        for (const { element, context } of application.controllers) {
+          const pattern = new RegExp(`data-${context.identifier}-\\w+-value`)
+
+          if (element === target) {
+            event.preventDefault()
+          }
+        }
+      })
+
+      application.register("test", class extends Controller {
+        static targets = ["output"]
+        static values = { state: String }
+
+        capture({ target }) {
+          this.stateValue = target.value
+        }
+
+        outputTargetConnected(target) {
+          target.textContent = "connected"
+        }
+      })
+    </script>
 
     <style>
         body {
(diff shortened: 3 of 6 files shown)


weaverryan:
This does ease handling morphing inside of a Stimulus controller. I keep going back-and-forth on whether Turbo should just trigger an event like this, or if the generic concept of "you were just morphed" should live inside of Stimulus.

Consider this: I create a 3rd party Stimulus controller that, on mutates the HTML in same way - maybe just adding a `hidden` class when the user clicks something. When Turbo morphs, that `hidden` class would be removed. My Stimulus controller needs to be aware of the morph and "re-add" the `hidden` class (based on some internal state the controller keeps track of that says "you should be in the hidden state").

This could be handled with this event - e.g. `<div data-controller="toggler" data-action="morph->toggler->resetState">`. However, that leaves 2 "not idea" things:

1) This `data-action` needs to live in userland code. I can't make my 3rd party Stimulus controller be "morph-ready" out-of-the-box (unless I decide to register the listener manually in the controller).
2) I believe... the `resetState` state would be called multiple times - if there were multiple elements morphed in the controller? If I'm correct, that's not... ideal, and it seems like it would be hard to de-dup that.

Also, since other libraries - HTMX, Livewire & Symfony Live Components - use morphing, it would be great to have a way that those could "hint" to Stimulus controllers that they are being morphed and then have the same behavior as when Turbo morphs.

Again, I go back-and-forth on this, but part of my thinks a `morphed()` method in Stimulus would be more usable (or perhaps this PR *and* that).

Thanks :)

klenis:
Thumbs up. The `turbo:before-morph` event is also a solution to be able to initialize Alpine.js on the new nodes.

jorgemanrubia (approved):
I like this a lot @seanpdoyle, thank you. I've also found this need where stimulus controllers need fine-grained control over what happens when a page refresh morph them, and `data-turbo-permanent` is not always an option.

This is a +1 from my side, but I'd love to hear what @afcapel thinks too.

brunoprietog:
I would like to bring @seanpdoyle's PR [hotwired/stimulus#460](https://github.com/hotwired/stimulus/pull/460) back to the table. This was something that was supposed to be added to Stimulus but never merged. In my opinion it is quite common to create observers in Stimulus controllers and that need will only increase with Turbo morphing. What do you think?

donnysim:
I've played around with the whole morphing thing and I think there going are to be some issues with current approach, that maybe should be accounted before hand? So the whole problem starts with scripts that modify DOM on execution and append content to the end of the body or after the morphed element. Firstly, the scripts are not re-evaluated because of morph, which I think will need a solution, but more importantly, if for say a stimulus controller appends elements to the body on connect, like a global tooltip or select dropdown, the morph still removes them because in order of processing the nodes it firstly processes the Stimulus controller that appends the elements, e.g. to the end of the body, then it continues forward to those elements and decide that they should be removed because they do not match the new content from the new body, e.g.:

before morph:
```html
<div data-controller="append"/>
<div appended-by-stimulus/>
```

on morph:
```html
<div data-controller="append"/>
<div appended-by-stimulus/> <!-- old -->
<div appended-by-stimulus/> <!-- new -->
<!-- or -->
<div data-controller="append"/>
<div appended-by-stimulus/> <!-- new -->
<!-- depending if stimulus cleans up the element -->
```

after morph (no matter if cleanup or not):
```html
<div data-controller="append"/>
```

if it makes sense. This would also be the problem with script tags if they start to re-evaluate that modify the dom. From hindsight probably the stimulus controllers need to reconnect after the whole DOM was fully morphed.

seanpdoyle:
@jorgemanrubia @afcapel I've expanded this change to include a `turbo:before-morph-attribute` event to afford stateful elements (like `HTMLDetailsElement`, `HTMLDialogElement`, `TrixEditorElement`, or a Stimulus controller with `[data-*-value]` attributes) an opportunity to partially opt out of the morph changes.

jorgemanrubia (commented):
@seanpdoyle I'd keep the `turbo:morph` focused at the standalone morphing operation.  I think it's a common need wanting to tie logic to a morph refresh as a whole, and with this change you will be forced to add additional code to handle that.

I like a lot the idea of letting fine gained control over how morphing happens, but I lean towards offering specific events:

- `turbo:before-morph-element`
- `turbo:morph-element`
- `turbo:before-morph-attribute`

I think that would make for a clearer API.

Thanks for working on this one 🙏!

seanpdoyle:
@jorgemanrubia prior to the introduction of these events, what strategies have you been using to support Trix and Stimulus controllers in a morph-compatible way?

seanpdoyle:
I've opened https://github.com/hotwired/turbo-site/pull/159 to document these new events.

jorgemanrubia (approved):
Thanks @seanpdoyle

jorgemanrubia:
> @jorgemanrubia prior to the introduction of these events, what strategies have you been using to support Trix and Stimulus controllers in a morph-compatible way?

We haven't found issues with Trix, I think because we aren't using a Trix instance in scenarios where a page refresh happens 🤔 . In the calendar, there is a single case where we used a stimulus value-change callback to prevent an unintended change in an attribute. It hasn't been a commong/recurring issue for us.

brendon:
Just wondering if there was ever any progress on a tighter integration with morphing in stimulus? I know I'll have many cases of incompatibility as I go through my controllers and test them out, but one simple one is a button that starts off hidden and the `connect()` method shows the button once everything has booted up because I'm using `static get shouldLoad(){}` to check if the controller should load at all.

Right now as soon as the page is morphed the button disappears again but the connect method never re-runs. It's almost as if there needs to be another version of `connect()` such as `morphed()` or even just a `reconnect()`. It's so hard because the logic will be application specific. Sometimes you might actually want the updated values coming in from the morph, but you might also want to keep the ones that the browser has calculated.

It'd almost be better to have some kind of 'in-use` flag on a controller and have the morph strategy either ignore updates from that controller node downward if it's in use, or otherwise mutate and call `connect()` again.

mnrlx:
@brendon if still relevant, then you can try a bit different approach
```
connect() {
  this.element.addEventListener("turbo:before-morph-element", this.reconnect.bind(this))
  ...
}
reconnect(event) {
  this.element.replaceWith(event.detail.newElement)
}
```
this also works nice for updated values and connects your element back again since it gets added to the dom.

brendon:
> @brendon if still relevant, then you can try a bit different approach
> 
> ```
> connect() {
>   this.element.addEventListener("turbo:before-morph-element", this.reconnect.bind(this))
>   ...
> }
> reconnect(event) {
>   this.element.replaceWith(event.detail.newElement)
> }
> ```
> 
> this also works nice for updated values and connects your element back again since it gets added to the dom.

That's an interesting approach :) I've managed to iron out all my controllers now but will try this out next time I have an issue.

I found that if I only blocked updates of certain attributes (like `class` if a certain class currently exists like `active`) then I could maintain the state of the controller (e.g. a user has opened a dropdown menu) while still allowing the morphs to update the actual content of the dropdown menu.

I'm sure some smart cookie will come up with a clean way to describe these interactions in the future if turbo sticks with morphing :D

PR #1098, opened 2023-12-04
Page refreshes: Don't render previews

### LINKED ISSUE #1080: Error when morphing a page when visiting it for the second time
With morphing enabled by default, when visiting a page with the 'replace' action, revisiting the previous page using the 'replace' action, and then revisiting the current page using the 'replace' action, an error occurs, and the body becomes null.

You can easily test the error at <https://turbo-morph-error.fly.dev/>

Visit the second page, then return to the first page by pressing the link, and then press the link to visit the first page again. You will see that the page remains blank.

From my perspective, what is happening is that in one of the renderings with morphing, the body is set to null after calling idiomorph. I tried debugging, but I only managed to notice that; I still don't understand the cause.

brunoprietog:
Fixes #1080

In #1080, when rendering the cached snapshot as preview, idiomorph is causing `document.body` to become `null`, I didn't quite understand why.

Anyway, In my opinion morphing should not render previews anyway. If the goal of morphing is to maintain state, why would we want to do that and show a previous version of the document for a moment?

So, this PR does just that. If it is a page refresh, the cached snapshot will not be rendered as preview.

### DIFF
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -84,6 +84,7 @@ export class Visit {
     this.snapshotHTML = snapshotHTML
     this.response = response
     this.isSamePage = this.delegate.locationWithActionIsSamePage(this.location, this.action)
+    this.isPageRefresh = this.view.isPageRefresh(this)
     this.visitCachedSnapshot = visitCachedSnapshot
     this.willRender = willRender
     this.updateHistory = updateHistory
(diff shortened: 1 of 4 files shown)


pfeiffer:
I think this would be really useful and the way to go - right now a previous version of the document is shown for a brief period of time while the before the refreshed page is being morphed leading to some potential weird flickering and defeats the niceness of the refreshing.

afcapel:
Makes sense, thanks @brunoprietog 🙏 

We [excluded page refreshes from the cache on async updates](https://github.com/hotwired/turbo/blob/f81f9c2172534ad1fdd85762a53fd82dca76fbcd/src/core/session.js#L111), but forgot to do the same in regular navigations.

Issue #1047, opened 2023-10-26
Promoted Frame Visits cause `tracked_element_mismatch` reloads


domchristie:
This only impacts `main` branch builds since https://github.com/hotwired/turbo/pull/887.

When promoting Frame Visits using `data-turbo-action="…"`, Turbo will reload the page. The `turbo:reload` event reason is: `tracked_element_mismatch`. This only appears to be an issue since: #887 

/cc @seanpdoyle

domchristie:
@afcapel I think this is quite a major issue, since it breaks all Turbo Frames visits with a data-turbo-action. Would be good to get fixed before v8 is released

afcapel:
Agreed, let's get it fixed before v8 👍

qinmingyuan:
Any progress for this issue?

seanpdoyle:
@domchristie I'm unable to reproduce this locally in the following test:

```js
test("navigating turbo-frame[data-turbo-action=advance] from within pushes URL state", async ({ page }) => {                                                                                                        
  await page.click("#add-turbo-action-to-frame")                                                                                                                                                                    
  await page.click("#link-frame")                                                                                                                                                                                   
  await nextEventNamed(page, "turbo:frame-load")                                                                                                                                                                    
  await noNextEventNamed(page, "turbo:reload")                                                                                                                                                                      
  await nextEventNamed(page, "turbo:load")                                                                                                                                                                          
  await noNextEventNamed(page, "turbo:reload")                                                                                                                                                                      
                                                                                                                                                                                                                    
  const title = await page.textContent("h1")                                                                                                                                                                        
  const frameTitle = await page.textContent("#frame h2")                                                  
                                                                                                                                                                                                                    
  assert.equal(title, "Frames")                                                                           
  assert.equal(frameTitle, "Frame: Loaded")                                                                                                                                                                         
  assert.equal(pathname(page.url()), "/src/tests/fixtures/frames/frame.html")                                                                                                                                       
})   
```

I've added some assertions related to `turbo:reload`:

```diff
diff --git a/src/tests/functional/frame_tests.js b/src/tests/functional/frame_tests.js
index b090aff..a796e44 100644
--- a/src/tests/functional/frame_tests.js
+++ b/src/tests/functional/frame_tests.js
@@ -642,7 +642,10 @@ test("navigating a frame with a form[method=get] that does not redirect still up
 test("navigating turbo-frame[data-turbo-action=advance] from within pushes URL state", async ({ page }) => {
[...]

PR #1101, opened 2023-12-06
Add InstantClick behavior

davidalejandroaguilar:
# Description

This PR adds [InstantClick](http://instantclick.io/)-like behavior to Turbo. For those not familiar with it, what it does is prefetch links that are likely to be clicked on.

>Before visitors click on a link, they **hover** over that link. Between these two events, 200 ms to 300 ms usually pass by ([test yourself here](http://instantclick.io/click-test)). 

The result is effectively instant navigation in most cases.

# Implementation

* An observer is added to the Turbo session. It adds 3 event listeners:
  * `mouseover`.
    * When the mouse is over a link, it will wait 100ms before starting a `fetch` request to prefetch it and save the request without `await`ing for it to a cache with an expiration of 10 seconds (configurable).
    * This delay is necessary to prevent a flurry of prefetch requests when casually scrolling down the page.
  * `touchstart`.
    * Same thing as above, but when the user finger touches the screen.
  *  `turbo:before-fetch-request`.
      * When a link is clicked, it will check if the location exists in the cache, and if so, take the `fetch` request out of it and put it on the `event.detail.response`.
* The `FetchRequest#perform` method will now accept overriding the response internally (not exposed as part of public API) by assigning it to `event.detail.response` on the `turbo:before-fetch-request` event added by the observer.
  * The cached response is then used here instead of creating a new `fetch` request.
  * This cached request will be `await`ed and `#receive`d as usual (most probably by this time, the `fetch` request promise has already resolved), resulting in an instant navigation.
  * This means that all the usual Turbo behavior still happens (browser history change, snapshot, page update without full reload, etc.).

# Configuration

* The amount of time a link is prefetched again after hovering can be configured via a `<meta name="turbo-prefetch-cache-time" content="1000">` tag, in milliseconds, and defaults to `10000`.

* This behavior is **disabled by default** and can be opted-in via a `<meta name="turbo-prefetch" content="true">` tag.

* You can opt-out a link from being prefetched on hover by adding a `data-turbo-prefetch="false"` attribute to it. 

* You can also opt-out/in descendants of an element this way (opt-out/in by container).

# Notes

>On mobile devices, preloading starts on “touchstart”, letting 300 ms (Android) to 450 ms (iOS) for preloading the page.[2](http://instantclick.io/#note-3g)

* The concept name I chose for this behavior is "prefetch", because [we already have the concept of "preloading"](https://turbo.hotwired.dev/handbook/drive#preload-links-into-the-cache) by using `data-turbo-preload` on specific links we want to preload on page load.
* A few more differences from the existing preloading concept and implementation, apart from the main difference of preloading on hover vs on initial page load:
  * Preloading awaits each `fetch` request sequentially in order to manually populate the `SnapshotCache`, whereas the prefetched ones are not `await`ed when created, and thus allows for non-blocking concurrent execution of requests. 
  * Prefetching doesn't manually populate the `SnapshotCache` like preloading does, but just lets Turbo do everything automatically by just using the cached `fetch` request instead of creating a new one. This means it's less intrusive on the internals of Turbo.

# Prior art 

In no specific order:

* Oleksandr and Pete from Docuseal (Open Source Document Signing built with Rails and Turbo) ([repo](https://github.com/docusealco/docuseal)), [added](https://github.com/docusealco/docuseal/commit/787df43ac025889372ad9411291bef3076107682) a Turbo version of InstantClick. The perceived speed benefits are simply _amazing_. They can be experienced on [their website](https://docuseal.co/).
  * This implementation is based on this one.
  * I added this to an app of mine, including support for `touchstart`, and have been running in production for a while with no hiccups.
* [hopsoft's](https://twitter.com/hopsoft) Turbolinks prefetching [implementation](https://gist.github.com/hopsoft/ab500a3b584e2878c83137cb539abb32).
* [phacks](https://github.com/phacks/phacks)' [implementation](https://gist.github.com/phacks/9be4a4ceb27e51f60f7670f28e7f5280).

# Demo

A scaffold Rails app with 0.5 delay on the `PostsController#show` action.

Noticeable delay:

https://github.com/hotwired/turbo/assets/15317732/2500bf9e-5fee-4e53-b64b-bd4910a21e64

Instant navigation:

https://github.com/hotwired/turbo/assets/15317732/adfbbdbd-f74d-4396-92bc-7c5cc625a53c

## Scrolling down a page with many links

1. No delay, ends up in a flurry when moving the mouse.

https://github.com/hotwired/turbo/assets/15317732/cf099c3d-52d2-407e-968c-ebf99ff72360

2. With the [recommended](http://instantclick.io/preloading#delay) 100ms max. Notice how there's no more flurry of requests, yet when we do a small intent pause before clicking, we get a prefetch.

https://github.com/hotwired/turbo/assets/15317732/e5286ce7-c984-4b3d-9024-a278dd8031f4


brandondrew:
This looks awesome—thanks for your work on it!

> * The concept name I chose for this behavior is "prefetch", because [we already have the concept of "preloading"](https://turbo.hotwired.dev/handbook/drive#preload-links-into-the-cache) by using `data-turbo-preload` on specific links we want to preload on page load.
>   
>   * I'm open to suggestions on this.

Regarding the name, let's consider what ideas are conveyed by the related words:
- "preload" and 
- "prefetch" 

One contains the verb "load" and the other contains "fetch", implying that preloading and prefetching perform different operations.

Both contain "pre" which is a temporal reference, suggesting doing something early, where "early" means before some other significant point in time.  Since both have the same temporal reference, it would imply that they both perform their operations at the same early point in time.

However the truth is the opposite:
- the perform (more or less) _the same operation_, 
- they perform their operation _at different points in time_.

Therefore it would be vastly preferable in my opinion if they shared the verb "load" but the new one had a different way of referring to _when_ it performs this loading operation. 

The clearest option I can think of is "**_hoverload_**".  _It loads the linked page on hover._  I think this will avoid lots of confusion when reading code, where we have to stop and ask ourselves "wait, what's the difference between preloading and prefetching again?"

davidalejandroaguilar:
@brandondrew Thanks for chiming in! 

Agree that the naming might lead to confusion and I like `hoverload`, though this behavior can happen on `mouseover` (hover), `mousedown` (start to press mouse button) and eventually `touchstart` (finger touches screen). So perhaps the prefix `hover` is no longer appropriate. 

Although if we base the name on the _default_ behavior, which is hovering, and think of touching the screen as the equivalent of "hovering" over a link on a mobile device, then it does make sense.

Happy to make the change, though would like to wait for a few more folks to chip in too!

brandondrew:
hmm... you're right, it might not be perfect.... but I can't think of anything better so far... 😄 

- `mouseload`? 
- `late_pre_load`?
- `triggerload` (where hovering or mousing down, or putting your finger on the screen is the "trigger")  

Maybe someone else will have a better idea...

brandondrew:
- `insantload`—is not a perfectly accurate description of what actually happens, but it corresponds to "InstantClick" and it does describe the user's experience.

airblade:
"PredictiveLoad"?

adrienpoly:
Thanks for this proposal that would be a great addition. 

I agree that the proposed naming is confusing with the current preload attribute.

What about `data-turbo-instant=true`

afcapel:
@davidalejandroaguilar thanks for bring this up, it is something we definitely want to add to Turbo! I'm going to give it a spin in one of our apps and and will come back with feedback after that.

brandondrew:
> Could we build on top of the existing data attribute?
> 
> * `data-turbo-preload` preloads the link on page load (backward compatibility)
> * `data-turbo-preload="page-load"` preloads the link on page load (new attribute value for existing way of link preload)
> * `data-turbo-preload="on-demand"` preloads the link on `mouseover`/`mousedown`/`touchstart` (this PR)


Okay, this suggestion has a lot of merit, and in many ways makes sense (I like the cohesive API suggested) but I have strong reservations about falling back to preloading everything on pageload for clients that can't handle loading on hover. 

To determine whether the question is moot or not, do we know that there are actually any browsers in this situation?  Would there actually be people who have to have everything load on page load because their browser can't handle the code in this PR that's been (at some future date) added to Turbo?

If there are no such browsers, then I like this suggestion a lot.

But if there *are* browsers like that, then I would not want to be forced to make those browsers fall back to preloading everything on pageload. Imagine a gallery home page with 100 links that all go to image-heavy pages.  Imagine a user on a slow cellular network opening the gallery home page.

brandondrew:
If the attribute is added to each link, then why does this section (below) seem to imply that it is controlled at the page level, and the links are opt-out only?

> # Configuration
...
> * This behavior is disabled by default and can be opted-in via a `<meta name="turbo-prefetch" content="true">` tag.
> * You can opt-out a link from being prefetched on hover by adding a `data-turbo-prefetch="false"` attribute to it. You can also opt-out descendants of an element this way.
> 

Did I misunderstand that 👆?

IMHO it would be preferable if the developer could _also_ opt-in at the element level.

brandondrew:
As far as names (if it is given a different name instead of a new value of the preload attribute)

I think [@adrienpoly's suggestion ](https://github.com/hotwired/turbo/pull/1101#issuecomment-1845037188) might be the catchiest name, and I like it a lot.  It describes the _user's_ experience (more or less) but it doesn't give a very clear description _to a developer_ of what's really happening.

@airblade's suggestion is the most accurate description of what it does, but `predictiveload` feels a little bit too long and awkward to me.   On the other hand, if it is determined that there is no problem with adding values to the `data-turbo-preload` attribute, then `data-turbo-preload="predictive"` seems perfect, since that's exactly what is happening.  It's a predictive preload.  This nails the terminology exactly.

(Also, `data-turbo-preload="predictive"` might be possible to use without falling back to preloading on page load for all of those links.)

My 2¢ in summary:
If a new attribute is introduced, my preference is slightly for for `hoverload` over `instant`, but only slightly.  If potential problems of falling back to preloading all designated links when the page loads can be resolved, then I think `data-turbo-preload="predictive"` is absolutely perfect.

pfeiffer:
A slightly different - but in my opinion way simpler solution in Turbo 'core' - would be to change the existing `Preloader` to use a `MutationObserver` observing any links with `data-turbo-preload` attribute, instead of just prefetching links on [DOMContentLoaded](https://github.com/hotwired/turbo/blob/db310ef15ed9e4d0d7b7990220a20abe165ea993/src/core/drive/preloader.js#L17-L19). 

Implementing this would allow developers to add their own behavior (ie. "instant click", preload only links above the fold, ..) by simply dynamically adding a `data-turbo-preload` attribute to the links as they see fit, eg on hover or scroll into the viewport or whatever custom logic they'd like.

The behavior described in this PR (hover, viewport, ..) would be very well suited as a separate add-on/package, providing reasonable default behavior as described in the PR description.

This PR introduces code that is [almost duplicate](https://github.com/hotwired/turbo/pull/1101/files#diff-1a0b7d769d04149117769a2bb8e5b27af85123be3c37e31f5790113c827a1962R190-R193) of the [existing preloading](https://github.com/hotwired/turbo/blob/db310ef15ed9e4d0d7b7990220a20abe165ea993/src/core/drive/preloader.js#L39-L41) behavior.

pfeiffer:
> The FetchRequest#perform method will now accept overriding the response internally (not exposed as part of public API) by assigning it to event.detail.response on the turbo:before-fetch-request event added by the observer.
>   * The cached response is then used here instead of creating a new fetch request.

I think this would lead to unexpected results and is a breaking change, as least when combined with the Turbo _preloaded_ links, where the behavior is that Turbo displays the preloaded snapshot from the cache *while performing a fetch for the url*.

In our application, we use the Turbo preload pattern quite extensively - all preloads fetches "skeleton pages" (that are extremely cachable with a `Vary: Sec-Purpose` header and with no database queries made) and upon navigation, we fetch the full page that contains dynamic content. 

This ensures that tapping a link navigates instantly and displays a skeleton page, and then the full page with dynamic and user-specific content is fetched by Turbo.

This PR assumes that the page returned by a prefetch request is the same as a full normally fetched page, which I think is not always the case as described above and is a breaking change from how Turbo behaves today.

pfeiffer:
> The amount of time a link is prefetched again after hovering can be configured via a `<meta name="turbo-prefetch-cache-time" content="1000">` tag, in milliseconds, and defaults to 10000.

I'm wondering why this would be needed? Why not rely on normal HTTP caching mechanism and headers to communicate TTL and let `fetch` handle that?

davidalejandroaguilar:
@marcelolx 

> Could we build on top of the existing data attribute?

> data-turbo-preload="on-demand" preloads the link on mouseover

While I think that building on top of existing solutions is ideal, and I did consider enhancing the `Preloader` class, I believe that needing to annotate every link with `data-turbo-preload="on-demand"` is a bit tedious and contrary to the idea of Turbo making your app feel faster out of the box, for every link, like Turbo Drive does by default, which to me is the ideal. Of course, you’d have to opt-in to this new behavior by adding the meta tag.

Unless you mean that, instead of adding this to every link, the metadata tag should be `<meta name="turbo-preload" content="on-demand">`, “reusing” the concept and name of “preload”, in which case I agree that could reduce confusion or cognitive load of adding a different name.

davidalejandroaguilar:
@brandondrew 

>If the attribute is added to each link, then why does this section (below) seem to imply that it is controlled at the page level, and the links are opt-out only?

This PR intends for the InstantClick behavior to be added at the global/page level, annotating sections or links with `data-turbo-prefetch="false"` to disable on a granular level.

Or simply not enabling this behavior by not opting-in at all, by not adding the  `<meta name="turbo-prefetch" content="true">` tag.

davidalejandroaguilar:
@pfeiffer 

>change the existing Preloader to use a MutationObserver observing any links with data-turbo-preload attribute

> This PR introduces code that is [almost duplicate](https://github.com/hotwired/turbo/pull/1101/files#diff-1a0b7d769d04149117769a2bb8e5b27af85123be3c37e31f5790113c827a1962R190-R193) of the [existing preloading](https://github.com/hotwired/turbo/blob/db310ef15ed9e4d0d7b7990220a20abe165ea993/src/core/drive/preloader.js#L39-L41) behavior.

> There's already a PR (https://github.com/hotwired/turbo/pull/911) changing the preloader to use MutationObserver ... Could it be built on that?

As I mentioned above, I do think that building on top of existing solutions is ideal, and very much considered enhancing the `Preloader` class. However, the intent of the InstantClick behavior is different than the core concept of that class:

  - The existing preloading behavior is to preload links that have been *specifically* and *manually* (hard-coded or dynamically added) annotated with `data-turbo-preload`.
  - The InstantClick behavior is to prefetch on hover for ALL links, and opting-out for specific links or sections of the page. (Please also remember this whole behavior is disabled by default, and only opted-in by adding `<meta name="data-turbo-prefetch" content="true">` to the pages you wish to add this behavior).

So I realized that *essentially*, they are different concepts, one to immediately preload links on load, and the other to make navigation feel faster by preloading on click intent, so it didn’t feel right to alter the `Preloader`.

Also, altering the `Preloader` seemed like altering an established default that would’ve stirred the waters a lot. Instead, adding a separate, opt-in behavior also made more sense.

In regards to introducing almost duplicate behavior, I have to say those lines are very much alike but it’s only because both are `fetch` requests in nature, and they both are used to load things in advance, so I’d expect them to be very similar. However, as mentioned above, *essentially*, they’re both different things.

The PR for adding a `MutationObserver` to `Preloader` to handle adding `data-turbo-preload` dynamically is a good call, however since this PR is adding the behavior to all the links in the page (if you opted-in), there’s no need for it since any dynamically added links will get automatically detected, because we’re using event delegation on `mouseover`.

I’d like to reiterate that the idea of this PR is to enhance Turbo’s *******general******* behavior, after opting in, by removing the burden on the developer to have to add custom events or data attributes other than a meta tag to get your app to feel even faster. Sort of what page refreshes are doing now: by simply adding a meta tag, you free the developer from the burden of having to write custom turbo streams for specific parts of the page, increasing both the developer happiness and the user experience.

davidalejandroaguilar:
@pfeiffer 

>I think this would lead to unexpected results and is a breaking change, as least when combined with the Turbo preloaded links

>In our application, we use the Turbo preload pattern quite extensively - all preloads fetch "skeleton pages" (that are extremely cachable with no database queries made and with a Vary: Sec-Purpose header) and upon navigation, we fetch the full page that contains dynamic content.

>This ensures that tapping a link navigates instantly and displays a skeleton page, and then the full page with dynamic and user-specific content is fetched by Turbo.

>I also believe the existing preloading behavior of Turbo supports not re-fetching the preloaded page, by setting cache headers

The fact that the preloaded snapshot from cache is displayed while performing a fetch from the URL is indeed a good thing. And the technique you use to offer the user a more instant navigation feel is interesting.

I’d first like to mention that I don’t believe this to be a breaking change for preloaded links.

- First, remember that this behavior is disabled by default and only opt-in.

- Second, the InstantClick behavior is completely compatible with preloading, and in fact they both work together very nicely:

    - By using the existing preload behavior, your links are loaded in advance, and stored in the `SnapshotCache`, then when the user hovers over one of those links before clicking, Turbo will now fetch that link ****in advance****. Meaning that the “refresh request” that Turbo did after showing you the snapshot cache is no longer done because Turbo already has the fresh copy.
    - So you get the best of both worlds: by using the existing preload behavior, you populate the snapshot cache in advance, so if the user hover and clicked on your link very fast, not giving the prefetch enough time to resolve, you’ll still get an instant navigation feel. And by using the InstantClick behavior, you remove the “refresh request” after they arrived to the destination.
    - However, I’d argue that the InstantClick behavior might render the preloading behavior unnecessary in most cases, since hovering before clicking usually takes ~400-500ms, which is more than enough for Turbo to load the next page and make navigation feel instant.
    - This could help you replace displaying loading skeletons to your users with actual content immediately!

Finally I’d like to mention that your existing cache headers will still work with the InstantClick behavior!

Here are 2 videos combining preloading and prefetching. It's the same demo app as the one in the description, I just increased the `PostsController#show` delay to 1 second to make this more clear.

Here's only preloading. Notice that the links are preloaded on load as usual, yet when the user clicks, even though the page navigation was "instant", there's still an extra "refresh request" to get the latest. 

I don't think is ideal and the technique you're using of prefetching a skeleton only for the "refresh request" to fetch the actual content might just be trying to work around this.

https://github.com/hotwired/turbo/assets/15317732/fcbc4756-a5b5-41fc-aa37-d0a78dd56ce0

Here's preloading with prefetching. Notice that the links are preloaded on load, but now also when the user hovers (I pause for 1 second on purpose for demonstration), this "refresh request" is prefetched, and when I click the link, there's no need to do that anymore!

https://github.com/hotwired/turbo/assets/15317732/eb5f56b6-c8f2-4d12-8737-92461d313c9c

davidalejandroaguilar:
@pfeiffer 

>I've added a gist here for inspiration

Thank you for taking the time to write the gist! I like the idea of dispatching an event to cancel prefetching (though you might just want to add `data-turbo-prefetch="false"` to that link instead).

I don’t think we’d like to delay preloading on hover, since we’re trying to take advantage of every millisecond we have from hover to click.

As mentioned above, however, I think the `Preloader` essence and the InstantClick behavior are 2 separate things, and in this case, for example, would mean that the snapshot cache is pre-populated, but on click, Turbo would need to make another “refresh request” once it arrives to the location, whereas if you left the preloading to do what it does, and introduce prefetching, you’d eliminate this “refresh request”.

davidalejandroaguilar:
>I'm wondering why this would be needed? Why not rely on normal HTTP caching mechanism and headers to communicate TTL and let fetch handle that?

It’s to entirely avoid making a request at all. If the HTTP caching headers are not set up correctly, there’s highly dynamic content and/or there’s network variability (bad connection), this TTL can provide a great out-of-the-box caching strategy.

I think it’s also an easy knob for devs to adjust than meddling with cache headers, which I believe it’s a very Rails-way to do things.

Of course, I’m happy to accommodate changes, though would like to give folks a chance to test this first and see how it feels in their apps.

davidalejandroaguilar:
> @davidalejandroaguilar thanks for bring this up, it is something we definitely want to add to Turbo! I'm going to give it a spin in one of our apps and and will come back with feedback after that.

Awesome, thanks for taking it out for a test drive, hope it's as smooth as it has been for me!

marcelolx:
> While I think that building on top of existing solutions is ideal, and I did consider enhancing the Preloader class, I believe that needing to annotate every link with data-turbo-preload="on-demand" is a bit tedious and contrary to the idea of Turbo making your app feel faster out of the box, for every link, like Turbo Drive does by default, which to me is the ideal. Of course, you’d have to opt-in to this new behavior by adding the meta tag.

> Unless you mean that, instead of adding this to every link, the metadata tag should be <meta name="turbo-preload" content="on-demand">, “reusing” the concept and name of “preload”, in which case I agree that could reduce confusion or cognitive load of adding a different name.

@davidalejandroaguilar Yeah, sorry, I focused too much on the name and forgot that the PR description mentioned it to be an option as a meta tag, my bad!

src/core/session.js
```diff
@@ -166,6 +171,50 @@ export class Session {
...
+
+  prefetchAndCacheRequestToLocation(link, location, cacheTtl) {
+    const requestOptions = {
+      credentials: "same-origin",
+      headers: {
+        "Sec-Purpose": "prefetch",
+        Accept: "text/html"
+      },
+      redirect: "follow"
+    }
```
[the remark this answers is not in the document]
seanpdoyle: The pattern of FetchRequest delegation is used in https://github.com/hotwired/turbo/pull/1034 to extend the `Preloader` class.

seanpdoyle:
> As mentioned above, however, I think the Preloader essence and the InstantClick behavior are 2 separate things, and in this case, for example, would mean that the snapshot cache is pre-populated, but on click, Turbo would need to make another “refresh request” once it arrives to the location, whereas if you left the preloading to do what it does, and introduce prefetching, you’d eliminate this “refresh request”.


I wonder if there is an opportunity to change the `Preloader` so that it doesn't write to the `SnapshotCache`. The [`Preloader` doesn't currently support Turbo Frames](https://github.com/hotwired/turbo/issues/857#issuecomment-1409507860), partly because of ongoing issues with [promoted Turbo Frame navigations driven by incomplete `<html>` document responses](https://github.com/hotwired/turbo/issues/1047).

I'm struggling to understand the end-user facing different between Preloading with `[data-turbo-preload]` and the Prefetching proposed in this PR. If the `Preloader` were modified to only issue and cache the `fetch` request in the same way as the `LinkPrefetchObserver` as its proposed here, I think we'd be able to re-use it for both use cases:

1. Exclude any element marked with `a[data-turbo-preload="false"]` from preloading or prefetching
2. When a page loads, query for all `a[data-turbo-preload="true"]` elements, then preload them
  2a. When an `a[data-turbo-preload="true"]` element is added to the page dynamically, preload it
3. Attach `mouseenter` (or similar `touch`-based event listeners) in a `document.body`-scoped observer. When InstantClick-like behavior is enabled (maybe with a `<meta>` in the `<head>`), handle those events with the same criteria as if they were marked with `a[data-turbo-preload="true"]`

@davidalejandroaguilar is exploring what it'd take to unify those two paths something you're interested in? If not, I'm happy to open a parallel PR to demonstrate what I have in mind.

davidalejandroaguilar:
@afcapel 

Glad your testing went well! 🚀

>This makes me think that we probably want a more flexible API that allows you to opt in/out on a per link basis. Or even better on a per-container basis, adding data-turbo-prefetch="true/false" to the container. This is similar to how data-turbo="true/false" works. If you want to add instaclicks to the whole page you can always add data-turbo-prefetch to the <body>.

Makes sense, I will make the change to be opt-in per link or container. This was also noted by @brandondrew  ✅

> we probably want to add some configurable delay before launching the prefetch request. 

I will add the configurable delay before launching the prefetch request, as you and @pfeiffer suggest ✅

I will also address the comments you left ✅

> If we want to use a per-container setting, we may want to use mousenter instead of mouseover. mouseenter doesn't bubble so we can check in the event handler whether the target is a link, and bail out if not, without having to use any DOM traversal.

Let me know if I'm misunderstanding, but we also currently don’t _need_ to use any DOM traversal to check whether the target is a link, I was just using `findLinkFromClickTarget` to reuse the matcher that function has: `"a[href]:not([target^=_]):not([download])"`. However, we can stop using that function, and change our approach to use `matches`:

```js
#tryToPrefetchRequest = (event) => {
    const target = event.target
    const isLink = target.matches("a[href]:not([target^=_]):not([download])")
    const link = target

    if (isLink && this.#isPrefetchable(link)) {
      // ...
    }
  }
```

This works with both `mouseover` and `mouseenter`.

However, we still have to use DOM traversal to check if the target is within a container with either `data-turbo="false"` or `data-turbo-prefetch="false"`.

Here's 2 videos using `mouseover` and the above implementation.

Bottom to top:

https://github.com/hotwired/turbo/assets/15317732/55af1347-cc5e-4985-b0e2-426d1fbe347a

Top to bottom:

https://github.com/hotwired/turbo/assets/15317732/a1ea124c-96ec-4ee9-85d0-d5fab1344042

A few more things to note:

* I'll push these changes so the PR can be ready to merge. After doing so, however, I'll work on @seanpdoyle 's [suggestion](https://github.com/hotwired/turbo/pull/1101#issuecomment-1847593910). If you merge before, I'll open a subsequent PR. These changes are:

Reuse the `Preloader` class for both the existing preloading and InstantClick behavior by having the `Preloader` class:
  - Preload links on load.
      - Those having `data-preload="true"`
  - Preload links on hover.
      - If opted-in.

* Based on @airblade and @brandondrew suggestion, the "predictive" concept name sounds good, e.g. `<meta name="turbo-predictive-preload" content="true/false">` and `data-turbo-predictive-preload="true/false"`. If you agree, I'd like to make this change when I work on the above one, since we'd now be reusing the `Preloader` class and thus reusing the "preload" concept.

davidalejandroaguilar:
@seanpdoyle 

That sounds like a good approach and definitely a change I can make. ✅ 

I just wasn’t sure if changing ***how*** the `Preloader` issues and caches `fetch` requests was desired, but seems it'd be better for Turbo.

The `Preloader` would then now:

- Preload links on load.
    - Those having `data-preload="true"`
- Preload links on hover.
    - If opted-in, as @afcapel suggests. 
    - Not sure what the name would then be, but I like the "predictive" concept as @airblade suggested, so for now I'm going with it, e.g. `<meta name="turbo-predictive-preload" content="true/false">` and `data-turbo-predictive-preload="true/false"`
    - Via the same mechanism as the prefetching proposed in this PR (placing the cached `fetch` request in the `event.detail.response` for `FetchRequest` to use)

This is also a good opportunity to use a `MutationObserver` as @pfeiffer suggested.

Let me know if I'm missing something.

However, I'd first address @afcapel changes as [mentioned here](https://github.com/hotwired/turbo/pull/1101#issuecomment-1848227012), to leave this PR ready to merge, and then work on this. If the PR is merged before, I'll open a subsequent PR.

davidalejandroaguilar:
@brandondrew 

> Why not allow both approaches?
> 
> opt-in globally, with the ability to opt-out on specific links
> opt-out globally (the default—opt-out by not opting in) and only opt-in on specific links. This is more flexible.
> If there's a need for a meta tag in the head before you can opt in, then https://github.com/hotwired/turbo/pull/2 could involve an explicit page-level opt-out, which could allow opting in on specific links.

.

> While I agree that you'll often want to apply predictive preloading to a large number of links and immediate preloading (on page load) to a smaller number, I don't think the ideal API would require remembering that they're handled in totally different ways.
> 
> Wouldn't developer flexibility be best? Allow the developer to specify either at a page-wide or a link-specific level.

I agree with this. I think @afcapel also thinks this is most appropriate. I will make the change to be opt-in on specific links or containers ✅

> referring to this sort of preloading as "predictive" is the clearest and most accurate (as per @airblade's suggestion). 

I agree with this, will make the change ✅

Since we'll be adding InstantClick behavior to the `Preloader` instead and changing to be opt-in. I'm going for `<meta name="turbo-predictive-preload" content="true/false">` and `data-turbo-predictive-preload="true/false"`.

However, before the name change, I will first address @afcapel changes as [mentioned here](https://github.com/hotwired/turbo/pull/1101#issuecomment-1848227012), to leave this PR ready to merge, and then work on it. If the PR is merged before, I'll open a subsequent PR.

davidalejandroaguilar:
@afcapel 

I have added your suggestions. The only one that is still pending is changing from:

* "global opt-in via `<meta>`, then per link or per container opt-out via `data-turbo-prefetch="false"` or `data-turbo="false"`"

To:

*  "only opt-in per link or per container" 

Though I wonder if we should support both.

However, with the new delay option, the flurry you talk about is _greatly_ mitigated! Because requests are cancelled if the user mouses-out and the prefetch wasn't done yet. Here's a video demonstrating that:

1. No delay, ends up in a flurry when moving the mouse.

https://github.com/hotwired/turbo/assets/15317732/cf099c3d-52d2-407e-968c-ebf99ff72360

2. Adding the [recommended](http://instantclick.io/preloading#delay) 100ms max. Notice how there's no more flurry of requests, yet when I do a small intent pause before clicking, I get a prefetch.

https://github.com/hotwired/turbo/assets/15317732/e5286ce7-c984-4b3d-9024-a278dd8031f4

This means that we could tentatively merge the PR to start getting feedback 🚀, then continue the pending suggestion mentioned above, as well as the follow-up change to reuse the `Preloader` and possibly rename to `predictive`.

Let me know what you think.

brandondrew:
> 2. Adding the [recommended](http://instantclick.io/preloading#delay) 100ms max. Notice how there's no more flurry of requests, yet when I do a small intent pause before clicking, I get a prefetch.

This is **spectacular**!  Although after watching the videos a few times, 100ms actually seems like a greater-than-desirable delay, but I'm not sure if others will agree with that.  I assume adjusting it should be simple?

davidalejandroaguilar:
>100ms actually seems like a greater-than-desirable delay

If you add a delay, InstantClick [recommends](http://instantclick.io/preloading#delay) between 50ms and 100ms.

>More than 100 ms may actually be worse than on mousedown, less than 50 ms likely won’t be perceptible.

Adjusting this value is easy:

* Adding a `<meta name="turbo-prefetch-delay" content="100">` tag, in milliseconds, defaults to none. Applies to all links.

* Tweak it per link, via `data-turbo-prefetch-delay`.

davidalejandroaguilar:
@seanpdoyle 

I have your suggestions ready on another branch, it's on a single commit and [can be seen here](https://github.com/davidalejandroaguilar/turbo/compare/davidramos-add-instantclick-behavior...davidalejandroaguilar:turbo:davidramos-reuse-preloader-on-instantclick-behavior?expand=1).

It's an alternate implementation of this PR to reuse the `Preloader` for both on-load and predictive preloading instead of adding a new prefetch observer:

* `Preloader` no longer writes to `SnapshotCache`, instead, caching is done at the HTTP level and by setting the request on `event.detail.fetchRequest` for `FetchRequest` to reuse (like this PR does).
* `Preloader` no longer "observes" for links to preload. Instead, that responsibility is offloaded to a new `LinkPreloadObserver`, which is responsible for checking for `data-turbo-preload` elements to preload on load.
  * It now has a `MutationObserver` to detect the `data-turbo-preload` attribute added dynamically.
  * If predictive preload is enabled, it will also observe links to preload on hover.
  * When either on-load or predictive preload happen, it notifies its delegate.
  * The delegate no longer fetches and caches, this is now the `Preloader` responsibility, so it calls it.
* Renames the "prefetch" concept to "predictive-preload".
  * e.g. `data-turbo-predictive-preload="true"` and `<meta name="turbo-predictive-preload" content="true">`

@afcapel 

Please let me know if you'd want me to cherry-pick that commit here.

davidalejandroaguilar:
Just a friendly reminder about this 😄 

* @afcapel [Please see here](https://github.com/hotwired/turbo/pull/1101#issuecomment-1848357462) for your changes.
* @seanpdoyle [Please see here](https://github.com/hotwired/turbo/pull/1101#issuecomment-1849332818) for yours.

Y'all are probably busy with life and other stuff, so I'll ping back after the holidays. Have a great time!

afcapel:
@davidalejandroaguilar I've been very busy with another project and I haven't had the bandwidth to give this the attention it deserves. But I'll come back to it as soon as I have a chance.

dhh:
Very excited for this! Have wanted it for the longest time. Thanks for putting in the work. Will review with @afcapel shortly, and we will definitely have this as part of Turbo 8 👌

afcapel (commented):
@davidalejandroaguilar I gave this another try and the delay solves the problem of the excessive number of requests when you casually scroll on a page with many links. Nice work 👌

I'm very excited about this, the speed improvement is very noticeable. So much so, that at some point we'd like to have it enabled by default in Turbo apps. It fits the Turbo philosophy where you just need to add the library to a website and it makes the navigation faster. But to do that we’ll need some sensible defaults that work for most people in most situations. In most cases, you shouldn’t have to think about which events you need to use or what’s the best delay value.

For the delay, we can start with an educated guess, say 100ms, and fine-tune it by testing on our applications. For the events, we can start listening to `mousenter`, on desktop, and `touchstart` on mobile. I’m not sure we need to expose the setting to choose the events as a public API. We should be able to reach some defaults that work generally.

Another thing we’ll need is better HTML scoping of the feature. If this is going to be generally enabled we need more granular control to disable/enable it on general containers, not just the whole page. I think we can have an data-turbo-prefetch=“true/false” attribute (similar to `data-turbo`), so you can do something like this:

```html
<body data-turbo-prefetch=“false”> <!— disable prefetching on the page by default —>
   …
   <div data-turbo-prefetch=“true”>
      <!— links in here are prefetched —>
  </div>
</body>
```

Lastly, I found another issue during my tests, where sometimes we can show stale content. For example:

1. You navigate to a message on a forum app
2. Hover the backlink to the messages index page, which prefetches it
3. Add a comment to the message
4. Go back to the messages index page, clicking on the link you initially hovered.
5. The messages page shows stale content, with the wrong number of comments on the message.

To fix this I think we’ll need to clear the cache, either when there’s a form submission, a navigation, or maybe on `mouseleave`. `mouseleave` is probably the simplest option.

afcapel:
Re. using the Preloader for both Instaclick and data-turbo-preload, I think both features are going to be different enough that it'll be better to keep the bulk of their implementation separated.

1. Instaclick should be enabled by default, data-turbo-preload is a per-link setting.
2. They will have different APIs if we allow enabling/disabling Instaclik on containers.
3. data-turbo-preload loads the page in the Turbo cache, Instaclick caches the HTTP request (which also makes me think data-turbo-prefetch is a good name).

I think handling all the possible combinations in the same place is going to be messy. We can extract, however, some smaller helpers, for example, to decide if a link is preloadable.

davidalejandroaguilar:
@afcapel Thank you for reviewing!

Got it, that all makes sense, will make the changes and ping you back soon.

src/tests/server.mjs
```diff
@@ -151,6 +153,127 @@ router.put("/messages/:id", (request, response) => {
...
+router.get("/posts", async (request, response) => {
+  const { worker_id } = request.query
+
+  const posts = await getPosts(worker_id)
+
+  const res = templateRenderer.render("./posts", { posts, worker_id })
+
+  response.type("html").status(200).send(res)
+})
+
```
davidalejandroaguilar, line 165: These routes are for app-like behavior (with persistence). They require a `worker_id` param to avoid flaky tests by ensuring that each test worker has its own database file. Since tests are running in parallel, a single database file would be shared between all workers.

This can still be accessed on your local browser by starting the server and visiting `http://localhost:9000/__turbo/posts?worker_id=0` (any `worker_id` is fine).

src/tests/server.mjs
```diff
@@ -151,6 +153,127 @@ router.put("/messages/:id", (request, response) => {
...
+const ensureDatabase = async (worker_id) => {
+  if (worker_id == null || worker_id == undefined) {
+    throw new Error("worker_id is required")
+  }
+
+  if (!fs.existsSync(postsDatabaseName(worker_id))) {
+    fs.writeFileSync(postsDatabaseName(worker_id), "[]")
+  }
+}
+
```
davidalejandroaguilar, line 231: I'm using `async` even when using synchronous functions, just in case we ever decide to use some other database implementation, like sqlite. In such case, we'd only modify these reader/writer functions.

davidalejandroaguilar (commented):
Ready for another review @afcapel, very excited to get this moving! 🚀 

Here's a summary of the changes:

1. Delay is no longer configurable. Defaults to 100ms as InstantClick [recommends](http://instantclick.io/preloading#delay).
2. Events are no longer configurable. Defaults to `mouseenter` on desktop and `touchstart` on mobile.
3. Clears cache after form submission (fixes stale content issue).
4. Fixes landing on a page with `turbo-prefetch="false"`, then navigating to a page where it's true, and prefetching not being enabled.
5. The more granular scoping control you mentioned was already possible, but added a test for this case:

```html
<div data-turbo-prefetch="false">
  <div data-turbo-prefetch="true">
    <!— Links in here should be prefetched —>
  </div>
</div>
```

dhh:
Excellent work, @davidalejandroaguilar 👏

afcapel:
Great work @davidalejandroaguilar. I'm currently testing this on Basecamp and will come back to you as soon as possible.

src/observers/link_prefetch_observer.js
```diff
@@ -0,0 +1,184 @@
...
+        event.detail.fetchRequest = cached.fetchRequest
+        prefetchCache.clear()
+      }
+    }
+  }
+
+  prepareRequest(request) {
+    const link = request.target
+
+    request.headers["Sec-Purpose"] = "prefetch"
```
pfeiffer, line 106: `Sec-Header` seems to be stripped from the actual request headers sent by `fetch(..)` - could be it's a reserved header in Chromium? See https://github.com/hotwired/turbo/issues/1107

pfeiffer: Relevant reference is [this](https://www.w3.org/TR/fetch-metadata/#sec-prefix):

> 4.2. The Sec- Prefix
Each of the headers defined in this document is prefixed with Sec-, which makes them all [forbidden response-header names](https://fetch.spec.whatwg.org/#forbidden-response-header-name), and therefore unmodifiable from JavaScript. This will prevent malicious websites from convincing user agents to send forged metadata along with requests, which should give sites a bit more confidence in their ability to respond reasonably to the advertised information.

> These are forbidden so the user agent remains in full control over them.
> [Header names](https://fetch.spec.whatwg.org/#header-name) starting with `Sec-` are reserved to allow new [headers](https://fetch.spec.whatwg.org/#concept-header) to be minted that are safe from APIs using [fetch](https://fetch.spec.whatwg.org/#concept-fetch) that allow control over [headers](https://fetch.spec.whatwg.org/#concept-header) by developers, such as [XMLHttpRequest](https://xhr.spec.whatwg.org/#xmlhttprequest). [[XHR]](https://fetch.spec.whatwg.org/#biblio-xhr

davidalejandroaguilar: * Chromium [is still working](https://chromestatus.com/feature/6247959677108224), on replacing `Purpose: prefetch` with `Sec-Purpose: Prefetch`.

* Firefox [has now](https://bugzilla.mozilla.org/show_bug.cgi?id=1836328) replaced `X-Moz: Prefetch` with `Sec-Purpose: Prefetch`.

So looks like it's just a matter of time before it's sent. As @rik [pointed out](https://github.com/hotwired/turbo/issues/924#issuecomment-1561480326), and @afcapel [too](https://github.com/hotwired/turbo/pull/925#issuecomment-1691788329), using this header instead of a custom one would be more forward looking.

---

By the way, the reference you link mentions _specific_ headers that are forbidden, and `Sec-Purpose` is not in that list.

Here are some specifications that do mention that `Sec-Purpose` with a value of `prefetch` is valid:

* [Fetch Standard](https://fetch.spec.whatwg.org/#sec-purpose-header)
* [WICG spec](https://wicg.github.io/nav-speculation/prefetch.html#sec-purpose-header)

Though you're probably already aware of them, since you made the change to `Sec-Purpose": "prefetch"`.

I think we should continue this conversation in [the issue you opened](https://github.com/hotwired/turbo/issues/1107) though, since it's a very particular topic. And we can follow-up with a PR to change the header if needed.

pfeiffer: The issues linked for [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1836328) and [Chromium](https://chromestatus.com/feature/6247959677108224) are both related to what header is sent when prefetching _link-rel prefetch requests_ and does not mention any changes in the `fetch` request.

In the [Fetch Standard](https://fetch.spec.whatwg.org/#forbidden-header-name) under the "Forbidden Header Name" section, the following paragraph explicitly mentions that any header starting with `Sec-` should not be allowed to be set by client (eg. via JS), but rather be fully controlled by the user agent:

> A [header](https://fetch.spec.whatwg.org/#concept-header) (name, value) is forbidden request-header if these steps return true:
> ...
> * If name when [byte-lowercased](https://infra.spec.whatwg.org/#byte-lowercase) [starts with](https://infra.spec.whatwg.org/#byte-sequence-starts-with) `proxy-` or `sec-`, then return true.

The linked spec also mentions _why_ these are reserved, even in the future:

> Note: These are forbidden so the user agent remains in full control over them.
> [Header names](https://fetch.spec.whatwg.org/#header-name) starting with `Sec-` are reserved to allow new [headers](https://fetch.spec.whatwg.org/#concept-header) to be minted that are safe from APIs using [fetch](https://fetch.spec.whatwg.org/#concept-fetch) that allow control over [headers](https://fetch.spec.whatwg.org/#concept-header) by developers, such as [XMLHttpRequest](https://xhr.spec.whatwg.org/#xmlhttprequest). [[XHR]](https://fetch.spec.whatwg.org/#biblio-xhr)

As I understand the above this means that:

1. The `Sec-Purpose` header is **intentionally** reserved for user agents only and **intentionally not** allowed to be set by XMLHttpRequest.
2. The Chromium and Firefox changes are unrelated to this, as they change the header being sent on `<link rel=prefetch>`
3. There's no future plan of making `Sec-Purpose` available to XMLHttpRequest, on the other hand any `Sec-` header will not be allowed via XMLHttpRequest.

davidalejandroaguilar: Thanks for sharing additional details, I'd be happy to continue the discussion on the issue you opened. This information would be more valuable there for others that want to chime in on this.

src/observers/link_prefetch_observer.js
```diff
@@ -0,0 +1,184 @@
...
+  getMetaContent,
+  findClosestRecursively
+} from "../util"
+import { prefetchCache, cacheTtl } from "../core/drive/prefetch_cache"
+
+export class LinkPrefetchObserver {
+  started = false
+  delayBeforePrefetching = 100
+  hoverTriggerEvent = "mouseenter"
+  touchTriggerEvent = "touchstart"
```
seanpdoyle, line 13: In addition to mouse and touch events, should the observer also monitor [focusin](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusin_event) events for keyboard navigators to also benefit from pre-fetching?

afcapel:
@davidalejandroaguilar I've given this a good try on Basecamp. It's a good use case because it's a large app with plenty of tests and being a bit older it also has more complicated use cases where Turbo needs to play nicely with older JavaScript.

In all those tests, I kept finding some cases where some stale content was being rendered. Clearing the cache on Turbo submission helps, but Basecamp 4 has plenty of screens that use older JavaScript and perform destructive actions without Turbo. I also found a case in which the cache was used in a redirection, after a form submission that didn't use Turbo.

But I think there's an easy way to sidestep this problem. Instead of keeping a cache of all the links we've hovered in the last 10 seconds, we can just keep a reference to a single request, for the link we are currently hovering. We clear up any existing cache entry each time we `mouseover` (or `touchstart`) a new link.

So let's say this happens:

1. You hover a link for more than 100ms. That launches a prefetch request.
2. You move away from the link and perform a destructive action.
3. Come back to the link and this time click on the link.

The cache entry created in 1. doesn't matter because we clear it up when we re-enter the link in 3. This is still very fast most of the time and avoids stale content problems. The cache may be a bit less efficient if you hover a link, move away and then go back to it. But trying to use the cache in this situation is what's introducing the stale content in the first place.

I've given this a try in fb58b6b and it solves all the stale content problems. With that in place, I think we can merge the PR and make any improvements, for example for keyboard navigation, in other PRs. What do you think?

src/observers/link_prefetch_observer.js
brunoprietog: This would trigger requests all the time when pressing tab, right? Or by browsing the page with the keyboard as a screen reader user does. I think it could be a machine gun of requests in that case.

However, if the focusin event is not considered, screen reader users cannot take advantage of this feature at all.

davidalejandroaguilar:
@afcapel That does make sense to me. We should prioritize compatibility with those cases over prefetching optimization. I have pushed that commit. 

I also like the `Session` now only checking if it's visitable and the prefetch observer creating the request to be used later 👍🏼 

Would like to note that I have re-added cancelling the prefetch if the link is no longer hovered, otherwise the flurry of requests when casually scrolling comes back.

Finally, I've been testing the new delay default of 100ms and I'm thinking we might want to allow this to be configured. There might be some apps that are fine with having no delay and would prefer to prefetch as soon as possible.  However, to avoid further delaying of this PR, we could add that on a subsequent PR if you agree on this.

Otherwise, I think we're in a good shape now for merging. Thanks again for the thorough testing on a big app like Basecamp! 💪🏼

src/observers/link_prefetch_observer.js
davidalejandroaguilar: That's a great idea, as @afcapel noted, let's add that on a follow-up PR!

davidalejandroaguilar: @brunoprietog The delay feature would prevent this. 

i.e. It would prevent a flurry of requests from happening if the user is fast navigating links using the Tab key. However, once the user stops at a particular link, we'd prefetch it.

brunoprietog: I'm not so sure that value is satisfactory enough, as an average screen reader user navigates more slowly. This is because when focusing on a control, you have to listen to what it contains before making a decision. It could even be seconds. I really think it's more complicated than I thought.

guillaumebriday:
Another check that could be done, is the ensure the connection if good before prefetching data

In my controller [stimulus-prefetch](https://github.com/stimulus-components/stimulus-prefetch), I checked if the connection isn't in `saveData` mode and if it's fast enough to prefetch new pages

See the code here 👉 https://github.com/stimulus-components/stimulus-prefetch/blob/master/src/index.ts#L38-L48

What do you think?

afcapel:
> Another check that could be done, is the ensure the connection if good before prefetching data

@guillaumebriday I like the idea 👍 We don't need it for this PR, but it's something we can add later.

afcapel:
> Finally, I've been testing the new delay default of 100ms and I'm thinking we might want to allow this to be configured.

I'm OK with that. The[ menu is omakase](https://rubyonrails.org/doctrine) but substitutions are still possible.

afcapel:
Great work @davidalejandroaguilar 👏

davidalejandroaguilar:
@afcapel Thank you! And thanks again for your time. Excited for all the awesomeness that's coming for everyone on Turbo 8! 🚀 

I'll open a follow-up PR for the configurable delay and another one for the connection health check.

afcapel:
@guillaumebriday I've just released v8.0.0.beta.3!

brandondrew:
> Finally, I've been testing the new delay default of 100ms and I'm thinking we might want to allow this to be configured. 

I strongly agree with this

Kagayakashi:
Is it possible to implement caching to prevent multiple requests for a single link? Currently, every time I hover over the same link, a new request is made. Why isn't this request cached? If the concern is that cached data might be outdated, we could consider making a second request upon click. Alternatively, is it better to use a delay? However, if the user moves the mouse over the link too quickly and clicks, wouldn't this result in no request being made?

davidalejandroaguilar:
@Kagayakashi Hey there! 👋🏼 The initial implementation cached requests internally for a configurable duration of 10 seconds.

However, during testing with a big app like Basecamp, @afcapel noticed this approach could lead to stale content, particularly in areas using older Javascript without Turbo. So for now, we opted to prioritize compatibility over optimization. Despite this, we're still able to harness the benefits of HTTP caching!

Regarding the delay, it was added to prevent a flurry of requests when casually scrolling over a list of links. There are demo videos showing a before and after for this on the PR description.

brandondrew:
>  @afcapel noticed this approach could lead to stale content particularly in areas using older Javascript without Turbo. 

Why disable it globally, instead of making it configurable?

> So for now, we opted to prioritize compatibility over optimization. 

Why not do that by making the default configuration as compatible as possible?

> Despite this, we're still able to harness the benefits of HTTP caching!

Are you referring to caching by proxy servers?  By the browser, by default?

doits:
Is there any chance to configure this to work like this?

- Prefetch and cache *once* on hover, even when mouse is moved out and in again
- When clicking the link, display the cached version but make a new get request in parallel to update the page (in case something changed)

The cache could be configured longer then, e.g. 1 minute or so, because there will be always a new fetch after visiting the cached link.

The current default behaviour fetches the same page multiple times, just scrolling an moving the mouse over the a page. Is this really a good default choice? I know the discussion was lengthy already, but maybe it can be reconsidered?

(For my case I will simply disable it, but I'd prefer to have a solution that works like above with one cached fetch and a second fetch on page visit.)

paulodeon:
This is a bizarre default!

jarvis-cochrane:
Just flagging that that the PR says:

> * This behavior is **disabled by default** and can be opted-in via a `<meta name="turbo-prefetch" content="true">` tag.

However, the *code* says:

`if (link.getAttribute("data-turbo-prefetch") === "false") return true`

Which suggests that this feature is actually **enabled** by default. It's no great effort to add the appropriate `meta` tag to globally disable this, but having prefetching disabled by default would probably have been a better choice, and consistent with the PR.

davidalejandroaguilar:
@jarvis-cochrane It was enabled by default in a subsequent PR https://github.com/hotwired/turbo/pull/1162!

jarvis-cochrane:
> @jarvis-cochrane It was enabled by default in a subsequent PR #1162!

Thank you!

PR #1102, opened 2023-12-06
Page refreshes: Replace remaining calls to nextAnimationFrame by nextRepaint

brunoprietog:
In #1042, `nextRepaint` util was created, which only calls `nextAnimationFrame` when the page is in the foreground, and when it is in the background it calls `nextEventLoopTick`. Then, `nextAnimationFrame` was replaced in the `StreamElement` `render` method to allow Turbo streams actions to run even if the page is in the background.

Now that a page refresh can also be triggered by Turbo streams, it is necessary to change the remaining calls to `nextAnimationFrame`, so that page refreshes and Turbo frame reloads can occur in the background as well.

So, this change ensures that page refreshes work properly when the page is in the background.

### DIFF
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -1,7 +1,7 @@
 import { FetchMethod, FetchRequest } from "../../http/fetch_request"
 import { getAnchor } from "../url"
 import { PageSnapshot } from "./page_snapshot"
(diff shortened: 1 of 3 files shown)


afcapel:
Thanks @brunoprietog!

Issue #1107, opened 2023-12-14
v8.0.0-beta1: Preloader: `Sec-Purpose` header is ignored by `fetch(..)`


pfeiffer:
Previously the header for indicating that a resource was preloaded was named `VND.PREFETCH`. The dot in the header name caused it to be stripped by reverse proxies, including CloudFlare, as discussed in #924 

It was discussed that it should be renamed, originally the PR I made in #925 suggested `x-purpose`. However, it was suggested that the header should instead be named `sec-purpose` to match a newer spec. This was done in #925 and merged and released as part of v8 beta.

However, it seems that `fetch(..)` - at least in Chrome - strips this header from the request being made.

To reproduce, fire a manual `fetch(..)` from the console:

```
fetch("https://.../", { headers: { "sec-purpose": "preview", "x-something": "foo" })
```

Inspecting the request headers, the `sec-purpose` is not sent, while the `x-something` is:

![image](https://github.com/hotwired/turbo/assets/195925/574f755b-d4c3-46c6-84bb-2972b47d14f0)

I guess the `sec-...` are reserved headers and that fetch strips these out before sending.

I propose that we rename the header to `x-purpose` which seems to be the header used by Safari and Chrome (for the preview on the launch pages), as well as Facebook in-app browser. I'll wrap up a PR.

pfeiffer:
As discussed in [#1101](https://github.com/hotwired/turbo/pull/1101#discussion_r1454780688) the `Sec-` prefixed headers are intentionally not allowed to be set via `fetch(..)`:

In the [Fetch Standard](https://fetch.spec.whatwg.org/#forbidden-header-name) under the "Forbidden Header Name" section, the following paragraph explicitly mentions that any header starting with `Sec-` should not be allowed to be set by client (eg. via JS), but rather be fully controlled by the user agent:

> A [header](https://fetch.spec.whatwg.org/#concept-header) (name, value) is forbidden request-header if these steps return true:
> ...
> * If name when [byte-lowercased](https://infra.spec.whatwg.org/#byte-lowercase) [starts with](https://infra.spec.whatwg.org/#byte-sequence-starts-with) `proxy-` or `sec-`, then return true.

The linked spec also mentions _why_ these are reserved, even in the future:

> Note: These are forbidden so the user agent remains in full control over them.
[...]

PR #1108, opened 2023-12-14
Preloader: Rename header to `x-purpose` to avoid it being stripped

pfeiffer:
This PR addresses the issue mentioned in #1107 by renaming the header to `x-purpose`, as the `Sec-Purpose` header is stripped by `fetch`.

In this PR, the value is also changed to `preview` which is consistent with how eg. [Safari and Chrome uses the `x-purpose` header for their "quick view" of eg. favorite pages](https://stackoverflow.com/questions/1483008/what-is-the-definition-of-http-x-purpose).

### DIFF
--- a/src/core/drive/preloader.js
+++ b/src/core/drive/preloader.js
@@ -37,7 +37,7 @@ export class Preloader {
     }
 
     try {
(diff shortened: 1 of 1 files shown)


afcapel:
Thanks @pfeiffer and @seanpdoyle!

PR #1110, opened 2023-12-18
Set aria-busy on the form element during a form submission

afcapel:
This is useful for styling the form while it is submitting.

Before this change, we were only setting aria-busy on frames while they were loading and on the html element while Turbo was processing a visit.

### DIFF
--- a/src/core/drive/form_submission.js
+++ b/src/core/drive/form_submission.js
@@ -1,6 +1,6 @@
 import { FetchRequest, FetchMethod, fetchMethodFromString, fetchEnctypeFromString, isSafe } from "../../http/fetch_request"
 import { expandURL } from "../url"
-import { dispatch, getAttribute, getMetaContent, hasAttribute } from "../../util"
+import { clearBusyState, dispatch, getAttribute, getMetaContent, hasAttribute, markAsBusy } from "../../util"
 import { StreamMessage } from "../streams/stream_message"
 
 export const FormSubmissionState = {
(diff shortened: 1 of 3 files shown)


dhh:
Wonder if we actually need to set it on the form element itself or if setting it on the body is enough? With frames, do we set it just on the frame or on the root document body? I guess if frames set it on their own element, setting busy on the form makes sense too.

afcapel:
Yes, frames set [it on their element](https://github.com/hotwired/turbo/blob/6788194d61a8feaba0fd5c6408bfa2990326a06a/src/core/frames/frame_controller.js#L203). I think it's better to set it on the form, just in case there are several forms in the page.

seanpdoyle:
According to the specification, elements with [role="form](https://www.w3.org/TR/wai-aria-1.0/roles#form) do support the `[aria-busy]` attribute.

The [aria-busy](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-busy) portion of the specification mentions that it

> Indicates whether an element, and its subtree, are currently being updated.

I think there's some room to interpret what "being updated" means to a `<form>` element. 

With that being said, setting `<body aria-busy="true">` might supersede anything that Turbo would do to elements deeper in the document, which has the potential to make this change moot from an assistive tech perspective.

afcapel:
@seanpdoyle as far as I can tell we're not setting `[aria-busy]` on the body on form submissions, only on page navigations.

dhh:
Given that we already have it scoped with frames, it seems best to follow that pattern with forms too. Especially, as @afcapel says, you could have multiple forms per page, and we want to be able to use aria-busy to style which one is submitting.

PR #1123, opened 2024-01-08
Check new snapshot (instead of previous) for refresh-method

scuml:
While trying to port the Turbo Morph Demo over to Django, I found that if a form submission returns a classic django error page, the morph renderer is still used, causing an undesired blend of the previous pages <head> tags with the error page's content.  

Since morphing is predicated on the idea of a full-page refresh, the tag `<meta name="turbo-refresh-method" content="morph">` should be on all responses that intend to be morphed, and if not (as the case of an error page), it would be ideal to fallback on the full page renderer.  

This PR checks both the current page AND the upcoming snapshot for the meta setting `refresh-method='morph'` to determine if the morph renderer should be used, instead of just the originating page.

https://github.com/hotwired/turbo/assets/1790447/96d7ec6e-e791-4fae-819b-bc73151a45e1

### DIFF
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -16,8 +16,7 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const shouldMorphPage = this.isPageRefresh(visit) && snapshot.shouldMorphPage;
-    console.info("shold morph", shouldMorphPage);
+    const shouldMorphPage = this.isPageRefresh(visit) && snapshot.shouldMorphPage
     const rendererClass = shouldMorphPage ? MorphRenderer : PageRenderer
 
     const renderer = new rendererClass(this.snapshot, snapshot, PageRenderer.renderElement, isPreview, willRender)
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -16,7 +16,7 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const shouldMorphPage = this.isPageRefresh(visit) && this.snapshot.shouldMorphPage
+    const shouldMorphPage = this.isPageRefresh(visit) && snapshot.shouldMorphPage
     const rendererClass = shouldMorphPage ? MorphRenderer : PageRenderer
 
     const renderer = new rendererClass(this.snapshot, snapshot, PageRenderer.renderElement, isPreview, willRender)
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -16,7 +16,8 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const shouldMorphPage = this.isPageRefresh(visit) && snapshot.shouldMorphPage && this.snapshot.shouldMorphPage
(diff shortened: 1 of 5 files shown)


scuml:
@seanpdoyle @afcapel Could you give this a quick review.  This is behavior a minor source of friction getting Turbo 8 adopted on Django projects.

src/core/drive/page_view.js
```diff
@@ -16,7 +16,7 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const shouldMorphPage = this.isPageRefresh(visit) && this.snapshot.shouldMorphPage
+    const shouldMorphPage = this.isPageRefresh(visit) && snapshot.shouldMorphPage && this.snapshot.shouldMorphPage
```
seanpdoyle, line 19: @scuml thank you for opening this PR.

Conceptually, I wonder if we should be checking the current snapshot at all. For example, when determining whether to initiate a [Visit or a Full Reload](https://turbo.hotwired.dev/handbook/drive#ensuring-specific-pages-trigger-a-full-reload), The `PageRenderer` only scans the _new_ snapshot head for a `<meta>` element:

* [PageRenderer.shouldRender](https://github.com/hotwired/turbo/blob/00527e50ed1a4298ce52f968f2a007f6cfd30776/src/core/drive/page_renderer.js#L14)
* [PageRenderer.reloadReason](https://github.com/hotwired/turbo/blob/00527e50ed1a4298ce52f968f2a007f6cfd30776/src/core/drive/page_renderer.js#L18)

Similarly, when using the same `<meta>` to determine whether to ["Break out" of a Frame](https://turbo.hotwired.dev/handbook/frames#“breaking-out”-from-a-frame), the `FrameController` scans the _new_ snapshot:

* [FrameController.loadResponse](https://github.com/hotwired/turbo/blob/00527e50ed1a4298ce52f968f2a007f6cfd30776/src/core/frames/frame_controller.js#L136)

@afcapel @jorgemanrubia I wonder, is this sufficient?

```suggestion
    const shouldMorphPage = this.isPageRefresh(visit) && snapshot.shouldMorphPage
```

scuml: Interesting you mention that, as that was my initial implementation: https://github.com/hotwired/turbo/pull/1123/commits/c3eacbc25be112adf8663f4c662dcf87bdcc5bcf

I put the other check in as a safety if I missed a use-case - but thinking of it along the lines of "breaking out of a frame", I think you are right that the 'current page check' can be removed.  

This would also offer the user an extra degree of freedom where they can now opt-out and opt-back in to a morph render.

scuml:
Tests passing and ready to merge.

pfeiffer:
Would it make sense to update the title of the PR to reflect the actual implemented behavior (ie. only check the new snapshot)?

scuml:
Title updated.  Could we get this merged in?

seanpdoyle-intercom:
@jorgemanrubia @afcapel are either of you able to review this change?

scuml:
@jorgemanrubia @afcapel -- another request for your review.  Should be quick.

jorgemanrubia (approved):
Thanks @scuml. I think this one is good to merge @afcapel!

scuml:
@afcapel Approved and ready to merge.

afcapel:
Thanks @scuml !

Issue #876, opened 2023-02-17
turbo:load not fired after turbo:click when using anchor to navigate on the same page


overdrivemachines:
In my rails app, I have a few links on a page that are used to navigate within the page using anchor tags' href="#idname" as shown:
```erb
<%= link_to "Home", root_path(anchor: "home") %>
<%= link_to "About", root_path(anchor: "about") %>
<%= link_to "Portfolio", root_path(anchor: "portfolio") %>
<%= link_to "Login", login_path %>
```
When I click on Login everything works as expected. The turbo:click event fires and then turbo:load fires, just as expected. However, when I click on Home or About when I'm already on the root_path, then things don't work as expected. The turbo:click event fires and nothing else fires. Can we change this behavior? In my app, whenever there is a turbo:click event a loader image appears and it disappears when the turbo:load event fires. With only turbo:click event firing, my app is stuck showing a loader image.

marcoroth:
Hey @overdrivemachines, thanks for opening this issue!

I don't think it 100% makes sense to also emit a `turbo:load` event in that case. The `turbo:load` event is just fired when a Turbo visit is being made, which is not the case here. 

In this case it's just operating on the current page, as no new data is being fetched. I think if anything, then the `turbo:click` event shouldn't even fire in that case, since Turbo isn't involved.

Does that make sense?

tannakartikey:
@overdrivemachines what if you disable `turbo` for these links? Will that solve your issue?


> In this case it's just operating on the current page, as no new data is being fetched. I think if anything, then the turbo:click event shouldn't even fire in that case, since Turbo isn't involved.

@marcoroth this is something to think about.

overdrivemachines:
> @overdrivemachines what if you disable `turbo` for these links? Will that solve your issue?

I could disable turbo for those links but what if I'm in the login url then I want turbo enabled because otherwise there will be a full page reload when the user wants to navigate to the home page.

overdrivemachines:
> In this case it's just operating on the current page, as no new data is being fetched. I think if anything, then the `turbo:click` event shouldn't even fire in that case, since Turbo isn't involved.

I agree. Maybe it is designed like that for a reason. In the mean time, how do I figure out if the user is operating on the current page and no new data is fetched. For eg if I'm on the root_url and I click on root_url#about. Here is some of the code:
```js
document.addEventListener("turbo:click", () => {
  // how to check if the user is operating on the current page and no new data is going to be fetched

});
```

[...]

Issue #539, opened 2022-02-16
Turbo should not perform a visit() when tapping same-page anchors


keithamus:
I'm pretty sure this is a regression, since this issue is effectively a dupe of https://github.com/hotwired/turbo/pull/298 and https://github.com/hotwired/turbo/issues/42.

When clicking a link with an href that is an anchor on the page, even if the anchor is present in the URL, Turbo will initiate a navigation, causing the page to "reload". Here's a minimal reproduction: 

```html
<!DOCTYPE html>
<html>
  <meta charset="UTF-8">
  <script type="module">
    import {session} from "https://cdn.skypack.dev/@hotwired/turbo@7.1.0";
    session.drive = !new URLSearchParams(location.search).has('nodrive')
    const a = document.createElement('a')
    a.href = '#foo'
    a.textContent = 'I should be visible!'
    document.body.append(a)
    setTimeout(() => a.click())
  </script>
  <body>
    <h1 name="foo">Below should be a visible link</h1>
  </body>
</html>
```

### Steps to reproduce

1. Copy the above code snippet and save as `index.html` on your local filesystem
2. Spin up a web server (I used `npx http-server`, you could also use `python -m http-server 8080`, or `ruby -run -e httpd -- -p 8080 .`)
3. Visit "http://127.0.0.1:8080/#foo"
4. Observe the link "I should be visible" **flashes for a second but disappears**.
    - Expected: The link is visible
    - Actual: The link disappears.
5. Visit "http://127.0.0.1:8080/?nodrive#foo"
6. Observe the browsers default behaviour, that the link is visible and the page did not reload.

It is also possible to create an infinite reload loop with this:

```html
<!DOCTYPE html>
<html>
  <meta charset="UTF-8">
  <script type="module">
    import {session} from "https://cdn.skypack.dev/@hotwired/turbo@7.1.0";
    session.drive = !new URLSearchParams(location.search).has('nodrive')
    customElements.define('lazy-click', class extends HTMLElement {
      connectedCallback() {
        setTimeout(() => this.querySelector('a').click())
      }
    })
  </script>
  <body>
    <h1 name="foo">This page will cause an infinite load loop!</h1>
    <turbo-frame id="infinite-loop">
      <lazy-click><a href="#foo">I get clicked!</a></lazy-click>
    </turbo-frame>
  </body>
</html>
```

### Steps to reproduce

1. Copy the above code snippet and save as `index.html` on your local filesystem
[...]

Issue #592, opened 2022-05-23
Turbo visits don't set `:target` to the targeted element


manuelpuyol:
When doing a Turbo visit to a link that has an anchor `#`, the targeted element does not have the `:target` pseudo-class.

### Steps to reproduce

1. In the Turbo project, visit http://localhost:9000/src/tests/fixtures/navigation.html
2. Search for elements with `:target` pseudo-class: `$(':target')` returns `null`
3. Reload page
4. Search for elements with `:target` pseudo-class: `$(':target')` returns element with id `element-id`

### Expected behavior

After clicking a link with an anchor, it's expected that the element which matches the URL hash has the `:target` pseudo-class.

This is important for apps that use that selector to add an indicator to the element - like navigating to a specific comment in a GitHub Issue:

![image](https://user-images.githubusercontent.com/11280312/169894341-f37da035-d64f-4d5f-a286-220d207cb9e0.png)

We are happy to look into this if we can get some pointers on where hashes may be getting interfered with :)

domchristie:
This might be related to https://github.com/hotwired/turbo/issues/539.

I'm not sure there is a way to manually set the `:target` element, and unfortunately browsers don't seem to update it after `pushState`/`replaceState`. [Turbo does focus the target element](https://github.com/hotwired/turbo/blob/daabebb0575fffbae1b2582dc458967cd638e899/src/core/view.ts#L52-L62), so you might want to experiment with that.

Another option might be to create a link element behind-the-scenes with an `href` corresponding to the target element and with `data-turbo=false`, and then call `click` on it to mimic a non-Turbo navigation. (I've not tried this, but it might work.)

domchristie:
> Another option might be to create a link element behind-the-scenes

This seems to work:

```js
addEventListener('turbo:load', target)
addEventListener('hashchange', target) // for same-page navigations

function target () {
  if (location.hash) {
    const a = document.createElement('a')
    a.href = `#${location.hash.slice(1)}`
    a.click()
  }
}
```

manuelpuyol:
Wouldn't that cause a full page navigation after the Turbo visit?

domchristie:
I don't think it will because the link is not contained in the document so Turbo won't handle the click.

straight-shoota:
I've encountered this problem as well when `:target` CSS selectors stopped working with turbo navigation.
[...]

PR #1125, opened 2024-01-09
Improve handling of anchor links and hashchange events

Matt-Yorkley:
When navigating around between multiple anchor links on the same page there's a number of weird behaviours that can come up with how the navigations are handled and the subsequent `hashchange` events that get fired:

- The `hashchange` event's value for `event.oldURL` often contains an incorrect URL (wrong hash)

- Using the back or forwards buttons (restoration visits) causes two `hashchange` events to be fired each time instead of one, and the duplicate contains incorrect information

- Moving between multiple anchors on the same page can trigger an unexpected page visit in some cases, including a network request and repainting the DOM (`turbo:load`) when it shouldn't

Related issues/discussions here: https://github.com/hotwired/turbo/issues/539 and here: https://github.com/hotwired/turbo/issues/876#issuecomment-1492878908

### DIFF
--- a/src/core/drive/navigator.js
+++ b/src/core/drive/navigator.js
@@ -129,13 +129,12 @@ export class Navigator {
 
   locationWithActionIsSamePage(location, action) {
     const anchor = getAnchor(location)
-    const currentAnchor = getAnchor(this.view.lastRenderedLocation)
     const isRestorationToTop = action === "restore" && typeof anchor === "undefined"
 
     return (
       action !== "replace" &&
       getRequestURL(location) === getRequestURL(this.view.lastRenderedLocation) &&
-      (isRestorationToTop || (anchor != null && anchor !== currentAnchor))
+      (isRestorationToTop || anchor != null)
     )
   }
 
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -349,7 +349,7 @@ export class Visit {
       } else {
         this.scrollToAnchor() || this.view.scrollToTop()
         if (this.isSamePage) {
-          this.delegate.visitScrolledToSamePageLocation(this.view.lastRenderedLocation, this.location)
+          this.delegate.visitScrolledToSamePageLocation(window.location.href, this.location)
         }
       }
 
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -348,9 +348,9 @@ export class Visit {
         this.scrollToRestoredPosition() || this.scrollToAnchor() || this.view.scrollToTop()
       } else {
         this.scrollToAnchor() || this.view.scrollToTop()
-      }
-      if (this.isSamePage) {
-        this.delegate.visitScrolledToSamePageLocation(this.view.lastRenderedLocation, this.location)
+        if (this.isSamePage) {
+          this.delegate.visitScrolledToSamePageLocation(this.view.lastRenderedLocation, this.location)
+        }
       }
 
       this.scrolled = true
--- a/src/tests/fixtures/one.html
+++ b/src/tests/fixtures/one.html
@@ -14,8 +14,8 @@
     <a name="named-anchor"></a>
     <div id="element-id" style="margin-top: 1em; height: 200vh">An element with an ID</div>
     <p><a id="redirection-link" href="/__turbo/redirect?path=/src/tests/fixtures/visit.html">Redirection link</a></p>
-    <a id="named-anchor" href="#named-anchor">Named1</a>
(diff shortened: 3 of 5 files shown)


domchristie (commented):
The whole navigate-to-fragment stuff is pretty complex. Not only Turbo's code, but also browser behaviours. I'm pretty sure there's good reason for the conditionals here as it was well tested at the time, however it still throws up bugs in some situations.

So I think it may result in whack-a-mole style development.

In terms of the behaviour, I think our best option is to do something like: https://github.com/hotwired/turbo/issues/592#issuecomment-1137827028 where we create an `<a>` with the hash and programatically click it. That way we get scrolling, element focussing, and `:target` styling.

We'll still need to spec out the different cases when this is needed (e.g. a clicked link vs a programatic visit, vs a restoration visit), but hopefully this might tidy thing up a bit?

Matt-Yorkley:
Interesting! I've never used the `:target` pseudo-class before. That's mostly for styling though, right?

The above examples are relevant also to just clicking simple anchor links on the same page (without doing the navigate-to-fragment across separate pages). It's a bit sticky due to the way `this.view.lastRenderedLocation` is used in the comparison in `locationWithActionIsSamePage`. 

https://github.com/hotwired/turbo/blob/2a638e037ceda39d94f29d9912f4947935bda448/src/core/drive/navigator.js#L130-L140

That value only gets set/updated when a full visit or page load happens, and it also captures the current hash, but it doesn't change if you click an anchor link. Eg: on `/page`, click `#hash1`, then refresh the page, `this.view.lastRenderedLocation` will be `<url>/page#hash1`. Click `#hash2`, and `this.view.lastRenderedLocation` will still be `<url>/page#hash1` even though the current URL is now `<url>/page#hash2` and the current hash is `#hash2`. Know what I mean? 

So using it here as "`currentAnchor`" on line 132 can produce inconsistent results, because it doesn't necessarily hold the _current_ anchor, but whatever the anchor was last time the page was fully rendered. This leads to the `event.oldURL` being wrong in `hashchange` events, depending on what value was last stored in `this.view.lastRenderedLocation`. Does that make sense?

Matt-Yorkley:
Minimal repro/testing page for anchor links and hashchange events:

```html
<!DOCTYPE html>
<html>
<head>
  <script src="https://cdn.jsdelivr.net/npm/@hotwired/turbo@8.0.0-beta.2/dist/turbo.es2017-umd.min.js"></script>
  <script type="text/javascript">
    window.addEventListener("hashchange", (event) => {
      console.log("hashchange: ", event)
      document.getElementById("events").insertAdjacentHTML("beforeend",
        `<p>Hashchange: { oldURL: ${event.oldURL}, newURL: ${event.newURL} }</p>`
      )
    });
    window.addEventListener("turbo:before-fetch-request", (event) => {
      console.log("turbo:before-fetch-request: ", event)
      document.getElementById("events").insertAdjacentHTML("beforeend",
        `<p>Fetch request: ${event.detail.url} (cancelled for demonstration)</p>`
      )
      event.preventDefault()
      history.replaceState({}, "", event.detail.url)
    });
    function clearEvents() { document.getElementById("events").innerHTML = "" }
  </script>
  <meta content="no-cache" name="turbo-cache-control" />
</head>
<body>

<h1>Same-page anchors</h1>
<ul>
  <li><a href="">No anchor</a></li>
  <li><a href="#one">Anchor #one</a></li>
  <li><a href="#two">Anchor #two</a></li>
  <li><a href="#three">Anchor #three</a></li>
</ul>

<h2>Event log:</h2>
<div id="events">
</div>

<button onclick="clearEvents()">Clear events</button>
</body>
</html>

```

Ambient-Impact:
> Interesting! I've never used the :target pseudo-class before. That's mostly for styling though, right?

It can be used for just about anything you can do with CSS, including layout stuff. For example, it's a crucial part of our mobile menu on [Omnipedia](https://omnipedia.app), which actually works if our JavaScript fails to execute because the link to open it just points to `#menu`, which is the in-page ID. [The selectors can get pretty ugly](https://github.com/search?q=repo%3Aneurocracy%2Fdrupal-omnipedia-site-theme+path%3A%2F%5Estylesheets%5C%2F%2F+%3Atarget&type=code) and require a very specific page structure, but it works - just try turning off JavaScript and resize the browser so it's narrow like a phone, and click the Menu link in the header.

Anyways, I'm actually working on integrating Turbo into our site and ran across some of these issues with 7.3.0, notably the double `hashchange` and the `turbo:load` on the same page when changing the hash under some circumstances. Glad to see this being worked on.

brunoprietog:
Thanks for your contribution, but the plan is simplify same page anchor visits in #1285.

PR #1126, opened 2024-01-09
Revert: Proposal: pass isPreview value into the event detail in render events (#926)

domchristie:
It is not necessary to pass along isPreview through various methods in order to determine whether a render is a preview since that can be determined via the data-turbo-preview attribute. (As discussed in https://github.com/hotwired/turbo/pull/926#issuecomment-1692292545)


afcapel:
Thanks @domchristie

PR #1128, opened 2024-01-12
Remove unused stylesheets when navigating

kevinmcconnell:
When navigating to another page, Turbo merges the `<head>` contents from the current and new pages, which results in a `<head>` containing the superset of both.

For certain items, like scripts, this makes sense. We have no way to remove a running script. But that's not the case for styles: styles can be unloaded easily, and for the page to display properly, we need to do so. Otherwise styles kept in scope from a previous page could cause a page to render incorrectly.

The common way to avoid this has been to use `data-turbo-track="reload"` to force a reload if styles change. This works, but it's a bit heavy-handed, causing full page reloads that could have been avoided.

There are a couple of common cases where updating styles on the fly would be useful:

- Deploying a CSS change. Clients should be able to pick up the change without having to reload.

- Allowing pages to include their own specific styles, rather than bundle them all together for the whole site. This can reduce the size of the loaded CSS, and make it easier to avoid style conflicts.

### DIFF
--- a/src/core/drive/page_renderer.js
+++ b/src/core/drive/page_renderer.js
@@ -1,5 +1,6 @@
-import { Renderer } from "../renderer"
 import { activateScriptElement, waitForLoad } from "../../util"
+import { Renderer } from "../renderer"
+import { ProgressBarID } from "./progress_bar"
 
 export class PageRenderer extends Renderer {
   static renderElement(currentElement, newElement) {
@@ -73,8 +74,13 @@ export class PageRenderer extends Renderer {
     const mergedHeadElements = this.mergeProvisionalElements()
     const newStylesheetElements = this.copyNewHeadStylesheetElements()
     this.copyNewHeadScriptElements()
+
     await mergedHeadElements
     await newStylesheetElements
+
+    if (this.willRender) {
+      this.removeUnusedHeadStylesheetElements()
+    }
   }
 
   async replaceBody() {
@@ -103,6 +109,12 @@ export class PageRenderer extends Renderer {
   copyNewHeadScriptElements() {
     for (const element of this.newHeadScriptElements) {
       document.head.appendChild(activateScriptElement(element))
+    }
+  }
+
+  removeUnusedHeadStylesheetElements() {
+    for (const element of this.unusedHeadStylesheetElements) {
+      document.head.removeChild(element)
     }
   }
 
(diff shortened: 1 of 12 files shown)


src/core/drive/page_renderer.js
```diff
@@ -73,8 +74,13 @@ export class PageRenderer extends Renderer {
     const mergedHeadElements = this.mergeProvisionalElements()
     const newStylesheetElements = this.copyNewHeadStylesheetElements()
     this.copyNewHeadScriptElements()
+
     await mergedHeadElements
     await newStylesheetElements
+
+    if (this.willRender) {
```
kevinmcconnell, line 81: @afcapel this condition is needed because of the way frame navigations reuse the page renderer to make snapshots. We should refactor the frame head handling to make this sort of thing unnecessary. But let's do that separately, as it has wider implications.

Also, given our recent discussions about potential impact of instaclick, we should see where that's heading before pulling on the snapshotting thread.

src/tests/functional/drive_stylesheet_merging_tests.js
```diff
@@ -0,0 +1,20 @@
...
+test("navigating removes unused style elements", async ({ page }) => {
+  await page.locator("#go-right").click()
+  await nextBody(page)
+
+  assert.ok(await hasSelector(page, 'link[rel=stylesheet][href="/src/tests/fixtures/stylesheets/common.css"]'))
+  assert.ok(await hasSelector(page, 'link[rel=stylesheet][href="/src/tests/fixtures/stylesheets/right.css"]'))
+  assert.notOk(await hasSelector(page, 'link[rel=stylesheet][href="/src/tests/fixtures/stylesheets/left.css"]'))
+
+  assert.equal(await getComputedStyle(page, "p", "font-weight"), "200")
+  assert.equal(await getComputedStyle(page, "p", "font-family"), "serif")
```
afcapel, line 18: This is only asserting the common styles, we could also assert the new style.

afcapel (approved):
Looks good, nice one! 👏

src/tests/functional/drive_stylesheet_merging_tests.js
kevinmcconnell: The `serif` is asserting the new style. And the `200` is asserting that the old style has been removed, allowing the common style to come through.

It's a bit cryptic though isn't it. I'll look for a clearer way to do this.

afcapel: I mean the styles in `left.css` vs `right.css`, the difference is the background color, isn't it?

kevinmcconnell: Ah I see, sorry I misunderstood what you meant.

I was checking those by asserting whether or not the stylesheet links were present. We could verify the computed styles of those as well, though. Doesn't hurt to do both.

kevinmcconnell: OK I've changed it to assert the presence of links, and of embedded CSS rules. And also to check the computed styles that should apply once only the common & right styles are applied. (Which is a bit awkward because it needs to test that the old ones are actually gone, not just overridden, so it uses a couple different properties to do so).

I think it seems a little clearer now, as well as more thorough. What do you think?

afcapel: Nice 👌

PR #1129, opened 2024-01-13
Render `<turbo-stream-source>` as `display: none`

seanpdoyle:
Since the `<turbo-cable-stream-source>` is a Custom Element, it has a default `style:` property of `display: inline`. As a visible element, it participates in the page's layout.

This commit changes the element to always be `display: none` so that it no longer participates in the layout.


stevepolitodesign (commented):
Would [`hidden`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) be an appropriate alternative?

brunoprietog:
Looks si mpler!

PR #1133, opened 2024-01-18
Keep Trix dynamic styles in the head

afcapel:
After https://github.com/hotwired/turbo/pull/1128 we are dynamically removing styles we're no longer using on page navigations. Unfortunately, Trix dynamically adds some styles to the head that we need to keep around. Otherwise, the editor will not render correctly after a page change.

We can detect those styles because Trix also adds a `data-tag-name` attribute to the style elements it adds.

Ideally, we would move those styles to Trix's CSS file, but for now we'll just skip removing them. We don't want to force everyone upgrading to Turbo v8 to also upgrade Trix.

Ref:

- https://github.com/basecamp/trix/blob/06d8b1db5fb682d007c5ca041884f6297674c8b7/src/trix/elements/trix_editor_element.js#L108-L162
- https://github.com/basecamp/trix/blob/06d8b1db5fb682d007c5ca041884f6297674c8b7/src/trix/core/helpers/custom_elements.js#L11

### DIFF
--- a/src/core/drive/page_renderer.js
+++ b/src/core/drive/page_renderer.js
@@ -114,7 +114,12 @@ export class PageRenderer extends Renderer {
 
   removeUnusedHeadStylesheetElements() {
     for (const element of this.unusedHeadStylesheetElements) {
-      document.head.removeChild(element)
+      // Trix dynamically adds styles to the head that we want to keep around
+      // Long term we should moves those styles to Trix's CSS file
+      // but for now we'll just skip removing them
+      if (!element.dataset.tagName) {
+        document.head.removeChild(element)
+      }
     }
   }
 
--- a/src/core/drive/page_renderer.js
+++ b/src/core/drive/page_renderer.js
@@ -1,6 +1,5 @@
 import { activateScriptElement, waitForLoad } from "../../util"
 import { Renderer } from "../renderer"
-import { ProgressBarID } from "./progress_bar"
 
 export class PageRenderer extends Renderer {
   static renderElement(currentElement, newElement) {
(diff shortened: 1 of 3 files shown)


src/core/drive/page_renderer.js
```diff
@@ -114,7 +114,12 @@ export class PageRenderer extends Renderer {
 
   removeUnusedHeadStylesheetElements() {
     for (const element of this.unusedHeadStylesheetElements) {
-      document.head.removeChild(element)
+      // Trix dynamically adds styles to the head that we want to keep around
+      // Long term we should moves those styles to Trix's CSS file
+      // but for now we'll just skip removing them
+      if (!element.dataset.tagName) {
+        document.head.removeChild(element)
```
kevinmcconnell, line 121: Just an idea, but given we're ignoring certain items based on a data attribute, I wonder if we should make a more official way to opt in to that, in case people want to have the same handling for other legacy code. We could make it so that anything with `data-turbo-permanent` doesn't get removed, and then the special case for Trix is that we also treat those with the `dataset.tagName` as if they had been marked `data-turbo-permanent`. Essentially just check 2 attributes instead of 1.

The code currently has to also special-case the styles for the Turbo progress bar. It does that by ID right now, but we could change that to have `data-turbo-permanent` too.

We could look at that in a separate PR of course, after landing this fix as-is.

kevinmcconnell (approved):
Looks good to me. Shame we have to do it! But it seems like a pragmatic move for now, and the new semantics that we have still make sense 👍

src/core/drive/page_renderer.js
kevinmcconnell: One other thing -- do we know what the specific values for that `tagName` attribute will be? Could we make the check stricter by using it?

seanpdoyle: Since Turbo manages the progress bar, I really like the idea of rendering it with `[data-turbo-permanent]`.

afcapel: 👍 makes sense, I've updated the branch so now we'll keep any styles marked with `data-turbo-permanent` and also use `data-turbo-permanent` to keep the progress bar styles around.

PR #1136, opened 2024-01-22
Introduce `FrameElement.getElementById(id: string)`

seanpdoyle:
The `FrameElement.getElementById(id: string)` static method unifies the internal logic to locate a `<turbo-frame>` element by its `[id]`, while providing that consistent behavior to consumer applications.

### DIFF
--- a/src/core/session.js
+++ b/src/core/session.js
@@ -162,9 +162,9 @@ export class Session {
     const frameTarget = element.getAttribute("data-turbo-frame")
     const frame = frameTarget == "_top" ?
       null :
-      document.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame:not([disabled])")
+      FrameElement.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame")
 
-    if (isUnsafe || isStream || frame instanceof FrameElement) {
+    if (isUnsafe || isStream || (frame && !frame.disabled)) {
       return false
     } else {
       const location = new URL(element.href)
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -432,7 +432,7 @@ export class FrameController {
 
   #findFrameElement(element, submitter) {
     const id = getAttribute("data-turbo-frame", submitter, element) || this.element.getAttribute("target")
-    return getFrameElementById(id) ?? this.element
+    return FrameElement.getElementById(id) ?? this.element
   }
 
   async extractForeignFrameElement(container) {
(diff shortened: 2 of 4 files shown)


src/core/session.js
```diff
@@ -162,9 +162,9 @@ export class Session {
     const frameTarget = element.getAttribute("data-turbo-frame")
     const frame = frameTarget == "_top" ?
       null :
-      document.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame:not([disabled])")
+      FrameElement.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame")
 
-    if (isUnsafe || isStream || frame instanceof FrameElement) {
+    if (isUnsafe || isStream || (frame && !frame.disabled)) {
```
seanpdoyle, line 167: @sfnelson since `FrameElement.getElementById` accounts for `instanceof FrameElement`, does this property check account for the concerns you've shared in https://github.com/hotwired/turbo/pull/1136/files/b750a8f2e44ef3133baf6328e5187c80ae5cf2e1#r1496721094?

seanpdoyle: For the sake of shrinking the diff, both of these lines could likely be reverted in part:

```suggestion
    if (isUnsafe || isStream || frame instanceof FrameElement) {
```

src/core/session.js
```diff
@@ -162,9 +162,9 @@ export class Session {
     const frameTarget = element.getAttribute("data-turbo-frame")
     const frame = frameTarget == "_top" ?
       null :
-      document.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame:not([disabled])")
+      FrameElement.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame")
```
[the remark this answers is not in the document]
seanpdoyle:
```suggestion
      FrameElement.getElementById(frameTarget) || findClosestRecursively(element, "turbo-frame:not([disabled])")
```

src/core/session.js
sfnelson: Yeah, that looks fine.

Would it make sense to add a static helper for finding the enclosing (non-disabled) FrameElement? In that case the `findClosestRecursively` could be replaced by e.g. `FrameElement.findClosest(element)` and the check would become `if (isUnsafe || isStream || frame) {`.

sfnelson: For the purposes of code reuse, I can see two additional utility methods that would be helpful:

- `FrameElement.findEnclosingFrame(element)` would walk up the tree to find the enclosing frame (if any)
- `FrameElement.findTargetFrame(element)` would expose `#findFrameElement` – given an element it would identify which frame should be targeted – for use with links/forms that target frames for navigation

I think these would bring general utility to downstream consumers of Turbo – both of these cases have come up in our use of Frames to implement modals.

PR #1138, opened 2024-01-23
Ensure that the turbo-frame header is not sent when the turbo-frame has a target of _top

afcapel:
With Instaclick enabled and given an HTML structure like this:

```html
<turbo-frame id="frame" target="_top">
  <a href="/src/tests/fixtures/prefetched.html">Hover to prefetch me</a>
</turbo-frame>
```

When the user hovers over the link, the turbo-frame header should not be sent in the prefetch request because the wrapping turbo-frame has a target of `_top`.

If we don't respect this turbo-rails sees the turbo-frame header and renders a turbo-frame response with a minimal layout that doesn't include any assets in the head. When turbo processes the response it may find a missmatch in tracked assets and cause a reload.

//cc @davidalejandroaguilar

### DIFF
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -100,14 +100,11 @@ export class LinkPrefetchObserver {
 
     request.headers["Sec-Purpose"] = "prefetch"
 
-    if (link.dataset.turboFrame && link.dataset.turboFrame !== "_top") {
-      request.headers["Turbo-Frame"] = link.dataset.turboFrame
-    } else if (link.dataset.turboFrame !== "_top") {
-      const turboFrame = link.closest("turbo-frame")
+    const turboFrame = link.closest("turbo-frame")
+    const turboFrameTarget = link.dataset.turboFrame || turboFrame?.getAttribute("target") || turboFrame?.id
 
-      if (turboFrame) {
-        request.headers["Turbo-Frame"] = turboFrame.id
-      }
+    if (turboFrameTarget && turboFrameTarget !== "_top") {
+      request.headers["Turbo-Frame"] = turboFrameTarget
     }
 
     if (link.hasAttribute("data-turbo-stream")) {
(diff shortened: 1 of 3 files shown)


src/observers/link_prefetch_observer.js
```diff
@@ -100,14 +100,11 @@ export class LinkPrefetchObserver {
 
     request.headers["Sec-Purpose"] = "prefetch"
 
-    if (link.dataset.turboFrame && link.dataset.turboFrame !== "_top") {
-      request.headers["Turbo-Frame"] = link.dataset.turboFrame
-    } else if (link.dataset.turboFrame !== "_top") {
-      const turboFrame = link.closest("turbo-frame")
+    const turboFrame = link.closest("turbo-frame")
+    const turboFrameTarget = link.dataset.turboFrame || turboFrame?.getAttribute("target") || turboFrame?.id
```
seanpdoyle, line 104: Could we rely on `getAttribute("data-turbo-frame")` instead of the `dataset` property?

Most other places in the application, we use the `getAttribute` method to fetch attributes. If we use `.dataset` here, grepping for `data-turbo-frame` won't include this line.

src/observers/link_prefetch_observer.js
```diff
@@ -100,14 +100,11 @@ export class LinkPrefetchObserver {
 
     request.headers["Sec-Purpose"] = "prefetch"
 
-    if (link.dataset.turboFrame && link.dataset.turboFrame !== "_top") {
-      request.headers["Turbo-Frame"] = link.dataset.turboFrame
-    } else if (link.dataset.turboFrame !== "_top") {
-      const turboFrame = link.closest("turbo-frame")
+    const turboFrame = link.closest("turbo-frame")
```
seanpdoyle, line 103: Is checking for an ancestor `<turbo-frame>` sufficient here? What if the `<a>` element has a `[data-turbo-frame]` attribute?

afcapel: That's what the `link.dataset.turboFrame` in the line below is doing. I'll update it to use `getAttribute("data-turbo-frame")` so that's hopefully clearer.

kevinmcconnell (approved):
Looks great. Nice job chasing this one down!

seanpdoyle (approved):

davidalejandroaguilar:
Sorry I couldn't chime in on time, but thanks for this @afcapel!

Issue #1139, opened 2024-01-24
8.0.0-beta.3 - Issue with stylesheets


meneerprins:
For some context, I am using a Ruby on Rails app with [Vite Ruby](https://github.com/ElMassimo/vite_ruby). Now with the latest beta version this PR is included: [Remove unused stylesheets when navigating](https://github.com/hotwired/turbo/pull/1128). That solves the issue of overlapping styles when switching between layouts (when you remove the `data-turbo-track="reload"` attribute) but it also causes some issues:

- When the Vite dev server is on, during a Turbo navigation, all the styles disappear (probably because the stylesheets gets injected)
- In my case, many stimulus controllers are lazy loaded and when you import the styles of a library inside the stimulus controller for example `import '@splidejs/splide/dist/css/splide-core.min.css';`, the styles also get removed.

Anyone has an idea for a solution for this? (I saw there is some patch for Trix here https://github.com/hotwired/turbo/pull/1133/files to add `data-turbo-permanent` but in many cases you have no control over the tags when they are injected by either tools or 3rd party libraries).

afcapel:
We've thought about this and, to avoid breaking changes, we're are going to make the behavior to remove unused stylesheets opt-in, instead of the new default. See https://github.com/hotwired/turbo/pull/1140

afcapel:
@meneerprins thanks for reporting. This was fixed in https://github.com/hotwired/turbo/pull/1140

We'll release a beta version soon.

PR #1140, opened 2024-01-25
Introduce data-turbo-track="dynamic"

afcapel:
To track stylesheets and styles that we can dynamically remove when navigating.

We introduced this behaviour in https://github.com/hotwired/turbo/pull/1128 and thought it would be a good default to avoid full page reloads when styles change.

However, it seems it's quite common for libraries to add stylesheets and styles to the head that they want to keep around. For example, see

- https://github.com/hotwired/turbo/pull/1133
- https://github.com/hotwired/turbo/issues/1139

So let's make this behaviour opt-in by adding a `data-turbo-track="dynamic"` attribute to stylesheets and styles that we want to dynamically remove when they are no longer in a new page after a navigation.

This avoids any breaking changes for existing apps.

### DIFF
--- a/src/core/drive/page_renderer.js
+++ b/src/core/drive/page_renderer.js
@@ -78,7 +78,7 @@ export class PageRenderer extends Renderer {
     await newStylesheetElements
 
     if (this.willRender) {
-      this.removeUnusedHeadStylesheetElements()
+      this.removeUnusedDynamicStylesheetElements()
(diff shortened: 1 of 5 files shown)


kevinmcconnell (approved):
Looks great, thanks for adding this. It does seem like making dynamic the default would lead to too much breakage.

PR #1143, opened 2024-01-29
Don't include Turbo-Frame header in promoted frame navigations

afcapel:
This PR changes the behavior of frame navigations promoted with the `[data-turbo-action]` attribute so that they don't include the `Turbo-Frame` header in their requests. This mimics the existing behavior of frame navigations promoted with `[data-turbo-target=_top]`.

If we don't do this turbo-rails will see the `Turbo-Frame` header and render the response as a frame update, with a minimal layout that doesn't include any assets. When Turbo receives the response it believes it's a full page response in which the assets have gone stale and it issues a full page reload with `tracked_element_mismatch` as the reason.

Fixes https://github.com/hotwired/turbo/issues/1047

This is similar to https://github.com/hotwired/turbo/pull/1138

//cc @seanpdoyle @domchristie

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -192,7 +192,9 @@ export class FrameController {
   // Fetch request delegate
 
   prepareRequest(request) {
-    request.headers["Turbo-Frame"] = this.id
+    if (!this.action) {
+      request.headers["Turbo-Frame"] = this.id
+    }
 
     if (this.currentNavigationElement?.hasAttribute("data-turbo-stream")) {
       request.acceptResponseType(StreamMessage.contentType)
--- a/src/tests/functional/frame_navigation_tests.js
+++ b/src/tests/functional/frame_navigation_tests.js
@@ -49,6 +49,30 @@ test("lazy-loaded frame promotes navigation", async ({ page }) => {
   assert.equal(pathname(page.url()), "/src/tests/fixtures/frames/frame_for_eager.html")
 })
 
+test("frame navigation sets the Turbo-Frame header", async ({ page }) => {
+  await page.goto("/src/tests/fixtures/frame_navigation.html")
+  assertNextRequesTurboFrameHeader(page, "frame")
+
+  await page.click("#inside")
+  await nextEventNamed(page, "turbo:load")
+})
+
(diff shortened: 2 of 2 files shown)


domchristie:
I think this still triggers a `tracked_element_mismatch` reload when testing with https://gist.github.com/domchristie/d143a6e7ae1807bd1f8323ecc8a4978b (and building Turbo from this branch).

> This mimics the existing behavior of frame navigations promoted with [data-turbo-target=_top].

I was under the impression that a frame visit with an action should only replace a part of the page and update the history (not target the whole page). Am I misunderstanding?

afcapel:
> I was under the impression that a frame visit with an action should only replace a part of the page and update the history (not target the whole page). Am I misunderstanding?

Oh, you're right. Not sure if that's a big issue, as far as I know we only use the `Turbo-Frame` header to skip the full layout rendering. But it'd be better if we could keep the header more semantic.

domchristie:
> The bug … sounds more like an error in treating a frame response as if it was a full page response.

If I'm understanding the behaviour correctly, Turbo processes the frame response as expected, but then performs another visit with `willRender: false, updateHistory: false` (https://github.com/hotwired/turbo/blob/b7187f13b216cee8c855ef1fb6ac790b40dde361/src/core/frames/frame_controller.js#L373). This then triggers the `tracked_element_mismatch`

> Turbo should know that the frame request is not responsible for populating the assets in head

Am I correct in thinking that frame requests _can_ populate content in `head` e.g. https://github.com/hotwired/turbo-rails/pull/428?

kevinmcconnell:
@domchristie yes, exactly. It's when performing the visit that Turbo incorrectly treats the frame response as a full page response with regard to the asset version checking. It's a step that shouldn't happen for frame responses, even those with actions. Avoiding that would solve this bug.

> Am I correct in thinking that frame requests can populate content in head

Frame requests shouldn't be able to render anything outside of their target frame, as I understand it. The responses can return content in `head` because there are cases where Turbo will make use of that in determining what to do with the response (things like triggering a reload via a meta tag). But that content isn't rendered onto the page, due to it being outside of the target frame.

afcapel:
I found a better fix in https://github.com/hotwired/turbo/pull/1144

Issue #468, opened 2021-11-23
changing turbo frame src manually doesn't update main url


alexgerber:
Hi, is this a bug or something intentionally left out? On latest turbo release.

When using a turbo frame with the new data-turbo-action="advance" directive ie:
```
<turbo-frame data-turbo-action="advance" data-search-target="frame">
  <a href="/articles?page=2" rel="next">Next page</a>
</turbo-frame>

```
When manually selecting the turbo frame and changing the src in stimulus controller
ie `frameTarget.src = '/'` it only changes the src of the frame, not url.

It would be nice to be able to change the src programmatically and have it sync with the main url as well, like normal link clicks. Thanks, let me know what you think, or if I missed something.

seanpdoyle:
This is a current limitation that is likely to stick around in `7.1.0`. 

In cases where you need to promote a Frame Visit to a Page Visit programmatically (like in a Stimulus Controller), you could synthesize an `<a>` element, declare the desired `frameTarget.src` as the `<a href="...">`, set `[data-turbo-action]` on the element, then click it.

```js
function promoteToFrameVisit(element /*: FrameElement */, src /*: string */, action = "advance") {
  const link = document.createElement("a")
  link.setAttribute("href", src)
  link.setAttribute("data-turbo-action", action) // you can skip this is the <turbo-frame> already declares [data-turbo-action]
  link.setAttribute("hidden", "")
  element.appendChild(link)
  link.click()
  element.removeChild(link)
}

promoteToFrameVisit(this.frameTarget, "/", "replace")
```

Alternatively, the `<turbo-frame>` could have a descendant `<a hidden>` element that the controller clicks.

To share a little extra context, the current internal structure Turbo Frames portion of the codebase makes it challenging to retain contextual information throughout a Frame navigation.

For example, once an `<a>` element is clicked, a `<form>` element is submitted (often times with `<button>` or `<input type="submit">` submitter), or a `<turbo-frame>` element's `[src]` attribute is changed programmatically, the methods that respond to that action cannot access the original elements' `[data-*]` attributes. This is a problem when trying to infer additional information. All that's available is the value of the `[src]`.

There are efforts in-progress to refactor and restructure the Turbo internals , but they're unlikely to ship until a stable `7.1.0` has been released.

PR #1141, opened 2024-01-25
Morph with `ignoreActiveValue: true`

seanpdoyle:
Morph with the [ignoreActiveValue: true][] option to morph the currently focused element's attributes, but preserve its value.

This behavior can be extremely helpful when paired with an auto-submitting `<form>` element, like a typeahead `[role="combobox"]`, or an auto-submitting [`<input type="search">`][search].

This commit depends on a fork of `idiomorph` that [fixes a bug related to `ignoreActiveValue: true`][bigskysoftware/idiomorph#34].

https://github.com/hotwired/turbo/assets/2575027/ca40524f-da6c-40fd-aacf-1bbe74be72f5

[ignoreActiveValue: true]: https://github.com/bigskysoftware/idiomorph#options
[search]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search
[bigskysoftware/idiomorph#34]: https://github.com/bigskysoftware/idiomorph/pull/34

PR #1135, opened 2024-01-22
Drive `turbo-frame` with `Turbo.visit(url, { frame:, action: })`

seanpdoyle:
Closes [#489][]
Closes [#468][]

First, add test coverage to exercise the `<turbo-frame id="frame">` and
`<turbo-frame id="hello" target="frame">` as outlined in a
[comment on #489][].

Next, add coverage to support driving a `<turbo-frame
data-turbo-action="advance">` through the `Turbo.visit` method. To pass
that test coverage, invoke
`frameElement.delegate.proposeVisitIfNavigatedWithAction(frameElement)`
prior to modifying the element's `[src]` attribute.

To support the implementation changes necessary to pass the tests, this
commit changes the `proposeVisitIfNavigatedWithAction(frame, action)`
interface by flattening the variable arguments into a single `action`
argument, then moving the `getVisitAction` call to the call sites that
require it. The call sites that know their `action` pass it in directly
when its available.

[#489]: https://github.com/hotwired/turbo/issues/489
[#468]: https://github.com/hotwired/turbo/issues/468
[comment on #489]: https://github.com/hotwired/turbo/issues/489#issuecomment-1503078653

Issue #1116, opened 2023-12-28
`turbo-visit-control` isn't respected with morphing


northeastprince:
In my app, I'm using SimpleCSS which adds styles to HTML elements themselves (not their classes). In development, when new code causes an error impacting the route I'm already on, Turbo mistakenly morphs to Rails' error info page, even though [Turbo visits are disabled there](https://github.com/rails/rails/blob/efce114ddea89110bdee7af7a8c65377a53182b7/actionpack/lib/action_dispatch/middleware/templates/rescues/layout.erb#L6), thus putting me in a weird limbo where exception backtraces are semantically styled 💀

brunoprietog:
Related #1105

This is because MorphRenderer doesn't inherit from PageRenderer, so all these things are lost.

PR #1144, opened 2024-01-29
Do not invalidate pages in visits that don't render

afcapel:
Fixes https://github.com/hotwired/turbo/issues/1047
Supersedes https://github.com/hotwired/turbo/pull/1143

Rendering a frame with a data-turbo-action was rendering the response twice.

First, the `FrameController` in charge of the frame renders the response using a `FrameRenderer`, while processing callback the `loadResponse` callback.

https://github.com/hotwired/turbo/blob/b7187f13b216cee8c855ef1fb6ac790b40dde361/src/core/frames/frame_controller.js#L314

But when we update a frame with a data-turbo-action, we also issue a visit to update the browser URL & history:

https://github.com/hotwired/turbo/blob/b7187f13b216cee8c855ef1fb6ac790b40dde361/src/core/frames/frame_controller.js#L373

This session visit was also triggering a `loadResponse` callback, which in turn was instantiating a `PageSnapshot` and rendering it.

This caused problems when the response for the frame didn't include all the tracked assets in the page. The page snapshot considered the assets stale and triggered a reload of the page with reason `tracked_element_mismatch`.

This commit fixes the issue by not rendering the page snapshot when navigating from a frame.

The commit includes a new test that reproduces the issue returning a turbo frame in an HTML document with an empty `<head>`. Before this fix, the response would trigger a full page reload.

This is the call stack before the fix. Notice the two calls to `loadResponse`.

Screenshot 2024-01-29 at 17 30 43

### DIFF
--- a/src/core/view.js
+++ b/src/core/view.js
@@ -56,7 +56,8 @@ export class View {
   // Rendering
 
   async render(renderer) {
-    const { isPreview, shouldRender, newSnapshot: snapshot } = renderer
+    const { isPreview, shouldRender, willRender, newSnapshot: snapshot } = renderer
+
     if (shouldRender) {
       try {
         this.renderPromise = new Promise((resolve) => (this.#resolveRenderPromise = resolve))
@@ -77,7 +78,7 @@ export class View {
         this.#resolveRenderPromise(undefined)
         delete this.renderPromise
       }
-    } else {
+    } else if (willRender) {
       this.invalidate(renderer.reloadReason)
     }
   }
--- a/src/core/view.js
+++ b/src/core/view.js
@@ -58,6 +58,10 @@ export class View {
   async render(renderer) {
     const { isPreview, shouldRender, willRender, newSnapshot: snapshot } = renderer
 
+    // A workaround to ignore tracked element mismatch reloads when performing
+    // a promoted Visit from a frame navigation
+    const shouldInvalidate = willRender
+
     if (shouldRender) {
       try {
         this.renderPromise = new Promise((resolve) => (this.#resolveRenderPromise = resolve))
@@ -78,7 +82,7 @@ export class View {
         this.#resolveRenderPromise(undefined)
         delete this.renderPromise
       }
-    } else if (willRender) {
+    } else if (shouldInvalidate) {
       this.invalidate(renderer.reloadReason)
     }
   }
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -63,6 +63,7 @@ export class Visit {
     const {
       action,
       historyChanged,
+      frame,
       referrer,
       snapshot,
       snapshotHTML,
@@ -79,6 +80,7 @@ export class Visit {
     }
     this.action = action
     this.historyChanged = historyChanged
+    this.fromFrame = frame
     this.referrer = referrer
     this.snapshot = snapshot
     this.snapshotHTML = snapshotHTML
@@ -209,8 +211,10 @@ export class Visit {
         if (this.view.renderPromise) await this.view.renderPromise
 
         if (isSuccessful(statusCode) && responseHTML != null) {
-          const snapshot = PageSnapshot.fromHTMLString(responseHTML)
-          await this.renderPageSnapshot(snapshot, false)
+          if (!this.fromFrame) {
+            const snapshot = PageSnapshot.fromHTMLString(responseHTML)
+            await this.renderPageSnapshot(snapshot, false)
+          }
 
           this.adapter.visitRendered(this)
           this.complete()
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -63,7 +63,6 @@ export class Visit {
     const {
       action,
       historyChanged,
-      frame,
       referrer,
       snapshot,
       snapshotHTML,
@@ -80,7 +79,6 @@ export class Visit {
     }
     this.action = action
     this.historyChanged = historyChanged
-    this.fromFrame = frame
     this.referrer = referrer
     this.snapshot = snapshot
     this.snapshotHTML = snapshotHTML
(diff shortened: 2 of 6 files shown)


src/core/frames/frame_snapshot.js
```diff
@@ -0,0 +1,10 @@
+import { parseHTMLDocument } from "../../util"
+import { Snapshot } from "../snapshot"
+import { expandURL } from "../url"
+import { HeadSnapshot } from "./head_snapshot"
+
+export class PageSnapshot extends Snapshot {
```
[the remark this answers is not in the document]
afcapel: @seanpdoyle sorry, commit mishap, that file is not necessary, I've removed it from the PR.

domchristie:
Nice! A couple of things: 

Is `fromFrame` set for the same purpose as `willRender`. As far as I can tell, willRender is only set to false for this frame visit case, in which case, could we just re-use that?

What is the frame visit actually doing? It doesn’t render and it’s instantiated with `updateHistory: false`. It looks like `FrameController#changeHistory` is already called and it doesn’t render, so is it needed at all?

domchristie:
> What is the frame visit actually doing? It doesn’t render and it’s instantiated with updateHistory: false. It looks like FrameController#changeHistory is already called and it doesn’t render, so is it needed at all?


😵‍💫 From the looks of the failing tests (any my own brief manual tests), it looks like the visit is doing a fair bit (on the caching front?)

afcapel:
>  As far as I can tell, willRender is only set to false for this frame visit case, in which case, could we just re-use that?

That's a good idea. I've updated the PR so now we use `willRender` to ensure we are not invalidating a page in a visit that wouldn't render anyway.

> What is the frame visit actually doing? It doesn’t render and it’s instantiated with updateHistory: false. It looks like FrameController#changeHistory is already called and it doesn’t render, so is it needed at all?

This was introduced in https://github.com/hotwired/turbo/pull/398/ to add navigation-like behavior to to frame navigations with `data-turbo-action`. It's a bit difficult to follow everything that's happening, and it'd be nice if we had some abstractions to do this without having to issue a new visit but that's a big change that would require changing many internal parts of Turbo.

For now this small patch fixes https://github.com/hotwired/turbo/issues/1047, which is the main blocker for a Turbo 8 final release. If we're happy with this I think are ready for the final release.

src/core/view.js
```diff
@@ -77,7 +78,7 @@ export class View {
         this.#resolveRenderPromise(undefined)
         delete this.renderPromise
       }
-    } else {
+    } else if (willRender) {
```
kevinmcconnell, line 81: I don't think it necessarily needs addressed here, but I find it confusing that we have `shouldRender` and `willRender` cases. It's not clear how they are different. If you *should* render, doesn't that imply that you *will*?

I think the easiest answer is probably to rename `willRender`. It really means that the view is being used in a context where rendering is not required. Or more specifically, it's being used to do all the other operations that normally accompany a page render, but after a frame update has already rendered.

But ideally we'd separate the responsibilities to avoid this sort of conditional check happening here. So maybe we're as well to stick with the existing language for now, and then look at refactoring in another PR which keeps the same behaviour?

kevinmcconnell (approved):

src/core/view.js
afcapel: Yes, agreed to all of that 👍

domchristie: I agree the `shouldRender` / `willRender` is confusing here.

Is there a plan to refactor before releasing v8? If not then I reckon we should clarify `willRender` by renaming and adding a comment. e.g.:

```js
// A workaround to ignore tracked element mismatch reloads when performing a promoted Visit from a frame navigation

const shouldInvalidate = willRender
```

brunoprietog:
> If we're happy with this I think are ready for the final release.

@afcapel Sorry to bother with this but, have you considered any of these concerns before releasing the stable version? #1080 (Turbo crashes completely, I opened #1098 to address it). Other concerns with full page reloading when assets are diferent #1105 and #1116. Thanks!

src/core/view.js
seanpdoyle: I'd opened https://github.com/hotwired/turbo/pull/741 to limit the occurrence of `willRender`.

I'd also opened https://github.com/hotwired/turbo/pull/430 to try and introduce some abstractions for Frame navigation. They've both been kept up to date originally, but I eventually stopped dragging them forward.

Would either be worthwhile to rebase for inclusion in 8?

pfeiffer:
>> If we're happy with this I think are ready for the final release.
>
> @afcapel Sorry to bother with this but, have you considered any of these concerns before releasing the stable version? #1080 (Turbo crashes completely, I opened #1098 to address it)

I second that closing #1080 (causing blank pages with replace navigations!), would need to be part of a stable release. The potential fix in #1098 makes a lot of sense. We're seeing flashes of unwanted content (stale snapshots) while performing refreshes with morphing. It kinda defeats the purpose of the new shiny morph refreshes.

Let me know if we can do anything to help out here?

afcapel:
Ok, I'll do a final round before publishing a RC 👍 

Aiming to fix:

- https://github.com/hotwired/turbo/issues/1047
- https://github.com/hotwired/turbo/issues/1080
- https://github.com/hotwired/turbo/issues/1105
- https://github.com/hotwired/turbo/issues/1116

All of them have PRs already open to address them, so it's a matter of validating and merging.

Anything we should consider a blocker?

brunoprietog:
@seanpdoyle has made a big effort with contributions and there are still many unmerged, several bug fixes and some very old ones. If you have time, it would be great if you could review as much as you can, obviously within your availability.

afcapel:
Sure, I'd also like to clean up the remaining PRs, but unfortunately right now I'm focused on another project and have limited availability. We could always do more, but for context, [we've already merged 72 PRs since v7.3.0](https://github.com/hotwired/turbo/pulls?q=is%3Apr+is%3Aclosed+merged%3A%3E%3D2023-03-01+sort%3Aupdated-desc).

seanpdoyle:
Out of the pull requests that I've opened at the moment, I consider two are worth merging into 8:

* https://github.com/hotwired/turbo/pull/1135 (closes #468 and #489)
* https://github.com/hotwired/turbo/pull/1141 (but depends on `idiomorph@main`)

The others either require additional troubleshooting, or are refactors and abstractions.

afcapel:
Cool, thanks @seanpdoyle I'll be sure to review those two as well before the RC.

As for other refactorings, they are of course welcome but they'll probably have to wait until after the v8 release.

Issue #1085, opened 2023-11-28
Page Refreshes: allow them to happen on an "advance" action


weaverryan:
Hi!

I've been testing out Turbo 8 because I'm super excited about the page refreshes. One use-case for "page refreshes" is a "data-tables" like setup, where you can search a table, click column links to sort, paginate, etc. Submitting the search form will work great with page refreshes.

But currently, the "column sorting" links and pagination links can't use page refreshes. That's the "action" must be `replace`:

https://github.com/hotwired/turbo/blob/5d7ff7cf5852d28cea6c5562a5fd76dbf3e0bc61/src/core/drive/page_view.js#L59

This was done on purpose in #1072 for sensible reasons. And there IS a workaround - `<a href="..." data-turbo-action="replace">`. However, this is a case where I DO want the history to `advance` so the user can click back to the previous page.

Would it be reasonable to introduce a new `data-turbo-morph` attribute for the link to opt into morphing? The URL of the page would still need to match, this would just remove the requirement for the action to be `replace`.

Cheers!

jorgemanrubia:
I understand the need but I have some concerns with using morphing for regular "advance" navigations.

There is an internal implementation one: current snapshot caching in Turbo won't work, because the current implementation relies on managing different `body` elements at rendering time (not the same morphed one). That's something we can solve, but, more conceptually, does browser history make sense with morphing? You are changing regions of the screen. Should Turbo create snapshots before each morph so that it can restore those? 

Also, this would also be an additional API in the library, more choice for the user, which is something that bothers me. 

We'll soon [support page refreshes with different params in the URL](https://github.com/hotwired/turbo/pull/1079), would that alleviate your use case here? You still wouldn't have browser history, but at least you could have meaningful URLs representing the state of the filter.

weaverryan:
Thanks for the reply!

> We'll soon https://github.com/hotwired/turbo/pull/1079, would that alleviate your use case here?

I saw that PR! That is great, yes ❤️ 

> You still wouldn't have browser history, but at least you could have meaningful URLs representing the state of the filter.

I think this will mostly work, I could live with it, though in a [data tables](https://datatables.net/) setup, you really should be able to navigate back.

> That's something we can solve, but, more conceptually, does browser history make sense with morphing? You are changing regions of the screen

Fair point! Though "you are changing regions of the screen" is case-by-case. In a [data tables](https://datatables.net/) setup, it's expected that navigation won't change the screen position.

[...]

PR #1145, opened 2024-01-29
Add support for `data-turbo-replace-method="morph"` to force a morph

krschacht:
This PR allows you to add an optional `data-turbo-replace-method="morph"` in the cases where a `data-turbo-action="replace"`. If you try to specify `morph` when the action is `advance` it will be ignored. It is an alternate, and more general, solution to #1079 so it also reverts that behavior. 

**Background:** After introducing the concept of morphing _page refresh_, Turbo detects when the current page is being refreshed again in order to decide if it should morph. It determines this by comparing the url of the previous location and the new location. This automatic detection works great, most of the time. However, there are some times when the href may include some tracking query string parameters so the href does not match and a refresh is not triggered. PR #1079 was an attempt to address this.

**Problem:** However, [this discussion](https://github.com/hotwired/turbo-ios/pull/178#discussion_r1501357383) pointed out that while query string parameters are sometimes unimportant, other times they completely change the behavior of the page (e.g. `?tab=overview`). In addition, there are some situations where we are doing a replace and that replace is effectively a refresh, even though the URL has changed. Imagine the user is on `/posts/new`, clicks "save draft", and redirect to `/posts/:id` to continue editing. You designate this as a `replace` action (not `advance`) and you want to morph this page since it's the same form and same template.

**Solution:** This PR introduces adds support for `data-turbo-replace-method="morph"`. There are explicit tests for all the relevant cases within `navigation_tests.js`, specifically:

* Within `a href` links
* Within `<form>` tags (get & post)
* Within the `<button>` tag within a `<form>` (get & post)

### DIFF
--- a/src/core/drive/navigator.js
+++ b/src/core/drive/navigator.js
@@ -1,4 +1,4 @@
-import { getVisitAction } from "../../util"
+import { getVisitAction, getVisitReplaceMethod } from "../../util"
 import { FormSubmission } from "./form_submission"
 import { expandURL, getAnchor, getRequestURL } from "../url"
 import { Visit } from "./visit"
@@ -10,6 +10,7 @@ export class Navigator {
   }
 
   proposeVisit(location, options = {}) {
+    console.log(`proposeVisit ${location}`, options)
     if (this.delegate.allowsVisitingLocationWithAction(location, options.action)) {
       this.delegate.visitProposedToLocation(location, options)
     }
@@ -69,6 +70,7 @@ export class Navigator {
   }
 
   async formSubmissionSucceededWithResponse(formSubmission, fetchResponse) {
+    console.log(`formSubmissionSucceededWithResponse`)
     if (formSubmission == this.formSubmission) {
       const responseHTML = await fetchResponse.responseHTML
       if (responseHTML) {
@@ -79,11 +81,15 @@ export class Navigator {
 
         const { statusCode, redirected } = fetchResponse
         const action = this.#getActionForFormSubmission(formSubmission, fetchResponse)
+        const replaceMethod = this.#getReplaceMethodForFormSubmission(formSubmission, fetchResponse)
+        console.log(`action = ${action} and replaceMethod = ${replaceMethod}`)
         const visitOptions = {
           action,
+          replaceMethod,
           shouldCacheSnapshot,
           response: { statusCode, responseHTML, redirected }
         }
+        console.log(`calling proposeVisit from formSubmission`)
         this.proposeVisit(fetchResponse.location, visitOptions)
       }
     }
@@ -161,5 +167,11 @@ export class Navigator {
   #getDefaultAction(fetchResponse) {
     const sameLocationRedirect = fetchResponse.redirected && fetchResponse.location.href === this.location?.href
     return sameLocationRedirect ? "replace" : "advance"
+  }
+
+  #getReplaceMethodForFormSubmission(formSubmission, fetchResponse) {
+    if (this.#getActionForFormSubmission(formSubmission, fetchResponse) !== "replace") return
+    const { submitter, formElement } = formSubmission
+    return getVisitReplaceMethod(submitter, formElement) || "body"
   }
 }
--- a/src/core/drive/navigator.js
+++ b/src/core/drive/navigator.js
@@ -10,7 +10,6 @@ export class Navigator {
   }
 
   proposeVisit(location, options = {}) {
-    console.log(`proposeVisit ${location}`, options)
     if (this.delegate.allowsVisitingLocationWithAction(location, options.action)) {
       this.delegate.visitProposedToLocation(location, options)
     }
@@ -70,7 +69,6 @@ export class Navigator {
   }
 
   async formSubmissionSucceededWithResponse(formSubmission, fetchResponse) {
-    console.log(`formSubmissionSucceededWithResponse`)
     if (formSubmission == this.formSubmission) {
       const responseHTML = await fetchResponse.responseHTML
       if (responseHTML) {
@@ -82,14 +80,12 @@ export class Navigator {
         const { statusCode, redirected } = fetchResponse
         const action = this.#getActionForFormSubmission(formSubmission, fetchResponse)
         const replaceMethod = this.#getReplaceMethodForFormSubmission(formSubmission, fetchResponse)
-        console.log(`action = ${action} and replaceMethod = ${replaceMethod}`)
         const visitOptions = {
           action,
           replaceMethod,
           shouldCacheSnapshot,
           response: { statusCode, responseHTML, redirected }
         }
-        console.log(`calling proposeVisit from formSubmission`)
         this.proposeVisit(fetchResponse.location, visitOptions)
       }
     }
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -16,7 +16,7 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const shouldMorphPage = this.shouldMorphPage(snapshot, visit)
+    const shouldMorphPage = snapshot.shouldMorphPage && this.isMorphableVisit(visit)
     const rendererClass = shouldMorphPage ? MorphRenderer : PageRenderer
 
     const renderer = new rendererClass(this.snapshot, snapshot, PageRenderer.renderElement, isPreview, willRender)
@@ -55,9 +55,8 @@ export class PageView extends View {
     return this.snapshotCache.get(location)
   }
 
-  shouldMorphPage(snapshot, visit) {
-    return snapshot.shouldMorphPage &&
-            (this.isPageRefresh(visit) || this.isReplaceMethodMorph(visit))
+  isMorphableVisit(visit) {
+    return this.isPageRefresh(visit) || this.isReplaceMethodMorph(visit)
   }
 
   isPageRefresh(visit) {
@@ -65,7 +64,7 @@ export class PageView extends View {
   }
 
   isReplaceMethodMorph(visit) {
-    return !visit || (visit.action === "replace" && visit.replaceMethod === "morph")
+    return visit.action === "replace" && visit.replaceMethod === "morph"
   }
 
   shouldPreserveScrollPosition(visit) {
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -17,7 +17,6 @@ export class PageView extends View {
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
     const shouldMorphPage = this.shouldMorphPage(snapshot, visit)
-    console.log(`should morph? ${shouldMorphPage}`)
     const rendererClass = shouldMorphPage ? MorphRenderer : PageRenderer
 
     const renderer = new rendererClass(this.snapshot, snapshot, PageRenderer.renderElement, isPreview, willRender)
@@ -57,7 +56,6 @@ export class PageView extends View {
   }
 
   shouldMorphPage(snapshot, visit) {
-    console.log(`shouldMorphPage snapshot? ${snapshot.shouldMorphPage}... ${visit.action} and ${visit.replaceMethod}`)
     return snapshot.shouldMorphPage &&
             (this.isPageRefresh(visit) || this.isReplaceMethodMorph(visit))
   }
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -61,7 +61,7 @@ export class PageView extends View {
   }
 
   isPageRefresh(visit) {
-    return !visit || (this.lastRenderedLocation.pathname === visit.location.pathname && visit.action === "replace")
+    return !visit || (this.lastRenderedLocation.href === visit.location.href && visit.action === "replace")
   }
 
   isReplaceMethodMorph(visit) {
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -16,7 +16,8 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const shouldMorphPage = this.isPageRefresh(visit) && this.snapshot.shouldMorphPage
+    const shouldMorphPage = this.shouldMorphPage(snapshot, visit)
+    console.log(`should morph? ${shouldMorphPage}`)
     const rendererClass = shouldMorphPage ? MorphRenderer : PageRenderer
 
     const renderer = new rendererClass(this.snapshot, snapshot, PageRenderer.renderElement, isPreview, willRender)
@@ -55,8 +56,18 @@ export class PageView extends View {
     return this.snapshotCache.get(location)
   }
 
+  shouldMorphPage(snapshot, visit) {
+    console.log(`shouldMorphPage snapshot? ${snapshot.shouldMorphPage}... ${visit.action} and ${visit.replaceMethod}`)
+    return snapshot.shouldMorphPage &&
+            (this.isPageRefresh(visit) || this.isReplaceMethodMorph(visit))
+  }
+
   isPageRefresh(visit) {
     return !visit || (this.lastRenderedLocation.pathname === visit.location.pathname && visit.action === "replace")
+  }
+
+  isReplaceMethodMorph(visit) {
+    return !visit || (visit.action === "replace" && visit.replaceMethod === "morph")
   }
 
   shouldPreserveScrollPosition(visit) {
--- a/src/tests/server.mjs
+++ b/src/tests/server.mjs
@@ -171,7 +171,6 @@ router.get("/messages", (request, response) => {
 function receiveMessage(content, id, target) {
   const data = renderSSEData(renderMessage(content, id, target))
   for (const response of streamResponses) {
-    console.log("delivering message to stream", response.socket?.remotePort)
     response.write(data)
   }
 }
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -62,6 +62,7 @@ export class Visit {
 
     const {
       action,
+      replaceMethod,
       historyChanged,
       referrer,
       snapshot,
@@ -78,6 +79,7 @@ export class Visit {
       ...options
     }
     this.action = action
+    this.replaceMethod = replaceMethod
     this.historyChanged = historyChanged
     this.referrer = referrer
     this.snapshot = snapshot
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -147,6 +147,7 @@ export class FrameController {
   // Appearance observer delegate
 
   elementAppearedInViewport(element) {
+    // TODO: Determine if I need to check replaceMethod() here — I need to find the corresponding tests
     this.proposeVisitIfNavigatedWithAction(element, getVisitAction(element))
     this.#loadSourceURL()
   }
--- a/src/core/index.js
+++ b/src/core/index.js
@@ -34,6 +34,10 @@ export function registerAdapter(adapter) {
  * @param options Options to apply
  * @param options.action Type of history navigation to apply ("restore",
  * "replace" or "advance")
+ * @param options.frame If specified, finds a turbo-frame element with an [id]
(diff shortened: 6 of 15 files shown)


morki:
Another use case is saving new entity and continuing editation, like saving draft, e.g.:

`/posts/new`  -> save draft -> `/posts/{id}`

It is other URL, but it is the same form and template.

morki:
@krschacht sorry, I never touched the codebase, I am primarily backend developer, so I am pure user of this library :)

pch:
I created my own proof-of-concept for this: https://github.com/hotwired/turbo/pull/1179 (issue: https://github.com/hotwired/turbo/issues/1177)

The solution isn't great, but the attached gifs demonstrate the difference morphing makes in this particular case.

@krschacht Your version looks much better, although I'm not sure that `refresh` accurately describes what's happening here. Maybe a new type of `turbo-action` is needed.

pfeiffer:
In my opinion, using a `refresh` action to fetch a URL that is different than the current feels somewhat off.

I think the `replace` visit action is the appropriate one here. Would it be an option to provide a granular data attribute to control _how_ the replacing would happen instead?

```html
<form action="/posts" method="post" data-turbo-action="replace" data-turbo-replace-with="morph">
  <input ...>  
</form>
```

```html
<a href="/posts/4" data-turbo-action="replace" data-turbo-replace-with="morph">Next post</a>
```

krschacht:
Thanks @pfeiffer Great suggestion on the naming. I'm happing to clean up this PR and get it ready for merging in if you're in support of it. I've been waiting for one of the project maintainers to give it a blessing since I know you all have the larger context about what makes sense in this library and what doesn't.

pfeiffer:
I'm wondering if this could address some of the issues that are outlined in https://github.com/hotwired/turbo-ios/pull/178#discussion_r1501357383 and https://github.com/hotwired/turbo/pull/1079

Essentially, I find it a bit off that we can `refresh` to a URL that's different than the current URL - it's not really a refresh then, right? Semantically, a refresh should be of the of the current page and while it _could_ be the same page even with different params (ie `&utm_...`) it could also not be (eg. `?tab=overview`) - it's implementation specific and could be different for every application.

To cover the use-case that @brunoprietog outlines in https://github.com/hotwired/turbo/pull/1079, I would think that a `replace` with morphing like this provided by PR would actually be a better solution than comparing pathnames as in #1079 to determine if two URLs could be morphed.

@krschacht, @brunoprietog - is there something that I've missed in terms of the refresh action to different URLs? When does it make sense to "refresh" a page, that is not same as the current page? Would a `replace` with `morph` enabled cover that use case?

brunoprietog:
I feel that both are complementary. Ideally it should be as magical as possible. As I argued in the PR, my way of thinking about it is in terms of a Rails controller. The search params don't generally affect what action each controller invokes, which translates to it being the same page simply with other parameters. And changing one parameter for another shouldn't make much difference to the page, otherwise you'd have another action.

Well, this would be changing replace to refresh. I think it's not bad either, but it would ignore the options set in the meta tags. And a new API. In essence I would expect this PR to be merged anyway, and that would imply that it should have support in native adapters as well.

pfeiffer:
> I feel that both are complementary. Ideally it should be as magical as possible. As I argued in the PR, my way of thinking about it is in terms of a Rails controller. The search params don't generally affect what action each controller invokes, which translates to it being the same page simply with other parameters. And changing one parameter for another shouldn't make much difference to the page, otherwise you'd have another action.

@brunoprietog - thanks for helping me trying to understand :-) I agree and my concern is that the 'magic' disappears if there's diverging behavior in native adapters and web adapters.

I agree that _sometimes_ query params do not matter. Sometimes they are used to filter the same collection type (eg. [Rails docs](https://guides.rubyonrails.org/action_controller_overview.html#parameters)) and sometimes they're simply tracking params. But other times they can drastically change the page (remember Turbo is backend agnostic!) - it depends on the backend implementation.

Navigating and 'morphing' between eg. a filtering query param, I believe could be made with a `replace` and `morph` (as this PR would add) or a Turbo Frame with `replace`.

I re-read the original PR and I still have a hard time thinking of a _real-life example_ of when we'd like to refresh to a URL with different query strings? I can imagine we'd like to _replace and morph_ to a URL with different query params and I have a feeling that we might be misusing the refresh action for something that this PR would cover.

@brunoprietog Could you help me with a practical example of when a refresh should morph between two URLs that have same path but different query params? How would the visit be triggered?

brunoprietog:
Well, you might be on a product checkout page and there is a form to enter a discount code with an apply button. When you press that button, you are redirected to the same path but now you have the code parameter with the discount code. This causes the backend to look for the discount code if there is one and update the prices accordingly. Magically, Turbo did morphing.

pfeiffer:
> Well, you might be on a product checkout page and there is a form to enter a discount code with an apply button. When you press that button, you are redirected to the same path but now you have the code parameter with the discount code. This causes the backend to look for the discount code if there is one and update the prices accordingly. Magically, Turbo did morphing.

Okay, so in this case, would you be able to use the potential fix in this PR like:

```html
<form action="/checkouts/new" method="get" data-turbo-action="replace" data-turbo-replace-method="morph">
  <input type="text" name="coupon" />
  <input type="submit" name="Apply coupon" />
</form>
```

I'm not sure the example you provided is a  page "refresh" in it's conceptual meaning, is it? It's revisiting the form, applying the coupon to the page; ie. the _content_ of the page depends on the query param (eg. "You've got 50% off!").

krschacht:
@pfeiffer Sorry I'm chiming in late. I actually spend a few hours on this PR yesterday, digging through the code much more deeply and thinking through scenarios. I can share some bigger picture thinking in a sec, but first to directly answer your question:

1) The case where you're on an `/post/new` and you save mid-stream but you want to continue editing so after the save the user is redirected to `/post/1/edit`. This is a case where `action=replace` and `replace-method=morph` would work.

2) Think about pagination. Often it's implemented as "/comments?page=1". I believe Turbo already handles this because the querystring is ignored and so the URL can be considered a match (i.e. a refresh). Am I correct about that?
* But there may be cases where you implement pagination without the query string. Imagine a case where you're paging through full photos, one per page. The URL may go from: `/photo/1` to `/photo/2`. It can be really nice to morph because there may be a bunch of other state on the page that you want to remain and the actual HTML replacement is really minimal.

What's notable about this case is that you wouldn't want it do it with a replace because you would want the history to keep working.

brunoprietog:
Having to do it that way is not so magical anymore. The beauty of page refreshes is just that it's magic, that you almost don't have to think about it.

In the example itself, the page is practically the same. Only the prices or a slight message change. Why wouldn't that be an appropriate candidate for a page refresh?

krschacht:
Bigger picture: I do agree that my original approach of calling something a "refresh" when it the URL was changing is bad naming. I think your proposal of `data-turbo-action="replace" data-turbo-replace-method="morph"` makes more sense conceptually.

However, as I began implementing this and as I consider the pagination case (my 2nd example in my recent comment) then I started wondering if it's only a `replace` method or if we want to declare morph even for `advance`. So last night I started thinking through:

`data-turbo-action="replace | advance" data-turbo-action-method="morph | full"`

For a href, the default action is `advance` and the default method would `full`.

pfeiffer:
> In the example itself, the page is practically the same. Only the prices or a slight message change. Why wouldn't that be an appropriate candidate for a page refresh?

Mostly because I think what you're doing in this example is _not_ a **refresh** - you are not fetching the latest state of the _current page_ shown in the browser and morphing it. Instead, the content differ due to it's constraints - a different query param. Therefor it shouldn't be considered a refresh, but a replace.

Maybe there's a better example?

I have a feeling that the use-case for #1079 would be much better solved with the solution in PR, avoiding the pitfalls of the #1079.

krschacht:
@brunoprietog I think you are using the word _**refresh**_ to mean "morphing" whereas Mattias is using the word refresh to mean "the same conceptual page, with an unchanged URL, has an updated state which needs to be retrieved".

@pfeiffer What do you think of my [updated proposal](https://github.com/hotwired/turbo/pull/1145#issuecomment-1964425039)?

pfeiffer:
> Bigger picture: I do agree that my original approach of calling something a "refresh" when the URL was changing is bad naming. I think your proposal of `data-turbo-action="replace" data-turbo-replace-method="morph"` makes more sense conceptually.

> I think you are using the word refresh to mean "morphing" whereas Mattias is using the word refresh to mean "the same conceptual page, with an unchanged URL, has an updated state which needs to be retrieved". Mattias and I are proposing an alternate means of telling the system that you want to morph the page, without it needing to be a refresh. That's what this PR is attempting to do.

Exactly. Conceptually, a `refresh` would be a "request to fetch the updated state of this resource". If the constraints change (think `?page=..` or `?coupon=..`) I would argue that it is indeed a `replace`. I do think that it'd be useful to be able to morph the replace action as suggested in this PR, though! 

> However, as I began implementing this and as I consider the pagination case (my 2nd example in my recent comment) then I started wondering if it's only a `replace` method or if we want to declare morph even for `advance`. So last night I started thinking through a variation which makes this independent of the `replace` action.

There has been [discussions](https://github.com/hotwired/turbo/issues/1085) [previously](https://github.com/hotwired/turbo/issues/1177) regarding this. 

I can imagine there'd be some issues with this, namely around `restore` - are they also supposed to morph 'back'?

As a side note, in our application, we've implemented a custom renderer to support morphing (of certain elements) when doing advance visits. That could also cover the use-case of eg. pagination. We can tag elements with an `id` and `[data-turbo-morph]` in which case the custom renderer morphs them _after_ replace to switch the bodies.

pfeiffer:
Added the renderer as a gist here in case it's of any interest: https://gist.github.com/pfeiffer/54e3bf929637cad42d534c84f65b1e99

krschacht:
@pfeiffer Does your app run on a fork of Turbo in order to support your MorphableRenderer and MorphableSnapshot, or are you able to wire those in as overrides? From your gist I couldn't see how Turbo would be made to use these at the appropriate time. I would like to try this out in my app.

I understand there are some issues with restore and morphing. That's probably a good enough reason to restrict morphing only to replace, for now. I can try to re-work this PR to do that. But I do think that we should eventually support cases where you want to advance with a morph and you would, therefore, want to fix the underlying restore / snapshot issue. The reasons for wanting to advance & morph is because (a) the page is changing minimally and there is a bunch of other state that you would like preserved, but (b) the changed state is meaningful enough that you want the URL to update and you want the user to be able to go "back" to the previous version.

A good example of wanting `advance` with `morph` is a full screen Next-pagination from `/photos/1` to `/photos/2`

brunoprietog:
@krschacht I don't want it to be misunderstood that I'm against this PR, quite the contrary! I also see the need to trigger morphing in other similar scenarios that aren't page refreshes.

The discussion with @pfeiffer goes the other way, I feel it is more related to the limitations found in native adapters. I have the feeling that practically the strongest argument for using `href` instead of `pathname` is only given by a limitation in the handling of those adapters. Wouldn't this solution have a similar problem in Turbo native?

Anyway, it makes sense to me the point you make about what a page refresh is conceptually. Still, I feel like going back to just using `href` for the review would be a step backwards in terms of how magical it is.

Maybe @jorgemanrubia or @afcapel would have a stronger opinion on this.

pfeiffer:
> @pfeiffer Does your app run on a fork of Turbo in order to support your MorphableRenderer and MorphableSnapshot, or are you able to wire those in as overrides? From your gist I couldn't see how Turbo would be made to use these at the appropriate time. I would like to try this out in my app.

Yeah, it's vanilla Turbo. We listen to the `turbo:before-render` which allows you to provide your [own `render` method](https://turbo.hotwired.dev/handbook/drive#custom-rendering). I've updated the gist to include how we do that: https://gist.github.com/pfeiffer/54e3bf929637cad42d534c84f65b1e99#file-turbo_render-js

pfeiffer:
> The discussion with @pfeiffer goes the other way, I feel it is more related to the limitations found in native adapters. I have the feeling that practically the strongest argument for using `href` instead of `pathname` is only given by a limitation in the handling of those adapters.

Yeah, well - sort of. The native adapters revealed a potential 'misuse' of the `refresh` concept to perform page changes with morphing. The use-case in #1079 and https://github.com/hotwired/turbo/pull/1145#issuecomment-1964380400 is absolutely valid, but I think it would be better solved by something like the suggested change in this PR, reserving the `refresh` to the action of getting the latest state of the current visited URL, not changing URLs with morphing.

Besides, the current behavior of assuming a `replace` visit with similar `pathname` is a page refresh with morphing, makes it really hard to opt-out of the morphing - imagine you'd like to _not_ morph a `replace` visit from `/new` to `/new?something` but still allow both of these to individually be refreshed eg. via broadcasts. You'd need to inspect the `request.referer` and output a `meta` tag disallowing morphing based on some params and a referer, which feels off.

brunoprietog:
Meta tags are supposed to determine whether morphing should be done or not. This as far as I understand is not being respected in native adapters either. Which makes me wonder even more if we are defining something just because of a limitation in the implementation from the native side or because of something more conceptual.

krschacht:
@brunoprietog and @pfeiffer Take a look at this PR now. It's not quite ready to merge in since I need to spend more time examining and testing the cases with turbo-frames, but it covers everything we've been discussing in this thread. However, I welcome any feedback on it.

Bruno — Notably, see the changed behavior from your #1079. Now a querystring change alone will cause a morph not to occur but you can explicitly specify the behavior in order to ensure it does:
https://github.com/hotwired/turbo/pull/1145/files#diff-ca04d88186bdf2d24a3a6a146366aac4be3561b4978380ee05066ce7d6ec70caR115

src/tests/server.mjs
```diff
@@ -171,7 +171,6 @@ router.get("/messages", (request, response) => {
 function receiveMessage(content, id, target) {
   const data = renderSSEData(renderMessage(content, id, target))
   for (const response of streamResponses) {
-    console.log("delivering message to stream", response.socket?.remotePort)
```
krschacht, line 174: While I was removing all my console.log() statements I found this old one left in the code.

src/core/drive/page_view.js
```diff
@@ -55,8 +55,17 @@ export class PageView extends View {
...
+            (this.isPageRefresh(visit) || this.isReplaceMethodMorph(visit))
+  }
+
   isPageRefresh(visit) {
-    return !visit || (this.lastRenderedLocation.pathname === visit.location.pathname && visit.action === "replace")
+    return !visit || (this.lastRenderedLocation.href === visit.location.href && visit.action === "replace")
+  }
+
+  isReplaceMethodMorph(visit) {
+    return !visit || (visit.action === "replace" && visit.replaceMethod === "morph")
```
pfeiffer, line 68: `!visit` reads a bit weird here; is it needed?

krschacht:
@brunoprietog and @pfeiffer Take a look at this PR now. It's not quite ready to merge in since I need to spend more time examining and testing the cases with turbo-frames, but it covers everything we've been discussing in this thread. I welcome any feedback on it.

Bruno — Notably, see the changed behavior from your #1079. Now a querystring change alone will cause a morph not to occur but you can explicitly specify the behavior in order to ensure it does:
https://github.com/hotwired/turbo/pull/1145/files#diff-ca04d88186bdf2d24a3a6a146366aac4be3561b4978380ee05066ce7d6ec70caR115

Regarding your comment:

> Meta tags are supposed to determine whether morphing should be done or not. 

This is true, but the issue is more subtle than this. The meta tags determine whether **_refresh_** should be done with morphing. Imagine you're a PHP developer, you're using Turbo, you set `<meta name="turbo-refresh-method" content="morph">`, and then you create a page which does `home.php?page=1` which does a replace link to `home.php?page=2`. You're surprised because you wanted to do a replace but you did **not** expect that to morph, since it's not a refresh — it's a navigation to a conceptually different URL and content. This is the problem with assuming that a replace visit with a similar pathname is a refresh.

With this PR you now explicitly specify that the replace-method should be morph. It does not automatically decide it for you.

src/core/drive/navigator.js
```diff
@@ -162,4 +164,10 @@ export class Navigator {
     const sameLocationRedirect = fetchResponse.redirected && fetchResponse.location.href === this.location?.href
     return sameLocationRedirect ? "replace" : "advance"
   }
+
+  #getReplaceMethodForFormSubmission(formSubmission, fetchResponse) {
+    if (this.#getActionForFormSubmission(formSubmission, fetchResponse) !== "replace") return
+    const { submitter, formElement } = formSubmission
+    return getVisitReplaceMethod(submitter, formElement) || "body"
```
pfeiffer, line 171: Wondering if there's a better value than `body` here. It does not have the same clear action as `morph` does. The corresponding renderer is called `PageRenderer` and the rendering is not isolated to changing the body; it also does some merging of `head` elements etc.

src/core/drive/page_view.js
```diff
@@ -16,7 +16,7 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const shouldMorphPage = this.isPageRefresh(visit) && this.snapshot.shouldMorphPage
+    const shouldMorphPage = this.shouldMorphPage(snapshot, visit)
```
pfeiffer, line 19: Since the `shouldMorphPage(..)` is only used here, what about removing the `snapshot` argument and naming it more in relation to a visit? Eg. `isMorphableVisit(visit)` or something like that and keep the `this.snapshot` clause on this line. This is a bit closer to the original code as well and would remove the variable name being the same as the method name :)

src/core/drive/navigator.js
krschacht: It's true that the contents of `<head>` also gets merged, but the body is what gets replaced and that's the thing in question; that's what differs between the morph & non-morph case. I was keying off this description from the docs:

> During rendering, Turbo Drive replaces the current <body> element outright and merges the contents of the <head> element. The JavaScript window and document objects, and the <html> element, persist from one rendering to the next.

But I initially had this as `full`. Do you think that's better? I don't feel strongly so I'm happy to go with `full` or something else.

src/core/drive/page_view.js
pfeiffer: I'm thinking that `isReplaceMethodMorph(null) === true` reads wrong. The method is used together with `isPageRefresh` which would return true if there's no visit, so I don't think it makes sense to have here.

krschacht: Good call, fixing

brunoprietog:
Good discussion here, these are all good points.

What I wanted to say about meta tags is that apparently native adapters are not considering them, so it's not just a matter of `href` or `pathname`. But this remains to be verified.

Personally I like where this is headed.

jorgemanrubia:
Thanks for the work here @pfeiffer and to everyone for the discussion around it.

I understand there are scenarios where fined-grained controls would be useful but not sure that `data-*` attributes to set the replacement methods are the way to go here. It's not just about offering a solution for the problem at hand but about Turbo as a whole and its programming model. Every new option we add adds complexity, and we need to be very mindful about this, because Turbo offers quite a few options already.

We intentionally left using morphing out of arbitrary navigations because needing those is more rare and to keep the programming model as simple as possible. I know there is a need for this, but I think we need to take a step back to see how to better fit the whole "replacement strategy" idea so that Turbo's story remain as cohesive as possible.

krschacht:
Hi @jorgemanrubia, no problem. I totally respect that. I'll close this PR; I won't do any further work on it. To be honest, I had already worked around my original need for this with a creative use of <turbo-frame>. I was only wrapping this up since it seemed like there was a deeper problem it was solving with the `pathname` comparison.

Check out my other PR which is ready. I'll tag you in it. And meanwhile, there are a couple other bugs in turbo that are bigger impacts for me so I'll chase those down. :)

pfeiffer:
@jorgemanrubia I'm also fine with closing this and agree that it's important to keep things streamlined and simple. My main motivation participating in PR was that the change introduced in #1079, treating a `replace` visit with equal `pathname`'s (and not `href`s) as morphable refresh feels off and this PR provided a way to rollback that change while still allowing control over if a replace action should be morphable.

The change in #1079 conceptually is not a refresh (it's a replace to different URLs!) and the inconsistency showed up while implementing morphing in Turbo Native (https://github.com/hotwired/turbo-ios/pull/178#discussion_r1501703274).

Issue #1105, opened 2023-12-09
Page refreshes: Reloading When Assets Change is not being handled when morphing


brunoprietog:
Page refreshes with morphing currently don't take into account if assets change. This can lead to a version of the page that doesn't match the styles or new JavaScript after a deploy.

An example of this could be the addition of a new Stimulus controller. If in a deploy the controller is added to the JavaScript assets and a user triggers a page refresh, there will be html making use of the new Stimulus controller, but the assets are still the old ones and therefore will not behave as expected. This could be solved for CSS styles but for JavaScript it's more difficult. Something similar applies to turbo frames.

I'm not sure how this could be solved. I thought about replicating the Turbo drive behavior in the morph renderer, but this could be quite nasty. You would lose scrolling, Turbo frames status, basically everything.

How could this be addressed? Thoughts?

afcapel:
This is the same problem we've had since forever when we push turbo-stream updates or other async content. The client can receive new markup before it has navigated to a new page and it will render it with stale assets.

I think for page refreshes the fix should be pretty simple, though. We just need to inherit the `MorphRenderer` from `PageRenderer` so it issues a full page reload if the assets signature changes.

afcapel:
Fix in https://github.com/hotwired/turbo/pull/1146

PR #1146, opened 2024-01-30
Ensure page refreshes cause a reload when assets change

afcapel:
Fixes https://github.com/hotwired/turbo/issues/1105
Fixes https://github.com/hotwired/turbo/issues/1116


kevinmcconnell (approved):
Looks great. Nice that the fix was straightforward! And logically, it makes sense that the morph renderer is a type of page renderer; I think it’s expected that they would have behaviour in common.

PR #1148, opened 2024-01-31
`LinkPrefetchObserver`: replace `dataset` with `getAttribute`

seanpdoyle:
First, import the `StreamMessage` module to utilize the `contentType`
instead of the hard-coded `"text/vnd.turbo-stream.html"` String.

Next, replace occurrences of [HTMLElement.dataset][] with the
`"data-"`-prefixed calls to [Element.getAttribute][] so that they are as
`grep`-able as their counterparts elsewhere in the codebase.

Finally, remove a `link` comparison conditional clause, since it's
repeated verbatim several lines prior to it.

[HTMLElement.dataset]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset
[Element.getAttribute]: https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute

### DIFF
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -5,6 +5,7 @@ import {
   findClosestRecursively
 } from "../util"
 
+import { StreamMessage } from "../core/streams/stream_message"
 import { FetchMethod, FetchRequest } from "../http/fetch_request"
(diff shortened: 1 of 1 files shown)


afcapel:
Thanks @seanpdoyle !

Issue #1151, opened 2024-01-31
`turbo_stream` request with updating url


loqimean:
I found a nice example on the Medium website that shows us that it is really nice to have the ability to request turbo_stream in a turbo_frame, it will update the `src` of a frame, but even when I USE `data-turbo-action="advance"` it doesn't change the browser URL.

```ruby
# app/controllers/home_controller.rb
class HomeController < ApplicationController
  def index
  end
end
```

```erb
<%# index.turbo_stream.erb %>
<%= turbo_stream.update(:test) do %>
  <%= link_to("Time now is #{Time.now}", home_path(time: Time.now.to_s.parameterize), data: { turbo_stream: true }) %>
<% end %>

<%# index.html.erb %>
<h1>Hello</h1>
<% turbo_frame_tag :test, data: { turbo_action: "advance" } do %>
    <%= link_to("Time now is #{Time.now}", home_path(time: Time.now.to_s.parameterize), data: { turbo_stream: true }) %>
<% end %>
```

loqimean:
Section #8
https://dev.to/chwistophe/rails-7-pagination-with-turbo-frame-streams-infinite-scrolling-stimulus-js-3fih

loqimean:
So, I gues this thing should be fixed or added feature instead of this workaround

brunoprietog:
Hi @loqimean, Turbo streams aren't intended to update a visit, they are only intended to perform more complex dom operations in a response.

Issue #792, opened 2022-11-09
Unable to manage history state programmatically (stream response)


pySilver:
My use case is fairly simple. I have a list of products and a filter POST form. Whenever a user triggers form submit I receive a stream response that updates the product list and current page URL to something shareable. Filter form itself is contained in a `<turbo-frame>` if that matters. 

My goal is to update the page URL every time user makes changes to a filter form while letting them go back in history and "unapply" filters by simply pressing the `Back` button in a browser. 

Here is my custom stream action:
```js
Turbo.StreamActions.my_push_state = function () {
  const url = this.getAttribute("url");
  history.pushState(history.state, "", url);
  Turbo.navigator.history.push(url);
  console.log("my_push_state", url);
};
```

And here is the stream response from the server after the filter form is submitted:

```html
<turbo-stream action="replace"><template>...list of products</template></turbo-stream>

<turbo-stream
  action="my_push_state"
  url="http://localhost:3000/?foo=bar"
></turbo-stream>
```

I cannot make it work. When the `Back` button is pressed the following error appears in a console:

```
Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at ProgressBar.uninstallProgressElement (webpack-internal:///./node_modules/@hotwired/turbo/dist/turbo.es2017-esm.js:1481:38)
    at eval (webpack-internal:///./node_modules/@hotwired/turbo/dist/turbo.es2017-esm.js:1455:22)
uninstallProgressElement @ turbo.es2017-esm.js?7ebb:1457
eval @ turbo.es2017-esm.js?7ebb:1431
setTimeout (async)
fadeProgressElement @ turbo.es2017-esm.js?7ebb:1453
hide @ turbo.es2017-esm.js?7ebb:1430
hideVisitProgressBar @ turbo.es2017-esm.js?7ebb:2070
visitRequestFinished @ turbo.es2017-esm.js?7ebb:2050
finishRequest @ turbo.es2017-esm.js?7ebb:1795
requestFinished @ turbo.es2017-esm.js?7ebb:1920
perform @ turbo.es2017-esm.js?7ebb:529
await in perform (async)
issueRequest @ turbo.es2017-esm.js?7ebb:1767
visitStarted @ turbo.es2017-esm.js?7ebb:2018
start @ turbo.es2017-esm.js?7ebb:1722
startVisit @ turbo.es2017-esm.js?7ebb:2283
[...]

PR #1152, opened 2024-01-31
Add ability to click link in a frame with a `data-turbo-stream="true"`

loqimean:
So, I guess it would be fine to have the ability to change the URL with turbo_stream request for some reason: when you have pagination or other type of request, like filters, you need to provide for user URL with its parameters, so in my case me too need. So, with this type of pages-request you'll update the page by parts with turbo_stream but it affects the URL, and the only way to solve it is to use JS code like stimulusUse with its mutate or create some workarounds that solve It by handling turbo events.
So, when contributors added the ability to create a `turbo_stream` with a GET request we also need the ability to show this GET request for the browser to add an ability for the user to share it, or find it, to be on the same page after page reloading. So, let me solve this problem, help or just complete this issue, please
solves #1151
solves #792

### DIFF
--- a/src/http/fetch_request.js
+++ b/src/http/fetch_request.js
@@ -124,6 +124,7 @@ export class FetchRequest {
     const event = await this.#allowRequestToBeIntercepted(fetchOptions)
     try {
       this.delegate.requestStarted(this)
+      this.delegate.changeHistory()
 
       if (event.detail.fetchRequest) {
         this.response = event.detail.fetchRequest.response


loqimean:
@afcapel @dhh check it please and let me now If I need something improve or change, it's just my first great contribution to a big repository, thanks!

brunoprietog:
I'm afraid that this is the expected behavior. Turbo streams are not intended, in principle, to alter the browsing history, only to make changes to the DOM. At least that's how I understand it.

So, it's possible to have a controller that reacts to a request that is in Turbo stream format, so that this controller responds with the Turbo stream actions that were requested by the call to the controller. But that controller is not necessarily prepared to receive normal HTML requests, so if you change the URL in the browser and a user, for example, refreshes the page, or simply clicks on a link and then goes back and caching is disabled on that page, will get a blank page. This could break many applications eventually.

loqimean:
> I'm afraid that this is the expected behavior. Turbo streams are not intended, in principle, to alter the browsing history, only to make changes to the DOM. At least that's how I understand it.
> 
> 
> 
> So, it's possible to have a controller that reacts to a request that is in Turbo stream format, so that this controller responds with the Turbo stream actions that were requested by the call to the controller. But that controller is not necessarily prepared to receive normal HTML requests, so if you change the URL in the browser and a user, for example, refreshes the page, or simply clicks on a link and then goes back and caching is disabled on that page, will get a blank page. This could break many applications eventually.

Sure, I got your point, what I should when I need to load some parts of html by turbo_stream, but it is filters page and I'd like to change url, 'cause when user refreshes page there is page without filters. So, what approach you recommend in this case?

afcapel:
You can use `data-turbo-action='advance'` to update the URL when loading a frame. Hope that helps.

loqimean:
> You can use `data-turbo-action='advance'` to update the URL when loading a frame. Hope that helps.

But it's not relevant because I don't need turbo-frames, or just give me an example app

brunoprietog:
The question is, why do you need Turbo streams for that? Isn't it a typical page navigation? Maybe with the replace action?

loqimean:
> The question is, why do you need Turbo streams for that? Isn't it a typical page navigation? Maybe with the replace action?

What do you mean? Like it's typical page navigation but instead of reloading whole page to be navigated to next filtered page version,because of using turbo streams

brunoprietog:
What I mean is that you could build these filters with typical form submissions and page navigations instead of Turbo streams using the replace action. That's why I'd like to better understand the use case you need. Even now with morphing there are more options to simplify it. Maybe with that we could suggest an alternative

PR #1160, opened 2024-02-04
Allow different methods of appending turbo-streams to document

cryptogopher:
Right now `<turbo-stream>` elements received in response are added all-at-once to the document. This results in rendering in order of apperance in response, without chance to modify the list of elements. Example response:

image

The problem arises, when you have some custom stream action (in above example: *click*), that needs to modify list of actions client-side. The action *click* in the example is responsible for invoking *onclick* handler - which itself is a call to _renderStreamMessage()_ and renders some template consisting of `<turbo-stream>`s to keep things DRY:

<pre>
Turbo.StreamActions.click = function() {
  this.targetElements.forEach((e) => { e.click() })
}
</pre>

Without modifications, this yields the following order of rendering turbo streams:

image

Notice how the additional elements will only be rendered at the end of the chain - which is not our intent. The order matters and we would like to have turbo streams resulting from *click* action rendered right after that action.

Solution:

Allow customization of the method of adding turbo stream templates to document, by separating `appendChild()` call inside `StreamMessageRenderer#render` into separate method (see attached commit). Later this method can be overridden by something like this, to keep order of actions proper in the example above:
<pre>
Turbo.session.streamMessageRenderer.appendFragment = async function (fragment) {
  for (let child of [...fragment.children]) {
    document.documentElement.appendChild(child)
    if (child.action == "click") {
      await new Promise((resolve) => {
        new MutationObserver((mutations, observer) => {
          mutations.forEach((m) => {
            if ([...m.removedNodes].includes(child)) {
              resolve(child)
              observer.disconnect()
            }
          })
        }).observe(document.documentElement, {childList: true})
      })
    }
  }
}
</pre>

That yields expected result:

image

### DIFF
--- a/src/core/streams/stream_message_renderer.js
+++ b/src/core/streams/stream_message_renderer.js
@@ -7,7 +7,7 @@ export class StreamMessageRenderer {
     Bardo.preservingPermanentElements(this, getPermanentElementMapForFragment(fragment), () => {
       withAutofocusFromFragment(fragment, () => {
         withPreservedFocus(() => {
-          document.documentElement.appendChild(fragment)
+          this.appendFragment(fragment);
         })
       })
     })
@@ -20,6 +20,10 @@ export class StreamMessageRenderer {
   }
 
   leavingBardo() {}
+
+  appendFragment(fragment) {
+    document.documentElement.appendChild(fragment);
+  }
 }
 
 function getPermanentElementMapForFragment(fragment) {


afcapel:
@cryptogopher even if we extract an `appendFragment` method, it will part of the internal API and we can not guarantee it won't change without any notice. It will be a brittle to depend on that.

Does handling `turbo:before-stream-render` work for your use case? In an event handler you can completely override the turbo stream `render` method, including how the stream element is appended to the DOM. That's in the official API and should be stable.

cryptogopher:
@afcapel I understand, that this is not the most future-proof solution. But it's the smallest change I can think of, that solves the problem.

Using `turbo:before-stream-render` does not solve the problem. Actually the 2nd screenshot in PR description above is generated using `turbo:before-stream-render` in following way (with `turbo` code unchanged):
<pre>
function beforeStreamRender(event) {
  console.log(event.target)
}
document.addEventListener('turbo:before-stream-render', beforeStreamRender)
</pre>
As you can see the order of events is unvaforable. The problem lays in how the order of rendering is determined. `StreamElement#render` is called by `StreamElement#connectedCallback`, i.e. the rendering order is determined by the order of connecting the stream elements to the DOM (not the _order of appearance of stream elements in the DOM_). At least it works like that in Firefox.

All turbo streams from server response are connected at once by `appendChild()`. Even if custom action handler - like *click* in my example - tries to insert some additional action in the middle, rendering happens only after all of the previous elements have been rendered. That results in out-of-order execution of handlers.

I took a look at all available turbo events. The other one that could potentially be used was `turbo:before-fetch-response`. It happens before anything is rendered/connected to DOM. Unfortunately it does not solve the problem as long as *click* action is not the first one. Otherwise - without rendering all of the actions that come before *click* - there is no way to target/targets properly determined for *click* action (as previous actions can change the DOM).

cryptogopher:
After some more work on the problem, it turns out that client-side inclusion of `turbo-stream`s during `turbo-stream` rendering is not an easy idea to implement properly. For example this creates problems with setting focus on `autofocus` elements.

So instead I moved out all unique information from `onclick`'s `renderStreamMessage()` handler into the DOM elements' attributes, and now it is possible to render the same handler server-side.

Closing this PR as problem has been solved without it.

PR #1162, opened 2024-02-05
Enable InstaClick by default

afcapel:
It will be the new default for Turbo 8. You can always opt out by setting `<meta name="turbo-prefetch" content="false">` in the head of your HTML.

To help older apps with the transition we don't prefetch any links that contain the `data-remote`, `data-behavior`, `data-method` or `data-confirm` attributes. These are used by UJS in older apps to trigger unsafe behavior on links. More modern apps tend to implement these as `<button>`s instead.

//cc @dhh @davidalejandroaguilar

### DIFF
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -58,7 +58,7 @@ export class LinkPrefetchObserver {
   }
 
   #tryToPrefetchRequest = (event) => {
-    if (getMetaContent("turbo-prefetch") !== "true") return
+    if (getMetaContent("turbo-prefetch") === "false") return
 
     const target = event.target
     const isLink = target.matches && target.matches("a[href]:not([target^=_]):not([download])")
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -138,6 +138,10 @@ export class LinkPrefetchObserver {
       return false
     }
 
+    if (link.hasAttribute("data-remote") || link.hasAttribute("data-behavior") || link.hasAttribute("data-method") || link.hasAttribute("data-confirm")) {
+      return false
+    }
+
     if (link.origin !== document.location.origin) {
       return false
     }
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -1,9 +1,9 @@
 import {
+  dispatch,
   doesNotTargetIFrame,
   getLocationForLink,
   getMetaContent,
-  findClosestRecursively,
-  isUJSLink
+  findClosestRecursively
 } from "../util"
 
 import { StreamMessage } from "../core/streams/stream_message"
@@ -18,7 +18,6 @@ export class LinkPrefetchObserver {
   constructor(delegate, eventTarget) {
     this.delegate = delegate
     this.eventTarget = eventTarget
-    this.preventLinkPrefetch = (link) => isUJSLink(link)
   }
 
   start() {
@@ -140,7 +139,12 @@ export class LinkPrefetchObserver {
       return false
     }
 
-    if (this.preventLinkPrefetch(link)) {
+    const event = dispatch("turbo:before-prefetch", {
+      target: link,
+      cancelable: true
+    })
+
+    if (event.defaultPrevented) {
       return false
     }
 
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -2,7 +2,8 @@ import {
   doesNotTargetIFrame,
   getLocationForLink,
   getMetaContent,
-  findClosestRecursively
+  findClosestRecursively,
+  isUJSLink
 } from "../util"
 
 import { StreamMessage } from "../core/streams/stream_message"
@@ -17,6 +18,7 @@ export class LinkPrefetchObserver {
   constructor(delegate, eventTarget) {
     this.delegate = delegate
     this.eventTarget = eventTarget
+    this.preventLinkPrefetch = (link) => isUJSLink(link)
   }
 
   start() {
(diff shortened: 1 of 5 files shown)


src/observers/link_prefetch_observer.js
```diff
@@ -134,7 +134,11 @@ export class LinkPrefetchObserver {
   #isPrefetchable(link) {
     const href = link.getAttribute("href")
 
-    if (!href || href === "#" || link.getAttribute("data-turbo") === "false" || link.getAttribute("data-turbo-prefetch") === "false") {
+    if (!href || href.startsWith("#") || link.getAttribute("data-turbo") === "false" || link.getAttribute("data-turbo-prefetch") === "false") {
+      return false
+    }
+
+    if (link.hasAttribute("data-remote") || link.hasAttribute("data-behavior") || link.hasAttribute("data-method") || link.hasAttribute("data-confirm")) {
       return false
```
kevinmcconnell, line 142: I'd be tempted to not build these checks in, but instead document the new behaviour, and have a cancellable event people can use to avoid prefetching according to their own needs. I can definitely see how this is helpful for people with older codebases, but it also feels like Turbo is having to carry knowledge about UJS, and once we have it, it's harder to remove later.

I don't feel very strongly about it, though. There's pros and cons either way.

kevinmcconnell: Also if we're keeping it, maybe worth extracting this to give it a helpful name, like `#hasUJSattributes` or similar.

kevinmcconnell (approved):
Love the idea of switching the default 👍 It's made such a noticeable difference in the places we've used it so far.

Just had a question about how we handle the UJS compatibility part, but it's your call.

src/observers/link_prefetch_observer.js
afcapel: @kevinmcconnell it is a trade-off, for sure, but I think in this case the pros outweigh the cons. Having this in place means that older applications can upgrade without needing any custom code. Without this the upgrade for an application like Basecamp was far from trivial. And adding a cancellable event as an extension point is also extending the public API in a way we'll have to maintain.

dhh: I'd like to make this as "drop-in replacement" as possible. Even for older apps. But I'm wondering if we can extract all those checks into a single function that just happens to be the default before filter on this. A filter that you can then overwrite with your own function, if you either need more/less/different on the opt-out.

afcapel: I've added an [extension point](https://github.com/hotwired/turbo/blob/5b866dba8e571fe4eb0a2ef3d076255514d29ce1/src/core/index.js#L111-L124) so now you can do something like:

```js
Turbo.preventLinkPrefetch((link) => link.hasAttribute("hx-post"))
```

But I find a bit confusing that it only overrides some checks, not all of them. We have [several different conditions](https://github.com/hotwired/turbo/blob/5b866dba8e571fe4eb0a2ef3d076255514d29ce1/src/observers/link_prefetch_observer.js#L136-L175), checking things such as whether the link has the same origin, if it has `data-turbo=false`, if it targets an `iframe`, etc. I don't think there's a practical use case to remove any of them, the prefetch is not going to work in those cases.

We could allow you to redefine the whole logic, including the defaults, but then you'll have to recreate all these checks in application logic, which may be error prone and also you'll miss out if the framework updates this behavior.

I think the most practical use case is adding more checks on top of the defaults. But in that case I think a cancellable event, as @kevinmcconnell suggested, makes more sense and feels more idiomatic.

dhh: I don't have a strong opinion on that. Only that this should be on by default, and we should account for these common caveats with earlier Rails apps.

afcapel: Cool, thanks. I've added a cancellable event in 7327a955ea93ef885ea157977acaeb96068f80ba. This can be used add custom logic to opt out of prefetching.

```javascript
document.body.addEventListener("turbo:before-prefetch", (event) => {
  if (isUJSLink(event.target)) event.preventDefault()
})

function isUJSLink(link) {
  return link.hasAttribute("data-remote") || link.hasAttribute("data-behavior") || link.hasAttribute("data-method") || link.hasAttribute("data-confirm")
}
```

We'll add this to the documentation so older apps have an easy migration path.

src/tests/functional/link_prefetch_observer_tests.js
```diff
@@ -62,6 +62,22 @@ test("it doesn't prefetch the page when link has data-turbo=false", async ({ pag
...
+    document.body.addEventListener("turbo:before-prefetch", (event) => {
+      if (event.target.hasAttribute("data-remote")) {
+        event.preventDefault()
+      }
+    })
+  })
+
+  await assertNotPrefetchedOnHover({ page, selector: "#anchor_with_remote_true" })
+})
+
```
davidalejandroaguilar, line 80: We might want to add a test for not prefetching when `data-turbo-prefetch` is set to `"false"`.

davidalejandroaguilar (approved):
Looks good, thank you! 😺 , just have a non-blocking comment.

Excited this is going to be enabled by default on Turbo 8!

src/tests/functional/link_prefetch_observer_tests.js
afcapel: That seems already covered in [its own test](https://github.com/hotwired/turbo/blob/7327a955ea93ef885ea157977acaeb96068f80ba/src/tests/functional/link_prefetch_observer_tests.js#L55-L58). The features are fairly independent, so no need to test their interaction.

navidemad:
is it possible to skip also the <a> with data-turbo-stream="true" ?

kevinmcconnell:
@navidemad if you always want to skip any link that has `data-turbo-stream="true"` you could also do so in a `turbo:before-prefetch` event handler, along the line of [this example](https://github.com/hotwired/turbo/pull/1162#discussion_r1478939930) that @afcapel posted earlier.

PR #1164, opened 2024-02-05
Allow customizing the delay for InstantClick behavior

davidalejandroaguilar:
# Description

This PR adds a configurable delay to the new InstantClick behavior.

The 100 ms default is a nice one. However, some apps might be fine with having no delay and would prefer to prefetch as soon as possible, so [let's provide sharp knives](https://rubyonrails.org/doctrine#provide-sharp-knives).

Configuration
* Via a `data-turbo-prefetch-delay` attribute on the `<a>` element.
* Via a `turbo-prefetch-delay` meta tag.

Both can be set to any numerical value, or to `"0"` for no delay. Note that the latter means that `setTimeout` won't be used at all, so it will be an instant prefetch.

Notes
I also thought about using `"false"` instead of `"0"` for no delay, let me know if that or something else is preferred.

### DIFF
--- a/src/core/drive/prefetch_cache.js
+++ b/src/core/drive/prefetch_cache.js
@@ -10,14 +10,16 @@ class PrefetchCache {
     }
   }
 
-  setLater(url, request, ttl) {
+  setLater(url, request, ttl, delay) {
     this.clear()
 
-    this.#prefetchTimeout = setTimeout(() => {
-      request.perform()
-      this.set(url, request, ttl)
-      this.#prefetchTimeout = null
-    }, PREFETCH_DELAY)
+    if (delay === "0") {
+      this.#setNow(url, request, ttl)
+    } else {
+      delay = Number(delay) || PREFETCH_DELAY
+
+      this.#enqueue(url, request, ttl, delay)
+    }
   }
 
   set(url, request, ttl) {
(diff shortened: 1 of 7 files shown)


kevinmcconnell:
@davidalejandroaguilar what would be a use case for someone needing to change the prefetch delay? Do you think there will be cases where the difference is quite noticeable, or the default won't provide a good experience?

I think it's generally good if we can keep Turbo's API simple and small, by only adding attributes/tags that many people are likely to need. If a customisation is not in high demand, it might be better to keep it out of that API. (Which is not to say there couldn't still be a way to affect it in code, if necessary.)

Perhaps this is one of those common needs! But I was struggling to think of a concrete example of needing it, so was curious if you had more context about why the setting needs to be exposed. Thanks!

afcapel:
I agree with @kevinmcconnell. I'd like to see some practical cases where we need this before expanding the public API. 

Another thing to note is that we may add InstaClick for keyboard navigation, but [that could need different timing](https://github.com/hotwired/turbo/pull/1101#discussion_r1458077922). It'd be useful to know what we're going to do with the keyboard events, and if we need different delays for different events, before committing to an API for this.

davidalejandroaguilar:
Thanks for chiming in, appreciate it! @kevinmcconnell @afcapel 

I agree — there's no real need for configuring the delay value. What I was really going for here is a way to turn off the delay completely. I've updated the code to reflect this.

The goal is to provide an option for cases where even 100ms can make a big difference. In other words, keeping a default 100ms delay for casual browsing, with the option to remove it for quicker interactions.

I'm thinking of 2 types of apps:

* **Fast-paced apps:** Apps where users point and click things very fast, and where speed is critical. e.g. a Point of Sale app where a waiter is navigating categories and items to add things to the order as fast as possible.
* **Relaxed apps:** For more casual browsing where users explore and click things in a more relaxed way. e.g. a CMS.

I'm sorry I wasn't more clear before with the intention of the PR.

MrConorAE:
My two cents: I'd find actually *increasing* the prefetch delay useful.
I've found that the prefetch seems too fast in some cases, especially with users who use other pointing devices (like trackballs) that can move slower/in more 'steps', so being able to increase it, even just to 200ms, would significantly reduce the number of prefetches on links that don't get clicked.

PR #1167, opened 2024-02-06
Disable InstantClick for touch events

afcapel:
While testing https://github.com/hotwired/turbo/pull/1147 I found an issue on touch devices that often leads to duplicated requests on mobile. The culprit seems a `mouseenter` event that Safari fires after a `touchend` for compatibility reasons. The `mouseenter` was overriding the current prefetch request launched on `touchstart` causing a second one to be triggered.

According to ChatGPT:

> Some browsers may synthesize mouse events (including `mouseenter`) after touch events to ensure compatibility with web content not designed for touch interfaces. This means that a `mouseenter `event might be fired on a touch device, usually after a `touchend` event, as part of the sequence to simulate mouse interaction. This behavior can vary between browsers and might not always be consistent.

Seems we'll need to re-think how we implement Instaclick touch events. In the meantime, let's disable the touch handlers and leave InstantClick for desktop devices where it's working fine.

This PR also removes the `mouseleave` event listener on specific links to avoid memory leaks, as [was the original intention](//github.com/hotwired/turbo/pull/1147#discussion_r1476162113) of https://github.com/hotwired/turbo/pull/1147.

//cc @seanpdoyle @davidalejandroaguilar

### DIFF
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -36,7 +36,7 @@ export class LinkPrefetchObserver {
       capture: true,
       passive: true
     })
-    this.eventTarget.removeEventListener("mouseleave", this.#checkIfPrefetchValidAfterMouseLeave, {
+    this.eventTarget.removeEventListener("mouseleave", this.#cancelRequestIfObsolete, {
       capture: true,
       passive: true
     })
@@ -50,7 +50,7 @@ export class LinkPrefetchObserver {
       capture: true,
       passive: true
     })
-    this.eventTarget.addEventListener("mouseleave", this.#checkIfPrefetchValidAfterMouseLeave, {
+    this.eventTarget.addEventListener("mouseleave", this.#cancelRequestIfObsolete, {
       capture: true,
       passive: true
     })
@@ -85,7 +85,7 @@ export class LinkPrefetchObserver {
     }
   }
 
-  #checkIfPrefetchValidAfterMouseLeave = (event) => {
+  #cancelRequestIfObsolete = (event) => {
     if (event.target === this.#prefetchedLink) this.#cancelPrefetchRequest()
   }
 
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -12,8 +12,7 @@ import { prefetchCache, cacheTtl } from "../core/drive/prefetch_cache"
 
 export class LinkPrefetchObserver {
   started = false
-  hoverTriggerEvent = "mouseenter"
-  touchTriggerEvent = "touchstart"
+  #prefetchedLink = null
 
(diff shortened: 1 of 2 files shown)


src/observers/link_prefetch_observer.js
```diff
@@ -78,12 +81,19 @@ export class LinkPrefetchObserver {
...
         prefetchCache.setLater(location.toString(), fetchRequest, this.#cacheTtl)
-
-        link.addEventListener("mouseleave", () => prefetchCache.clear(), { once: true })
       }
     }
   }
 
+  #checkIfPrefetchValidAfterMouseLeave = (event) => {
+    if (event.target === this.#prefetchedLink) this.#cancelPrefetchRequest()
+  }
```
davidalejandroaguilar, line 90: I'd rename this to `#tryToCancelPrefetchRequest` and leave it in a single method:

```js
#tryToCancelPrefetchRequest = (event) => {
  if (event.target === this.#prefetchedLink) {
    prefetchCache.clear()
    this.#prefetchedLink = null
  }
}
```

Reason is that it's more evident that, on `mouseleave`, the callback is actually trying to cancel the prefetch request, rather than just "checking if it's valid".

Also, the reason I suggest adding "try", is because if the user doesn't leave the element within 100ms, we WILL make the prefetch request, and so this won't really cancel anything. So being explicit with "try" makes this more evident.

seanpdoyle (approved):
Are there tap related tests we need to remove?

* https://github.com/hotwired/turbo/blob/3c3eeb8471cf7ac6be3240dc5ee36722137f844f/src/tests/functional/link_prefetch_observer_tests.js#L267-L285
* https://github.com/hotwired/turbo/blob/3c3eeb8471cf7ac6be3240dc5ee36722137f844f/src/tests/functional/link_prefetch_observer_tests.js#L243-L255

davidalejandroaguilar (approved):
Great work and thank you for digging to find this! Seems like we will need to dig into the nuances of mobile and similar devices to add InstantClick functionality to them.

The change looks good, though I have a couple suggestions to make the code intent more clear.

afcapel:
> Are there tap related tests we need to remove?

Yes, good catch! Removed in 209fbffa882a8c8a369e824526ab86a98c3bbd62

src/observers/link_prefetch_observer.js
afcapel: `tryToCancelPrefetchRequest` also sound like you're going to try to cancel it, no matter what. But we only cancel if the user left the link. What about `cancelRequestIfObsolete`? I've implemented that in a9ec60d8e80462acb8862d8253492e2923bf6a7a.

davidalejandroaguilar: That sounds good, thank you!

PR #1175, opened 2024-02-08
Morphing + complete frame issue : Stop reloading turbo frames when complete attribute changes

Intrepidd:
I encountered the following problem : 

consider a turbo frame like so on index.html

```html
<turbo-frame id="foobar">
  <a href="/some/path">Click me</a>
</turbo-frame>
```

When the link is clicked, the HTML will look like this

```html
<turbo-frame id="foobar" src="/some/path" complete>
  Some content
</turbo-frame>
```

If the page is then refreshed with morphing, my expectation was that everything will return to the initial state.
However, in practice, the frame first appears to be reset, but then is reloaded and ends up in the second state.

This is because we have a callback that reloads the frame when there is any change on the `complete` attribute : 

https://github.com/hotwired/turbo/blob/e2e5782776e565e80358c5ecd34e05c3b3bcf28c/src/core/frames/frame_controller.js#L101-L105

The morphing causes this attribute to change thus the frame to reload.

I don't really understand why we want to reload a frame when the `complete` attribute changes, isn't it supposed to be only a read-only value ?

This PR removes the callback on the `complete` attribute, this is probably very naive, happy to explore another route if you point me in the right direction !

Repro and video in this comment : https://github.com/hotwired/turbo/pull/1175#issuecomment-1935561849

### DIFF
--- a/src/elements/frame_element.js
+++ b/src/elements/frame_element.js
@@ -25,7 +25,7 @@ export class FrameElement extends HTMLElement {
   loaded = Promise.resolve()
 
   static get observedAttributes() {
-    return ["disabled", "complete", "loading", "src"]
+    return ["disabled", "loading", "src"]
   }
 
   constructor() {
@@ -48,11 +48,9 @@ export class FrameElement extends HTMLElement {
   attributeChangedCallback(name) {
     if (name == "loading") {
       this.delegate.loadingStyleChanged()
-    } else if (name == "complete") {
-      this.delegate.completeChanged()
     } else if (name == "src") {
       this.delegate.sourceURLChanged()
-    } else {
+    } else if (name == "disabled") {
       this.delegate.disabledChanged()
     }
   }
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -90,18 +90,10 @@ export class FrameController {
 
   sourceURLReloaded() {
     const { src } = this.element
-    this.#ignoringChangesToAttribute("complete", () => {
-      this.element.removeAttribute("complete")
-    })
+    this.element.removeAttribute("complete")
     this.element.src = null
     this.element.src = src
     return this.element.loaded
-  }
-
-  completeChanged() {
-    if (this.#isIgnoringChangesTo("complete")) return
-
-    this.#loadSourceURL()
   }
 
   loadingStyleChanged() {
@@ -528,13 +520,11 @@ export class FrameController {
   }
 
   set complete(value) {
-    this.#ignoringChangesToAttribute("complete", () => {
-      if (value) {
-        this.element.setAttribute("complete", "")
-      } else {
-        this.element.removeAttribute("complete")
-      }
-    })
+    if (value) {
+      this.element.setAttribute("complete", "")
+    } else {
+      this.element.removeAttribute("complete")
+    }
   }
 
   get isActive() {
--- a/src/tests/fixtures/frame_refresh_after_navigation.html
+++ b/src/tests/fixtures/frame_refresh_after_navigation.html
@@ -1,0 +1,3 @@
+<turbo-frame id="refresh-after-navigation">
+  <h2 id="refresh-after-navigation-content">Frame has been navigated</h2>
+</turbo-frame>
--- a/src/tests/fixtures/page_refresh.html
+++ b/src/tests/fixtures/page_refresh.html
@@ -90,6 +90,11 @@
       <h2>Frame to be reloaded</h2>
     </turbo-frame>
 
+    <turbo-frame id="refresh-after-navigation">
+      <h2>Frame to be navigated then reset to its initial state after reload</h2>
+      <a id="refresh-after-navigation-link" href="/src/tests/fixtures/frame_refresh_after_navigation.html">Navigate</a>
+    </turbo-frame>
+
     <div id="preserve-me" data-turbo-permanent>
       Preserve me!
 
--- a/src/tests/functional/loading_tests.js
+++ b/src/tests/functional/loading_tests.js
@@ -119,17 +119,6 @@ test("navigating away from a page does not reload its frames", async ({ page }) 
   assert.equal(requestLogs.length, 1)
 })
 
-test("removing the [complete] attribute of an eager frame reloads the content", async ({ page }) => {
-  await nextEventOnTarget(page, "frame", "turbo:frame-load")
-  await page.evaluate(() => document.querySelector("#loading-eager turbo-frame")?.removeAttribute("complete"))
-  await nextEventOnTarget(page, "frame", "turbo:frame-load")
-
-  assert.ok(
-    await hasSelector(page, "#loading-eager turbo-frame[complete]"),
-    "sets the [complete] attribute after re-loading"
-  )
-})
-
 test("changing [src] attribute on a [complete] frame with loading=lazy defers navigation", async ({ page }) => {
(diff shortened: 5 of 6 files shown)


src/elements/frame_element.js
```diff
@@ -48,11 +48,9 @@ export class FrameElement extends HTMLElement {
   attributeChangedCallback(name) {
     if (name == "loading") {
       this.delegate.loadingStyleChanged()
-    } else if (name == "complete") {
-      this.delegate.completeChanged()
     } else if (name == "src") {
       this.delegate.sourceURLChanged()
-    } else {
+    } else if (name == "disabled") {
```
Intrepidd, line 53: figured it's a bit safer

jorgemanrubia:
@Intrepidd if you mark the `turbo-frame` with `refresh="morph"`, when a page refresh happens:

1) It won't get replaced.
2) It will reload, and its contents will be reloaded with morphing.

Would that cover your use case?

Intrepidd:
No, I want the frame to be replaced with the content from the page that is being refreshed.

Imagine a page where you have a list of record as rows, and by clicking on an edit button you transform the row to a form.

When submitting the form, you expect it to "close" and render as if you refresh the page ( this behavior happens naturally )

But with a morph refresh, this will happen then a few milliseconds later the form will "open" again.

I can try to write a repro if that helps understanding the issue ?

Intrepidd:
@jorgemanrubia here is a repository with a repro : https://github.com/Intrepidd/turbo_8_bug_demo

And here is a video of the issue 

https://github.com/hotwired/turbo/assets/803765/96b51843-7c48-4a5a-b0c4-000bd7e8bf84

As you can see, the frame first resets with the content of the index (expected) but then since the morphing removes the `complete` attribute, the callback is executed and the frame is loaded from its `src` attribute

jorgemanrubia:
@Intrepidd changing the `complete` callback behavior would be a breaking change, so I don't love that. 

As a workaround here, you could prevent the morphing of the attribute relying on the [event](https://turbo.hotwired.dev/reference/events#turbo%3Abefore-morph-attribute). Is not ideal, but that should do it.

```javascript

  prevent(event) {
    event.preventDefault()
  }
}
```

To use like:

```ruby
<%= turbo_frame_tag "frame", data: { controller: "event-default", action: "turbo:before-morph-attribute->event-default#prevent" }  do %>
# ...
```

As a longer term solution, I wonder if we could use some convention around the current `refresh=morph` attribute. For example, if the new `turbo-frame` arrives without a URL, instead of reloading the frame morph the contents 🤔 .

Intrepidd:
@jorgemanrubia Thanks for having a look !

I understand, however this callback is not documented as far as I know, and doesn't seem to make sense in the first place, I fail to understand when such behaviour is expected or wanted. Do you know why it is here in the first place ?

I would argue that now more than ever is the perfect time to make a "breaking" change as people are going to upgrade from 7 to 8 and expect some breaking changes.

> As a longer term solution, I wonder if we could use some convention around the current refresh=morph attribute. For example, if the new turbo-frame arrives without a URL, instead of reloading the frame morph the contents 🤔 .

There is definitely going to be some improvements around the `refresh` attribute of turbo frames I believe, as many behaviours can be desired ( I was toying with the idea of having `refresh="none"` )

But this would still leave the broken behaviour when no refresh attribute is specified, wouldn't it ?

jorgemanrubia:
@afcapel @seanpdoyle what do you think about removing that callback? I'm trying to think of real scenarios where one wants to mutate the `complete` but I can't think of any example.  That was added [here](https://github.com/hotwired/turbo/commit/4a61c68044099636bc6e7683267e053a7bd8bee6). Could you add some context there @seanpdoyle?

seanpdoyle:
> Do you know why it is here in the first place 

> I'm trying to think of real scenarios where one wants to mutate the complete and get this behavior, but I can't think of any example.

From the commit message (the `[loaded]` attribute was renamed to `[complete]` before merging that PR):

> The act of "reloading" involves the removal of the `[loaded]` attribute,
which can be done either by `FrameElement.reload()` or
`document.getElementById("frame-element").removeAttribute("loaded")`.

jorgemanrubia:
Thanks for the context @seanpdoyle, I had missed that.

It seems that this is an internal implementation detail. No reference to this behavior in [the docs](https://turbo.hotwired.dev/reference/frames). So it seems that @Intrepidd original approach is good: it's a callback we can remove. Of course, we need to make sure that `reload()` reloads the frame as expected.

Intrepidd:
Good ! Thanks to you both 🙏 

What are the next steps to make it a go ?

We should probably write a regression test but it's a bit tricky to test that the frame does not reload after a bit of time : 

```
- navigate inside the frame
- trigger a morph
- make sure the content is the one we expect
- wait a bit?
- make sure the content has not changed ? 
```

jorgemanrubia:
@Intrepidd why not try to imitate the example you used in the description for the test: after morphing, the turbo frame contents – which changed through the test – should be reset to the initial state?

Intrepidd:
@jorgemanrubia the problem is that it does get changed to the initial state but shortly after switched to the navigated state once the frame has been reloaded. So I'm afraid such test will pass even if the frame gets reloaded in the end.

I'll draft up something and we can iterate if needed

Intrepidd:
@jorgemanrubia I managed to write a spec that seems to fail on `main` and to pass on this branch ! However it appears to be flaky, although I can't seem to reproduce the flakyness with `--headed` on my machine, do you have any insights ? Thanks !

benwalks:
Hi @jorgemanrubia, @Intrepidd, any updates on this PR? This issue is affecting my projects as well, am happy to help.

Intrepidd:
Waiting as well regarding the specs 🙏

jorgemanrubia:
@Intrepidd sorry what's the problem with the specs? The referred flakiness? Is that happening in CI? We definitely need to sort that out, if you want to help with that @benwalks.

Intrepidd:
@jorgemanrubia my specs were indeed flaky but I think I managed to fix it. So I don't think there are any blockers anymore for this ?

Intrepidd:
Thanks a lot @jorgemanrubia for merging ! 

Any chance we can get a release soon ? As I believe this issue is a blocker for some to update to turbo 8. Thanks !

afcapel:
@Intrepidd I've just cut a new release https://github.com/hotwired/turbo/releases/tag/v8.0.4

jonsgreen:
I actually was relying on removing the 'complete' attribute as a convenient way to be able to reload a turbo_frame. I was doing it using [turbo_power](https://github.com/marcoroth/turbo_power) and sending turbo_streams like `turbo_stream.remove_attribute('#comp-packages', 'complete')`. Is there an official way to trigger a turbo_frame to reload from its src?

alexgerber:
> I actually was relying on removing the 'complete' attribute as a convenient way to be able to reload a turbo_frame. I was doing it using [turbo_power](https://github.com/marcoroth/turbo_power) and sending turbo_streams like `turbo_stream.remove_attribute('#comp-packages', 'complete')`. Is there an official way to trigger a turbo_frame to reload from its src?

You may use the reload method.

FrameElement.reload() is a function that reloads the frame element from its src.

jonsgreen:
Thanks @alexgerber. I have discovered that turbo_power has `turbo_stream.turbo_frame_reload` which is perfect!

ghiculescu:
Thanks for fixing this @Intrepidd 😍

Issue #1174, opened 2024-02-08
InstantClick is incompatible with `data-turbo-stream` and `data-turbo-confirm`


Slotos:
Reproduction is available [here](https://github.com/Slotos/turbo-instantclick-stream-issue)

When prefetching `data-turbo-stream="true"` links, Instantclick evaluates `turbo-stream` tags immediately, effectively turning hover event into a click one.

When hovering over `data-turbo-confirm` links, the request is being made immediately, bypassing the confirmation requirement.

### My use case

I'm using `data-turbo-stream="true"` to open selected forms in modal `<dialog>`. The page needs to open a dialog when link is engaged directly, but render a separate usable form page when link is pasted or is otherwise taken out of the page lifecycle (e.g. middle click). The problem boiled down to either discerning `turbo-frame` as html variant, which seems to be justifiably discouraged, or introducing modals as a streamed enhancement. I chose the latter.

### General argument

With `data-turbo-stream` and `data-turbo-confirm` being official parts of the framework, InstantClick should be aware and compatible.

"GET" requests are idempotent on server side, but responses to those requests have effect on browser state. Regular navigation takes care of replays by effectively cleaning the slate, but in case of Turbo Drive, we're skipping that phase, which allows for effects to accumulate.

Both `data-turbo-stream` and `data-turbo-confirm` imply that important or at least noticeable client-side effects are probable, and as such InstantClick should either be disabled by default for such links, or contain prefetched data without triggering its effects (stream processing, connecting stimulus controllers, etc.) until it is actually requested by the user.

------------------

This is similar to #1170 but contains a reproduction code, an additional issue, and an attempt to provide minimal reasonable expectations with justification.

brentkirby:
Ran into this problem today when upgrading turbo/turbo-rails.  For anyone else coming across this issue, the temporary solution is to add `data-turbo-prefetch="false"` to any link that triggers a turbo stream.  Disabling prefetch causes streams and confirmations to work as they previously did. 

This issue only seems to impact links which would trigger a GET request.  If  `data-turbo-method` is set to any other method, they work as expected.

Issue #1170, opened 2024-02-08
Links with UJS like behavior are prefetched for InstantClick


ur5us:
Technically, this is documented in https://turbo.hotwired.dev/handbook/drive#instantclick which also provides instructions how to opt out of InstantClick behavior for UJS enabled links. However, given how prevalent this is in Rails apps as well as https://turbo.hotwired.dev/handbook/drive#requiring-confirmation-for-a-visit the current default to opt in all links seems like a bad choice with potentially unintentional side effects. Thus, wouldn’t it be better to require the developer to enable (opt in) InstantClick behavior for such cases (s. `isUJS` function to check for various known `data-*` attributes)?

PR #1178, opened 2024-02-08
Add more prefetch conditions

afcapel:
This PR adds some more conditions where a link is not considered safe to be prefetched.

The risk/reward ratio favors a more conservative approach. If we don't prefetch a link the interaction can be a bit slower, but if we prefetch a link that is not safe, we may trigger a bug.

The conditions are quite specific, in any case, so they shouldn't trigger often.

The new conditions are:

- If the link is a UJS link (i.e. it has a `data-remote` attribute).
- If the link has a data-turbo-confirm attribute.
- If the link has a data-turbo-stream attribute.

The PR also cleans up the code a bit, extracting each of the conditions into helper methods.

Fixes https://github.com/hotwired/turbo/issues/1170
Fixes https://github.com/hotwired/turbo/issues/1174

### DIFF
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -176,8 +176,17 @@ const linkOptsOut = (link) => {
 
 const nonSafeLink = (link) => {
   const turboMethod = link.getAttribute("data-turbo-method")
-  if (turboMethod !== "get") return true
+  if (turboMethod && turboMethod.toLowerCase() !== "get") return true
+
+  if (isUJS(link)) return true
+  if (link.hasAttribute("data-turbo-confirm")) return true
+  if (link.hasAttribute("data-turbo-stream")) return true
+
   return false
+}
+
+const isUJS = (link) => {
+  return link.hasAttribute("data-remote") || link.hasAttribute("data-behavior") || link.hasAttribute("data-confirm") || link.hasAttribute("data-method")
 }
 
 const eventPrevented = (link) => {
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -1,6 +1,5 @@
 import {
   dispatch,
-  doesNotTargetIFrame,
   getLocationForLink,
   getMetaContent,
   findClosestRecursively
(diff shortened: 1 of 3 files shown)


src/observers/link_prefetch_observer.js
```diff
@@ -145,50 +139,52 @@ export class LinkPrefetchObserver {
...
-      return false
-    }
+    if (!href) return false
 
-    const event = dispatch("turbo:before-prefetch", {
-      target: link,
-      cancelable: true
-    })
+    if (unfetchableLink(link)) return false
+    if (linkToTheSamePage(link)) return false
```
kevinmcconnell, line 145: As we discussed IRL, I'm not sure we should always exclude the current page. Sometimes clicking navigation to the same page is a valid way of getting latest content on a page, and in that case prefetching seems valid.

Perhaps there are edge cases to think about when fetching them, though. In any case, probably fine to leave for another PR, just wanted to note is since we're touching this part.

src/observers/link_prefetch_observer.js
```diff
@@ -145,50 +139,52 @@ export class LinkPrefetchObserver {
...
+    if (linkOptsOut(link)) return false
+    if (nonSafeLink(link)) return false
+    if (eventPrevented(link)) return false
```
kevinmcconnell, line 148: This is much nicer with these checks extracted and given meaningful names, btw! 👏

kevinmcconnell (approved):
Looks great!

seanpdoyle:
Should we also open PR to remove https://github.com/hotwired/turbo-site/blob/694d442e6e57208814803cb5b2536e93bb398cd6/_source/handbook/02_drive.md#L369-L374 from the examples for canceling a prefetch?

src/observers/link_prefetch_observer.js
ur5us: You could probably argue that if the current page is prefetched it might as well morph it to the latest version. Though there are probably edge cases as you say.

ur5us (commented):
Thanks for addressing this so quickly.

afcapel:
@seanpdoyle good point, I've removed the example in https://github.com/hotwired/turbo-site/pull/162

Issue #1177, opened 2024-02-08
Morphing between different URLs


pch:
There's a scenario where I think morphing could be useful between different URLs.

Consider the following example (a 9mb gif screen recording below):

![CleanShot 2024-02-08 at 16 32 21](https://github.com/hotwired/turbo/assets/9818/9b8301ac-797d-4af4-9a37-ebb1e73fba5f)

Let's say we have an image browser like the one above. The page structure is basically the same, only certain parts change when you select a different image. Because each image has a unique URL, morphing doesn't apply and scrollbars reset.

In this case, it would be great to be able to set something like `<meta name="turbo-page-id" content="images#show">` and have Turbo use it to apply morphing between different URLs.

krschacht:
@pch I think turbo-frames solve your situation.

If you put the center column (or center + right column) within it's own `<turbo-frame id="content">` then the thumbnails on the left can direct target that frame: `<a href="/..." data-turbo-frame="content" data-turbo-action="advance">`

This way the left column retains it's scroll position because only the `<turbo-frame>` DOM node will be replaced.

PR #1179, opened 2024-02-08
Allow morphing between different URLs

pch:
Allows morphing between similar types of pages that have different URLs.

A proof-of-concept-fix for the problem described here: #1177

For example, if you set this meta tag:

`<meta name="turbo-morph-url-prefix" content="/images/">`

Then morphing will be triggered for:

- `/images/1`
- `/images/12`
- `/images/123`

**BEFORE:**

CleanShot 2024-02-08 at 16 32 21

**AFTER:**

CleanShot 2024-02-08 at 20 37 13

This solution has some downsides (e.g, won't work with nested resources). A regex-based alternative could address that.

Please let me know if you'd be willing to consider this change and I'll go ahead and add tests.

### DIFF
--- a/src/core/drive/page_snapshot.js
+++ b/src/core/drive/page_snapshot.js
@@ -82,6 +82,10 @@ export class PageSnapshot extends Snapshot {
     return this.getSetting("refresh-scroll") === "preserve"
   }
 
+  get morphURLPrefix() {
+    return this.getSetting("morph-url-prefix") ?? ""
+  }
+
   // Private
 
   getSetting(name) {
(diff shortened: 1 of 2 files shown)


pch:
A more elegant (and flexible) alternative to the above would be a per-link setting:

`<a href="/image/123" data-turbo-morph="true">Visit with morph</a>`

But I haven't explored this path yet...

websymphony:
👋 @pch Not sure if you have seen but there is already a [PR](https://github.com/hotwired/turbo/pull/1145), trying to solve same problem.

pch:
Closing this, I'll move the discussion to #1145

Issue #1176, opened 2024-02-08
InstantClick always fetches page when hovering same link within the turbo-prefetch-cache-time window.


jean-francois-labbe:
Hi, thanks for this amazing Turbo 8.
Regarding InstantClick.

The default `turbo-prefetch-cache-time` is 10000 (I'm not sure of the unit, seems to be milliseconds).

I've noticed that the current InstantClick behavior is to fetch the page on each link hover (even when hovering multiple times within the 10 seconds cache window).

![CleanShot 2024-02-08 at 16 41 37](https://github.com/hotwired/turbo/assets/1558372/2e9d196a-ddf0-462a-8f0b-2104bdde3892)

I thought on link hover InstantClick would fetch the page from the server store it to cache for 10 seconds.
If the user hovers the link again within 10 seconds it would not fetch a new page (because it's in cache, and why would we cache it if we'd fetch a new page on each link hover).

I may misunderstand the feature.

janklimo:
Observing the same behavior.

marcelolx:
> The feature is also missing documentation.

Missing something here https://turbo.hotwired.dev/handbook/drive#instantclick?

janklimo:
@marcelolx it's missing any info about the `turbo-prefetch-cache-time` meta tag

https://github.com/hotwired/turbo/blob/fccb3a4e5b2a47e63b50efe8333ec7a3265006c1/src/observers/link_prefetch_observer.js#L136

mansakondo:
From my understanding, this behavior was introduced in this [commit](https://github.com/hotwired/turbo/commit/52c8533c83bf02b428ddab6140ddfb663208cb06) where the cache is considered obsolete as soon as a `mouseleave` event is triggered. So, I guess the intent was to clear the cache if a user has hovered over a link but decided not to click it.
https://github.com/hotwired/turbo/blob/fccb3a4e5b2a47e63b50efe8333ec7a3265006c1/src/observers/link_prefetch_observer.js#L51

jean-francois-labbe:
Behavior has been explained in this comment https://github.com/hotwired/turbo/pull/1181#issuecomment-1936505362

PR #1181, opened 2024-02-09
Test for demonstrating: InstantClick always fetches page when hovering same link within the turbo-prefetch-cache-time window

jean-francois-labbe:
Demonstrating https://github.com/hotwired/turbo/issues/1176 with a test case.

I've noticed that the current InstantClick behavior is to fetch the page on each link hover (even when hovering multiple times within the 10 seconds cache window).
I thought on link hover InstantClick would fetch the page from the server store it to cache for 10 seconds.

Here is a test case showing the issue.

I'll try to come up with a code change that would make it pass.

### DIFF
--- a/src/tests/functional/link_prefetch_observer_tests.js
+++ b/src/tests/functional/link_prefetch_observer_tests.js
@@ -168,6 +168,27 @@ test("it caches the request for 1 millisecond when turbo-prefetch-cache-time is 
   await assertPrefetchedOnHover({ page, selector: "#anchor_for_prefetch" })
 })
 
+test("it caches the request for 10 second when turbo-prefetch-cache-time is set to 10000 and doesn't fetches from server", async ({ page }) => {
+  await goTo({ page, path: "/hover_to_prefetch_custom_cache_time.html" })
+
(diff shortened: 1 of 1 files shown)


jean-francois-labbe:
Thanks @afcapel for taking the time to explain the reason behind this behavior.

What if :
1. On link hover page is cached with ttl = turbo-prefetch-cache-time
2. On any non GET request clear the prefetch cache.

jean-francois-labbe:
I'm closing this as I don't see a path forward.

coorasse:
Thanks for explaining! This was really helpful to understand the reasoning behind it. It would be amazing to have some examples on how we can change this behavior: I agree to have a "safe default", but for some links I'd be totally fine in caching the page for 2 or 3 seconds. How could I achieve that?
Also, what about expiring if the mouse clicks somewhere else? 😄

inspire22:
This is preventing me from using this feature, it causes too many requests.  

What's the point of turbo-prefetch-cache-time if it's only the time the user has not left the link element?

Also it'd be great to be able to set turbo-prefetch-cache-time on the link element (or ideally a parent element)

Issue #1163, opened 2024-02-05
Add `morph` Turbo Stream Action


danecando:
Would be nice to support a `morph` turbo stream action for users who maybe aren't ready to opt in or transition to full page morphing. 

```javascript
Turbo.StreamActions.morph = function() {
  this.targetElements.forEach((target) => {
    Idiomorph.morph(target, this.templateContent)
  })
}
```

krschacht:
@danecando what would be an example use case for this? In my usage so far, you either use Turbo Stream -or- you morph. I can’t imagine using a stream to trigger a morph.

When I’m doing a turbo stream, it’s because I’m updating the specific piece of the page that I care about. Are you using Turbo Stream for a large section?

danecando:
I think the use cases are probably less than I had suspected initially. Probably replacing larger chunks of content or things containing images to prevent reloads or potential flashes

adrienpoly:
I had a use case when I wanted to animate a list of item using https://auto-animate.formkit.com/
When an action creates a new item I could not just append it to the list as the list was sorted. So I decided to render the full list back as a stream. A standard update would not play the animation. Using a morph stream only the new item got appended to the list at the correct location and with the animation. They are probably other ways to do this with lifecycle event but here that made the solution pretty straight forward.

krschacht:
@adrienpoly Are you suggesting that there is already a turbo stream action for triggering a morph?

krschacht:
Overall, I don’t think it’s right to conceive of morph as an action to combine with turbo stream.

Page Refresh w/ morph and Turbo Stream actions are really alternatives to each other. We are intended to pick one or the other.

If you want to be surgical about your updates to the page, you wire up a bunch of Turbo Stream actions which makes just the right update/append/insert as needed. Or if you don’t want to be surgical because that’s too much work, you can instead trigger a full Page Refresh and tell it to morph what it gets back. It’s a little heavier handed (e.g. two HTTP responses and a larger payload) but not too much so and it should just magically work.

adrienpoly:
@krschacht no there is no official Morph but adding a custom action is fairly easy now. https://marcoroth.dev/posts/guide-to-custom-turbo-stream-actions. They are even a few packages to get it out of the box

I was just giving a use case where Stream morph has been useful for me in the past. Maybe today I would build it with a full page morph 🤷‍♂️

krschacht:
[...]

PR #1185, opened 2024-02-15
Add Turbo stream morph action

omarluq:
This PR introduces a morph Turbo Stream action. The motivation for this addition stems from the recent discussions and suggestions in issue [#1163](https://github.com/hotwired/turbo/issues/1163). There is a clear interest in providing users with more flexible options for DOM manipulation, especially for those not ready or looking to fully transition to using Turbo's full page morphing capabilities.

turbo-rails PR https://github.com/hotwired/turbo-rails/pull/583
documentation PR https://github.com/hotwired/turbo-site/pull/169

### DIFF
--- a/src/core/streams/actions/morph.js
+++ b/src/core/streams/actions/morph.js
@@ -1,0 +1,68 @@
+import { Idiomorph } from "idiomorph/dist/idiomorph.esm"
+import { dispatch } from "../../../util"
+
+export default function morph(streamElement) {
+  const morphStyle = streamElement.hasAttribute("children-only") ? "innerHTML" : "outerHTML"
+  streamElement.targetElements.forEach((element) => {
+    try {
+      Idiomorph.morph(element, streamElement.templateContent, {
+        morphStyle: morphStyle,
+        ignoreActiveValue: true,
+        callbacks: {
+          beforeNodeAdded,
+          beforeNodeMorphed,
+          beforeAttributeUpdated,
+          beforeNodeRemoved,
+          afterNodeMorphed,
+        },
+      })
+    } catch (e) {
+      console.error(e)
+    }
+  })
+}
+
+function beforeNodeAdded(node) {
+  return !(node.id && node.hasAttribute("data-turbo-permanent") && document.getElementById(node.id))
+}
+
+function beforeNodeRemoved(node) {
+  return beforeNodeAdded(node)
+}
+
+function beforeNodeMorphed(target, newElement) {
+  if (target instanceof HTMLElement && !target.hasAttribute("data-turbo-permanent")) {
+    const event = dispatch("turbo:before-morph-element", {
+      cancelable: true,
+      detail: {
+        target,
+        newElement,
+      },
+    })
+    return !event.defaultPrevented
+  }
+  return false
+}
+
+function beforeAttributeUpdated(attributeName, target, mutationType) {
+  const event = dispatch("turbo:before-morph-attribute", {
+    cancelable: true,
+    target,
+    detail: {
+      attributeName,
+      mutationType,
+    },
+  })
+  return !event.defaultPrevented
+}
+
+function afterNodeMorphed(target, newElement) {
+  if (newElement instanceof HTMLElement) {
+    dispatch("turbo:morph-element", {
+      target,
+      detail: {
+        newElement,
+      },
+    })
+  }
+}
--- a/src/core/streams/actions/morph.js
+++ b/src/core/streams/actions/morph.js
@@ -4,21 +4,16 @@ import { dispatch } from "../../../util"
 export default function morph(streamElement) {
   const morphStyle = streamElement.hasAttribute("children-only") ? "innerHTML" : "outerHTML"
   streamElement.targetElements.forEach((element) => {
-    try {
-      Idiomorph.morph(element, streamElement.templateContent, {
-        morphStyle: morphStyle,
-        ignoreActiveValue: true,
-        callbacks: {
-          beforeNodeAdded,
-          beforeNodeMorphed,
-          beforeAttributeUpdated,
-          beforeNodeRemoved,
-          afterNodeMorphed,
-        },
-      })
-    } catch (e) {
-      console.error(e)
-    }
+    Idiomorph.morph(element, streamElement.templateContent, {
+      morphStyle: morphStyle,
+      callbacks: {
+        beforeNodeAdded,
+        beforeNodeMorphed,
+        beforeAttributeUpdated,
+        beforeNodeRemoved,
+        afterNodeMorphed
+      }
+    })
   })
 }
 
@@ -34,10 +29,10 @@ function beforeNodeMorphed(target, newElement) {
   if (target instanceof HTMLElement && !target.hasAttribute("data-turbo-permanent")) {
     const event = dispatch("turbo:before-morph-element", {
       cancelable: true,
+      target,
       detail: {
-        target,
-        newElement,
-      },
+        newElement
+      }
     })
     return !event.defaultPrevented
   }
@@ -50,8 +45,8 @@ function beforeAttributeUpdated(attributeName, target, mutationType) {
     target,
     detail: {
       attributeName,
-      mutationType,
-    },
+      mutationType
+    }
   })
   return !event.defaultPrevented
 }
@@ -61,8 +56,8 @@ function afterNodeMorphed(target, newElement) {
     dispatch("turbo:morph-element", {
       target,
       detail: {
-        newElement,
-      },
+        newElement
+      }
     })
   }
 }
--- a/src/core/streams/stream_actions.js
+++ b/src/core/streams/stream_actions.js
@@ -1,7 +1,6 @@
 import { session } from "../"
-import { morph } from "idiomorph"
+import { Idiomorph } from "idiomorph/dist/idiomorph.esm.js"
 import { dispatch } from "../../util"
-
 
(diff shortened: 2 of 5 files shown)


src/core/drive/morph_renderer.js
```diff
@@ -57,9 +57,9 @@ export class MorphRenderer extends PageRenderer {
         })
 
         return !event.defaultPrevented
-      } else {
-        return false
       }
+      
+      return false
```
omarluq, line 62: The `else` was redundant here 🤧

jorgemanrubia: Hey @omarluq, you are right it's redundant but I'd rather keep the PR focused on its purpose without performing cosmetic changes around 🙏.

jorgemanrubia (changes requested):
Hey @omarluq, thanks for the PR:

I think a morph stream action makes sense but I am not sure about those `data-` attributes. It's quite a new API the one we would be introducing. I'd aim to start with something simple, like what [`turbo-morph`](https://github.com/marcoroth/turbo-morph) does, where it only supports a "children only" option via a regular attribute in the stream action, which matches morphdom. I'd start without any option here. 

This would also need to be consistent with the [events](https://turbo.hotwired.dev/reference/events#page-refreshes) we dispatch during morphing, both in the events it emits and in how they let you cancel the operation. We should probably extract the morphing operation out to a common place we can reuse.

Finally, this would need its counterpart helper in [`turbo-rails`](https://github.com/hotwired/turbo-rails) and documentation in [`turbo-site`](https://github.com/hotwired/turbo-site).

omarluq:
Hey @jorgemanrubia, thank you for the speedy review sir! I've dropped the cosmetic change 🫡. 
Extracting the morph operation into a common place sound like a great idea; how about helper function in `util.js`? open to other suggestions as well. 
Also, I opened a related PR in turbo-rails, but there are a few specs that aren't passing. I plan to tackle those tomorrow along with updates you suggested to simplify the API! Ty much 🙏🏻

jorgemanrubia:
> Extracting the morph operation into a common place sound like a great idea; how about a helper function in util.js? open to other suggestions as well.

Checking the code, this might not be the best path, because the current renderer has some parts that are tied to a Turbo page rendering operation. I'd see how the code looks without doing this first.

src/core/streams/stream_actions.js
```diff
@@ -36,5 +38,87 @@ export const StreamActions = {
...
+          newElement: this.templateContent,
+        })
+      } catch (error) {
+        console.error(error)
+      }
+    })
+  },
+}
+
+const dispatchEvent = (eventName, detail, cancelable = false) => {
```
marcelokanzaki, line 69: `dispatchEvent` is a js API https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent

Also, you might want to follow their conventions and use `dispatch` directly ☺️

omarluq: Hey @marcelokanzaki, I am aware of that, but this is scoped locally, so it won't collide with the global `dispatchEvent`. The file is small and clean, and I'm trying to keep it as tight as possible. I didn't want to repeat the same object all over the place; it just doesn't look as good 😄 however if there is a strong objection against this pattern Im open to changing it.

omarluq: @marcelokanzaki I took some time to consider your feedback, ended up refactoring to your suggestion 👍🏻

omarluq:
Hey @jorgemanrubia, I believe I've addressed all of your feedback. Additionally, I added tests for the data-turbo-morph-style="innerHTML" option. I would appreciate a review. 🫡

src/tests/unit/stream_element_tests.js
```diff
@@ -196,3 +196,33 @@ test("test action=refresh discarded when matching request id", async () => {
...
+  target.setAttribute("data-turbo-morph-style", "innerHTML")
+
+  subject.append(element)
+
+  await nextAnimationFrame()
+
+  assert.ok(subject.find("div#hello"))
+  assert.ok(subject.find("div#hello > h1#hello-child-element"))
+  assert.equal(subject.find("div#hello > h1#hello-child-element").textContent, "Hello Turbo Morphed")
+})
```
jorgemanrubia, line 227: We'd also need tests covering how Stimulus listeners can handle events. See tests in `page_refresh_tests.js`.

omarluq: Hey @jorgemanrubia, could you expand a little on what exactly we are looking to test, targeting a controller element with a morph stream action? or just a controller listening to the morph events?

omarluq: Hey @jorgemanrubia, just checking in to see if you've had a chance to look at my previous question. It's the final piece I need to wrap up this PR. Thanks a lot!

jorgemanrubia: Sorry about the delay here @omarluq. What we need to test are two things:

- That the stream action triggers the events we expect. You can see an example of how we do that in the general case here:

https://github.com/hotwired/turbo/blob/e8d25b29460d745955dc9c0f11f046eac54996dc/src/tests/functional/page_refresh_tests.js#L38-L45

- That the `before-` events that the stream action emits give the listener a chance to cancel the action. Example here:

https://github.com/hotwired/turbo/blob/e8d25b29460d745955dc9c0f11f046eac54996dc/src/tests/functional/page_refresh_tests.js#L47-L60

The PR is looking great. Thanks for your work here.

omarluq: @jorgemanrubia No worries, thank you for your guidance, I pushed new browser test to cover both. Please lmk if you have anymore feedback.

src/core/streams/actions/morph.js
```diff
@@ -0,0 +1,67 @@
...
+        morphStyle: morphStyle,
+        callbacks: {
+          beforeNodeAdded,
+          beforeNodeMorphed,
+          beforeAttributeUpdated,
+          beforeNodeRemoved,
+          afterNodeMorphed
+        }
+      })
+    } catch (e) {
```
jorgemanrubia, line 18: Omar could we remove the wrapping `try/catch`, please? I don't think this should raise any exception so we would be silencing bugs we don't know about.

jorgemanrubia (commented):
Thanks @omarluq. Another little comment about the removing the wrapping try/catch and we'll get this merged.

cc @afcapel for another pair of eyes here.

src/core/streams/actions/morph.js
omarluq: @jorgemanrubia removed

src/tests/functional/morph_stream_action_tests.js
```diff
@@ -0,0 +1,49 @@
...
+
+  await nextEventOnTarget(page, "message_1", "turbo:before-morph-element")
+  await nextEventOnTarget(page, "message_1", "turbo:morph-element")
+  await expect(page.locator("#message_1")).toHaveText("Morphed")
+})
+
+test("preventing a turbo:before-morph-element prevents the morph", async ({ page }) => {
+  await page.goto("/src/tests/fixtures/morph_stream_action.html")
+
+
```
afcapel, line 27: Please remove the extra CR here.

afcapel (commented):
This looks good to me. Just a minor formatting issue.

src/tests/functional/morph_stream_action_tests.js
omarluq: @afcapel good eye for details 🔥 removed 🫡

src/core/streams/actions/morph.js
```diff
@@ -0,0 +1,63 @@
...
+
+function afterNodeMorphed(target, newElement) {
+  if (newElement instanceof HTMLElement) {
+    dispatch("turbo:morph-element", {
+      target,
+      detail: {
+        newElement
+      }
+    })
+  }
```
seanpdoyle, line 62: @jorgemanrubia duplicating the `MorphRenderer` logic feels like a big risk. Would generalizing the details of how we invoke Idiomorph (through something like https://github.com/hotwired/turbo/pull/1192) help here?

omarluq: I tend to agree with @seanpdoyle I think [`morph_elements.js`](https://github.com/hotwired/turbo/blob/07c35779f63b4261f37542eaba6ff62fe741c927/src/core/morph_elements.js) can be a drop in replacement for this class

jorgemanrubia: @seanpdoyle the problem is that the current implementation of the page renderer has logic interweaved to deal with morphing frames. I think we need to untangle that in https://github.com/hotwired/turbo/pull/1192, but I still don't see a clear separation there, and I haven't had proper time for looking into that yet. Agree that, ultimately, we should invoke the same morphing logic from everywhere, but we need to do some deep refactor to accommodate that.

seanpdoyle: > to deal with morphing frames

How does the exclusion of the Frame handling logic from the MorphRenderer impact the Stream's behavior?

Are there downsides to handling Frames consistently across Drive navigation morphs and Stream morphs?

seanpdoyle: @jorgemanrubia In response to https://github.com/hotwired/turbo/pull/1185#discussion_r1527139119, I've opened https://github.com/hotwired/turbo/pull/1234 to share the bulk of the morphing logic across Pages, Frames, and Streams.

Issue #1150, opened 2024-01-31
Detecting if `turbo:load` event was caused by a refresh


pfeiffer:
We're testing out the new Turbo 8 refresh feature, which is a great new addition!

One issue we're facing is that we have quite a few Stimulus controllers which react to eg. `turbo:load` events and we'd like to be able to differentiate between a refresh (eg. which triggered via a Turbo stream) and a normal visit triggered by a user action.

An example could be an `analytics` controller which tracks page views like `<body .. data-action="turbo:load@window->analytics#trackPageView">`. When a page is refreshed and morphed, we do not want to track a page view, as the action was not triggered by a user interaction.

Inspecting the event details received in `turbo:load` doesn't really provide is any granularity about the `visit` that triggered that action.

Any ideas or suggestions?

Issue #175, opened 2021-02-12
When I click a link during cache preview (and loading) it loads the page but then reloads it from scratch!


frederikhors:
I'm able to reproduce an issue with Turbo NOT present with Turbolinks:

https://codesandbox.io/s/hotwire-turbo-reload-if-click-during-cache-preview-iu6cg

(Please do not try this in codesandbox because it doesn't work. Export as zip and download locally.)

If I'm seeing the preview page from cache (the progress bar is loading) and I click another link Turbo loads that page but then reloads page from scratch:

With Turbolinks this does NOT happen.

**It's my fault?**

I'm using this for Turbo:

```html
<script type="module">
    addEventListener("load", async () => {
    const hotwiredTurbo = await import("https://cdn.skypack.dev/@hotwired/turbo");
    });
</script>
```

and this for Turbolinks in that pages:

```html
<script defer src="https://www.unpkg.com/turbolinks@5.2.0/dist/turbolinks.js"></script>
```

![issue](https://user-images.githubusercontent.com/41120635/107805448-a9a1a680-6d65-11eb-9e46-18546a806500.gif)

PR #1186, opened 2024-02-15
Introduce `refresh` visit action and method to refresh the current page

pfeiffer:
Related issues:
* https://github.com/hotwired/turbo-ios/issues/136#issuecomment-1944578518
* https://github.com/hotwired/turbo-ios/issues/175
* https://github.com/hotwired/turbo/issues/1150

This PR adds a `refresh` visit action and changes `Session#refresh` to use the current URL if not provided explicitly.

## Add `refresh` visit action
The motivation behind it is to support more granular behavior when doing refreshes. Today, a 'refresh' is essentially a `replace` action to the current URL. There's no easy way to know if this action was triggered by a user interaction or by a turbo stream refresh. By adding a granular `refresh` action, developers can handle refreshes (eg. broadcasted as streams) differently from replace actions (form submissions, drive navigations, ..).

An example of this would be in Turbo iOS, where there's no way to differentiate between a user-triggered `replace` action (ie. pull-to-refresh) or a programatically triggered refresh (ie. broadcasts) (see eg. [Turbo iOS #136](https://github.com/hotwired/turbo-ios/issues/136#issuecomment-1944578518) and [Turbo iOS #175](https://github.com/hotwired/turbo-ios/issues/175)).

With this change, developers (and library maintainers of eg. Turbo iOS) would able to differentiate between `replace` actions and `refresh` actions and handle eg. showing loading indicator behavior differently. As an example we'd like show a more subtle loading indicator on refresh actions than on replace actions.

## Make `Session#refresh` refresh current page unless specified
Secondly, the `Turbo.session.refresh(..)`  method is updated to use the current URL of the page. This allows developers to programatically refresh the current page simply by calling `Turbo.session.refresh()` without having to explicitly provide the current page's URL. 

A common use-case could be refreshing (and morphing) a page when it becomes re-focused as a tab, to always ensure that the content is up to date:

```js
addEventListener("visibilitychange", () => {
  if (document.visibilityState === "visible") {
    Turbo.session.refresh()
  }
})
```

I think this would also be usable for Turbo iOS and Turbo Android apps with multiple sessions, where other sessions could be _refreshed_ instead of _reloaded_ when a mutating request has been made in a single session. See: https://github.com/hotwired/turbo-ios/pull/150


pfeiffer:
Just noticed that there's a slightly similar PR approach-wise in https://github.com/hotwired/turbo/pull/1145 although that PR deals with a different issue (navigating to another URL than the current page with morphing). I think the two are unrelated in that this PR explicitly not allows morphing between two different path names.

src/core/index.js
```diff
@@ -33,7 +33,7 @@ export function registerAdapter(adapter) {
  * @param location Location to visit (a URL or path)
  * @param options Options to apply
  * @param options.action Type of history navigation to apply ("restore",
- * "replace" or "advance")
+ * "replace", "refresh" or "advance")
```
pfeiffer, line 36: I think the `refresh` action should be reserved for internal use, just as the `restore` action is.

pfeiffer:
Closing as morphing has landed in Turbo Native without this 🎉

tylerklose:
@pfeiffer any idea what the best upgrade path is?

For example, if I publish a new version of iOS with the updated Turbo Native, there will be some users with the change and others without the change.

I probably wouldn't want users on the older version of the app to see the previous behavior

pfeiffer:
The Turbo Native adapters are compatible with both Turbolinks and Turbo 8.

Your users using an older version of the app would still be able to do that; they would not get the benefit of refreshes though. You could conditionally subscribe to broadcasts on your website, so that users of older app versions without morph support, would not receive any broadcasted refreshes.

PR #1188, opened 2024-02-15
Cache snapshot before doing page refresh

pfeiffer:
This PR aims to fix issues related to white flickering and loading indicators on Turbo Native:
* https://github.com/hotwired/turbo-ios/issues/175
* https://github.com/hotwired/turbo-ios/issues/136

Previously, any `refresh` visit would exempt itself from the snapshot cache before issuing a `visit(..., action: "replace")`. In a browser, this keeps the existing body in-place while refreshing the page and then performing morphing. The visit would have `hasSnapshot: false`.

In Turbo Native, when a visit is performed without a snapshot, a spinner will be shown and in some cases the WebView will be blanked as well while the page is fetched. This is especially painful when refresh stream actions are broadcasted, as the WebView will at random become unresponsible and/or flickr.

This PR changes so that a snapshot of the page is cached _before_ issuing a refreshing visit. The cached snapshot will be of the recent state of the page and the native adapters will not blank the page while fetching the refreshed page.

I think it makes sense that page refreshes are storing the latest version of the page as a snapshot before refreshing; it is the most recent version of the content.

### DIFF
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -114,10 +114,15 @@ export class Visit {
     return this.isSamePage
   }
 
-  start() {
+  async start() {
     if (this.state == VisitState.initialized) {
       this.recordTimingMetric(TimingMetric.visitStart)
       this.state = VisitState.started
+
+      if (this.isPageRefresh) {
+        await this.cacheSnapshot()
+      }
+
       this.adapter.visitStarted(this)
       this.delegate.visitStarted(this)
     }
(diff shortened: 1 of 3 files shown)


pfeiffer:
This is how a refresh visit behaves in the native apps today:

https://github.com/hotwired/turbo/assets/195925/51f9f33f-4372-4118-9083-fb52813ca2f3

With the changes in this PR:

https://github.com/hotwired/turbo/assets/195925/d51c935b-8ce5-45ce-9516-4d5971d28487

Ultimately, the native apps will still show a short flash of white content, as they are swapping out the WebView for a snapshot. To fix this I believe that the changes proposed in #1186 is needed so that the native adapters would be able to distinguish refresh actions.

jayohms:
@pfeiffer can this be closed now that https://github.com/hotwired/turbo-ios/pull/178 has been merged?

pfeiffer:
Closing as not relevant.

Issue #1161, opened 2024-02-04
refresh="morph" in a turbo-frame morphs on a full Page Reload but fails to morph with frame.reload()


krschacht:
@jorgemanrubia acknowledge the issue over on this PR comment: https://github.com/hotwired/turbo/pull/1068#issuecomment-1925877557

To elaborate on the issue for whomever may tackle this. Imagine this page setup:

```
<body>
   Contents outside of frame
   
   <turbo-frame id="post" refresh="morph" src="/post/1">
      Loading... (preload contents)
   </turbo-frame>
</body>
```

If something triggers a full Page Reload causing <body> to be morphed rather than replaced, the presence of `refresh="morph"` is smart and it also re-requests the src and *morphs the response from src* so that the contents within side turbo-frame only updates what is needed.

However, if you instead trigger a refresh to the frame using javascript, e.g. `document.getElementById('post').reload()` the turbo-frame ignores the `refresh="morph"` and does a full replacement within the turbo-frame.

**The ideal behavior:**
When `refresh="morph"` is present on a turbo-frame then anything that triggers a refresh of that turbo-frame (e.g. `element.reload()`) should be a morphing refresh.

krschacht:
Hi @jorgemanrubia I went ahead and tee'd up a PR to fix this issue. Check #1192

seanpdoyle:
@krschacht was this issue resolved by https://github.com/hotwired/turbo/pull/1192 and https://github.com/hotwired/turbo/pull/1028? 

If it has been resolved, should it be closed?

PR #1192, opened 2024-02-19
Lift the frame morphing logic up to FrameController.reload

krschacht:
Fixes Issue #1161 

**Background:**

`turbo-frame` tags support an optional property of `refresh="morph"`. It's used like this:

```
   <turbo-frame id="post" refresh="morph" src="/post/1">
      Loading... (preload contents)
   </turbo-frame>
```

When the full page morphs, this turbo-frame src is reloaded, and the presence of `morph` ensures the contents is morphed rather than replaced.

**Problem:**

However, if you trigger an update of this frame by another means such as `document.getElementById('post').reload()` then the contents of the frame is replaced without morphing. @jorgemanrubia acknowledged in this [comment](https://github.com/hotwired/turbo/pull/1068#issuecomment-1925877557) that it would be nice for this behavior to be consistent.

**Solution within this PR:**

~~This PR creates a new `MorphFrameRenderer` which inherits from `FrameRenderer`. (Note: `MorphRenderer` is renamed to `MorphPageRenderer` for clarity.) And just like `PageView` decides when to use `MorphPageRenderer`, now `FrameController` makes the appropriate decision of when to use `MorphFrameRenderer`.~~

~~Because `MorphFrame*` and `MorphPage*` share so much logic this has been extracted into a utility class `core/morph_elements.js`~~ 

Most of this was extracted into PR #1234. All that remains in this PR is (1) a new conditional to determine whether `FrameRenderer` or `MorphingFrameRenderer` should be called on this frame, and (2) good test coverage for these new cases.

**I made a corresponding update to the turbo-frame docs here: https://github.com/hotwired/turbo-site/pull/170**


src/core/drive/morph_renderer.js
```diff
@@ -3,51 +3,32 @@ import { dispatch } from "../../util"
...
-  async render() {
-    if (this.willRender) await this.#morphBody()
-  }
-
-  get renderMethod() {
-    return "morph"
-  }
 
-  // Private
+  // Static methods — these helper methods are all static because morphing is done by turbo frames, which is outside of a PageView
```
krschacht, line 7: There is no change to the logic of any of these methods. They're simply converted to static methods so that they can be used by FrameController.

src/core/drive/morph_renderer.js
```diff
@@ -80,35 +61,42 @@ export class MorphRenderer extends PageRenderer {
...
     })
-    this.#morphElements(currentElement, newElement.children, "innerHTML")
   }
 
-  #isFrameReloadedWithMorph(element) {
-    return element.src && element.refresh === "morph"
+  #reloadRemoteFrames() {
+    this.#remoteFrames().forEach((frame) => {
+      frame.reload()
+    })
```
krschacht, line 99: This class no longer needs to check if the frame has morph set or not. It can blindly trigger a `reload()` on each frame and trust that the frame will be smart — it knows if it needs to morph itself.

src/core/frames/frame_controller.js
```diff
@@ -114,6 +115,10 @@ export class FrameController {
   }
 
   async #loadSourceURL() {
+    if (this.#isFrameRefreshMorph()) {
+      this.#prepareFrameToMorph()
+    }
```
krschacht, line 120: Right before a frame is about to be updated, we check to see if it's supposed to morph.

src/tests/fixtures/frame_refresh_morph.html
```diff
@@ -1,3 +1,4 @@
 <turbo-frame id="refresh-morph">
   <h2>Loaded morphed frame</h2>
+  <h3>Does not change</h3>
```
krschacht, line 3: Adding a piece of HTML which does not change makes it easier to observe whether morphing is occurring or not within chrome inspector

krschacht:
@seanpdoyle You reviewed my documentation PR yesterday, related to this change. I wanted to ping on this one in case you could also look at the bug fix.

jorgemanrubia (commented):
@krschacht thanks for working on this.

Instead of reusing `MorphRenderer`, which is a `PageRenderer` for doing this, I'd consider adding a specialized `FrameRenderer` that uses morphing. It could kick in when `[refresh="morph"]`. I think this would also remove some clunkiness in the current `MorphRenderer`, where it relies on an event to use morphing to refresh frames.

krschacht:
@jorgemanrubia Sure thing! I'm still trying to fully understand the architecture, but I think I follow what you're saying. That event logic was clunky. I'll get this cleaned up so we can get it merged in.

Just to preview my plan in case you want to weigh in. I will plan to:

* Rename `MorphRenderer` to `MorphPageRenderer`
* Create a new `MorphFrameRenderer` which inherits from `FrameRenderer`
* Create a utility class which contains the core morphing logic so that these two Renderers don't duplicate so much code. Maybe `src/core/morph_elements.js` so that the Renders will call `MorphElements.morph(currentElement, newElement, morphStyle = "outerHTML")`

krschacht:
@jorgemanrubia This PR is ready to go. I confirmed that all tests are passing and added good test coverage for the new cases.

I also made a corresponding update to the turbo-frame docs here: https://github.com/hotwired/turbo-site/pull/170

src/core/morph_elements.js
```diff
@@ -0,0 +1,66 @@
+import { Idiomorph } from "idiomorph/dist/idiomorph.esm.js"
+import { dispatch } from "../util"
+
+export class MorphElements {
```
seanpdoyle, line 4: Since these are all `static` methods, is there any benefit to defining exporting a `class`? Would a collection of exported functions serve the same purpose?

src/tests/server.mjs
```diff
@@ -23,6 +23,20 @@ router.use((request, response, next) => {
...
+    if (err) {
+      console.error(err)
+      return res.status(500).send('Error reading file')
+    }
+    const randomNumber = Math.floor(Math.random() * 10000000)
+    const processedContent = data.replace('##randNum##', randomNumber.toString())
+    res.send(processedContent)
+  })
+})
+
```
seanpdoyle, line 39: The test suite has generally avoided purpose-built server behavior in favor of general endpoints that either render fixtures directly or redirect.

What is being exercised by replacing `##randNum##`? Are there other ways to cover the intended behavior without introducing a new server response?

src/core/frames/morph_frame_renderer.js
```diff
@@ -0,0 +1,16 @@
+import { FrameRenderer } from "./frame_renderer"
+import { MorphElements } from "../morph_elements"
+import { dispatch } from "../../util"
+
+export class MorphFrameRenderer extends FrameRenderer {
+  static renderElement(currentElement, newParentElement) {
+    const newElement = newParentElement.children
+
+    dispatch("turbo:before-frame-morph", {
```
seanpdoyle, line 9: What are we hoping to achieve by introducing a new event that exists separately from the other Morph events? What differs between this event and a `turbo:before-morph-element` event that references the `FrameElement` instance with its `target` and `currentElement` properties?

src/core/morph_elements.js
krschacht: @seanpdoyle Unlike some of the other utility files, I did it as a class because they're tightly coupled methods so on any use you'd need to import all methods. None of them stand alone. In this way, it's similar to the Cache utility class. But there is no actual state that is being instantiated so I did them static methods.

But I'm happy to change them to independent functions which are each exported. Would you like me to make that change?

src/tests/server.mjs
krschacht: @seanpdoyle Yes, this is crufty. I didn't love it. I can safely remove this and all tests will still pass since the tests are just asserting that various events are firing.

I personally find it helpful to be able to actually observe with my own eyes with a morph is happening (i.e. loading localhost:9000 and opening inspector and watching which elements flash). Without a single element changing on reload I could not visually confirm that a morph was occurring and the pattern of our tests so far simply asserts that an event was fired. That's all this was for.

I'll go ahead and remove it.

src/core/morph_elements.js
seanpdoyle: > I did it as a class because they're tightly coupled methods so on any use you'd need to import all methods

Reading through the diff in its current state, the `MorphElements` class is only ever interacted with by importing the class and invoking `MorphElements.morph` directly:

```js
import { MorphElements } from "../morph_elements"

// ...
MorphElements.morph(...)
```

Since its consistently invoked with a single method, the surface area of the module's interface could be reduced to **only** include `morph`:

```js
import { morph } from "../morph_elements"

morph(...)
```

Then, the module could be simplified to be a collection of module-private functions supporting a single exported `morph` function.

src/core/frames/morph_frame_renderer.js
krschacht: @seanpdoyle That's existing behavior in main. I preserved it so as not to break anyone's usage:
https://github.com/hotwired/turbo/pull/1192/files#diff-04bb702ad80fc80a9ba0ddee175f0ad2527604ef65f45c78da9c379e5f63035eL102

seanpdoyle: Thank you for clarifying @krschacht, I'd missed that.

@jorgemanrubia currently the `turbo:before-frame-morph` event is not yet [publicly documented](https://turbo.hotwired.dev/reference/events#page-refreshes). 

Are we willing to omit it from the public collection of events, and instead encourage `turbo:before-morph-element` events? I think it pre-dated the introduction of those events in https://github.com/hotwired/turbo/pull/1097.

src/elements/frame_element.js
```diff
@@ -82,6 +82,10 @@ export class FrameElement extends HTMLElement {
     return this.getAttribute("refresh")
   }
 
+  get shouldReloadWithMorph() {
+    this.src && this.refresh === "morph"
```
seanpdoyle, line 86: Should this have a `return` statement?

seanpdoyle: Also, is it important to exclude initial `<turbo-frame>` elements that don't have a `[src]` attribute? Wouldn't morphing an initial HTTP request have a benefit?

krschacht: You're too fast for me! :) I actually mistakenly pushed before I was done. I'm doing a little more cleanup and making sure all tests pass before I went ahead and tagged anyone.

However, I don't understand your second comment. My mental model was that morphing for turbo frames only applied when there was a src. And I didn't introduce that extra condition, that's currently present within main.

When there is no src, then the full contents of the HTML must be supplied within the tags so that means it's part of the overall page. And so if the overall page morphs, then there is no special handling for turbo-frames — the contents within them just morphs along with all the rest of the HTML of that page. But I could be misunderstanding that.

That was my read of this original logic:
https://github.com/hotwired/turbo/blob/main/src/core/drive/morph_renderer.js#L95

The event binding hack was only being added when there was a src so I assume without that it's just regular HTML morphing of the page.

src/core/morph_elements.js
```diff
@@ -0,0 +1,64 @@
...
+      beforeNodeMorphed: shouldMorphElement,
+      beforeAttributeUpdated: shouldUpdateAttribute,
+      beforeNodeRemoved: shouldRemoveElement,
+      afterNodeMorphed: didMorphElement
+    }
+  })
+}
+
+function isFrameReloadedWithMorph(element) {
+  return element.src && element.refresh === "morph"
```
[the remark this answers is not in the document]
krschacht: Good suggestion, thanks. I just made that change.

src/core/morph_elements.js
krschacht: @seanpdoyle As I was doing some additional cleanup, I discovered that this change broke a random test in the test suite. Specifically, moving from a class to a collection of methods.

I spent a little while trying to track down the source of the bug and the only way I could solve it was to bring the class back.

The issue has something to do with `this` context as shared between methods. The `morph()` method sets `this.isMorphingTurboFrame` which is later referenced by `shouldMorphElement()`, but this is intentionally an arrow function because it's provided as a callback to Idiomorph. As I was banging my head against the wall trying to figure out another way around this or threading the context through as an argument, it occurred to me: this is the problem that classes solve! :) I just mean: sharing context between discrete method calls.

For now I reverted back to the class with static methods. The full test suite passes again. If you really think it shouldn't be a class, the solution needs to involve eliminating this shared state and somehow threading it through, but I don't understand the innerworkings of Idiomorph to easily do that.

seanpdoyle: If the morphing is stateful (uses `this`), you're correct: separate functions won't be  suitable.

The original feedback was around the over-use of the `static` keyword. A class that's composed of all `static` keywords is functionally equivalent to a group of functions. Any "state" preserved through assigning to a `this` will be global. That's equivalent to a module-local variable.

Sharing a model-local variable across all morphing will lead to sharing that state across potentially concurrent morphs. For example, if `this.isMorphingTurboFrame` is true for a frame but false for a page-wide morph, that contention could cause unexpected behavior.

As a balance, what do you think of this set of changes:

```diff
diff --git a/src/core/drive/morph_page_renderer.js b/src/core/drive/morph_page_renderer.js
index 3e2dbf8..9418f44 100644
--- a/src/core/drive/morph_page_renderer.js
+++ b/src/core/drive/morph_page_renderer.js
@@ -1,6 +1,6 @@
 import { dispatch } from "../../util"
 import { PageRenderer } from "./page_renderer"
-import { MorphElements } from "../morph_elements"
+import { morphElements } from "../morph_elements"
 
 export class MorphPageRenderer extends PageRenderer {
   async render() {
@@ -14,7 +14,7 @@ export class MorphPageRenderer extends PageRenderer {
   // Private
 
   async #morphBody() {
-    MorphElements.morph(this.currentElement, this.newElement)
+    morphElements(this.currentElement, this.newElement)
     this.#reloadRemoteFrames()
 
     dispatch("turbo:morph", {
diff --git a/src/core/frames/morph_frame_renderer.js b/src/core/frames/morph_frame_renderer.js
index 83d8bf1..b8404e2 100644
--- a/src/core/frames/morph_frame_renderer.js
+++ b/src/core/frames/morph_frame_renderer.js
@@ -1,5 +1,5 @@
 import { FrameRenderer } from "./frame_renderer"
-import { MorphElements } from "../morph_elements"
+import { morphElements } from "../morph_elements"
 import { dispatch } from "../../util"
 
 export class MorphFrameRenderer extends FrameRenderer {
@@ -11,6 +11,6 @@ export class MorphFrameRenderer extends FrameRenderer {
       detail: { currentElement, newElement }
     })
 
-    MorphElements.morph(currentElement, newElement, "innerHTML")
+    morphElements(currentElement, newElement, "innerHTML")
   }
 }
diff --git a/src/core/morph_elements.js b/src/core/morph_elements.js
index 0761486..7b74248 100644
--- a/src/core/morph_elements.js
+++ b/src/core/morph_elements.js
@@ -2,8 +2,14 @@ import { Idiomorph } from "idiomorph/dist/idiomorph.esm.js"
 import { dispatch } from "../util"
 import { FrameElement } from "../elements/frame_element"
 
-export class MorphElements {
-  static morph(currentElement, newElement, morphStyle = "outerHTML") {
+export function morphElements(currentElement, newElement, morphStyle = "outerHTML") {
+  const renderer = new Renderer()
+
+  renderer.morph(currentElement, newElement, morphStyle)
+}
+
+class Renderer {
+  morph(currentElement, newElement, morphStyle) {
     this.isMorphingTurboFrame = this.isFrameReloadedWithMorph(currentElement)
 
     Idiomorph.morph(currentElement, newElement, {
@@ -18,15 +24,15 @@ export class MorphElements {
     })
   }
 
-  static isFrameReloadedWithMorph(element) {
+  isFrameReloadedWithMorph(element) {
     return (element instanceof FrameElement) && element.shouldReloadWithMorph
   }
 
-  static shouldAddElement = (node) => {
+  shouldAddElement = (node) => {
     return !(node.id && node.hasAttribute("data-turbo-permanent") && document.getElementById(node.id))
   }
 
-  static shouldMorphElement = (oldNode, newNode) => {
+  shouldMorphElement = (oldNode, newNode) => {
     if (!(oldNode instanceof HTMLElement)) return
 
     if (oldNode.hasAttribute("data-turbo-permanent")) return false
@@ -44,17 +50,17 @@ export class MorphElements {
     return !event.defaultPrevented
   }
 
-  static shouldUpdateAttribute = (attributeName, target, mutationType) => {
+  shouldUpdateAttribute = (attributeName, target, mutationType) => {
     const event = dispatch("turbo:before-morph-attribute", { cancelable: true, target, detail: { attributeName, mutationType } })
 
     return !event.defaultPrevented
   }
 
-  static shouldRemoveElement = (node) => {
+  shouldRemoveElement = (node) => {
     return this.shouldMorphElement(node)
   }
 
-  static didMorphElement = (oldNode, newNode) => {
+  didMorphElement = (oldNode, newNode) => {
     if (newNode instanceof HTMLElement) {
       dispatch("turbo:morph-element", {
         target: oldNode,
```

That diff removes the `static` lines in favor of a bonafide Class with instances. It also exports a single `morphElements` function for callers to use without any knowledge that it's implemented with a class instance.

krschacht: @seanpdoyle Good call, I made this change. I confirmed that the tests all still pass.

My recommendation would be that we merge in this PR and circle back to the open "turbo:before-frame-morph" element later in a follow-up since this PR is not making any changes to events.

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/1234, which overlaps with the extractive portion of this PR. It doesn't incorporate the rest of the behavior change that accounts for `turbo-frame[refresh="morph"]`. 

If #1234 merges ahead of this, you'll likely need to rebase. I'm happy to assist with that effort if that ends up being the case.

robinboening:
I am excited this is being worked on and is brought thus far! It seems to need a rebase already, but besides that what's the status and can we expect to see it getting merged soon?

krschacht:
@seanpdoyle I can rebase this PR now that you've extracted `morphElements()`, I just didn't want to do that if you weren't supportive of this PR. What do you think? If I finish that cleanup, can we get this one merged in?

gjtorikian:
👋 Is it just the rebasing that needs to be done here, or is further support for the PR necessary?

jorgemanrubia:
Hey @krschacht yes please, it would be great if you could rebase and reuse the new extracted logic. We'll merge it 👍.

krschacht:
@jorgemanrubia I have updated this PR and it's description. It's ready to merge. Be sure to merge the change to the docs as well over here: https://github.com/hotwired/turbo-site/pull/170 . Then you can close out the Issue which started all this: https://github.com/hotwired/turbo/issues/1161

While I have your attention, I am using turbo-frames extensively in my app. I have discovered a bug where the back button in one situation where you target a turbo frame. I have opened a new issue regarding this, and I've created a simple repro case, but I'm stuck in attempting to create a PR for you all in `hotwired/turbo` which fixes it. If you can take a quick look at this issue: https://github.com/hotwired/turbo/issues/1300 I can potentially help fix this bug next. Teach a person to fish... :)

brunoprietog:
Hi @krschacht, we need to take advantage of these changes in `MorphingPageRenderer`.

https://github.com/hotwired/turbo/blob/664695762f152b744002f83f76d196d458f9030f/src/core/drive/morphing_page_renderer.js#L16

src/core/frames/frame_controller.js
```diff
@@ -299,7 +301,14 @@ export class FrameController {
 
     if (newFrameElement) {
       const snapshot = new Snapshot(newFrameElement)
-      const renderer = new FrameRenderer(this, this.view.snapshot, snapshot, FrameRenderer.renderElement, false, false)
+      let renderer
+
+      if (this.element.shouldReloadWithMorph) {
+        renderer = new MorphingFrameRenderer(this, this.view.snapshot, snapshot, MorphingFrameRenderer.renderElement, false, false)
```
brunoprietog, line 307: Actually, here you are changing the renderer for all cases. We only need to modify the renderer when calling the `reload` method, otherwise, any Turbo frame that has the `refresh="morph"` attribute would be rendered with morphing always, not only with page refreshes or when calling the `reload` method. Probably the simplest approach here is to reuse logic that already exists, moving it to `FrameController`.

```js
  sourceURLReloaded() {
    if (this.element.shouldReloadWithMorph) {
      this.element.addEventListener("turbo:before-frame-render", ({ detail }) => {
        detail.render = MorphingFrameRenderer.renderElement
      }, { once: true })
    }

    const { src } = this.element
    this.element.removeAttribute("complete")
    this.element.src = null
    this.element.src = src
    return this.element.loaded
  }
```

brunoprietog (changes requested):

src/core/frames/frame_controller.js
krschacht: Thanks for the feedback @brunoprietog I was hoping that would be a feature rather than a bug. :) You don't think we should support morphing for more cases? I know that for the overall page we're being really conservative right now and only morphing when a redirect happens back to the same URL. I was hoping that if we explicitly set reload="morphing" on a turbo-frame then we could morph in all cases even when the turbo-frame transitions in a way other than `.reload()`. In all my testing, the morphing logic is good enough to transition even complex page transitions. But maybe there are some limitations of the idiomorph which we're trying to stay within?

brunoprietog: Yes, I also think it's a good idea to support morphing in Turbo frames in more cases, but here it's more a matter of naming and expectation. `refresh="morph"`, means that when you refresh the frame you expect it to be morphed. It doesn't mean that when navigating that frame it will render with morphing necessarily. We can't assume that it works well because we've done tests. We've seen that with morphing all sorts of strange situations can occur, so I'd be a little cautious with this change.

src/tests/functional/page_refresh_tests.js
```diff
@@ -180,12 +180,18 @@ test("uses morphing to update remote frames marked with refresh='morph'", async
...
+      frame.textContent = "Frame to be preserved"
+    }
+  }) // The final assertion cannot be noNextEventOnTarget because the assertion passes even when it reloads the frame.
+
   await page.click("#form-submit")
   await nextEventNamed(page, "turbo:render", { renderMethod: "morph" })
   await nextBeat()
 
-  // Only the frame marked with refresh="morph" uses morphing
-  expect(await noNextEventOnTarget(page, "refresh-reload", "turbo:before-frame-morph")).toBeTruthy()
```
krschacht, line 188: This assertion was not working. When I commented out code and convinced myself the permanent frame WAS reloading, this test would still pass. I think it's a race condition where the assertion breezes through too quickly.

I tried multiple different strategies to make this test durable and the solution you see here is the only one that worked.

src/core/frames/frame_controller.js
krschacht: Ah, that makes sense.

I have incorporated your feedback in this PR. It now only morphs when an explicit .reload() is called on the page (or when the page is morphing). I think I did this all correctly. I've updated my test cases and ensured I didn't inadvertently break some other tests. In the process I discovered a mistaken test which I've fixed (and left a comment on).

src/tests/functional/page_refresh_tests.js
```diff
@@ -180,12 +180,18 @@ test("uses morphing to update remote frames marked with refresh='morph'", async
 test("don't refresh frames contained in [data-turbo-permanent] elements", async ({ page }) => {
   await page.goto("/src/tests/fixtures/page_refresh.html")
 
+  await page.evaluate(() => {
+    const frame = document.getElementById("remote-permanent-frame");
+    if (frame) {
+      frame.textContent = "Frame to be preserved"
+    }
+  }) // The final assertion cannot be noNextEventOnTarget because the assertion passes even when it reloads the frame.
+
```
seanpdoyle, line 189: Could this assertion rely on Playwright to find the `#remote-permanent-frame` using its synchronization mechanisms instead of `document.getElementById` and a conditional?

```suggestion
  const frame = await page.locator("#remote-permanent-frame")
  
  // The final assertion cannot be noNextEventOnTarget because the assertion passes even when it reloads the frame.
  await frame.evaluate((frame) => frame.textContent = "Frame to be preserved")
```

seanpdoyle:
@brunoprietog could you approve the CI workflow for this PR so that the test suite can execute?

src/tests/functional/page_refresh_tests.js
krschacht: Good call, yes. I just tested and that successfully worked to catch the broken state. I committed this change.

krschacht:
I see the CI lint is failing with this. But the import is being used on line 44 of that file.

```
/home/runner/work/turbo/turbo/src/core/drive/morphing_page_renderer.js
Error:   2:[10](https://github.com/hotwired/turbo/actions/runs/10636867002/job/29489628603?pr=1192#step:8:11)  error  'MorphingFrameRenderer' is defined but never used  no-unused-vars
```
Also, locally I see when I run the full browser test suite locally `yarn build; yarn test:browser --project=chrome` I get this failure:

```
    [chrome] › functional/visit_tests.js:47:5 › skip programmatically visiting a cross-origin location falls back to window.location 
```

But I this test also fails for me in "main" when I run it locally.

(I see one other failing test case in my branch which I'm investigating right now)

brunoprietog:
The lint failure is legit, the file is only 40 lines long.

krschacht:
@brunoprietog Ah yes, you're right. I've fixed. But let me finish investigating this other test failure that I discovered around navigated frame refresh. I didn't touch the test and didn't expect it to break this behavior, but I'm still wrapping my mind around it to debug. Tbd

krschacht:
@brunoprietog I could use your help on this test failure. I fully understand what's going on, and it's actually touching on a really interesting edge case. I think I need ya'll to make a clear product decision and then I can implement appropriately. Let me explain:

I believe the current intended logic for turbo-frames is as follows:

> When a page morph is triggered for a page because the rules are met (refresh=morph & redirect back to the same URL), as we are potentially morphing all page elements we make special decisions about how to handle the turbo-frames. If a turbo-frame has a `src` and it has a `refresh="..."` specified then we follow this directive and either `reload` or `morph`. So the overall page may be morphing, but `refresh="reload"` might be on a turbo-frame and we'd reload that. When the turbo-frame does not have a src, we don't do any special treatment of this frame element so as the overall parent page is undergoing a morph, if this page includes html within the turbo-frame tags then it'll be morphed along with the rest of the page, if needed.

Is this understanding all accurate? I'm not attempting to propose any changes in this paragraph, I'm just trying to articulate how I think it's supposed to work. Assuming that is correct, here is the edge case:

> What happens if a turbo-frame does not have a src at the time a parent page is loaded, but something dynamically updates the turbo-frame to have a src? In fact, something could dynamically update the src & change `refresh="morph"`. Actually, something could even dynamically insert a turbo-frame where one did not exist on page load. So, at the time the parent page is refreshing with a morph, do we respect the current src & refresh attributes of the turbo-frames on the page and handle them as indicated by the logic I summarized above. Or do we ignore the current src & refresh attributes and instead look at the src, refresh, and turbo-frame contents that was just returned to us by the parent URL and update the page state so that it respects those? This would even include removing a turbo-frame from the page if the parent URL provided us new HTML which did not contain that turbo-frame.

brunoprietog:
That's a good question. The morphing process has two stages. In the first, you go through all the elements. If we find a Turbo frame susceptible to be reloaded with morphing, we ignore it completely, i.e., it's not touched, no morphing is applied on it (yet). This manifests itself here:

https://github.com/krschacht/turbo/blob/a0bf9a3f7929f7727842386761ec3f1b75f34bc2/src/core/drive/morphing_page_renderer.js#L10

The other case where morphing is not done is if the element has the `data-turbo-permanent` attribute.

Once that process is complete, what we have are all new elements returned by the server, except for Turbo frames that are susceptible to morphing reloads and permanent elements, which remain in the state prior to the server's response. This is because we are interested in preserving the URL where the Turbo frame is located if the user has navigated through it. Otherwise, each page refresh would reset the Turbo frame to the initial state of the page, and that isn't what we want.

Then comes the second stage. We go through all the Turbo frames, and if there's one that can be reloaded with morphing, we refresh and replace the current content with the new content of the Turbo frame returned by the server using morphing. This is the process we are refactoring here.

So, to answer your question, yes, you can add Turbo frames dynamically and change their src. These won't be touched if they have the conditions for reloading using morphing. Only the `reload` method will be called on them to refresh them, based on the current URL they are at. In fact, you may have Turbo frames that initially have no src attribute, but when navigating through them have it due to having performed a navigation action on them. We are interested in preserving that state.

Even on HEY Calendar we do something like this for pagination. We dynamically add Turbo frames for each new page. On each page refresh, the server doesn't actually return these Turbo frames, but since they already exist on the client, they remain there, and are reloaded with morphing.

src/tests/functional/page_refresh_tests.js
```diff
@@ -204,7 +204,7 @@ test("frames marked with refresh='morph' are excluded from full page morphing",
   await expect(page.locator("#refresh-morph")).toHaveText("Loaded morphed frame")
 })
 
-test("navigated frames without refresh attribute are reset after morphing", async ({ page }) => {
```
brunoprietog, line 204: I'm sorry if I explained wrong. Turbo frames that aren't reloaded by morphing have to be reset correctly. This test should be fine.

Turbo frames that don't have the conditions to be reloaded with morphing are affected in the first round of morphing, therefore, they are morphed according to the changes coming from the server. Turbo frames that can be reloaded with morphing due to the presence of the `src` attribute and `refresh="morph"` are ignored in the first round, and then refreshed using morphing.

krschacht: I was just about to send a long final comment saying I finished. :) Happy to do this either way. This test is a bit confusing so it took me a bit to fully make sense of. In your longer explanation, you said what my expectation would be. Specifically:

> This is because we are interested in preserving the URL where the Turbo frame is located if the user has navigated through it. Otherwise, each page refresh would reset the Turbo frame to the initial state of the page, and that isn't what we want.

This test was enforcing the opposite, so I flipped it to reflect what you said here. ^ And it fits my intuition that if a user navigations a turbo frame, and it's implicitly reload="refresh" (not morph) then I would expect that navigation to stay. But you think we should revert it back since it **started** with no src when the page loaded?

brunoprietog: Yes, that's correct, but only for Turbo frames marked with `refresh="morph"` attribute. Sorry if I wasn't clear enough.

brunoprietog: What I mean is that the goal of the test as it was before was correct. Turbo frames that don't contain the `refresh="morph"` attribute are reset. We only retain the state of Turbo frames that are reloaded with morphing. The rest are morphed in the first round according to what the server returns in the request made at the time of the page refresh.

krschacht: The test has been reverted (so that it fails again) and the logic has been corrected (so now the test passes).

I think that's it and things are now ready. 🤞

krschacht:
This PR should be ready to merge in

jorgemanrubia (approved):
Great work here @krschacht 👏, I'll merge shortly.

test-results/.last-run.json
brunoprietog, line 1: One last thing! Can you remove this one please?

krschacht:
@brunoprietog I reverted the `last-run` file and also some stray changes to yarn.lock. Be sure to merge the change to the docs as well over here: https://github.com/hotwired/turbo-site/pull/170 . Then you can close out the Issue which started all this: https://github.com/hotwired/turbo/issues/1161

Guys — While I have your attention, I am using turbo-frames extensively in my app. I have discovered a bug with history and turbo frames. I have opened a new issue regarding this, others have confirmed the bug and there is a reliable repro and fix. I can't tell if the core team has had eyes on this yet so it's worth a quick look as people are sharing monkey patch hacks and teeing up different PRs attempting to fix it: https://github.com/hotwired/turbo/issues/1300

jorgemanrubia:
Thanks @krschacht. We'll have a look at that one.

seanpdoyle:
Thank you for pushing this through @krschacht!

I've modified https://github.com/hotwired/turbo/pull/1028 to incorporate the changes made here. It also addresses the underlying issues highlighted by https://github.com/hotwired/turbo/pull/1303.

jordancoil:
When I saw this PR get merged I was excited, but then I realized the original idea to expose the `MorphingPageRenderer` and `MorphingFrameRenderer` got scrapped 😢 .

https://github.com/hotwired/turbo/pull/1234 pointed out that by exposing things like `MorphingPageRenderer` and `MorphingFrameRenderer` that we could enable custom morphing behaviour across URL changes. For example, we have a page where a difference in URL means that a sidebar is open or not with certain content. In order to fit this use case, we have had to replicate the morphing functions. We were planning on scrapping our custom implementation and using the exposed renderers once this was merged.

Is there any plans to eventually expose `MorphingPageRenderer` and `MorphingFrameRenderer`, or not?

botandrose:
@jordancoil While I can't speak to your question about exporting those classes, I wonder if you could accomplish what you describe with less coupling to Turbo's internals (and less of a public API commitment from Turbo) by hooking into the existing rendering events: https://turbo.hotwired.dev/reference/events

krschacht:
@jordancoil I'm not sure that you've lost any flexibility. I think it should work for you to catch the event:

```
  frame.addEventListener("turbo:before-frame-render", ({ detail }) => {
    detail.render = CustomFrameRenderer.renderElement
  }, { once: true })
}
```

You might need to fiddle with `capture: true/false` to get the order of firing correct, but by default your event listener should fire last so you'll get to set the final `detail.render` value which will be used.

jordancoil:
@botandrose @krschacht thanks for the suggestions. Using the `turbo:before-render` and `turbo:before-frame-render` is essentially what we have done. The thing we are missing out on is having to manually update our custom frame renderer with upgrades and bug fixes in the turbo repo. We are pretty much using the exact same morphing behaviour, so it would have been nice to just call something like `Turbo.MorphingFrameRenderer.renderElement`, and not have to worry about updating our custom implementation.

Issue #1194, opened 2024-02-21
Morphing: impossible to "clear" a form due to ignoreActiveValue: true


weaverryan:
Hi there!

In #1141, Idiomorph was changed to use `ignoreActiveValue: true`. That maybe makes sense... but it prevents a form from being reset/cleared.

For example, in the following form, after submitting, the page redirects to the same URL, but with an empty form. The behavior varies based on whether an input is active or not. In the video:

1) Submit 1 (`:00-:04`) hitting enter while inside the `<input>` results in that field NOT being cleared (wrong behavior).
2) Submit 2 (`:05-:10`) clicking the submit button makes it the `activeElement` and then all fields *are* cleared.

https://github.com/hotwired/turbo/assets/121003/8a91474a-d858-4c9f-89e6-66038cf693ba

I'm not sure how to handle this. In #1141, @seanpdoyle added `ignoreActiveValue: true` to support things like auto-submitting forms while typing into an input. What's not clear in that example is this: was `ignoreActiveValue: true` added to preserve the *value* of the input or something with its active state? To ask differently: did the submitted form HTML used for morphing in that example contain the updated `value="another search term"` attribute value on the `<input type="search">`? If not, perhaps it should have... and the example was flawed (i.e. if the new `<input type="search">` does not have a value, then the value *should* be cleared)? If yes, was `ignoreActiveValue: true` done to preserve some sort of focus or active state problem?

Thanks!

seanpdoyle:
There were two major motivations behind adding `ignoreActiveValue: true`.

In #1141, test coverage was added for maintaining focus while filling in and submitting a form. That's useful for typeahead style searches, but also for searches where you submit through <kbd>enter</kbd>.

Another motivation that wasn't codified in a test involves wanting to avoid potential confusion or disorientation caused by resetting the form control with focus when receiving a broadcast refresh.

Imagine starting to fill out a form on a page, receiving a broadcast `<turbo-stream action="refresh">`, then losing that form control's value while maintaining focus within the element. 

If we were to rollback the addition of `ignoreActiveValue: true`, the typeahead form submission behavior could be implemented by adding `[data-turbo-permanent]` when the form control dispatches `focusin`, then removing it when the form control dispatches `focusout`. That is a pattern we can codify with documentation.

[...]

PR #1195, opened 2024-02-21
Omit `ignoreActiveValue: true` Morph option

seanpdoyle:
Closes [#1194][]
Rolls back [#1141][]

Don't pass the `ignoreActiveValue: true` option when Morphing. To restore that behavior, applications can set `[data-turbo-permanent]` when form control receives focus (through a `focusin` event listener), then remove it if necessary when the form control loses focus (through a `focusout` event listener).

[#1141]: https://github.com/hotwired/turbo/pull/1141/
[#1194]: https://github.com/hotwired/turbo/issues/1194


jorgemanrubia (approved):
Thanks a lot @seanpdoyle 🙏

PR #1196, opened 2024-02-21
Ensure that page refreshes do not trigger a snapshot cache

afcapel:
Fixes a bug introduced in https://github.com/hotwired/turbo/pull/1146

`exemptPageFromPreview()` adds a `<meta>` tag to the `<head>` setting `turbo-cache-control` to `no-preview`. However, since the MorphRenderer now inherits from the PageRenderer, it can update meta tags in the head and remove the `turbo-cache-control` tag. This means that the snapshot cache will be used for the next visit, which is not what we want.

Specifying `shouldCacheSnapshot: false` in the `visit` options ensures that the snapshot cache is not used for the refresh visit.

### DIFF
--- a/src/core/session.js
+++ b/src/core/session.js
@@ -110,8 +110,7 @@ export class Session {
   refresh(url, requestId) {
     const isRecentRequest = requestId && this.recentRequests.has(requestId)
     if (!isRecentRequest) {
-      this.cache.exemptPageFromPreview()
-      this.visit(url, { action: "replace" })
+      this.visit(url, { action: "replace", shouldCacheSnapshot: false })
     }
   }
 
--- a/src/tests/functional/page_refresh_tests.js
+++ b/src/tests/functional/page_refresh_tests.js
@@ -33,6 +33,7 @@ test("async page refresh with turbo-stream", async ({ page }) => {
 
   await expect(page.locator("#title")).not.toHaveText("Updated")
   await expect(page.locator("#title")).toHaveText("Page to be refreshed")
+  expect(await noNextEventNamed(page, "turbo:before-cache")).toBeTruthy()
 })
 
 test("dispatches a turbo:before-morph-element and turbo:morph-element event for each morphed element", async ({ page }) => {


pfeiffer:
Hey, sorry if hijacking, but related to this is a [few other issues ](https://github.com/hotwired/turbo-ios/issues/136)and PRs that I've been exploring around page refreshes, in particular how they behave in Turbo Native with Turbo 8. 

Specifically refreshes in Turbo Native will make the native app unresponsive as the native apps will display a spinner and blank the WebView during the refresh action which makes the new refreshes more or less break any native apps that subscribe to refresh broadcasts:

https://github.com/hotwired/turbo/assets/195925/5f41f74a-70c4-472b-9b46-d1173ad8fd71

I'm wondering what the reason is for exempting refreshes from the snapshot cache? I've explored actually making a snapshot of the page before a refresh in #1188 which significantly improves the Turbo Native experience, as at least the WebViews aren't blanked out while refreshing. This PR seems to re-iterate the wish to exempt refreshes from snapshots, but it's not clear to me why?

afcapel:
@pfeiffer we've found a few issues in which rendering a cache snapshot interferes with the morphing algorithm. In particular we don't want a cached preview rendering on a morph refresh because that can override any changes to the current page, which is what we want to avoid with morphing.

For example, we found that rendering a cached snapshot may result in closing a modal that was added dynamically to the page after the cache snapshot was taken.

pfeiffer:
> In particular we don't want a cached preview rendering on a morph refresh because that can override any changes to the current page, which is what we want to avoid with morphing.
> 
> For example, we found that rendering a cached snapshot may result in closing a modal that was added dynamically to the page after the cache snapshot was taken.

Wouldn't that be solved by explicitly making a new snapshot before refreshing as introduced in #1188? In my tests, I've seen the issues you describe if there was snapshot of the page taken way earlier (ie. as part of a `advance` -> `restore` -> `refresh` cycle). In that case I've seen eg. a modal being closed during the refresh visit, as the snapshot was taken of the page while navigating with the advance action.

By explicitly taking a snapshot in #1188 before refreshes, we ensure that snapshots being shown are up to date with the current state (ie modals open).

afcapel:
@pfeiffer that's good point, it may be an implementation artifact. I'll give #1188 a try, to see if it solves the problems we've seen, and will come back to you 👍

jayohms:
It looks like the iOS-specific issue is due to the `adapter.visitStarted()` callback after a refresh is initiated and how the `turbo-ios` adapter works: https://github.com/hotwired/turbo-ios/blob/c9d060fcdec136c3808cd0764e956d375012c6fb/Source/Session/Session.swift#L173

I'll look at this separately, but we should be able to stop displaying the activity/progress indicator in the iOS library during same-page refreshes/visits.

PR #1201, opened 2024-02-23
Add `Navigator#locationWithActionIsPageRefresh` for native adapter use

pfeiffer:
Related to https://github.com/hotwired/turbo-ios/pull/178 in Turbo iOS.

In order to support refreshes in Turbo Native, the native adapters need to determine if a proposed visit is a refresh visit. This PR adds a new method `locationWithActionIsPageRefresh` to `Navigator`, which the native adapters can call instead of implementing the logic on their own to determine if a proposed visit is a refresh action or not.

This way, the native adapters do not need to care about _how_ to determine if a visit is refresh; this is kept inside Turbo and can be updated without having to publish new native modules.

This PR also adds tests for the methods on `Navigator` which the native adapters use today, so that we make sure that whatever refactoring of Turbo still keep the required interface intact for supporting the native adapters.

See discussion in: https://github.com/hotwired/turbo-ios/pull/178#discussion_r1500266398

There will be a corresponding PR in Turbo iOS using this method.

### DIFF
--- a/src/core/drive/navigator.js
+++ b/src/core/drive/navigator.js
@@ -139,6 +139,10 @@ export class Navigator {
     )
   }
 
+  locationWithActionIsPageRefresh(location, action) {
+    return this.view.lastRenderedLocation.pathname === location.pathname && action === "replace"
+  }
+
   visitScrolledToSamePageLocation(oldURL, newURL) {
     this.delegate.visitScrolledToSamePageLocation(oldURL, newURL)
   }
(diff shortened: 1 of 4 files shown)


pfeiffer:
I've opened a corresponding PR in Turbo iOS using this new method: https://github.com/hotwired/turbo-ios/pull/179

CC @jayohms @afcapel

pfeiffer:
Closing as discussed in https://github.com/hotwired/turbo-ios/pull/178#discussion_r1501155443

PR #1204, opened 2024-02-23
Trigger an event when adding a new node

coorasse:
While implementing some animations on our page, we realized that we had no access to an event before an element was added to the DOM.
With this PR I'd like an early feedback from maintainers if this is the right direction and I can continue.
In particular regarding the event name.

If I get your OK, I'll:
* add tests
* update documentation to include this new event

Let me know 🙏


coorasse:
I'd love to get a feedback on this one. Thanks!

PR #1205, opened 2024-02-24
Add `turbo:before-prefetch` test coverage

seanpdoyle:
Expand the link prefetching functional test coverage to exercise prefetching in relation to the events dispatched during the lifecycle. For example, include coverage for:

* `turbo:before-prefetch` being dispatched
* `turbo:before-prefetch` being canceled
* `turbo:before-fetch-request` dispatched only once during a successful prefetch


brunoprietog:
Thanks @seanpdoyle!

PR #1208, opened 2024-02-28
Add `[method]` and `[scroll]` attributes for Refresh Stream

seanpdoyle:
> I have a page where I am using Turbo morph. when I submit a form and
> redirect I would like to reset the scroll, but if a refresh is triggered
> by a broadcast I would like to preserve teh scroll. Is that possible ?
>
> [#turbo Discord][discord]

This commit expands the set of attributes for `<turbo-stream action="refresh">` to include `[method]` and `[scroll]` (in addition to `[request-id]`). These attributes correspond directly to the [`turbo-refresh`-prefixed `<meta>` element][meta] elements that control morphing and scroll preservation.

When present on the `<turbo-stream action="refresh">`, their values are forward along to the `Session.refresh` method call, which in turn encodes them into Visit options under the `refresh` key. Those options are then used during `Visit` instantiation, and transformed into `.refresh` properties.

At render time, the `PageRenderer` attempts to read the refresh method and scroll preservation settings with the following precedence:

1. read from the corresponding `Visit.refresh` property (possibly null)
2. read from the corresponding `<meta name="turbo-...">` element (possibly null)

If no value is provided, fallback to the default (`{ method: "replace",
scroll: "reset" }`).

[discord]: https://discord.com/channels/988103760050012160/1044659721229054033/1212443786270212167
[meta]: https://turbo.hotwired.dev/handbook/page_refreshes#morphing

### DIFF
--- a/src/core/drive/navigator.js
+++ b/src/core/drive/navigator.js
@@ -99,7 +99,7 @@ export class Navigator {
       } else {
         await this.view.renderPage(snapshot, false, true, this.currentVisit)
       }
-      if(!snapshot.shouldPreserveScrollPosition) {
+      if (snapshot.refreshScroll !== "preserve") {
         this.view.scrollToTop()
       }
       this.view.clearSnapshotCache()
--- a/src/core/drive/page_snapshot.js
+++ b/src/core/drive/page_snapshot.js
@@ -74,12 +74,12 @@ export class PageSnapshot extends Snapshot {
     return this.headSnapshot.getMetaValue("view-transition") === "same-origin"
   }
 
-  get shouldMorphPage() {
-    return this.getSetting("refresh-method") === "morph"
+  get refreshMethod() {
+    return this.getSetting("refresh-method")
   }
 
-  get shouldPreserveScrollPosition() {
-    return this.getSetting("refresh-scroll") === "preserve"
+  get refreshScroll() {
+    return this.getSetting("refresh-scroll")
   }
 
   // Private
--- a/src/core/drive/page_view.js
+++ b/src/core/drive/page_view.js
@@ -16,7 +16,8 @@ export class PageView extends View {
   }
 
   renderPage(snapshot, isPreview = false, willRender = true, visit) {
-    const shouldMorphPage = this.isPageRefresh(visit) && this.snapshot.shouldMorphPage
+    const refreshMethod = visit?.refresh?.method || this.snapshot.refreshMethod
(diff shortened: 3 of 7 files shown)


seanpdoyle:
@adrienpoly I've opened this in response to our conversation in Discord.

seanpdoyle:
The changes to the `Visit` options has potential to provide programatic resolutions to:

* https://github.com/hotwired/turbo/pull/1040
* https://github.com/hotwired/turbo/issues/1177

I feel that the `[method]` and `[scroll]` attributes matching the `turbo-refresh`-prefixed meta element `[name]` attributes is appropriate.

I'm less confident about the Visit options key (and resulting property name). The "public API" boundaries of Visit options are currently unclear. If we intend for these keys to be private and for internal use only, the name isn't important to decide upon. If they're intended to be public, I think it's important to workshop and appropriate name and structure.

@jorgemanrubia does this seem like a conservative enough expansion of Morph Refresh support?

adrienpoly:
> @adrienpoly I've opened this in response to our conversation in Discord.

Thanks, I am trying to understand how I could leverage in my Rails app and it is a bit unclear.

A simplify version of what I have is 

```ruby
class Post 
  broadcast_refreshes
  ...
end
````

and a view view 

```erb
<%= turbo_stream_from @post %>
<%= turbo_refreshes_with method: :morph, scroll: :reset %>

<form ....></form>
```

The form redirect to this page with a scroll reset and I would like to preserve the scroll  when the refresh is triggered by a broadcast.

Looking at your proposal I am not sure how I should set the turbo-stream attributes?

Would it be something like this?

```erb
<%= turbo_stream_from @post, scroll: :preserve %>
```

or at the model level?

Thanks

seanpdoyle:
> or at the model level?

To support scroll resets from redirects during a form submission, you could continue to render the `<meta name="turbo-refresh-scroll" content="reset">`, or omit the element entirely.

To support scroll preservation from broadcasts, the idea would be to broadcast the `<turbo-stream>` element with the configuration encoded directly into the element.

Since this change must be made separate from the Rails code, that support is still unclear and to be determined. One possibility could be to extend `broadcast_refreshes` to accept HTML attribute options:

```ruby
broadcast_refreshes scroll: :preserve
```

Another approach might be to keep `broadcast_refreshes` generic and simple like it currently is, but change `broadcast_refresh_to` and `broadcast_refresh_later_to` to accept the attributes in the same way that the other stream broadcast methods accept `:target`, `:targets`, etc.

seanpdoyle:
@jorgemanrubia are you able to review this proposal?

Issue #1209, opened 2024-03-01
Turbo stream refresh takes precedence over user interaction


rbclark:
I am running into a bit of an interesting race condition when using Turbo 8 with broadcasted `refresh` actions. It seems that the streaming refresh action takes precedence over user interactions on the page, meaning a user could click a link on a page and instead receive the refresh response of the previous page. I've attached a video of the behavior and a simple repository to reproduce the issue. In this video I have simulated an app that is a bit slower to respond by adding a 2 second delay to the index route, and then I am clicking on the headings in order to sort the users list.

https://github.com/hotwired/turbo/assets/2308869/753b2e7f-b5e8-47f0-82c9-6b2d5c947c4c

In the console at the bottom of the screen you can see my click action fire and then be cancelled by the turbo stream refresh streaming from the server. Also note that the list maintains it's original order.

The demo app in the video can be found at: https://github.com/rbclark/turbo-8-stimulus-bug-demo. I've tried playing around with different ways of disabling turbo refreshes once the user has interacted with the page using the different `turbo:before-*` events but so far to no avail.

seanpdoyle:
Thank you for opening this issue, and thank you for sharing a bug report repository.

If the root of the issue is that receiving a `<turbo-stream>` while a `Visit` is in-flight, I've opened https://github.com/hotwired/turbo/pull/1213 to try and resolve that race condition.

Are you able to consume that fork's branch and test whether or not it resolves your application's issues?

rbclark:
Thank you for the quick reply! I've gone ahead and applied the changes locally and it seems that `this.navigator.currentVisit` is being set and persisting, causing refreshes to stop working after the first refresh. 


https://github.com/hotwired/turbo/assets/2308869/9ed6566a-7f39-4797-9be3-ec2123b4faec


I went ahead and threw a breakpoint in and it looks like `this.navigator.currentVisit` is being set after morph.

https://github.com/hotwired/turbo/assets/2308869/f625d54f-dfb4-4b5a-ad5e-3e04dc47ec40

Full disclosure: I was having a little bit of trouble getting a local build of Turbo that I could point to in my rails project with actioncable working so I just went ahead and directly modified `app/assets/javascripts/turbo.js` in my local copy of the `turbo-rails` project in order to test this change. I don't believe that accounts for the new behavior I'm seeing though.

PR #1213, opened 2024-03-01
Prevent Refresh from interrupting ongoing Visit

seanpdoyle:
Closes [#1209]

Defers the `Session.refresh` calls while handling `<turbo-stream action="refresh">` elements if there is a `Visit` that's already initiated and being handled by the `Navigator`.

[#1209]: https://github.com/hotwired/turbo/issues/1209

### DIFF
--- a/src/core/drive/navigator.js
+++ b/src/core/drive/navigator.js
@@ -125,6 +125,7 @@ export class Navigator {
 
   visitCompleted(visit) {
     this.delegate.visitCompleted(visit)
+    delete this.currentVisit
   }
 
   locationWithActionIsSamePage(location, action) {
--- a/src/core/session.js
+++ b/src/core/session.js
@@ -109,7 +109,7 @@ export class Session {
 
   refresh(url, requestId) {
     const isRecentRequest = requestId && this.recentRequests.has(requestId)
-    if (!isRecentRequest) {
+    if (!isRecentRequest && !this.navigator.currentVisit) {
       this.visit(url, { action: "replace", shouldCacheSnapshot: false })
     }
   }
--- a/src/tests/fixtures/page_refresh.html
+++ b/src/tests/fixtures/page_refresh.html
@@ -81,6 +81,7 @@
     <h1 id="title">Page to be refreshed</h1>
 
     <a href="/src/tests/fixtures/page_refresh.html" id="reload-link">Reload</a>
+    <a href="/__turbo/delayed_response" id="delayed_link">Navigate with delayed response</a>
 
     <turbo-frame id="refresh-morph" src="/src/tests/fixtures/frame_refresh_morph.html" refresh="morph">
       <h2>Frame to be morphed</h2>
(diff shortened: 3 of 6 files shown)


rbclark:
Thank you for this! I've been unable to reproduce #1209 with these changes applied and everything else seems to still be functioning properly. Really appreciate the quick response!

seanpdoyle:
@afcapel @jorgemanrubia are either of you available to review this change?

src/core/drive/navigator.js
```diff
@@ -125,6 +125,7 @@ export class Navigator {
 
   visitCompleted(visit) {
     this.delegate.visitCompleted(visit)
+    this.stop()
```
jorgemanrubia, line 128: I see that, before, it was invoking `stop` only when starting visits and submitting forms. This looks correct but, since this is a pretty important path, any concerns about introducing unwanted side effects here? I can't think of anything in particular, but concerned about interrupting form submissions we were not interrupting before.

jorgemanrubia (approved):
@seanpdoyle thanks for fixing this one!

src/core/drive/navigator.js
seanpdoyle: > This looks correct but, since this is a pretty important path, any concerns about introducing unwanted side effects here?

Now that you mention it, a `Visit` initiated by a link click is guaranteed to either do nothing (when canceled) or navigated. A form, on the other hand, has the potential to complete and succeed without navigating when handling stream responses. The same is true for a link click with `[data-turbo-stream]`.

Acknowledging that, what *is* the desired order of operations for a Page Refresh that's competing with a Stream form submission?

It feels like the Visit should always win out, since that's requesting the most (and most up-to-date) content. Is that correct?

seanpdoyle: @jorgemanrubia I've pushed up a change to remove the call to `this.stop()`, and replace it with `delete this.currentVisit`.

seanpdoyle-intercom: @jorgemanrubia are you available to re-review this?

jorgemanrubia (approved):
Thanks @seanpdoyle.

jorgemanrubia:
This one is good to merge @afcapel from my side.

klevo:
I'm not sure this resolves the issue. I have reported a [similar issue](https://github.com/hotwired/turbo-rails/issues/622) separately and after @seanpdoyle 's notification tested this patch, but it does not prevent the interruption in my test app.

klevo:
> I'm not sure this resolves the issue. I have reported a [similar issue](https://github.com/hotwired/turbo-rails/issues/622) separately and after @seanpdoyle 's notification tested this patch, but it does not prevent the interruption in my test app.

Actually these are most likely two separate issues. The issue I was referring to and [opened a PR for](https://github.com/hotwired/turbo/pull/1250) is a scenario where user navigation succeeds but is then overridden by a debounced refresh from previous page. This one is about interrupting in-progress visits. So this is well worth merging 👍

Issue #774, opened 2022-10-21
Accessibility issues with Turbo navigation


khiga8:
Hi Turbo team, 

Our team migrated to using turbo drive and turbo frame in our Rails app around 3 months ago. We have received reports and discovered that turbo navigation poses accessibility issues for our screen reader users. Some of these issues, we can alleviate on our end through better guidance and additional accessibility feedback updates. Other issues, we do not see any workarounds for. We wanted to share our investigation and the feedback we’ve gotten with you.

### Expected behavior

With a standard, hard page navigation, there are default mechanisms built in that allow a browser to communicate to a screen reader user that page navigation has taken place. For instance, when page navigation happens, the page title will be announced by the screen reader. 

In addition, some screen readers have settings that can be toggled to specify what happens when loading a new web page like “Speak webpage summary” or “Automatically speak the webpage”. When these settings are toggled, a screen reader user can expect certain announcements to be made on page load.

### Turbo frame

When a turbo frame navigation happens, there is no feedback given to screen reader users to indicate anything has happened.

### Turbo drive

- When a turbo drive navigation happens, there is inconsistent feedback given to screen reader users to indicate anything has happened. We saw inconsistent results across different screen reader and browser combinations.
  - With VoiceOver/Safari, the new page title is announced somewhat frequently.
  - With NVDA/Chrome, the announcement is much more infrequent or non-existent. The most commonly used screen readers are NVDA and JAWS so announcements being made in VoiceOver is not enough. (Source: [WebAim: Screen Reader / Browser combinations](https://webaim.org/projects/screenreadersurvey9/#browsercombos))
The screen reader settings for behavior that should happen at page load like,  “Speak webpage summary” or “Automatically speak the webpage” are not respected, even in scenarios where the page title is announced.

### Workarounds and Caveats

- We worked around the lack of page navigation announcements by populating the  [aria-live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) with the page title. We use this approach for both turbo frame and turbo drive navigations. However there are still caveats to this.
  - We noted above that with turbo drive, page title announcements are inconsistently made. As a result, we end up with duplicate page title announcements at times. We are not sure if there’s anything within turbo drive that results in page title announcements, but it is inconsistent and need to ensure it happens.
[...]

PR #1216, opened 2024-03-02
Configure Submitter disabling

seanpdoyle:
Depends on [#1217][]
===

Follow-up to [#386][]

While Turbo's support for disabling a Form Submissions `<input
type="submit">` or `<button>` element is rooted in [Rails UJS][], it
degrades (and has always degraded) the accessibility of those
experiences.

To learn more about the risks involved, read the [Don't Disable
Submits][] section of Adrian Roselli's [Don't Disable Form Controls][]
along with the additional resources mentioned therein.

The risk of degrading accessibility is especially true for Morph-enabled
Form Submissions. If a form submission will trigger a morphing Page
Refresh with the submitter focused, it's likely that the focus *is
intended to* remain on the submitter.

With the current `[disabled]` behavior, that is not possible without a
bespoke event handler like:

```js
addEventListener("submit", ({ target, submitter }) => {
  if (submitter) {
    target.addEventListener("turbo:submit-start", () => {
      submitter.disabled = false
      submitter.focus()
    }, { once: true })
  }
})
```

This commit introduces a `Turbo.config.forms.submitter` object with two
pre-defined keys: `"disabled"` (the default until we can deprecate it),
and `"aria-disabled"`.

When applications specify either `Turbo.config.forms.submitter =
"disabled"` or `Turbo.config.forms.submitter = "aria-disabled"`, they
will be able to leverage those pre-packed hooks. Otherwise, they can
provide their own object with `beforeSubmit(submitter)` and
`afterSubmit(submitter)` functions.

[#386]: https://github.com/hotwired/turbo/pull/386
[#1217]: https://github.com/hotwired/turbo/pull/1217
[Rails UJS]: https://guides.rubyonrails.org/v6.0/working_with_javascript_in_rails.html#automatic-disabling
[Don't Disable Form Controls]: https://adrianroselli.com/2024/02/dont-disable-form-controls.html
[Don't Disable Submits]: https://adrianroselli.com/2024/02/dont-disable-form-controls.html#Submit

### DIFF
--- a/src/core/config/drive.js
+++ b/src/core/config/drive.js
@@ -1,0 +1,4 @@
+export const drive = {
+  enabled: true,
+  progressBarDelay: 500
+}
--- a/src/core/config/forms.js
+++ b/src/core/config/forms.js
@@ -1,0 +1,3 @@
+export const forms = {
+  mode: "on"
+}
--- a/src/core/config/forms.js
+++ b/src/core/config/forms.js
@@ -1,3 +1,41 @@
-export const forms = {
-  mode: "on"
+import { cancelEvent } from "../../util"
+
+const submitter = {
+  "aria-disabled": {
+    beforeSubmit: submitter => {
+      submitter.setAttribute("aria-disabled", "true")
+      submitter.addEventListener("click", cancelEvent)
+    },
+
+    afterSubmit: submitter => {
+      submitter.removeAttribute("aria-disabled")
+      submitter.removeEventListener("click", cancelEvent)
+    }
+  },
+
+  "disabled": {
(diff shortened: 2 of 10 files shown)


seanpdoyle:
@afcapel @jorgemanrubia this is inspired by https://github.com/hotwired/turbo/issues/774, although it not mentioned directly.

seanpdoyle:
@jorgemanrubia @brunoprietog are either of you able to review this proposal (as a follow-up to https://github.com/hotwired/turbo/pull/1217)?

jorgemanrubia (approved):
Thanks @seanpdoyle

PR #1217, opened 2024-03-03
Introduce `Turbo.config` object

seanpdoyle:
The initial implementation aims to replace the ad hoc configurations spread across `window.Turbo` and `Turbo.session`. Those objects shouldn't be considered part of the public interface, especially for extension.

This commit exports a single object from `src/core/config` to serve as a centralized, user-facing repository for configuration values.

The main `src/core/config` object is composed of two constituent parts:

* `src/core/config/drive`
* `src/core/config/forms`

In the future, the directory structure can be expanded upon to suit the needs of the project.

Any user-facing configuration outside of `Turbo.config` is deprecated.

### DIFF
--- a/src/core/config/drive.js
+++ b/src/core/config/drive.js
@@ -1,0 +1,4 @@
+export const drive = {
+  enabled: true,
+  progressBarDelay: 500
+}
--- a/src/core/config/forms.js
+++ b/src/core/config/forms.js
@@ -1,0 +1,3 @@
+export const forms = {
+  mode: "on"
+}
--- a/src/core/config/index.js
+++ b/src/core/config/index.js
@@ -1,0 +1,7 @@
+import { drive } from "./drive"
+import { forms } from "./forms"
+
+export const config = {
+  drive,
+  forms
+}
--- a/src/core/drive/form_submission.js
+++ b/src/core/drive/form_submission.js
@@ -3,6 +3,7 @@ import { expandURL } from "../url"
 import { clearBusyState, dispatch, getAttribute, getMetaContent, hasAttribute, markAsBusy } from "../../util"
 import { StreamMessage } from "../streams/stream_message"
 import { prefetchCache } from "./prefetch_cache"
+import { config } from "../config"
 
 export const FormSubmissionState = {
   initialized: "initialized",
@@ -22,7 +23,7 @@ export const FormEnctype = {
 export class FormSubmission {
   state = FormSubmissionState.initialized
 
-  static confirmMethod(message, _element, _submitter) {
+  static confirmMethod(message) {
     return Promise.resolve(confirm(message))
   }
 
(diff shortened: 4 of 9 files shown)


seanpdoyle:
@afcapel @jorgemanrubia Trix [utilizes a similar style of Object](https://github.com/basecamp/trix/blob/f7e676d65cd3615747dc18dceab3123774910d63/src/trix/config/index.js) that consists of [multiple modules scoped to particular concerns](https://github.com/basecamp/trix/tree/main/src/trix/config).

seanpdoyle:
@jorgemanrubia @brunoprietog are either of you able to review this proposal?

OutlawAndy:
Hey @seanpdoyle.  I was kind of surprised to learn of a change to Turbo's public API in a patch release.  But seeing as that ship has sailed, I have another question for you.

We maintain a [package](https://github.com/RoleModel/turbo-confirm) that integrates with `Turbo.setConfirmMethod` and wondering if you have an opinion on the best way to handle this change?

I'd rather not introduce branching logic based on the existence of a Turbo.config object. e.g.

```js
  if (window.Turbo.config) {
    window.Turbo.config.forms.confirm = confirmationHandler
  } else {
    window.Turbo.setConfirmMethod(confirmationHandler)
  }
```

However, I'd also prefer not to entirely drop support for previous versions of Turbo.  Especially considering that I'm not offering any other reasons for upgrading our package other than you can't, unless you also upgrade Turbo itself.

I'm concerned that adding a peerDependency requirement only emits a warning that is easily overlooked, but will actually crash our package if left unmet.

I would appreciate any feedback you have on the matter.

Thank you,

- Andy

our internal discussion: https://github.com/RoleModel/turbo-confirm/pull/31

seanpdoyle:
@OutlawAndy thank you for sharing those details here. Personally, I am not involved in Turbo's release schedule, so I cannot provide any insights into that process. I will say that the deprecations in this diff were intended to buy time and provide hints to consuming packages.

This diff proposed the `Turbo.config` object to serve as the interface for configuring Turbo. Until the `setConfirmMethod` is removed, you can continue to call it instead of interacting with `Turbo.config`, but it is will emit a deprecation warning.

As far as the impacts to your package are concerned, I understand your distaste for adding a conditional to cover this case. However, if you are committed to maintaining support for all historical Turbo version, that decision involves supporting Turbo's full range of interfaces as they evolve. Turbo does not programmatically expose version information to the JavaScript runtime, so packages are left with graceful degradation-style conditionals to serve as safety nets.

Again, while I cannot speak to Turbo's release schedule or maintenance policies, I do believe that as the package and its interfaces evolve, there will be an evolving minimal threshold version that will serve as a baseline of backwards compatibility. For now, the deprecations keep `Turbo.setConfirmMethod` on the supported side of that threshold, but eventually that won't be true anymore.

> I'm concerned that adding a peerDependency requirement only emits a warning that is easily overlooked

I think this ultimately boils down to your package's compatibility goals, but a `peerDependency` requirement feels one of the few concrete ways to encode your support goals to the consumers.

OutlawAndy:
@seanpdoyle Thank you for the insightful response!  That's quite helpful!

Issue #1154, opened 2024-02-01
Autofocus tests are flaky


afcapel:
The often fails on CI, for example https://github.com/hotwired/turbo/actions/runs/7723792974/job/21054601474

afcapel:
Another failure https://github.com/hotwired/turbo/actions/runs/7755348058/job/21150514185

seanpdoyle:
I've opened https://github.com/hotwired/turbo/pull/1219 to try and address this.

PR #1219, opened 2024-03-03
Use Playwright assertions for autofocus

seanpdoyle:
Closes [#1154][]

Replace bespoke CSS selector-based assertions with Playwright's built-in [toBeFocused](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-focused) assertion.

Like other assertions, `toBeFocused` will wait and retry for a period of time, which makes `nextBeat` calls unnecessary.

[#1154]: https://github.com/hotwired/turbo/issues/1154


afcapel:
Nice, thanks @seanpdoyle !

PR #1225, opened 2024-03-18
Bugfix/enable turbo stream morph on text

hcdeng:
I had the chance to test out @omarluq 's fantastic addition of the Turbo Stream morph action in https://github.com/hotwired/turbo/pull/1185. The feature works as expected, except fails to morph text-only changes. This appears to be due to a misreading of the Turbo-Rails implementation of the `beforeNodeMorph`/`#shouldMorphElement` callback: https://github.com/hotwired/turbo-rails/blob/102a491754d46f7dd924201fcfaf879a0f04b11c/app/assets/javascripts/turbo.js#L3830-L3844.

This PR updates the `beforeNodeMorph` callback to match the Turbo-Rails implementation -- with a tweak to hopefully increase readability -- and enable text-only morphing.

### DIFF
--- a/src/core/streams/actions/morph.js
+++ b/src/core/streams/actions/morph.js
@@ -26,19 +26,20 @@ function beforeNodeRemoved(node) {
 }
 
 function beforeNodeMorphed(target, newElement) {
-  if (target instanceof HTMLElement) {
-    if (!target.hasAttribute("data-turbo-permanent")) {
-      const event = dispatch("turbo:before-morph-element", {
-        cancelable: true,
-        target,
-        detail: {
-          newElement
-        }
-      })
-      return !event.defaultPrevented
-    }
-    return false
+  if (!(target instanceof HTMLElement)) {
+    return
   }
+  if (!target.hasAttribute("data-turbo-permanent")) {
+    const event = dispatch("turbo:before-morph-element", {
+      cancelable: true,
+      target,
+      detail: {
+        newElement
+      }
+    })
+    return !event.defaultPrevented
+  }
+  return false
 }
 
 function beforeAttributeUpdated(attributeName, target, mutationType) {
--- a/src/core/streams/actions/morph.js
+++ b/src/core/streams/actions/morph.js
@@ -26,20 +26,19 @@ function beforeNodeRemoved(node) {
 }
 
 function beforeNodeMorphed(target, newElement) {
-  if (!(target instanceof HTMLElement)) {
-    return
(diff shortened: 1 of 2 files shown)


src/core/streams/actions/morph.js
```diff
@@ -26,7 +26,10 @@ function beforeNodeRemoved(node) {
 }
 
 function beforeNodeMorphed(target, newElement) {
-  if (target instanceof HTMLElement && !target.hasAttribute("data-turbo-permanent")) {
+  if (!(target instanceof HTMLElement)) {
+    return
+  }
+  if (!target.hasAttribute("data-turbo-permanent")) {
```
omarluq, line 32: The issue with this implementation is that the `turbo:before-morph-element` event will not be dispatched for text morphs, instead I would change the if condition to a logical or that way events will be consistently dispatched
```suggestion
  if () {
    return
  }
  if (!(target instanceof HTMLElement) || !target.hasAttribute("data-turbo-permanent")) {
```

omarluq (changes requested):
Nice catch, one note about the implementation.

omarluq (approved):

omarluq:
CC: @jorgemanrubia for visibility

src/core/streams/actions/morph.js
```diff
@@ -26,7 +26,10 @@ function beforeNodeRemoved(node) {
 }
 
 function beforeNodeMorphed(target, newElement) {
-  if (target instanceof HTMLElement && !target.hasAttribute("data-turbo-permanent")) {
+  if (!(target instanceof HTMLElement)) {
```
jorgemanrubia, line 29: Could you please change it to wrap with the condition instead of a clause guard:

```js
if (target instanceof HTMLElement) {
...
}
```

I prefer to keep things consistent there, and, also, I think Turbo's codebase tends to avoid guard conditions (that's a just a matter of style preference).

jorgemanrubia (commented):
Thank you so much for fixing @hcdeng! A very minor comment below:

These bugs are one of the risks that @seanpdoyle pointed out for having two pieces of code implementing the same logic. This will be a nice refactor to do, although it's not as as simple and extracting and reusing the function: we need to rework the internals of how we are handling turbo frames too.

src/core/streams/actions/morph.js
hcdeng: Updated in 9c2bd18ff21d49bd9fc3d6df98822689d49c4b49, thanks!

Issue #886, opened 2023-03-02
Frames with loading lazy not working when src changes after restoration visit


night91:
I have been having troubles with frames with lazy loading after a restoration visit. It doesn't load when src changes (using a link with a `data-turbo-frame`). Removing the lazy in the inspector, it makes the frame to work again.

https://github.com/hotwired/turbo/blob/4593d06ce58d17af5b17495ad8524eaa9bc2f5d2/src/core/frames/frame_controller.ts#L116
`if (this.loadingStyle == FrameLoadingStyle.eager || this.hasBeenLoaded) {`

I have been digging into the code and after a restoration visit the `loadSourceURL()` doesn't get called because of loading lazy and the status of the `hasBeenLoaded` is false (lost value after the cache is loaded).

I am not sure about this because I don't have the full scope. Can someone verify this?

Version: v7.3.0

kevinmcconnell:
@night91 yes, I can reproduce this. It does look like a bug to me.

With markup like the following, the link stops working after a restoration visit:
```html
<body>
  <turbo-frame id="frame" src="/frame1.html" loading="lazy"></turbo-frame>

  <a href="/frame2.html" data-turbo-frame="frame">another frame</a>
</body>
```

And as you say, the problem is only when `loading="lazy"`. Verified on both 7.2.0 & 7.3.0.

Thanks for reporting! 🙏

dineshpanda:
It doesn't seem to be working with eager loaded frames as well:

```
<body>
  <turbo-frame id="frame" src="/articles">
    <a href="/articles?sort=author">Sort by author name</a>
    <%= render @articles %>
  </turbo-frame>
</body>
```

PedroAugustoRamalhoDuarte:
> Just ran into this issue, thought I might share my quick fix until things get ironed out in the library.
> 
> Before the lazy loaded turbo-frame gets cached, remove the loading attribute but only if the frame is marked as complete. This allows frames that weren't yet lazy loaded to continue to function as expected.
> 
> ```
> addEventListener('turbo:before-cache', () => {
>   document
>     .querySelectorAll('turbo-frame[loading="lazy"][complete]')
>     .forEach(frame => frame.removeAttribute('loading'))
> })
> ```

I had a similar issue, this code that work for me:

```js
// Handles turbo bug with lazy loading
// https://github.com/hotwired/turbo/issues/886
document.addEventListener('turbo:load', () => {
  document
[...]

PR #1227, opened 2024-03-19
[fix] loading after completed lazy loading and restored visit

yasu551:
## Description

There is a problem with loading after lazy loading is completed and then visit restored. The frames that have completed lazy loading do not execute their loading processes. This problem occurs during visit restoration because the status information indicating that lazy loading has already been completed is not being fully restored. To reproduce this successfully, we need to set `#hasBeenLoaded` to true when connecting to the frame.

https://github.com/hotwired/turbo/compare/main...yasu551:turbo:fix/frame-loading-lazy-restoration-visit#diff-c202bf76718319f756622258d06264c47dd74d0611ab50b8a28d7b6de0fe3ff5R52-R57

## Relations
Closes https://github.com/hotwired/turbo/issues/886

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -50,7 +50,11 @@ export class FrameController {
     if (!this.#connected) {
       this.#connected = true
       if (this.loadingStyle == FrameLoadingStyle.lazy) {
-        this.appearanceObserver.start()
+        if (this.complete) {
+          this.#hasBeenLoaded = true
+        } else {
+          this.appearanceObserver.start()
+        }
       } else {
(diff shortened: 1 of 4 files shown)


yasu551:
@seanpdoyle could you review this PR?

adrianthedev:
This is PR is fixing a few pretty grim issues for us as well. 
I tested the code locally and all seems sorted.
We spoke about it on https://github.com/avo-hq/avo/issues/2607.

@jorgemanrubia would you be able to have a quick look over this?
We can chip in and further test/troubleshoot/write some code.

Thank you!

Issue #1171, opened 2024-02-08
Turbo should not handle forms that target="_blank"


jcoyne:
Turbo seems to intercept these and render them in the same window.

Given this HTML
```html
<html>
  <head>
    <script type="module">
      import hotwiredTurbo from 'https://cdn.skypack.dev/@hotwired/turbo@7';
    </script>
  </head>
  <body>
    <form target="_blank" action="/foo" method="get">
      <button>Go</button>
    </form>
  </body>
</html>
```

when I click the button, a new window is not opened.

PR #1235, opened 2024-03-31
Ignore links and forms that target `"_blank"`

seanpdoyle:
Closes [#1171][]

Forms that target [_blank][form-target-blank] along with anchors that target [_blank][] navigate new tabs, and are therefore incompatible with Turbo's long-lived session.

This commit ensures that both `<a>` and `<form>` submissions that target `_blank` are ignored.

[#1171]: https://github.com/hotwired/turbo/issues/1171
[form-target-blank]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#target
[anchor-target-blank]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target

### DIFF
--- a/src/observers/form_submit_observer.js
+++ b/src/observers/form_submit_observer.js
@@ -1,3 +1,5 @@
+import { doesNotTargetIFrame } from "../util"
+
 export class FormSubmitObserver {
   started = false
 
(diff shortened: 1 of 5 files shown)


brunoprietog:
Thanks @seanpdoyle!

PR #1240, opened 2024-04-04
Re-structure `turbo-stream[action=morph]` support

seanpdoyle:
Closes https://github.com/hotwired/turbo/issues/1163
Follow-up to https://github.com/hotwired/turbo/pull/1185

[turbo-site PR](https://github.com/hotwired/turbo-site/pull/186)

This commit re-structures the new support for
`turbo-stream[action="morph"]` elements introduced in [#1185][].

Since the `<turbo-stream action="morph">` hasn't yet been part of a release, there's an opportunity to rename it without being considerate of backwards compatibility.

As an alternative to introduce a new Stream Action, this commit changes existing actions to be more flexible.

For example, the `<turbo-stream method="morph">` element behaves like a specialized version of a `<turbo-stream method="replace">`, since it operates on the target element's `outerHTML` property.

Similarly, the `<turbo-stream method="morph" children-only>` element behaves like a specialized version of a `<turbo-stream method="update">`, since it operates on the target element's `innerHTML` property.

```diff
-<turbo-stream action="morph">
+<turbo-stream action="replace" method="morph">
   <template>Replace me with morphing</template>
 </turbo-stream>

-<turbo-stream action="morph" children-only>
+<turbo-stream action="update" method="morph">
   <template>Update me with morphing</template>
 </turbo-stream>
```

This commit removes the `[action="morph"]` support entirely, and re-implements it in terms of the `[action="replace"]` and `[action="update"]` support.

By consolidating concepts, the "scope" of the modifications is more clearly communicated to callers that are familiar with the underlying DOM interfaces (`Element.replaceWith` and `Element.innerHTML`) that are invoked by the conventionally established Replace and Update actions.

This proposal also aims to reinforce the "method" terminology introduced by the Page Refresh `<meta name="turbo-refresh-method" content="morph">` element.

[#1185]: https://github.com/hotwired/turbo/pull/1185


seanpdoyle:
I started to open a PR related to https://github.com/hotwired/turbo-rails/pull/583 to remove `turbo_stream.morph` with extension to its `turbo_stream.replace` and `turbo_stream.update` counterparts, but wasn't sure whether or not it'd be worth removing those methods.

For example, the `turbo_stream.morph` method could serve as an abstraction on top of building a `<turbo-stream action="replace" method="morph">` in the same way as a `turbo_stream.morph(..., children_only: true)` call could be an abstraction on top of building a `<turbo-stream action="update" method="morph">`.

If this PR's proposal is well-received, I'll open a corresponding PR to `turbo-rails`.

seanpdoyle:
After doing some more research, I see that the original implementation introduced in https://github.com/hotwired/turbo/pull/1185 intentionally overlaps with the `turbo-stream[action="morph"]` element provided by the [turbo-morph](https://github.com/marcoroth/turbo-morph) package.

Since that package integrates with [morphdom](https://github.com/patrick-steele-idem/morphdom), support for the `[children-only]` attribute is a natural corollary to the `childrenOnly: Boolean` option that Morphdom supports.

Since Turbo 8's morphing is powered by Idiomorph, drawing inspiration from the Morphdom API isn't as compelling as it is for the `turbo-morph` package. 

Instead, utilizing existing Turbo terminology (instead of something like `[morph-style]` to mirror Idiomorph's `morphStyle: "innerHTML" | "outerHTML"` option) encapsulates that implementation detail in a way that makes it possible to change the underlying morphing library in the future without affecting downstream applications.

jonsgreen:
Not sure that my 2 cents are of value here but I do like the idea of having a different api for stream morphing that would not conflict with the `turbo-power` library. There are other features in `turbo-power` that are of value but I would prefer to use `turbo` with Idiomorph for morphing on a page and I believe that if `turbo_rails` were to use `turbo_stream.morph` it would collide with `turbo-power`.  

@seanpdoyle, I am curious if you have a sense of when this feature might be released? As other folks have mentioned in #1185 I have a pattern that I use in an app in which I post back to the same action (e.g. new or edit) to update the state of an object but without persisting it to the database so a redirect is not possible. I would like the UI to adjust as the state of the form object changes without doing a full replace since that would lose focus so morphing would be ideal. 

I think I did at least finally figure out how to get the full page render to trigger a morph refresh by passing a status code other than 200. Does that sound right? Tricky.

jorgemanrubia:
I love this @seanpdoyle. It makes the API tighter and more consistent. Thanks 🙏 . I would imagine this will require some adjustments after merging https://github.com/hotwired/turbo/pull/1234. Happy to tackle this one next. Thanks.

jorgemanrubia:
@seanpdoyle if you can rebase and adapt to latest changes I'll merge. Thanks!

seanpdoyle:
@jorgemanrubia I've rebased. I've also opened https://github.com/hotwired/turbo-site/pull/186 to document this change.

omarluq:
@seanpdoyle Im happy to take on the refactor on the turbo-rails side if you haven't started it yet

cc: @jorgemanrubia

jorgemanrubia (approved):
Thanks, @seanpdoyle. This is a great change 👍 .

searls:
Reading the play-by-play, this change makes a lot of sense to me. Update and Replace are the actions. `morph` vs `replace` are the methods by which that's accomplished. Comingling those interests into a single category (turbo stream actions) would have resulted in a bunch of confusion, option duplication, etc.

marcoroth:
Would it make sense to keep the `morph` action around as a shim and emit a warning saying that this action was "deprecated" in favor of using `replace` with a `method=morph`?

brunoprietog:
I don't think so, because no version was released with the Turbo stream morph action

searls:
@marcoroth, my understanding was it was never actually included in a release (the fact it didn't work is how I found these issues initially)

marcoroth:
Oh right, nevermind then. I guess I confused it with the `refresh` action which has nothing to do with this.

jorgemanrubia:
I just released [v8.0.5](https://github.com/hotwired/turbo/releases/tag/v8.0.5). I'll merge those site docs next.

searls:
Thanks! I'll try incorporating this next week and may [update this blog post](https://justin.searls.co/posts/recipe-swapping-out-a-model-div-with-turbo-and-stimulus/) I recently wrote

PR #1254, opened 2024-05-03
Read the csp meta tag nonce attribute and fall back to content

codergeek121:
This PR allows Turbo to support https://github.com/rails/rails/pull/51729. This changes Turbo to read the meta tag's `nonce` attribute first and fall back to the `content` attribute.

As described in https://github.com/rails/rails/issues/51580#issue-2246912613 this makes it harder to extract the nonce value.


codergeek121:
Great, thank you :)

PR #1256, opened 2024-05-06
Fix devcontainer configuration

radanskoric:
I wanted to try some changes to turbo source on to get myself started quickly I tried using the devcontainer setup but it didn't work. So instead I spent the time debugging that issue.

The devcontainer was no longer working since chrome version 115 because location of installation files changed. See comment on: https://chromedriver.chromium.org/downloads

However we don't need to build it ourselves since Microsoft maintains docker images with playwright preinstalled with chrome, firefox and webkit.

This means that we can simplify the setup a lot while still keeping the goal of allowing people to get an environment in which they can run tests by simply starting a devcontainer or running a Github codespace.

I've just tested this now by launching a github codespace from this here branch on my fork. After waiting a bit, I was able to just write `yarn test` and it worked. (~Strangely there were a few tests that failed.~ They were flaky, the runner re-ran them and they passed and where flagged as flaky by the runner)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,24 +1,28 @@
 // For format details, see https://aka.ms/devcontainer.json. For config options, see the README at:
 // https://github.com/microsoft/vscode-dev-containers/tree/v0.191.1/containers/typescript-node
 {
-  "name": "Node.js & TypeScript",
+  "name": "Node.js & Playwright",
   "build": {
     "dockerfile": "Dockerfile",
-    // Update 'VARIANT' to pick a Node version: 12, 14, 16
+    // Change to pick a different tag, see https://mcr.microsoft.com/en-us/product/playwright/tags
+    //
+    // IMPORTANT: The playwright image version must match @playwright/test version from package.json.
+    // Otherwise it will not work, the test version will look for the browser versions which it won't
+    // find on the image.
+    //
+    // pick '...-arm64' version if running on Apple Silicon.
     "args": {
-      "VARIANT": "16"
+      "PLAYWRIGHT_TAG": "v1.30.0-focal"
     }
   },
-  // Set *default* container specific settings.json values on container create.
-  "settings": {},
   // Add the IDs of extensions you want installed when the container is created.
-  "extensions": [
-    "dbaeumer.vscode-eslint"
-  ],
+  "customizations": {
+    "vscode": {
+      "extensions": ["dbaeumer.vscode-eslint"]
+    }
+  },
   // Use 'forwardPorts' to make a list of ports inside the container available locally.
-  // "forwardPorts": [],
+  "forwardPorts": [9000],
   // Use 'postCreateCommand' to run commands after the container is created.
-  "postCreateCommand": "yarn install && yarn build",
-  // Comment out connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
-  "remoteUser": "node"
+  "postCreateCommand": "yarn install && yarn build"
 }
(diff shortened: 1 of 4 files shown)


robzolkos (approved):

radanskoric:
@robzolkos is there anything else needed to merge this?

robzolkos:
> @robzolkos is there anything else needed to merge this?

Jut the attention of a maintainer.  This cleans things up a fair bit.

.devcontainer/devcontainer.json
```diff
@@ -1,24 +1,22 @@
...
   "build": {
     "dockerfile": "Dockerfile",
-    // Update 'VARIANT' to pick a Node version: 12, 14, 16
+    // Change to pick a different tag, see https://mcr.microsoft.com/en-us/product/playwright/tags
+    //
+    // IMPORTANT: The playwright image version must match @playwright/test version from package.json.
+    // Otherwise it will not work, the test version will look for the browser versions which it won't
+    // find on the image.
+    //
+    // pick '...-arm64' version if running on Apple Silicone.
```
brunoprietog, line 13:
```suggestion
    // pick '...-arm64' version if running on Apple Silicon.
```

.devcontainer/devcontainer.json
```diff
@@ -1,24 +1,22 @@
...
     "args": {
-      "VARIANT": "16"
+      "PLAYWRIGHT_TAG": "v1.28.0-focal"
     }
   },
-  // Set *default* container specific settings.json values on container create.
-  "settings": {},
-  // Add the IDs of extensions you want installed when the container is created.
-  "extensions": [
-    "dbaeumer.vscode-eslint"
```
brunoprietog, line 16: Any reason to remove this?

brunoprietog (changes requested):
Thanks for your contribution, love the simplicity! Just a couple of minor comments

.devcontainer/devcontainer.json
radanskoric: Thanks, fixed. :)

.devcontainer/devcontainer.json
radanskoric: It's a good thing you reminded me. The syntax for extensions has changed and VSCode was complaining about it so I removed it but meant to bring it back. 

I've now restored it with the correct syntax.

radanskoric:
@brunoprietog I've updated the PR. 

I also had to slightly bump @playwright/test to 1.30.0 because we already had playwright-core at both 1.28.0 and 1.30.0 through different dependency chains, causing issues. Unit test were trying to run with 1.28 and browser tests with 1.30 which is a problem when the docker image has just one playwright build.

Playwright should probably also be bumped to latest, 1.45.3. I'm happy to do that as well but I concluded it's best left for next PR, to limit the scope of this one.

.devcontainer/devcontainer.json
```diff
@@ -1,24 +1,28 @@
...
   },
-  // Set *default* container specific settings.json values on container create.
-  "settings": {},
   // Add the IDs of extensions you want installed when the container is created.
-  "extensions": [
-    "dbaeumer.vscode-eslint"
-  ],
+  "customizations": {
+    "vscode": {
+      // "extensions": ["dbaeumer.vscode-eslint"]
```
brunoprietog, line 21: Why would we leave this commented?

```suggestion
      "extensions": ["dbaeumer.vscode-eslint"]
```

brunoprietog:
Happy to see a PR updating Playwright! Maybe it will be helpful in reducing some flaky tests

brunoprietog (approved):
Thanks @radanskoric!

PR #1257, opened 2024-05-10
Fix test to verify no network request on prefetched link click

davidalejandroaguilar:
# Description

The test aimed to verify that no network request occurs when clicking a link that had already been prefetched by hovering. However, instead of performing a click, the test erroneously attempted to hover over the link again, which had already been hovered over. Therefore, it failed to test the intended behavior.

This PR makes sure that the test is correctly set up to prevent any regression in this functionality.

### DIFF
--- a/src/tests/functional/link_prefetch_observer_tests.js
+++ b/src/tests/functional/link_prefetch_observer_tests.js
@@ -246,11 +246,8 @@ test("it resets the cache when a link is hovered", async ({ page }) => {
 
 test("it does not make a network request when clicking on a link that has been prefetched", async ({ page }) => {
   await goTo({ page, path: "/hover_to_prefetch.html" })
-  await hoverSelector({ page, selector: "#anchor_for_prefetch" })
-
-  await sleep(100)
-
-  await assertNotPrefetchedOnHover({ page, selector: "#anchor_for_prefetch" })
+  await assertPrefetchedOnHover({ page, selector: "#anchor_for_prefetch" })
+  await assertRequestNotMadeOnClick({ page, selector: "#anchor_for_prefetch" })
 })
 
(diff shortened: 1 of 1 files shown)


src/tests/functional/link_prefetch_observer_tests.js
```diff
@@ -296,6 +293,13 @@ const assertNotPrefetchedOnHover = async ({ page, selector, callback }) => {
   assert.equal(requestMade, false, "Network request was made when it should not have been.")
 }
 
+const assertRequestNotMadeOnClick = async ({ page, selector }) => {
+  let requestMade = false
+  page.on("request", async (request) => (requestMade = true))
```
afcapel, line 298:
```suggestion
  page.on("request", async () => requestMade = true)
```

src/tests/functional/link_prefetch_observer_tests.js
```diff
@@ -296,6 +293,13 @@ const assertNotPrefetchedOnHover = async ({ page, selector, callback }) => {
...
+  await clickSelector({ page, selector })
+  assertRequestNotMade(requestMade)
```
afcapel, line 300: Can we add a callback to the `assertRequestMade` / `assertRequestNotMade` helpers so we can move the `requestMade` and event handler declaration to the helpers, so this looks like:

```js
assertRequestNotMade(async () => {
  await clickSelector({ page, selector })
})
```

davidalejandroaguilar: Sorry, just got free right now to tackle this.

Good call, love the resulting abstraction. LMK if this is what you had in mind.

afcapel:
Thanks @davidalejandroaguilar !

Issue #1258, opened 2024-05-13
The documentation does not mention that Stream Action tags can be rendered inside any HTML 


radanskoric:
The way Turbo is currently implemented, stream tags will be processed and executed when they are rendered anywhere in HTML. 

Here is a minimal example with a static HTML page:
```html
<html>
  <head>
    <script src="https://unpkg.com/@hotwired/turbo"></script>
  </head>
  <body>
    <turbo-stream action="append" target="list">
      <template>
        <li>list element</li>
      </template>
    </turbo-stream>

    This list has an element added via the stream action:
    <ul id="list">
    </ul>
  </body>
</html>
``` 

This will work anywhere where stream tags are rendered into the page DOM:
- Anywhere inside the initial page HTML.
- As part of the turbo frame response **as long as it is rendered inside the frame tag**. Since content from outside the frame is not rendered into the actual document, the tag needs to be *inside the frame* for this to work. 
- If any javascript at any point rendered a turbo stream element anywhere inside the document DOM.
- If a stream action renders HTML content (like it does in the example above) **and that content in turn also has a stream action tag inside it**, the second action tag will also work.

Not all of those are smart ways to use it but I was surprised that the documentation makes no mention of the feature. At least I couldn't find any mention anywhere in the documentation or old release notes. 

I am guessing that the reason is that this is a side-effect of how `StreamElements` are implemented: They are defined as [customElements](https://developer.mozilla.org/en-US/docs/Web/API/Window/customElements) with the stream tag execution being triggered by the `connectedCallback` implementation of the `StreamElement` class.

Perhaps it's just me and most other people find this clear without being mentioned but I think it would still be worth adding a few sentences and a documentation and a test to cover it. This can be a very useful feature in some cases.

I'd be very happy to mention this in the documentation but wanted to first check if there is something I am missing? 

Can someone confirm that this is an intentional Turbo feature that should be mentioned explicitly in the documentation? If not, can we make it an official feature. Considering how it's implemented, one would have to do add explicit hacks just to stop it from working.

seanpdoyle:
> Can someone confirm that this is an intentional Turbo feature that should be mentioned explicitly in the documentation?

I've personally relied on the behavior that you've described. For example, you can nest a `<turbo-stream>` element inside a `<template>` element, then append that content to the page after a `<button>` click ([try it out on JSFiddle](https://jsfiddle.net/ptq8sw6c/)):

```html
<script>
  addEventListener("click", ({ target }) => {
[...]

PR #1263, opened 2024-05-24
Add a test that stream actions rendered into the HTML are executed

radanskoric:
Resolves: https://github.com/hotwired/turbo/issues/1258
It's paired with: https://github.com/hotwired/turbo-site/pull/192

As a consequence of the fundamental way in which Stream actions are implemented, they can also be executed by rendering them within any HTML that's included on the dom.

Since this is already being used as the feature in the community, adding this test will ensure this keeps working.

### DIFF
--- a/src/tests/functional/stream_tests.js
+++ b/src/tests/functional/stream_tests.js
@@ -226,6 +226,25 @@ test("preventing a turbo:before-morph-element prevents the morph", async ({ page
   await expect(page.locator("#message_1")).toHaveText("Morph me")
 })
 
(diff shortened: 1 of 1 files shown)


brunoprietog (approved):
Thanks @radanskoric!

radanskoric:
Thank you @brunoprietog!  

What do you think are the chances of the matching documentation update being merged? 
That's the PR linked from this one: https://github.com/hotwired/turbo-site/pull/192 (it had a merge conflict which I resolved now)

brunoprietog:
Merged!

Issue #1264, opened 2024-05-24
turbo-confirm doesn't work on links without a turbo-method


JaceBayless:
### Expected behavior
According to the documentation [here](https://turbo.hotwired.dev/handbook/drive#requiring-confirmation-for-a-visit) you can use `data-turbo-confirm` on an `a` tag to prevent navigation without a confirmation. 

>```<a href="/articles" data-turbo-confirm="Do you want to leave this page?">Back to articles</a>```

### Actual behavior
Trying to confirm on navigation doesn't work unless a `turbo-method` is specified. Here's a codepen replicating [this](https://codepen.io/jacebayless/pen/WNBGrOe).

tpaulshippy:
Looking at fixing this. Reproduced with tests here: https://github.com/tpaulshippy/turbo/tree/issue-1264

tpaulshippy:
Care to comment on my PR with your particular use case?

PR #1266, opened 2024-05-26
Fix turbo-confirm for links without a turbo-method

tpaulshippy:
Resolves https://github.com/hotwired/turbo/issues/1264

It appears that the turbo-confirm feature implementation relies on the link being captured by the FormLinkClickObserver rather than the LinkClickObserver. Without a data-turbo-method (or data-turbo-stream) attribute, the FormLinkClickObserver will ignore the link and the confirm feature will not fire.

This small bug fix adds the data-turbo-confirm attribute to the other two attributes that indicate to the FormLinkClickObserver that it should handle the link.

### DIFF
--- a/src/observers/form_link_click_observer.js
+++ b/src/observers/form_link_click_observer.js
@@ -30,7 +30,7 @@ export class FormLinkClickObserver {
   willFollowLinkToLocation(link, location, originalEvent) {
     return (
       this.delegate.willSubmitFormLinkToLocation(link, location, originalEvent) &&
-      (link.hasAttribute("data-turbo-method") || link.hasAttribute("data-turbo-stream"))
+      (link.hasAttribute("data-turbo-method") || link.hasAttribute("data-turbo-stream") || link.hasAttribute("data-turbo-confirm"))
     )
   }
 
--- a/src/tests/fixtures/drive.html
+++ b/src/tests/fixtures/drive.html
@@ -11,6 +11,8 @@
 
     <div>
       <a id="drive_enabled" href="/src/tests/fixtures/drive.html">Drive enabled link</a>
+      <a id="drive_enabled_with_confirm_without_method" href="/src/tests/fixtures/drive.html?confirmed=true" data-turbo-confirm="Are you sure?">Drive enabled link with confirm</a>
+      <a id="drive_enabled_with_confirm_and_method" data-turbo-method="GET" href="/src/tests/fixtures/drive.html?confirmed=true" data-turbo-confirm="Are you sure?">Drive enabled link with confirm</a>
       <a id="drive_enabled_external" href="https://example.com">Drive enabled external link</a>
     </div>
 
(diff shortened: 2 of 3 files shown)


tpaulshippy:
Seeing one flaky test failure. 

  [firefox] › functional/navigation_tests.js:432:1 › does not fire turbo:load twice after following a redirect 

But it looks like this is happening in the main branch as well.

tpaulshippy:
https://github.com/hotwired/turbo/pull/379#issuecomment-913811333
Looks like this issue was intended and by design. If this is still the case, perhaps we just need an update to https://github.com/hotwired/turbo-site/blob/main/_source/handbook/02_drive.md with this detail.

pinzonjulian:
The original intention of the confirm dialog as I understand if is to assist the user about operations that will modify the state of the system. In the case of the web that's requests with a PUT/PATCH/DELETE verb. 

A GET request, which is how links work by default, shouldn't modify your system so a confirmation seems odd. 

When you explicitly tell a link to use another method (a technique that's more and more uncommon these days) then you are allowed to ask the user for confirmation. 

What's your use case for asking a user for confirmation on a GET request?

tpaulshippy:
I don't personally have a use case. I'm sure @JaceBayless does. 

But the one documented at https://turbo.hotwired.dev/handbook/drive#requiring-confirmation-for-a-visit seems reasonable enough to me: "Do you want to leave this page?"

tpaulshippy:
Actually a very common use case just happened for me. I started to edit my comment, and then decided to make a new comment. When I clicked "Cancel," Github asked me to confirm that I wanted to discard my changes.

pinzonjulian:
You're right! That's a great use case.

tpaulshippy:
The more I think about it, the more I think this implementation doesn't really fit that particular use case. For a couple of reasons: 1) You generally want to do a dirty check in these cases which would usually require additional Javascript before deciding to confirm. 2) You would want the confirm to fire on close of the window/tab as well as any navigation away so a beforeunload event would better cover that. Would be nice if Turbo could handle that but I'm not sure how.

There still could be other use cases for a confirm on navigate though. Maybe you have a set of filter controls and the user has built up a set of filters. Then you have a Clear Filters button that just navigates back to the list action without query strings. You might only show that button if query strings exist (the "dirty check"). In that case you probably wouldn't want to confirm on close of the window because there isn't really a potential for data loss.

I'm sure there are others.

JaceBayless:
In what I was doing, I didn't have a great use case. I was testing some things out and noticed that the documentation didn't match the actual behavior. 

FWIW I think one of the main uses for this would be if you're trying to prevent people from navigating to external sites, generally they do this with a warning message.

marcoroth:
FWIW, this is a deplicate of https://github.com/hotwired/turbo/pull/874

tpaulshippy:
True. The implementation approach is quite different though. How would you compare them?

marcoroth:
I'm wondering if your approach also works for form submissions

tpaulshippy:
Were form submissions broken? I thought it was just links.

tpaulshippy:
I think my approach just turns the links into form submissions if they have a confirmation.

tpaulshippy:
@marcorMar still curious about your opinion of this approach.

tpaulshippy:
@marcoroth 
Still curious about your opinion of this approach.

brendon:
This approach seems the cleanest. Using `data-turbo-confirm` implies that the user wants to invoke a form based approach to visiting the link. Providing `turbo-method="get"` is a good workaround but this is the default method for a link so I guess it's about whether people are surprised by needing to state it explicitly for the confirm to kick in.

PR #1267, opened 2024-05-30
Don't lose focus due to autofocus when morphing pages

jorgemanrubia:
Before this change, Turbo would always focus on the first element with `[autofocus]` when a page refresh with morphing happened  (default behavior inherited from `PageRenderer`). With this change, it will never autofocus when a page refresh with morphing happens.

This is meant to solve the problem where you are writing on a form, and it loses the focus because a broadcasted page refresh arrives.

### DIFF
--- a/src/core/drive/morph_renderer.js
+++ b/src/core/drive/morph_renderer.js
@@ -11,6 +11,10 @@ export class MorphRenderer extends PageRenderer {
     return "morph"
   }
 
+  get shouldAutofocus() {
+    return false
+  }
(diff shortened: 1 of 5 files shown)


jorgemanrubia:
I'll look into those tests...

afcapel (approved):
Looks good to me, nice one @jorgemanrubia! 🙌

jorgemanrubia:
Thanks for helping with those tests @afcapel 🙏

PR #1277, opened 2024-06-12
Update 422 status name to "Unprocessable Content"

janko:
I don't know when this changed, but it's now the official name for this response status. I found out about this when Rack 3.1+ stopped supporting `:unprocessable_entity` in favor of `:unprocessable_content`.


janko:
OK, I see now that autocorrect kicked in in my text editor, I will undo this next week.

dhh:
Please reopen when/if you get time to do a clean PR 👍

PR #1281, opened 2024-06-19
Support back button (restore) even if page loaded in frame does not contain matching HEAD

### LINKED ISSUE #1241: Turbo frame history not navigating properly w/ back button after upgrading from 7 to 8
**Context:**

Application has a simple live search via form:
- Form uses "get" method (just showing data; no updates)
- Form targets a "results" turbo frame
- Form uses data-turbo-action="advance" to update URL (to support users being able to copy/paste w/ params for the search and filter)

**What's happening:** Live search is working fine. When the form is submitted (automatically as a user types), the turbo frame is updated with the correct results and the URL displays the correct query parameters. However, when navigating with the back button the turbo frame does not navigate backwards, it only shows the most recent set of results (despite the url and associated query params appropriately changing backwards).

**What should be happening:** As the user navigates backwards the URL query params should change (which is happening) and the turbo frame should reflect the history from that point in time.

**When did this start happening:** This happened on Turbo 8.0.4. and Turbo Rails 2. If I revert to Turbo < 8 and Turbo Rails < 2, the behavior works as expected.

tpaulshippy:
## Purpose

Resolves #1241 

## Approach

Set the location on the session even if the `PageView` does not render because the `trackedElementsAreIdentical` is false.

## Test

This test illustrates the issue, once you remove line 6 from `three.html`
`npx playwright test -g "promoted frame navigations are cached"`

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -270,6 +270,8 @@ export class FrameController {
 
   viewRenderedSnapshot(_snapshot, _isPreview, _renderMethod) {}
 
+  setLastRenderedLocation() {}
+
   preloadOnLoadLinksForView(element) {
     session.preloadOnLoadLinksForView(element)
   }
--- a/src/core/session.js
+++ b/src/core/session.js
@@ -327,8 +327,12 @@ export class Session {
   }
 
   viewRenderedSnapshot(_snapshot, _isPreview, renderMethod) {
-    this.view.lastRenderedLocation = this.history.location
+    this.setLastRenderedLocation()
     this.notifyApplicationAfterRender(renderMethod)
+  }
+
+  setLastRenderedLocation() {
+    this.view.lastRenderedLocation = this.history.location
   }
 
   preloadOnLoadLinksForView(element) {
--- a/src/core/view.js
+++ b/src/core/view.js
@@ -84,6 +84,8 @@ export class View {
       }
     } else if (shouldInvalidate) {
       this.invalidate(renderer.reloadReason)
+    } else {
+      this.delegate.setLastRenderedLocation();
     }
   }
 
--- a/src/tests/fixtures/tabs/three.html
+++ b/src/tests/fixtures/tabs/three.html
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8">
     <title>Frame</title>
-    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
     <script src="/src/tests/fixtures/test.js"></script>
   </head>
   <body>


tleish:
Would a better option for testing this would be the `src/tests/functional/frame_navigation_tests.js`. 

Something like: https://github.com/hotwired/turbo/compare/main...tleish:turbo:turbo-frame-back-navigation?expand=1

tpaulshippy:
> promoted frame navigations are cached

It looks like you added `goBack` testing to `frame navigation with data-turbo-action` whereas I removed the `<script>` tag from `three.html` and tested `promoted frame navigations are cached`. 

I think I do like your testing approach a bit better. Will try to switch to it when I get a chance. The only piece I didn't quite understand is where you removed lines 114-117. Maybe you could help me understand that change?

Have you found a better way to resolve the issue?

tleish:
> The only piece I didn't quite understand is where you removed lines 114-117. Maybe you could help me understand that change?

Ignore those changes,I must’ve cut-paste instead of copy-paste. 
 
> Have you found a better way to resolve the issue?

The test is as far as I got.

tpaulshippy:
Confirmed that your testing approach accurately reproduced the issue and was resolved by the proposed implementation. Pushed. Thanks @tleish
Looking forward to seeing any alternative implementation ideas. I'm not convinced this is the best possible way to resolve the issue.

tleish:
I believe the fix is a bit simpler.  It appears with a turbo navigation that includes an action, turbo does the following:

1. Calls `FrameController#loadFrameResponse` where it loads and updates the frame on page
2. During this method, it calls `await this.fetchResponseLoaded(fetchResponse);` which is defined inside of `FrameController#proposeVisitIfNavigatedWithAction`
3. `frame.delegate.fetchResponseLoaded` defined within `FrameController#proposeVisitIfNavigatedWithAction`simulates a "Turbo#visit" (last line) wherein the PageSnapshot is updated with the HTML from the mocked `response` object.  Instead of setting the HTML to be the frame html, we could set it to be the current/updated page HTML (which has the frame contents loaded at this point).  Doing this means that the PageSnapshot.headSnapshot.trackedElementSignature will always match when the logic compares the history for back navigation.

see: https://github.com/hotwired/turbo/compare/main...tleish:turbo:turbo-frame-back-navigation?expand=1

tpaulshippy:
That's a direct revert of https://github.com/hotwired/turbo/commit/ed72ba16bc23dcccd88e0ab76ba72cfef82cd599 
I guess I wonder why that change was made.

tpaulshippy:
I'm happy to switch to that implementation in this PR if you think it's a better approach.

tleish:
> That's a revert of part of [ed72ba1](https://github.com/hotwired/turbo/commit/ed72ba16bc23dcccd88e0ab76ba72cfef82cd599) 

Hah!  You mentioned this before on the issue thread.  I didn't look at it, but you are right.  It's a revert of this commit.

tpaulshippy:
Looks like this was discussed here: https://github.com/hotwired/turbo/pull/887#discussion_r1376602627

fonji:
This seems to continue to be a problem, what can we do to get this merged?

PR #1284, opened 2024-07-09
Allow setting the request method for Turbo Frame requests to avoid 414 Long URL errors

ghiculescu:
Currently the request that a Turbo Frame makes to load a page is hardcoded to [a GET request](https://github.com/hotwired/turbo/blob/14284e6a8d2ffdf8c6beac61d26373f18491ba72/src/core/frames/frame_controller.js#L317). We came across an edge case where [HTTP 414s](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/414) were being thrown due to a very large request URL, in a complex form that heavily uses frames with URLs to hold form state.

The general approach to handling 414s is to switch to a POST. This PR adds the ability to declaratively do that on the frame element:

```html
<turbo-frame id="stuff" src="/my/url" method="post" params='{"required":true,"type":"string"}'></turbo-frame>
```

You need to provide: 

- `method`= `"post"`
- `params` = Something that [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) can interpret: stringified JSON, or form params.

This will make a POST to `/my/url`, with the provided `params` in the POST body.

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -314,7 +314,7 @@ export class FrameController {
   }
 
   async #visit(url) {
-    const request = new FetchRequest(this, FetchMethod.get, url, new URLSearchParams(), this.element)
+    const request = new FetchRequest(this, this.visitMethod, url, this.visitRequestBody, this.element)
 
     this.#currentFetchRequest?.cancel()
     this.#currentFetchRequest = request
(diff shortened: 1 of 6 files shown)


brunoprietog:
This is beyond the scope of what a Turbo frame should be. Thanks for your contribution, but I'm closing for now.

PR #1287, opened 2024-07-24
Re-connect Stream Source when attribute change

seanpdoyle:
Related to [hotwired/turbo-rails#638][]

Recent changes to integrate with morphing have altered the mental model for some Turbo custom elements, including the
`<turbo-stream-source>` element.

Custom Elements' `connectedCallback()` and `disconnectedCallback()` (along with Stimulus' `connect()` and `disconnect()`) improved upon invoking code immediately, or listening for `DOMContentLoaded` events.

There are similar improvements to be made to integrate with morphing. First, [observe attribute changes][] by declaring their own `static observedAttributes` properties along with
`attributeChangedCallback(name, oldValue, newValue)` callbacks. Those callbacks execute the same initialization code as their current `connectedCallback()` and `disconnectedCallback()` methods.

That'll help resolve this issue. In addition to those changes, it's important to emphasize this pattern for consumer applications moving forward. JavaScript code (whether Stimulus controller or otherwise) should be implemented in a way that' resilient to both asynchronous connection and disconnection *as well as* asynchronous modification of attributes.

[hotwired/turbo-rails#638]: https://github.com/hotwired/turbo-rails/issues/638
[observe attribute changes]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Components/Using_custom_elements#responding_to_attribute_changes

### DIFF
--- a/src/elements/stream_source_element.js
+++ b/src/elements/stream_source_element.js
@@ -1,19 +1,30 @@
 import { connectStreamSource, disconnectStreamSource } from "../core/index"
 
 export class StreamSourceElement extends HTMLElement {
+  static observedAttributes = ["src"]
+
   streamSource = null
 
   connectedCallback() {
     this.streamSource = this.src.match(/^ws{1,2}:/) ? new WebSocket(this.src) : new EventSource(this.src)
 
     connectStreamSource(this.streamSource)
+    this.setAttribute("connected", "")
   }
 
   disconnectedCallback() {
+    this.removeAttribute("connected")
     if (this.streamSource) {
       this.streamSource.close()
 
       disconnectStreamSource(this.streamSource)
+    }
+  }
+
+  attributeChangedCallback() {
+    if (this.streamSource) {
+      this.disconnectedCallback()
+      this.connectedCallback()
     }
   }
 
--- a/src/tests/functional/stream_tests.js
+++ b/src/tests/functional/stream_tests.js
@@ -2,6 +2,7 @@ import { expect, test } from "@playwright/test"
 import { assert } from "chai"
 import {
   hasSelector,
+  nextAttributeMutationNamed,
   nextBeat,
   nextEventNamed,
   nextEventOnTarget,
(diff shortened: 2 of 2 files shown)


seanpdoyle:
@jorgemanrubia I've opened this in response to https://github.com/hotwired/turbo-rails/issues/638. In addition to re-connected when the `[src]` changes, this commit also communicates connection state for the `<turbo-stream-source>` element through a `[connected]` attribute in the same style as `<turbo-cable-stream-source>` element.

freund17:
Hi, I was about to create an issue about this, but it seems you beat me to it. 😉 

Keep in mind, that both, `connectedCallback` _and_ `attributeChangedCallback`, will be triggered on first render - so your implementation would connect, disconnect and connect again in that case. This may break use-cases (such as mine) where the SSE URL is single-use.

This is my solution: (feel free to use, MIT licence, yada yada)

```ts
class StreamSourceElement extends HTMLElement {
  static get observedAttributes() {
    return ["src"];
  }

  #streamSource: EventSource | WebSocket | undefined;

  connectedCallback() {
    const src = this.getAttribute("src");

    if (src !== null) {
      this.#connectStreamSource(src);
    }
  }

  attributeChangedCallback() {
    const src = this.getAttribute("src");

    if (src === null) {
      this.#disconnectStreamSource();
    } else {
      this.#connectStreamSource(src);
    }
  }

  disconnectedCallback() {
    this.#disconnectStreamSource();
  }

  #connectStreamSource(src: string) {
    if (this.#streamSource?.url !== src) {
      this.#disconnectStreamSource();

      this.#streamSource = src.match(/^wss?:/)
        ? new WebSocket(src)
        : new EventSource(src);

      connectStreamSource(this.#streamSource);
    }
  }

  #disconnectStreamSource() {
    if (this.#streamSource) {
      this.#streamSource.close();

      disconnectStreamSource(this.#streamSource);

      this.#streamSource = undefined;
    }
  }

  get src() {
    return this.getAttribute("src") || ""
  }
}
```

PR #1302, opened 2024-08-23
Fix temporary elements disappearing on subsequent GET requests that redirect to a page with data-turbo-temporary elements.

### LINKED ISSUE #1301: Temporary elements are removed after subsequent visits of a GET redirect
[This change](https://github.com/hotwired/turbo/commit/db310ef15ed9e4d0d7b7990220a20abe165ea993#diff-49232cc7df722abcf7e1e45ef4e2598d022895ae6647a0d0ccf3db7d9705b20cL421-R421) introduced a regression. If you have a `data-turbo-temporary` element appear on a page after a GET request that redirects to another page (such as a Rails flash message on a redirect), the temporary element will be shown on the first view, then will be hidden on subsequent visits of the same redirected GET request.

This can be demonstrated in [this sample Rails app](https://github.com/mikepayready/investigate_turbo_cache).

I have a fix with a test case incoming, though I don't know if it causes regressions on [this issue](https://github.com/hotwired/turbo/pull/1102) (referenced from the original commit that introduced this regression).

mikepayready:
Fixes #1301

### DIFF
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -420,7 +420,8 @@ export class Visit {
   async render(callback) {
     this.cancelRender()
     await new Promise((resolve) => {
-      this.frame = requestAnimationFrame(() => resolve())
+      this.frame =
+        document.visibilityState === "hidden" ? setTimeout(() => resolve(), 0) : requestAnimationFrame(() => resolve())
     })
     await callback()
     delete this.frame
--- a/src/core/drive/visit.js
+++ b/src/core/drive/visit.js
@@ -1,7 +1,7 @@
 import { FetchMethod, FetchRequest } from "../../http/fetch_request"
 import { getAnchor } from "../url"
 import { PageSnapshot } from "./page_snapshot"
-import { getHistoryMethodForAction, uuid, nextRepaint } from "../../util"
+import { getHistoryMethodForAction, uuid } from "../../util"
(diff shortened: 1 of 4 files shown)


src/core/drive/visit.js
```diff
@@ -419,7 +419,9 @@ export class Visit {
 
   async render(callback) {
     this.cancelRender()
-    this.frame = await nextRepaint()
+    await new Promise((resolve) => {
+      this.frame = requestAnimationFrame(() => resolve())
```
brunoprietog, line 423:
```suggestion
      if (document.visibilityState === "hidden") {
        this.frame = setTimeout(() => resolve(), 0)
      } else {
        this.frame = requestAnimationFrame(() => resolve())
      }
```

brunoprietog (changes requested):
Good catch. In order not to introduce a regression, we should incorporate the bifurcation here as well.

src/core/drive/visit.js
brunoprietog: Or shorter:

```suggestion
      this.frame =
        document.visibilityState === "hidden" ? setTimeout(() => resolve(), 0) : requestAnimationFrame(() => resolve())
```

brunoprietog (approved):
Thanks @mikepayready!

PR #1303, opened 2024-08-25
Bugfix: Don't use Bardo during morphing frame render (its only for non-morphing renders)

botandrose:
Using the turbo frame morph renderer with permanent elements results in incorrect morphs. On v8.0.5, I'm seeing duplicate elements, missing elements, and other strange behavior. This took a while to track down!

It turns out that the `data-turbo-permanent` functionality is implemented via two different strategies: `Bardo` for non-morphing renderers, and `DefaultIdiomorphCallbacks` for morphing renders.

`MorphingFrameRenderer`'s parent class `FrameRenderer` uses `Bardo`, so `MorphingFrameRenderer` continues to use that _in addition_ to `DefaultIdiomorphCallbacks`, resulting in incorrect morphs.

Looking at how the morphing and non-morphing page renderers handle this strategy switch, `MorphingPageRenderer` works correctly, because its overwriting its `preservingPermanentElements` function with a no-op, so let's just copy that strategy in `MorphingFrameRenderer`, et voila.

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -93,6 +93,7 @@ export class FrameController {
     if (this.element.shouldReloadWithMorph) {
       this.element.addEventListener("turbo:before-frame-render", ({ detail }) => {
         detail.render = MorphingFrameRenderer.renderElement
+        detail.preservingPermanentElements = MorphingFrameRenderer.preservingPermanentElements
       }, { once: true })
     }
 
@@ -266,11 +267,16 @@ export class FrameController {
 
     const {
       defaultPrevented,
-      detail: { render }
+      detail: { render, preservingPermanentElements }
     } = event
 
-    if (this.view.renderer && render) {
-      this.view.renderer.renderElement = render
+    if (this.view.renderer) {
+      if (render) {
+        this.view.renderer.renderElement = render
+      }
+      if (preservingPermanentElements) {
+        this.view.renderer.preservingPermanentElements = preservingPermanentElements
+      }
     }
 
     return !defaultPrevented
--- a/src/core/frames/morphing_frame_renderer.js
+++ b/src/core/frames/morphing_frame_renderer.js
@@ -11,4 +11,7 @@ export class MorphingFrameRenderer extends FrameRenderer {
 
     morphChildren(currentElement, newElement)
   }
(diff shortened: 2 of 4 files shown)


seanpdoyle (changes requested):
Thank you. This implementation change makes sense. 

Could you add a test case to ensure that permanent elements are not duplicated during frame morphing?

brunoprietog:
Are you sure that this really resolves the error? Unless I'm missing something, this doesn't produce any change, since no instance of MorphingFrameRenderer is created yet. Only the static method is used separately here:

https://github.com/hotwired/turbo/blob/f48b0de1d541055f959aa6d37d96f2441cffb9a6/src/core/drive/morphing_page_renderer.js#L42-L48

Or am I missing something?

botandrose:
@seanpdoyle Yes, I looked into adding a test, but it looks like no morphing frame rendering tests exist at all at the moment, so there didn't seem to be a reasonable place to add one. If you want to gate merging this on a test, let's wait until  #1192 lands, and then I'll rebase and add one.

botandrose:
@brunoprietog Right you are, I've been manually creating an instance of MrophingFrameRender to refresh a turbo frame with morphing. I'm really looking forward to #1192 landing so I don't have to do that anymore!

botandrose:
@seanpdoyle @brunoprietog Okay, now that #1192 has landed, I've rebased and added a test to assert that a permanent element remains after a `frame[refresh=morph]` reload. This test exposes the issue described above by failing after refresh, complaining that the permanent element now has a duplicate: `locator('#permanent-input') resolved to 2 elements:`. This is because Bardo and DefaultIdiomorphCallbacks are both cooking in the kitchen. I'll push the fix as a separate commit now.

botandrose:
@seanpdoyle @brunoprietog Okay, I got the test passing by disabling Bardo for morphing frame renders, as described above. However, I've had to completely change the implementation strategy here, because #1192 has since turned the `MorphingFrameRenderer` subclass into a bag of methods to monkeypatch onto the current renderer, rather than using an instance of it to perform the render. This isn't the approach I would choose, personally, but I'm sure there are good reasons. I'm guessing this is a transitionary state of affairs?

botandrose:
Closed in favor of #1028

PR #1305, opened 2024-08-26
Upgrade playwright to 1.46.1

radanskoric:
It upgrades it in package.json and in the devcontainer setup. 

It's now set to a specific version in package.json because it needs to exactly match the playwright version which is declared in the devcontainer setup.


radanskoric:
If someone is looking at this and would like to help merge this: it's almost done but there are a few failing tests after the upgrade. I've fixed one but I'm a bit stuck with the others. If you have an idea of what it might be please share to help me get this update finished.

brunoprietog:
Thanks @radanskoric for your efforts! This was finally done in #1400

PR #1306, opened 2024-08-28
Delegate `Turbo.session` properties to `Turbo.config`

seanpdoyle:
Follow-up to comment on [238ec2688b2f4907e05ad1eaeb011e93bae75995][]

[hotwired/turbo#1216][] unintentionally removed support for the `Turbo.session.drive` boolean property and the `Turbo.session.formMode` string property.

This commit re-introduces that support by delegating reads and writes to those properties to their corresponding `Turbo.config` versions.

[hotwired/turbo#1216]: https://github.com/hotwired/turbo/pull/1216
[238ec2688b2f4907e05ad1eaeb011e93bae75995]: https://github.com/hotwired/turbo/commit/238ec2688b2f4907e05ad1eaeb011e93bae75995#commitcomment-145954235

### DIFF
--- a/src/core/session.js
+++ b/src/core/session.js
@@ -144,6 +144,22 @@ export class Session {
     return config.drive.progressBarDelay
   }
 
+  set drive(value) {
+    config.drive.enabled = value
+  }
+
+  get drive() {
+    return config.drive.enabled
+  }
+
+  set formMode(value) {
+    config.forms.mode = value
+  }
+
+  get formMode() {
+    return config.forms.mode
+  }
+
   get location() {
     return this.history.location
   }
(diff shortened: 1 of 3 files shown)


seanpdoyle:
@jorgemanrubia and @brunoprietog, could one of you review this change? @botandrose caught it and reported it on the original commit.

brunoprietog:
Right! Could you add deprecation warnings please?

seanpdoyle:
@brunoprietog I'm happy to do that. 

It's worth mentioning that the original PR only deprecated the *methods* like `setProgressBarDelay` but did not deprecate the new properties like `Turbo.session. progressBarDelay`.

I think deprecating the `Turbo.session` properties in favor of `Turbo.config` properties makes sense, but could that be done separately and across all properties?

brunoprietog:
Ah, yes, makes sense. Thanks @seanpdoyle!

PR #1311, opened 2024-09-15
Reloading a morphing frame should trigger reloads on its child morphing frames recursively

botandrose:
Hi folks, here's what seems to me to be the next obvious step after #1192 changed `reload()` to morph a morphing frame.

## Example scenario
Imagine the following page structure: a morphing page with a morphing frame that itself includes another morphing frame, i.e.
```
Morphing Page > Morphing Frame Child > Morphing Frame Grandchild
```

## Status Quo
Refreshing `Morphing Page` already triggers morphing `reload()`s on both the `Morphing Frame Child` and on `Morphing Frame Grandchild`. However, manually triggering `reload()` on `Morphing Frame Child` does NOT trigger a `reload()` on `Morphing Frame Grandchild`, but rather does a full non-morphing render. This seems inconsistent and surprising.

## With this PR
Morphing pages now only trigger `reload()`s on their immediate morphing frame children, and then leaves it up to them to trigger `reload()`s on their own immediate morphing frame children, and so on down. This makes the reload behavior consistent no matter where the reload is initially triggered.

What do you think?

### DIFF
--- a/src/core/drive/morphing_page_renderer.js
+++ b/src/core/drive/morphing_page_renderer.js
@@ -6,17 +6,15 @@ export class MorphingPageRenderer extends PageRenderer {
   static renderElement(currentElement, newElement) {
     morphElements(currentElement, newElement, {
       callbacks: {
-        beforeNodeMorphed: element => {
-          return !super.shouldRefreshChildFrameWithMorphing(null, element)
+        beforeNodeMorphed: (node, newNode) => {
+          if (super.shouldRefreshChildFrameWithMorphing(null, node, newNode)) {
+            node.reload()
+            return false
+          }
+          return true
         }
       }
     })
-
-    for (const frame of currentElement.querySelectorAll("turbo-frame")) {
-      if (super.shouldRefreshChildFrameWithMorphing(null, frame)) {
-        frame.reload()
-      }
-    }
 
     dispatch("turbo:morph", { detail: { currentElement, newElement } })
   }
--- a/src/core/drive/morphing_page_renderer.js
+++ b/src/core/drive/morphing_page_renderer.js
@@ -1,13 +1,16 @@
 import { PageRenderer } from "./page_renderer"
 import { dispatch } from "../../util"
-import { morphElements } from "../morphing"
+import { morphElements, shouldRefreshFrameWithMorphing, closestFrameReloadableWithMorphing } from "../morphing"
 
 export class MorphingPageRenderer extends PageRenderer {
   static renderElement(currentElement, newElement) {
     morphElements(currentElement, newElement, {
       callbacks: {
         beforeNodeMorphed: (node, newNode) => {
-          if (super.shouldRefreshChildFrameWithMorphing(null, node, newNode)) {
+          if (
+            shouldRefreshFrameWithMorphing(node, newNode) &&
+              !closestFrameReloadableWithMorphing(node)
+          ) {
             node.reload()
             return false
           }
--- a/src/core/drive/morphing_page_renderer.js
+++ b/src/core/drive/morphing_page_renderer.js
@@ -35,5 +35,6 @@ function canRefreshFrame(frame) {
   return frame instanceof FrameElement &&
     frame.src &&
     frame.refresh === "morph" &&
-    !frame.closest("[data-turbo-permanent]")
+    !frame.closest("[data-turbo-permanent]") &&
+    !frame.parentElement.closest("turbo-frame[src][refresh=morph]")
 }
--- a/src/core/drive/morphing_page_renderer.js
+++ b/src/core/drive/morphing_page_renderer.js
@@ -1,4 +1,3 @@
-import { FrameElement } from "../../elements/frame_element"
 import { PageRenderer } from "./page_renderer"
 import { dispatch } from "../../util"
 import { morphElements } from "../morphing"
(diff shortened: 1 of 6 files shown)


botandrose:
Just pushed another two commits that expose and fix an adjacent bug that's been in Turbo for a while. Page refreshes didn't check to see if the new version of the page actually has the frame before calling `reload()` on it. So if a page refresh's new version would have different frames, the original frames would be "stuck" in place. This is fixed by checking to see if the replacement element actually is the same turbo frame before calling `reload()`. Otherwise, it just gets morphed as normal.

botandrose:
@jorgemanrubia And this is the other morphing-related improvement PR!

I have more in the can than just these two, but these two are the big ones that really unlocked Turbo for my app.

P.S. I think the red test is flaky and unrelated. I can't get it to fail locally, and it seems to be unrelated to this change. I'm guessing a rerun would flip it.
```
    [firefox] › functional/frame_tests.js:458:5 › submitting a form that redirects to a page with a <turbo-frame recurse> which lazily loads a matching frame 
```

src/core/renderer.js
```diff
@@ -7,6 +9,19 @@ export class Renderer {
...
   }
 
+  static shouldRefreshChildFrameWithMorphing(parentFrame, currentFrame, newFrame) {
+    return currentFrame instanceof FrameElement &&
+      // newFrame cannot yet be an instance of FrameElement because custom
+      // elements don't get initialized until they're attached to the DOM, so
+      // test its Element#nodeName instead
+      newFrame instanceof Element && newFrame.nodeName === "TURBO-FRAME" &&
+      currentFrame.shouldReloadWithMorph &&
+      currentFrame.id === newFrame.id &&
```
jorgemanrubia, line 19: Isn't this ID check redundant? idiomorph won't match nodes with different IDs, right?

src/core/renderer.js
```diff
@@ -7,6 +9,19 @@ export class Renderer {
...
+      (!newFrame.getAttribute("src") || urlsAreEqual(currentFrame.src, newFrame.getAttribute("src"))) &&
```
jorgemanrubia, line 20: Why the "urls are equal" condition?

jorgemanrubia: Asking because, if we could get away without comparing URL and IDs, then we could just rely on making the check a single-param function (currentFrame).

src/core/drive/morphing_page_renderer.js
```diff
@@ -7,14 +6,16 @@ export class MorphingPageRenderer extends PageRenderer {
   static renderElement(currentElement, newElement) {
     morphElements(currentElement, newElement, {
       callbacks: {
-        beforeNodeMorphed: element => !canRefreshFrame(element)
+        beforeNodeMorphed: (node, newNode) => {
+          if (super.shouldRefreshChildFrameWithMorphing(null, node, newNode)) {
```
jorgemanrubia, line 10: I don't love passing `null` here. At first glance, it's surprising and it makes you wonder what this logic is about.

What do you think of splitting the logic in two functions: Here it could be `shouldReloadFrameWithMorphing(...) && !closestFrameReloadableWithMorphing(...)` and, in the frame renderer, it could be: `shouldReloadFrameWithMorphing(...) && !closestFrameReloadableWithMorphing(...) == currentElement`

jorgemanrubia (commented):
Somme comments / questions but this looks good @botandrose, thanks!

src/core/renderer.js
botandrose: This check is still addressing a valid concern:
`<turbo-frame id="a" src="/a">` => `<turbo-frame id="a">` Reload!
`<turbo-frame id="a" src="/a">` => `<turbo-frame id="a" src="/a">` Reload!
`<turbo-frame id="a" src="/a">` => `<turbo-frame id="a" src="/b">` Do NOT reload. Instead, morph as usual so that the `src` attribute gets mutated, and thus the turbo-frame loads the new `/b` page, rather than being stuck on `/a`.

Maybe there's a better way to do this, or a better place to do it in, or both! But this is what made sense to me.

src/core/renderer.js
botandrose: Confirmed. We can ditch this as of Idiomorph v0.4

botandrose: We should now also be able to remove line 17, now that I think about it.

botandrose:
@jorgemanrubia this PR isn't morphing-related, specifically. But I am always rebasing my dev branches on this branch, and then unrebasing them to push, because the status quo of devving on main is so painful whenever a test fails. FWIW.

src/core/renderer.js
botandrose: @jorgemanrubia Ah, turns out we DO need these lines after all, because Turbo is reusing the same callback for both morphed AND removed hooks, and thus `newFrame` can be `undefined`:
```js
class DefaultIdiomorphCallbacks {
...
  beforeNodeRemoved = (node) => {
    return this.beforeNodeMorphed(node)
  }
```

botandrose:
@jorgemanrubia Okay, the refactoring you suggested is done. Ready for another look!

jorgemanrubia (approved):
Thanks for addressing the feedback @botandrose

botandrose:
Awesome! Thank you @jorgemanrubia !

PR #1316, opened 2024-09-22
Enable frame morphing for data-turbo-frame links and forms if the URL doesn't change

botandrose:
Hi folks, this PR is aimed at enabling another specific use-case for morphing `turbo-frame[refresh=morph]` that isn't currently supported.

## Status quo:
Right now, if one clicks a link or submits a form that contains `data-turbo-frame`, that target frame is replaced without morphing, even if `refresh=morph`. This makes sense if the frame is navigating to an entirely new URL, because its not a "refresh" operation at all.

## Motivating use-case:
However, I propose that in some common idiomatic situations these operations are indeed refreshes, and therefore should be morphed. Imagine a turbo-frame with a list of blog posts, and an external new post form, which targets the blog post list. I submit for your consideration that in this situation, submitting the form is indeed refreshing the frame.

```html
<turbo-frame id="posts" src="/posts" refresh="morph">
  <ul>
    <li><a href="/posts/3">Post 3</a></li>
    <li><a href="/posts/2">Post 2</a></li>
    <li><a href="/posts/1">Post 1</a></li>
  </ul>
</turbo-frame>

<form action="/posts" data-turbo-frame="posts">
  <input name="title">
  <input type="submit">
</form>
```

## Proposed solution
If a morphing frame's URL changes from the result of a link click or form submission, keep the current behavior and render it with the normal replacement rendering. If the URL does NOT change, however, consider it a refresh and use morphing.

## Final notes
Please let me know if this PR is an idea worth considering for merging, and if so I'll do a bit more work here. I think it could use some more tests to fully nail down all the permutations it implies, and also an extract method refactoring... setting the `FrameController`'s private `#shouldMorphFrame` property from the outside reeks of Feature Envy, and I'm surprised its even legal JavaScript!

Thanks for your time, and continued work on Turbo!

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -91,10 +91,9 @@ export class FrameController {
   }
 
   sourceURLReloaded() {
-    const { refresh, src } = this.element
+    const { src } = this.element
 
-    this.#shouldMorphFrame = src && refresh === "morph"
-
+    this.#shouldMorphFrame = this.element.shouldReloadWithMorph
     this.element.removeAttribute("complete")
     this.element.src = null
     this.element.src = src
@@ -234,6 +233,9 @@ export class FrameController {
     const frame = this.#findFrameElement(formSubmission.formElement, formSubmission.submitter)
 
     frame.delegate.proposeVisitIfNavigatedWithAction(frame, getVisitAction(formSubmission.submitter, formSubmission.formElement, frame))
+    if (frame.src === response.response.url && frame.shouldReloadWithMorph) {
+      frame.delegate.#shouldMorphFrame = true
+    }
     frame.delegate.loadResponse(response)
 
     if (!formSubmission.isSafe) {
@@ -343,6 +345,9 @@ export class FrameController {
     frame.delegate.proposeVisitIfNavigatedWithAction(frame, getVisitAction(submitter, element, frame))
 
     this.#withCurrentNavigationElement(element, () => {
+      if (frame.src === url && frame.shouldReloadWithMorph) {
+        frame.delegate.#shouldMorphFrame = true
+      }
       frame.src = url
     })
   }
--- a/src/tests/functional/frame_tests.js
+++ b/src/tests/functional/frame_tests.js
@@ -267,7 +267,7 @@ test("following a link to a page with a matching frame does not dispatch a turbo
   )
 })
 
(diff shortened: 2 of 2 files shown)


src/core/frames/frame_controller.js
```diff
@@ -234,6 +233,9 @@ export class FrameController {
     const frame = this.#findFrameElement(formSubmission.formElement, formSubmission.submitter)
 
     frame.delegate.proposeVisitIfNavigatedWithAction(frame, getVisitAction(formSubmission.submitter, formSubmission.formElement, frame))
+    if (frame.src === response.response.url && frame.shouldReloadWithMorph) {
```
seanpdoyle, line 236: Determining whether a page-wide Visit is to be handled as a refresh [involves more than an exact comparison of URL values](https://github.com/hotwired/turbo/blob/v8.0.10/src/core/drive/page_view.js#L58-L60).

At the Page level, it involves comparing the [URL.pathname](https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname) values (ignoring differences in query parameters) *as well as* checking for whether or not the `Visit.action` is `"replace"`.

This change could make a similar comparison at the `[src]` and `URL` level, but driving Frame elements does not have a concept of a `Visit`, and `[data-turbo-action="replace"]` has Page-wide navigation implications.

Does it make sense to compare pathnames and ignore query parameters?

botandrose: @seanpdoyle Thanks for taking a look at this, and for pointing out how page refreshes already handle this situation regarding query params. I definitely want to try to keep things consistent between the frame and page systems! Not to mention that it will make this feature just that little bit more useful. I'll push a commit for this shortly.

Regarding #430, I love it. Initially wrapping my head around Turbo's internals has not been easy, and that additional abstraction would help to compartmentalize and clarify some of the complexity. I hope it finally gets merged soon.

adam-h:
This patch resolves the same issue we're hitting with turbo morph and forms in turbo frames. Anything I can do to help unblock it?

botandrose:
@jorgemanrubia This is one of the morphing-related improvement PRs that I mentioned to you earlier. Now that we're on Idiomorph v0.7.2, I'd love to see if we can get this (or something similar) merged. Just rebased on latest `main` for your review.

botandrose:
@jorgemanrubia Lastly, ping on this one. If this gets merged, I could finally get off of my fork of Turbo and back onto the release channel, which would be amazing. I really appreciate your time and attention!

PR #1317, opened 2024-09-24
Better testing timeouts for less painful development

botandrose:
## Background
* Many tests assert that certain browser events occurred using a helper named `nextEventNamed`. To deal with asynchronicity, this helper waits until  the specified event occurs. If the event never occurs, it loops forever until the the default Playwright test timeout is reached.
* The default Playwright test timeout is 30 seconds.
* Playwright is configured to retry failing tests for a max of 3 total runs.
* Playwright is configured to run tests on both Chromium and Firefox.

## Pain
When developing Turbo and running tests frequently, if even one of these event assertion fails, we're looking at a _minimum of 3 minutes_ before the test run completes! And that's the best-case scenario of running only one individual test... often we're running many, and if one test fails, often other tests are failing too.
 
## Proposed resolution
Configure Playwright's global default to 10 seconds. This still seems high to me, to be honest, but I figured it was best to be conservative. If we're okay with lowering it further to perhaps 5 or even 3 seconds, I'm happy to do this. Note that if one or two tests truly need a longer timeout, they can be set individually with a simple `test.setTimeout(ms)` call at the beginning of the test.

## Additional notes
While I was in the Playwright config file, I lowered some other seemingly excessive timeouts, also to 10 seconds. I figure if Chrome or Firefox doesn't start up within 10 seconds, its probably not going to start in 60 seconds either. The tiny node fixture testing server had a 2 minute startup timeout!

### DIFF
--- a/playwright.config.js
+++ b/playwright.config.js
@@ -7,7 +7,7 @@ const config = {
       use: {
         ...devices["Desktop Chrome"],
         contextOptions: {
-          timeout: 60000
+          timeout: 10000
         },
         hasTouch: true
       }
@@ -17,20 +17,21 @@ const config = {
       use: {
         ...devices["Desktop Firefox"],
         contextOptions: {
-          timeout: 60000
+          timeout: 10000
         },
         hasTouch: true
       }
     }
   ],
-  browserStartTimeout: 60000,
+  timeout: 10000,
+  browserStartTimeout: 10000,
   retries: 2,
   testDir: "./src/tests/",
   testMatch: /(functional|integration)\/.*_tests\.js/,
   webServer: {
     command: "yarn start",
     url: "http://localhost:9000/src/tests/fixtures/test.js",
-    timeout: 120 * 1000,
+    timeout: 10000,
     // eslint-disable-next-line no-undef
     reuseExistingServer: !process.env.CI
   },


botandrose:
As an alternative, or even as a follow-up, I think it makes sense to replace the infinite loop in `nextEventNamed` with a default timeout, perhaps 2 seconds to match Capybara's default. We could make this configurable via an optional argument, too.

botandrose:
Well, 10 seconds was still too painful (still have to wait a full minute for failing test results), so I just went ahead and did what I described above with adding 2 second timeouts to the individual event helpers, and I've pushed it here, too. Renaming to _Better testing timeouts for less painful development_. I think this is a good combo of approaches. Sane global timeouts, and finer-grained timeouts for assertions.

What do you think?

src/tests/helpers/page.js
seanpdoyle, line 1: Changing the global timeout is definitely an improvement. I'm skeptical that making event reading timeouts configurable is worthwhile. Adding a fourth positional argument to the `next`-prefixed methods awkward, and I think it's less likely to require one-off re-configurations.

Instead, what do you think about adding a `2000ms` timeout directly to the `readEventLogs` and `readBodyMutationLogs` functions? If that works, maybe it'd be worthwhile to source the `2000ms` value from a global configuration.

brunoprietog:
How about just going for the 10 seconds in the Playwright configuration as a first step?

botandrose:
@jorgemanrubia this PR isn't morphing-related, specifically. But I am always rebasing my dev branches on this branch, and then unrebasing them to push, because the status quo of devving on main is so painful whenever a test fails. FWIW.

brunoprietog:
If we just change the Playwright timeouts, I can merge right now. The other part we can look at later. Does that sound good?

botandrose:
Hi @brunoprietog, all right. Waiting a minimum of 0:60 for a failure result is undeniably better than waiting a minimum of 3:00 for a failure result. I've removed the last commit, so this should be GTG. I'll reopen a new PR with further improvements after this gets merged. Thank you!

botandrose:
@brunoprietog Would you mind giving me some specific feedback on what you would want to see in order to feel inclined to merge a PR that contains https://github.com/hotwired/turbo/commit/0c1a6fa7029fceb2ca8a1a47e9d33b163500d1af ?

With this commit, I'm seeing the minimum wait for a failure result drop from 0:60 to 0:15, at the cost of one specific timeout-related test needing to know about the new optional timeout parameter in the test helper. This seems clearly worth it to me, but what would you want to see?

brunoprietog (approved):
Thanks @botandrose!

brunoprietog:
It makes me uncomfortable to see that argument within the tests, but I understand that it soothes a rather annoying pain. We could start with that for now, only if the timeout needs to be changed at that specific location. Also, I'd move the default timeout to a constant.

botandrose:
Gotcha. Thanks! I'll see if I can't figure out a way to accomplish this without adding a new parameter.

PR #1321, opened 2024-09-26
Fix bug where Idiomorph sometimes ignores data-turbo-permanent

botandrose:
Hi folks, this is a redux of #1308 with an alternative strategy: fixing the issue in Idiomorph.

## Motivating use-case
I'm building a collaborative issue tracking app, and one of the issue tracker app's primary use-cases has exposed a bug in the integration between Idiomorph and the `data-turbo-permanent` attribute. Each ticket contains a `data-turbo-permanent` checkbox to keep track of the client-side state of whether or not the ticket is currently expanded for that client or not. We also allow tickets to be reordered via drag-and-drop. The issue is that this `data-turbo-permanent` checkbox is not always preserved across morphs involving ticket reorders. There are other more serious ramifications of this bug (data loss!), but this is the simplest case to illustrate the issue.

## Diagnosis and resolution
I have reduced this scenario down to the failing test case in this PR, and identified the issue as being within Idiomorph, which I've opened a PR for here: https://github.com/bigskysoftware/idiomorph/pull/61 . However, much smaller "no-brainer" PRs from other Turbo contributors have gone ignored for months, so I'm not optimistic at its odds of a timely merge and release. Perhaps Turbo should maintain its own Idiomorph branch? At the moment, this PR is pointing Idiomorph to my own botandrose Idiomorph branch, which I'm sure is not ideal.

Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "chai": "~4.3.4",
     "eslint": "^8.13.0",
     "express": "^4.18.2",
-    "idiomorph": "https://github.com/bigskysoftware/idiomorph.git",
+    "idiomorph": "~0.7.1",
     "multer": "^1.4.2",
     "rollup": "^2.35.1"
   },
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "chai": "~4.3.4",
     "eslint": "^8.13.0",
     "express": "^4.18.2",
-    "idiomorph": "~0.7.1",
+    "idiomorph": "~0.7.2",
     "multer": "^1.4.2",
     "rollup": "^2.35.1"
   },
--- a/src/core/morphing.js
+++ b/src/core/morphing.js
@@ -1,4 +1,4 @@
-import { Idiomorph } from "idiomorph/dist/idiomorph.esm.js"
+import { Idiomorph } from "idiomorph"
 import { dispatch } from "../util"
 
 export function morphElements(currentElement, newElement, { callbacks, ...options } = {}) {
--- a/src/core/morphing.js
+++ b/src/core/morphing.js
@@ -9,7 +9,7 @@ export function morphElements(currentElement, newElement, { callbacks, ...option
 }
 
 export function morphChildren(currentElement, newElement) {
-  morphElements(currentElement, newElement.children, {
+  morphElements(currentElement, newElement.childNodes, {
     morphStyle: "innerHTML"
   })
 }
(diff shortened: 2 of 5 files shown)


brunoprietog:
I see https://github.com/bigskysoftware/idiomorph/pull/72 was finally merged. Would incorporating the use of `twoPass` help here?

botandrose:
Hi @brunoprietog, yes, but not quite yet! There's one outstanding bug in v0.4.0 that breaks Turbo data-turbo-permanent: https://github.com/bigskysoftware/idiomorph/pull/83 . Once that gets merged and released, upgrading Idiomorph will become possible for a much-improved morphing experience in Turbo. I'm hoping we'll get a v0.4.1 or a v0.5.0 out the door very soon. Just gotta wait for Carson to carve out the time to review and release.

1cg:
I am going to review the PR tomorrow and will update this issue

brunoprietog:
Thanks!

botandrose:
@brunoprietog After months of work, we've just released Idiomorph v0.7.1 which is a big leap forward in many many things, including fixing this motivating bug. Morphing should be much, much more correct now across the board.

So I've force-pushed this PR with three commits:
1. A small bugfix to the Turbo->Idiomorph integration
2. The motivating failing test
3. The bump to Idiomorph v0.7.1 that fixes it.

Lastly, note that there is one additional feature in v0.7.1 that may be of interest here. Active focus/selection restoration after morph. I know Turbo already does this for non-morphing renders, but now the morphing render does it too via Idiomorph. If you want to evaluate that as a separate addition, it can be disabled by passing `restoreFocus: false` to `Idiomorph.morph`.

Finally, Carson has designated me as the lead of Idiomorph moving forward, so I'll happy to work with Turbo on any morphing-related issues.

P.S. main is red right on `main` so we're getting false failures on this PR. I'll rebase after https://github.com/hotwired/turbo/pull/1368 or something similiar is merged.

botandrose:
And, we're green! Ready for review.

botandrose:
Bumped Idiomorph from v0.7.1 to v0.7.2 which contains two bugfixes:
  * Prevent unnecessary selection restoration when it wasn't actually lost
  * Prevent focus & selection loss in more situations

jorgemanrubia:
This works great on my tests @botandrose, I'll merge. Thanks for the fantastic work!

jorgemanrubia:
The update was released in [`turbo` 8.0.13](https://github.com/hotwired/turbo/releases)  [`turbo-rails` 2.0.12](https://github.com/hotwired/turbo-rails/releases/tag/v2.0.12).

botandrose:
@jorgemanrubia Amazing, thank you! It feels great to get this out the door to the world. I'll push some more morphing-related follow-up PRs later today.

Issue #906, opened 2023-04-14
[Proposal] turbo-stream action 'versioning'


mbell697:
I'm using turbo-streams for a project and have noticed that in certain circumstances stream actions can be delivered out of order, an older update arriving after a newer update. 

A concrete example would be replying from a rails controller directly with a turbo-stream action, but there also being a background job enqueued during that request that sends turbo-stream actions over actioncable. We've see the update from the background job arrive first with fresher data, only to have the controller turbo stream then arrive later replacing the dom content with stale data.

I think it would be a great feature to be able to provide an ordering key (`version`) along with the each action and have the update code reject the update if the value currently in the dom is smaller.

For example, if this sequence of actions arrived in the browser in this order:

```
<turbo-stream action="append" target="item" version="1"></turbo-stream>
<turbo-stream action="replace" target="item" version="3"></turbo-stream>
<turbo-stream action="replace" target="item" version="2"></turbo-stream>
```

The final action with version `2` would be rejected because `3` is the current version in the `dom`. 

In practical applications, `version` would often be an `updated_at` timestamp.

I think this feature could be implemented with a simple stimulus controller for my particular situation, but I suspect it's common enough to belong in turbo-stream. 

If interested in this feature - I'm happy to open a PR for it.

phylor:
@brunoprietog Is there any plan to include versioning in turbo stream actions, to avoid out-of-order updates? In my experience, running stale actions happens more often than one might think. It's especially noticeable when implementing progress updates, as the "finished" message might arrive before the last "loading" message, thus not updating correctly. Having a versioning builtin would be awesome!

dhh:
I like this idea. Please do explore an implementation.

jorgemanrubia:
Closing. See https://github.com/hotwired/turbo/pull/1324#issuecomment-2683234802

PR #1324, opened 2024-10-09
[WIP] Stream Versioning

mbell697:
This still needs some additional more tests and docs prior to merging but I wanted to open this for comment on the general approach.

This is a proposed solution for #906

This PR introduces the concept of versioning for Turbo stream `update` and `replace` actions, see the above issue for more background.

Functionality:

* Add the attribute `data-steam-version` on stream target element. 
* Add the attribute `version` on the `turbo-stream` element. 
* During `update` and `replace`: If versions are present, they are parsed as numeric values and compared, actions are rejected unless the new version is greater than the existing version. 

The current API presented here does not attempt to automatically manage the versions within template content, i.e. an update looks like this:

```html
<turbo-stream action="update" target="item" version="1">
  <template>
    <div id="item" data-turbo-version="1"></div>
  </template>
</turbo-stream>
```

Versioning information ends up duplicated both in the `turbo-stream` and the template content. 

In theory we could search the template content for the target selector and propagate the version down from the `turbo-stream` element. But I think this is "too smart" and likely to have edge cases. I suspect most usage of `turbo` is via higher level libraries, like `turbo-rails`, and I think those layers should be able to mask the duplication given more domain knowledge.

Let me know what ya think and I'll wrap this up.

### DIFF
--- a/src/core/streams/stream_actions.js
+++ b/src/core/streams/stream_actions.js
@@ -1,5 +1,24 @@
 import { session } from "../"
 import { morphElements, morphChildren } from "../morphing"
+import { attributeToNumber } from "../../util"
+
+export function versionCheck(streamElement, targetElement) {
+  const targetVersion = attributeToNumber(targetElement.getAttribute("data-turbo-version"))
+  const streamVersion = streamElement.version
+
+  const streamVersionPresent = streamVersion != null
+  const targetVersionPresent = targetVersion != null
+
+  if (!streamVersionPresent && targetVersionPresent) {
+    return false // Don't allow an unversioned element to replace a versioned element
+  } else if (streamVersionPresent && !targetVersionPresent) {
+    return true // Do allow a versioned element to replace an unversioned element
+  } else if (streamVersionPresent && targetVersionPresent) {
+    return streamVersion > targetVersion
+  } else {
+    return true
+  }
+}
 
 export const StreamActions = {
   after() {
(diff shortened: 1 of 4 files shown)


MatheusRich:
@dhh @jorgemanrubia could you please give a first review on this approach? We'd greatly benefit from this on some of the apps I'm working on. Thank you!

jorgemanrubia:
After some consideration, we won't merge this. We don't think this belongs to the official Turbo API. This mechanism would need some server counterpart to avoid race conditions (like some session-global counter), and we don't feel the benefits here justify the added complexity (at least, not to be sanctioned officially). 

Thanks for your work on this one.

phylor:
That's very sad to here! It would be very helpful when rendering progress bars. This is a pattern I often encounter with Turbo:

1. Have a background job running, updating progress in a database model
2. On every update, broadcast a Turbo stream via web sockets

As messages aren't necessarily received in the same order on the client, it often happens that the 100% progress is received **before** the 99% progress. For the user the progress hence never completes.

To fix this, one could add client side polling (which defeats the purpose), or have a manual versioning solution involving custom JavaScript on the client.

I very much like the proposal in this PR. The version could be a number out of the database (no race conditions), or even the current nanoseconds (unlikely a race condition).

Is there an official opinion on how to address the above problem?

mbell697:
> After some consideration, we won't merge this. We don't think this belongs to the official Turbo API. This mechanism would need some server counterpart to avoid race conditions (like some session-global counter), and we don't feel the benefits here justify the added complexity (at least, not to be sanctioned officially).
> 
> Thanks for your work on this one.

Totally understand the concerns, but could you elaborate a bit more about the server side concerns? I’m not entirely following and it may inform a 3rd party implementation.

PR #1331, opened 2024-10-24
Remove unused `getHistoryMethodForAction` function in Visit

nanaya:
It's extracted to util file in #618 but the origin was left behind.

I'd like to run the test locally but playright doesn't play well with alpine so it's not tested 😬 The changes do look relatively straightforward though.


brunoprietog (approved):
Thanks @nanaya!

PR #1341, opened 2024-11-21
Fix navigating to the turbo-root does not use Turbo Drive

### LINKED ISSUE #912: Navigating to the `turbo-root` doesn't use Turbo Drive

leofeyer:
Fixes #912

This is just a copy of #913 with the merge conflicts resolved, so credits should go to @ tobyzerner.

### DIFF
--- a/src/core/url.js
+++ b/src/core/url.js
@@ -55,9 +55,5 @@ function getLastPathComponent(url) {
 }
 
 function getPrefix(url) {
-  return addTrailingSlash(url.origin + url.pathname)
-}
-
-function addTrailingSlash(value) {
-  return value.endsWith("/") ? value : value + "/"
+  url.origin + url.pathname
 }
--- a/src/tests/functional/root_test.js
+++ b/src/tests/functional/root_test.js
@@ -1,0 +1,27 @@
+import { test } from "@playwright/test"
+import { assert } from "chai"
+import { nextBody, pathname, visitAction } from "../helpers/page"
+
+test("test visiting a location inside the root", async ({ page }) => {
+  page.goto("/src/tests/fixtures/root/index.html")
+  page.click("#link-page-inside")
+  await nextBody(page)
+  assert.equal(pathname(page.url()), "/src/tests/fixtures/root/page.html")
+  assert.notEqual(await visitAction(page), "load")
+})
+
+test("test visiting the root itself", async ({ page }) => {
+  page.goto("/src/tests/fixtures/root/page.html")
+  page.click("#link-root")
+  await nextBody(page)
+  assert.equal(pathname(page.url()), "/src/tests/fixtures/root/")
+  assert.notEqual(await visitAction(page), "load")
+})
+
+test("test visiting a location outside the root", async ({ page }) => {
+  page.goto("/src/tests/fixtures/root/index.html")
+  page.click("#link-page-outside")
+  await nextBody(page)
+  assert.equal(pathname(page.url()), "/src/tests/fixtures/one.html")
+  assert.equal(await visitAction(page), "load")
+})
--- a/src/tests/fixtures/root/index.html
+++ b/src/tests/fixtures/root/index.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html id="html">
+  <head>
+    <meta charset="utf-8">
+    <title>Turbo</title>
+    <script src="/dist/turbo.es2017-umd.js" data-turbo-track="reload"></script>
+    <script src="/src/tests/fixtures/test.js"></script>
+    <meta name="turbo-root" content="/src/tests/fixtures/root">
+  </head>
+  <body>
+    <section>
+      <h1>Root</h1>
+      <p><a id="link-page-inside" href="/src/tests/fixtures/root/page.html">Link to page inside root</a></p>
+      <p><a id="link-page-outside" href="/src/tests/fixtures/one.html">Link to page outside root</a></p>
+    </section>
+    <div id="messages"></div>
+  </body>
+</html>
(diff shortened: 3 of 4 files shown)


fritzmg (approved):

src/core/url.js
```diff
@@ -25,7 +25,7 @@ export function getExtension(url) {
 }
 
 export function isPrefixedBy(baseURL, url) {
-  const prefix = getPrefix(url)
+  const prefix = getPrefix(url) || '/'
```
brunoprietog, line 28:
```suggestion
  const prefix = getPrefix(url) || "/"
```

src/tests/functional/root_test.js
```diff
@@ -0,0 +1,27 @@
+import { test } from "@playwright/test"
+import { assert } from "chai"
+import { nextBody, pathname, visitAction } from "../helpers/page"
+
+test("test visiting a location inside the root", async ({ page }) => {
```
brunoprietog, line 5:
```suggestion
test("visiting a location inside the root", async ({ page }) => {
```

src/tests/functional/root_test.js
```diff
@@ -0,0 +1,27 @@
...
+  page.goto("/src/tests/fixtures/root/index.html")
+  page.click("#link-page-inside")
+  await nextBody(page)
+  assert.equal(pathname(page.url()), "/src/tests/fixtures/root/page.html")
+  assert.notEqual(await visitAction(page), "load")
+})
+
+test("test visiting the root itself", async ({ page }) => {
```
brunoprietog, line 13:
```suggestion
test("visiting the root itself", async ({ page }) => {
```

src/tests/functional/root_test.js
```diff
@@ -0,0 +1,27 @@
...
+  page.goto("/src/tests/fixtures/root/page.html")
+  page.click("#link-root")
+  await nextBody(page)
+  assert.equal(pathname(page.url()), "/src/tests/fixtures/root/")
+  assert.notEqual(await visitAction(page), "load")
+})
+
+test("test visiting a location outside the root", async ({ page }) => {
```
brunoprietog, line 21:
```suggestion
test("visiting a location outside the root", async ({ page }) => {
```

brunoprietog (changes requested):
Hi @leofeyer, thanks. I did a quick first pass and left minor styling suggestions.

m-vo (commented):
Changes look good to me, this has annoyed me before as well  🙂.

@packagethief Wdyt? Could we maybe get this one merged as well for the upcoming release?

packagethief:
> Changes look good to me, this has annoyed me before as well 🙂.
> 
> @packagethief Wdyt? Could we maybe get this one merged as well for the upcoming release?

Ha, this affects Basecamp as well 😄 

Happy to see this resolved. I see a few of problems, though:

1. `getPrefix` is missing a `return`, so it won't work
2. `getPrefix` will always return a value, so we'll never fall through to the "or" branch in `isPrefixedBy`
3. The test file is named in the singular `root_test.js`, so it won't run with the test runner. It needs to be named `root_tests.js`.
4. I'm not convinced the implementation will work. If the root URL is `https://example.com/app` and the requested URL is `https://example.com/application` it will match on the prefix, which isn't right. I think we need to be appending the trailing slash to both URLs in order to compare accurately.

packagethief:
This (untested) diff against the branch should work, if anyone wants to run with it (and add a test that asserts that a link to `/rootlet.html` is not considered inside a root of `/root`) 😄

```diff
diff --git i/src/core/url.js w/src/core/url.js
index 57bbca1..9c92d60 100644
--- i/src/core/url.js
+++ w/src/core/url.js
@@ -25,8 +25,8 @@ export function getExtension(url) {
 }
 
 export function isPrefixedBy(baseURL, url) {
-  const prefix = getPrefix(url) || "/"
-  return baseURL.href === expandURL(prefix).href || baseURL.href.startsWith(prefix)
+  const prefix = addTrailingSlash(url.origin + url.pathname)
+  return addTrailingSlash(baseURL.href) === prefix || baseURL.href.startsWith(prefix)
 }
 
 export function locationIsVisitable(location, rootLocation) {
@@ -54,6 +54,6 @@ function getLastPathComponent(url) {
   return getPathComponents(url).slice(-1)[0]
 }
 
-function getPrefix(url) {
-  url.origin + url.pathname
+function addTrailingSlash(value) {
+  return value.endsWith("/") ? value : value + "/"
 }
```

m-vo:
Here we go: https://github.com/hotwired/turbo/pull/1426

Issue #1253, opened 2024-05-02
Server-side redirect from turbo stream to file download fails in Turbo 8


asilano:
### Steps to reproduce

Have HTML containing:
```html
<a href="/maybe_a_file" data-turbo-stream="true">Download</a>
```
Backed by server code:
```ruby
def maybe_a_file
  if error?
    # File doesn't exist - render a turbo-stream to display an error message
    return render "error_message"
  end

  redirect_to "/actual_file.txt"
end
```
Click the link (when `error?` returns false)

### Expected behaviour
The redirect is followed, and the browser downloads and saves `actual_file.txt`

### Actual behaviour
The redirect is followed, the browser requests `actual_file.txt`, and the server responds 200 OK with the contents of the file; but the file is not saved by the browser

### Cause
Via `git bisect`, determined to be caused by #968 (thanks to @pythonandchips for digging)

### Workaround
Instead of server-side redirecting to the file, return a turbo-stream action to visit the actual file URL in JS (both `Turbo.visit` and `window.location.assign` work fine)

brunoprietog:
If you mark a link with `data-turbo-stream`, Turbo performs a Turbo stream action, not a download. I suggest implementing this with a Turbo frame. Follow the link, if it doesn't find it, replace the Turbo frame with an error message, and if it finds it you can download the file.

PR #1348, opened 2024-12-06
Reproduce redirect to file download fails in turbo frame

DanielPechersky:
Reproduce https://github.com/hotwired/turbo/issues/1253 in tests

Interestingly, "following a link to a non html file changes page src" has different behaviour across chrome and firefox
Also, "following a link to an unknown non html file changes page src" is not directly related to reproducing this issue but it's a disparity between turbo drive and turbo frame behaviour.


DanielPechersky:
Actually, this is not exactly the same issue, but is probably related. I'm experiencing an issue with turbo frames in my project, but I'm guessing a link with `data-turbo-stream="true"` behaves similarly

PR #1355, opened 2024-12-17
rename unprocessable entity to unprocessable content to match Rack

reedrolemodel:
This is an alternative to https://github.com/hotwired/turbo/pull/1277 which looks abandoned.


brunoprietog (approved):
Thanks @reedrolemodel!

Issue #1374, opened 2025-03-04
Control view transition direction?


jbarrieault:
The view transition support is awesome. With just a little CSS, you can have rich, _directional_ page transitions that slide pages in and out, in a directional manner.

However, a UI may link to pages that are _logically_ back despite not always being a location in the history stack. Breadcrumb navigation, for example. As far as I can tell, _control of the direction_ of a view transition is not currently supported—it's hard-coded to be tied to the direction of the Turbo navigation direction.

**What if Turbo allowed for the View Transition direction to be set independently of the _visit_ direction?**

### Workaround 1
`Turbo.visit(url, { action: 'restore' })`

This works, even when the url is not in the history stack.

However, the handbook [specifically](https://turbo.hotwired.dev/handbook/drive#restoration-visits) advises against this:
> Restoration visits have an action of restore and Turbo Drive reserves them for internal use. You should not attempt to annotate links or invoke Turbo.visit with an action of restore.

I didn't find anything actually unsafe about doing so, but perhaps it's more about internal vs public API.

### Workaround 2
Hook into the render lifecycle:

```javascript
addEventListener('turbo:before-render', (event) => {
  event.preventDefault()
  
  // set by an anchor tag click handler, perhaps 
  const direction = document.querySelector('html').getAttribute('data-view-transition-direction')

  if (direction === 'back') {
    // override the 'forward' data attribute Turbo's internal handling added at the start of navigation
    document.querySelector('html').setAttribute('data-turbo-visit-direction', 'back')
  }

  event.detail.resume()
  document.querySelector('html').removeAttribute('data-view-transition-direction')
})
```

## Stick with a workaround or add support?

I like the simplicity of work around 1. It is clearly discouraged but for vague reasons. Perhaps it is safe enough?

If not, would you be open a pull request to add support for controlling view transition direction? If so, opinions on the API are welcome?

4lllex:
There is a way to do it with turbo events: 

```js
// app/javascript/application.js

document.addEventListener("turbo:click", (event) => {
  if (event.target.dataset.turboVisitDirection === "back") {
    window.visitDirection = "back"
  }
});
document.addEventListener("turbo:visit", (event) => {
  // this way you don't override browser back button navigation
  if (window.visitDirection) {
    document.documentElement.dataset.turboVisitDirection = window.visitDirection
  }
  window.visitDirection = undefined;
});
```
```ruby
<%= link_to "Cancel", post, data: {turbo_visit_direction: "back"} %>
[...]

PR #1375, opened 2025-03-04
Support Directional View Transitions

jbarrieault:
Addresses https://github.com/hotwired/turbo/issues/1374

This adds support for setting the direction of visit's view transition, and applies the logical direction when navigating stack history.

### Example
Annotate an anchor tag with its logical direction, which will be applied to the view transition.

```html
<!-- /cars/1/offers.html -->

<!-- equivalent to: Turbo.visit("/cars/1", { direction: "back" }) -->
<a href="/cars/1" `data-turbo-visit-direction="back"` />
```

After visiting the above link, subsequently using the browser's back button will apply a _forward_ view transition. The end result is that the original as well as subsequent back/forward button visits will perform a view transition using the intuitive direction of the navigation.

### TODO
- [ ] more testing
- [ ] write tests?

### DIFF
--- a/src/core/drive/history.js
+++ b/src/core/drive/history.js
@@ -87,13 +87,8 @@ export class History {
         this.location = new URL(window.location.href)
         const { restorationIdentifier, restorationIndex } = turbo
         this.restorationIdentifier = restorationIdentifier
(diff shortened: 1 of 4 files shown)


jbarrieault:
This needs more work, closing until I can give it some attention.

PR #1377, opened 2025-03-05
Ensure response content is loaded before requestFinish is called.

armstrjare:
Previously the response content was loaded asynchronously within the FetchRequest delegate. Consequently, the requestFinished callback was called prior this finishing. The requestFinished callback triggers the toggling of the busy attribute on `<html>/<turbo-frame>/<form>`, the triggering of the afterSubmit callback which toggles the disabled state of form submitters, and the the turbo:submit-end event.

For clients with high latency, this could result in janky behaviour when the busy/disabled state is removed a noticeable time before the response is loaded and rendered.

This change removes this delay by ensuring the response is loaded and available prior to notifying the delegate and continuing execution of the Turbo behaviour.

### DIFF
--- a/src/http/fetch_request.js
+++ b/src/http/fetch_request.js
@@ -154,11 +154,16 @@ export class FetchRequest {
     })
     if (event.defaultPrevented) {
       this.delegate.requestPreventedHandlingResponse(this, fetchResponse)
-    } else if (fetchResponse.succeeded) {
-      this.delegate.requestSucceededWithResponse(this, fetchResponse)
     } else {
-      this.delegate.requestFailedWithResponse(this, fetchResponse)
+      await fetchResponse.responseHTML
+
+      if (fetchResponse.succeeded) {
+        this.delegate.requestSucceededWithResponse(this, fetchResponse)
+      } else {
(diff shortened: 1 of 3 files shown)


armstrjare:
Fixed the `eslint` failure – added teardown to eslint globals

armstrjare:
For anyone that wants this fixed for their app before this gets merged, here is a monkey-patch that gives effectively the same outcome:

```js
// Ensure that the responseHTML is loaded before the fetchRequest is declared finished
const originalReceive = Turbo.FetchRequest.prototype.receive
Turbo.FetchRequest.prototype.receive = async function(response) {
  let fetchResponse = await originalReceive.call(this, response)
  await fetchResponse.responseHTML
  return fetchResponse
}

```

armstrjare:
@jorgemanrubia is this something you guys would consider merging? let me know if there is any feedback on this

skmedix:
We tested this in our project and, unfortunately, found that this change is causing some side effects. Specifically, when we use POST in a form, and our location doesn't change (we stay on the same page). Some events are not triggered, and some of our components simply break. I don't have much time to create a reproduction right now, but hopefully this helps.

PR #1378, opened 2025-03-06
Add 2 second timeouts to infinitely-looping assertions

botandrose:
Following up upon https://github.com/hotwired/turbo/pull/1317 , this time without adding a timeout parameter!

With this commit, I'm seeing the minimum wait for a failure result drop from 0:60 to 0:15 when one of these assertions is failing.


botandrose:
@brunoprietog WDYT?

brunoprietog (approved):
Thanks @botandrose!

Issue #1379, opened 2025-03-06
Rename the view transition meta tag ?


Intrepidd:
The meta tag used to enable view transitions (`view-transition` with `same-origin` value) was chosen to mimic the one in the spec for MPA view transitions. See #935

However, it has now been removed in favor of a configuration directly in the css :https://developer.chrome.com/docs/web-platform/view-transitions/cross-document

With the meta gone, I'm wondering if we should rather not try to piggy back on the mechanism for MPA view transition and rather be explicit, maybe with a meta named `turbo-view-transition` ? Similar to the ones we already have for morphing / scroll / cache control.

Intrepidd:
Fabulous, I opened https://github.com/hotwired/turbo/pull/1380 that does just that, I just have some questions on testing strategy.

PR #1380, opened 2025-03-06
Rename meta tag used to enable view transitions to turbo-view-transition

Intrepidd:
See #1379 

I guess I should add tests for the backwards compatibility, could they be in a different file, with different fixtures, for easier removability when we decide to drop it ?


kevinmcconnell (commented):
This looks good! I agree we shouldn't hold on to the meta tag, now that the standards have moved on (but also that we should keep it around for a while, for compatibility).

I would extend the tests to cover both cases, as you say. Probably easiest to make two versions of the fixtures, and then extend the current view transition tests to cover both sets.

The only other suggestion I have is: it would be even nicer if we could detect the behavior from the `@view-transition` CSS. Then we wouldn't need a setting at all, and the Turbo-enabled behavior would match the default browser behavior.

The rule itself should be present as a `CSSViewTransitionRule` in the stylesheets, so we'd at least have access that way. But I don't know if there are practical problems with detecting it in all cases. What do you think, is it something you've looked into?

Intrepidd:
> The only other suggestion I have is: it would be even nicer if we could detect the behavior from the @view-transition CSS. Then we wouldn't need a setting at all, and the Turbo-enabled behavior would match the default browser behavior.

The CSS rule describes the behavior for MPA view transitions, on the one hand to me this feels a little weird to piggy back this config but for a SPA, because ultimately we use the SPA api (`Document.startViewTransition()`) 

on the other hand it kinda makes sense as Turbo is designed to be dropped on a regular MPA and enhance it instantly without much special config.

However I have no idea if it is even possible to detect the presence of the rule from JS, I need to look into it

Intrepidd:
Ah yes indeed inside `document.styleSheets[x].rules` it is possible to find the `CSSViewTransitionRule` and it's value.

If you confirm we should go this direction rather than the meta config, I will update the PR accordingly

kevinmcconnell:
@Intrepidd I definitely think that's worth a shot, yes.

For me, one of the nicest parts of Turbo is that, as you say, you just drop it in and it can make things better. With browsers supporting MPA view transitions, this would be a case where dropping in Turbo would make something "worse" until you manually opt-back in with the setting. But if we can reliably detect that option from the stylesheet, then the MPA view transitions will continue to work alongside the rest of Turbo's features, without having to explicitly set anything. The fact that Turbo uses the SPA to do it is just an aspect of how Turbo is making page navigations more SPA-like already.

Intrepidd:
@kevinmcconnell So one problem I'm seeing is that in order to check for the presence of the rule it seems I need to access the `document`.

This doesn't play well with page snapshots because we have the current document, and the current and incoming snapshots, but obviously no document yet for the incoming page.

I've tried playing with [document.implementation.createDocument](https://developer.mozilla.org/En/DOM/DOMImplementation.createDocument) but the created document `styleSheets` method is empty even though the styles are inline in the page.

Any idea ?

Intrepidd:
@kevinmcconnell I added a spec to make sure backward compatibility is preserved, and gentle ping on my previous question

Intrepidd:
I just rebased this branch with the latest changes on main

PR #1399, opened 2025-04-15
Prevent slow turbo frame requests from resetting cookies

domchristie:
This pull request fixes an issue where slow `turbo-frame` requests can inadvertently reset cookies.

For example, let's say we have a form that changes the `session` and a `<turbo-frame>` that loads a slow endpoint:

```html
<form method="post" action="/session">
  <button type="submit">Change Session</button>
</form>

<turbo-frame id="slow" src="/slow"></turbo-frame>
```

Here's a potential flow:

1. Page loads, `<turbo-frame>` starts loading
2. User taps "Change Session"
3. That request completes and the session cookie is updated
4. `<turbo-frame>` request completes later and resets the session cookie to what it was in step 1.

Reproduction: https://github.com/domchristie/turbo_frame_cookies_test

The fix is to cancel the frame's current fetch request when the `<turbo-frame>` disconnects.

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -65,6 +65,7 @@ export class FrameController {
   disconnect() {
     if (this.#connected) {
       this.#connected = false
+      this.#currentFetchRequest?.cancel()
       this.appearanceObserver.stop()
       this.formLinkClickObserver.stop()
       this.linkInterceptor.stop()


domchristie:
Recursive `<turbo-frames>` look like they rely on the current behaviour. I've updated the `disconnect` flow to account for this, but it's not the prettiest.

packagethief:
Thanks @domchristie. I think this is an inherent problem with any stale response that sets a cookie. It's not specific to Turbo, but rather a timing issue with how browsers automatically process `Set-Cookie` headers without giving the client a chance to intervene or check for staleness.

That said, I agree we should mitigate it where possible. Your proposal to cancel a frame's request when it's removed from the DOM makes sense regardless. If the requesting frame is gone, there's no reason to process its response, and doing so can cause unintended side effects like the one you've described. I'm +1.

One thing to consider: this handles the case where the frame is explicitly removed, but the race could still occur if the frame remains in the DOM. I'm not sure there's much Turbo can do about that scenario, but I think eliminating the disconnect case is still valuable.

packagethief:
@domchristie, can you think of a way to test this behavior? Seems kind of tricky.

domchristie:
@packagethief thanks for reviewing.

> If the requesting frame is gone, there's no reason to process its response, and doing so can cause unintended side effects like the one you've described. 

Yes, exactly. From a user perspective I think the main issue is that the session gets reverted even when they've navigated away from the page containing the turbo-frame. E.g. a user taps "Log out", they're navigated to a different page, the turbo-frame request then completes, and they're logged in again without them knowing.

> One thing to consider: this handles the case where the frame is explicitly removed, but the race could still occur if the frame remains in the DOM. I'm not sure there's much Turbo can do about that scenario, but I think eliminating the disconnect case is still valuable.

When working on a workaround I considered:

1. setting the `disabled` attribute, which unfortunately doesn't abort in-flight requests
2. removing/setting the `src` attribute*, which will only abort a request if the `src` is present
3. manually cancelling the frame's current request (e.g. `frameElement.delegate.currentFetchRequest.cancel()`), however `currentFetchRequest` is a private property

I think it'd make sense that in-flight requests are aborted if the frame becomes `disabled` or if the `src` is blank.

> can you think of a way to test this behavior? Seems kind of tricky.

Yes, it's tricky! It might be handy for testing if `currentFetchRequest` were made public, then we could assert that `cancel()` is called on it?

---

*In the end I set up a `/noop` endpoint which does not set cookies, and set the `src` of any in-flight turbo-frames before the next page renders:

```js
addEventListener('turbo:before-render', () => {
  document.querySelectorAll('turbo-frame[busy]').forEach((element) => {
    element.src = '/noop'
  })
})
```

packagethief:
> I think it'd make sense that in-flight requests are aborted if the frame becomes disabled or if the src is blank.

I agree. Do you want to make that change in this PR, or start a new one?

domchristie:
👍 Happy to make the change on this PR

src/core/frames/frame_controller.js
```diff
@@ -69,18 +69,29 @@ export class FrameController {
       this.formLinkClickObserver.stop()
       this.linkInterceptor.stop()
       this.formSubmitObserver.stop()
+      if (!this.element.hasAttribute('recurse')) {
+        this.#currentFetchRequest?.cancel()
+      }
```
packagethief:
```suggestion

      if (!this.element.hasAttribute("recurse")) {
        this.#currentFetchRequest?.cancel()
      }
```

packagethief (approved):

PR #1400, opened 2025-04-17
Update playwright

silva96:
Updates playwright version to 1.51.1 and sets the same version in devcontainer and Docker configs.

I had to fix some tests:

- Fix tests that used `page.reload()` that was broken due to a known bug in playwright. See https://github.com/microsoft/playwright/issues/28264 for details.
- Fixed a test related to named anchor navigation, when an element is too small (0 pixels in the case of a naked anchor), `scrollingIntoView` would put the scroll position in a place where the element is not completely visible, so focusing it would re-scroll to the middle of it, messing up the `scrollIntoView` position.  I solved it by shifting the order of both method calls, first `focusElement` and finally `scrollToElement`.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,5 @@
 # See comments in devcontainer.json for details of setting the playwright tag.
-ARG PLAYWRIGHT_TAG="v1.30.0-focal"
+ARG PLAYWRIGHT_TAG="v1.51.1-focal"
 FROM mcr.microsoft.com/playwright:${PLAYWRIGHT_TAG}
 
 WORKDIR /app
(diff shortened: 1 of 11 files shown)


silva96:
I need to dig into the build failures, they are passing in my machine.


```bash
yarn test:browser src/tests/functional/navigation_tests.js:245

Running 2 tests using 2 workers

  ✓  1 [firefox] › src/tests/functional/navigation_tests.js:245:5 › following a same-origin link to a named anchor (1.2s)
  ✓  2 [chrome] › src/tests/functional/navigation_tests.js:245:5 › following a same-origin link to a named anchor (846ms)

  2 passed (3.6s)
```

silva96:
Update: I noticed that in Firefox in ubuntu the 

```js
scrollToAnchor(anchor) {
    const element = this.snapshot.getElementForAnchor(anchor)
    if (element) {
      this.scrollToElement(element) // added a breakpoint here, the screen is scrolled to the named anchor
      this.focusElement(element) // this one resets the scroll back to the top => need to investigate why / how to solve it
    } else {
      this.scrollToPosition({ x: 0, y: 0 })
    }
  }
 ```

brunoprietog (approved):
Thanks @silva96!

jorgemanrubia:
Thanks for handling this @silva96!

Issue #1408, opened 2025-06-02
[Feature]: view-transition - Respect prefers-reduced-motion


indykoning:
Currently we only listen to the view-transition meta tag in the html to activate view-transitions.

Users can pass their preference on animations with [prefers-reduced-motion](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)

We might want to respect it and disable view-transitions if that is the case

This check would see if motion is allowed
```js
window.matchMedia("(prefers-reduced-motion: no-preference)");
```

This check would enable reduced motion:
```js
window.matchMedia("(prefers-reduced-motion: reduce)");
```

PR #1409, opened 2025-06-02
Respect prefers-reduced-motion

indykoning:
Closes: https://github.com/hotwired/turbo/issues/1408

This PR checks wether the user has requested reduced motion via their OS/browser and disables view-transitions if that is the case, see: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

The way it is currently made it will assume [no-preference](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion#no-preference) by default in case a browser does not support the media query, matchMedia, or does not have window available

### DIFF
--- a/src/core/drive/page_snapshot.js
+++ b/src/core/drive/page_snapshot.js
@@ -71,7 +71,7 @@ export class PageSnapshot extends Snapshot {
   }
 
   get prefersViewTransitions() {
-    return this.headSnapshot.getMetaValue("view-transition") === "same-origin"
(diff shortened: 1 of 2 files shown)


brunoprietog (commented):
Nice @indykoning!

brunoprietog:
This is great for accessibility, we should certainly respect that. Do you think you can create a test too?

indykoning:
I've added a test which should handle this 🙂

brunoprietog (approved):
Thanks @indykoning!

Intrepidd:
Hijacking this to remind the meta tag used to check for view transition support is outdated, I started a PR to rename it in https://github.com/hotwired/turbo/pull/1380, but there are open questions, would love some feedback

Issue #1344, opened 2024-12-01
Scroll preservation persists globally after visiting a page with data-turbo-action="advance"


Jamie-Vandenbalck:
### Description:
When navigating to a page (e.g., the blog index) that uses data-turbo-action="advance", Turbo starts preserving scroll positions globally for all subsequent navigations, even on pages that do not use data-turbo-action="advance". The issue only occurs after visiting the page with this attribute—if the page is never visited, scroll behavior works as expected.

### Steps to Reproduce:

1. Add data-turbo-action="advance" to a link on the blog index page (/blog):
```
<turbo-frame id="blog_posts" src="{{ path('app.blog.index', routeParams) }}"data-turbo-action="advance">
```
2. Visit the /blog page.
3. Navigate to an unrelated page (e.g., /about) using a standard Turbo-enabled link: `<a href="/about">About</a>`
4. Observe that the scroll position on the /about page is preserved from the /blog page, even though data-turbo-action="advance" is not used on the /about page.

### Expected Behavior:
Scroll preservation should only apply:

- Within the context of the /blog page where data-turbo-action="advance" is explicitly used.
- Scroll position should reset appropriately when navigating to unrelated pages (e.g., /about).

### Actual Behavior:
- After visiting the /blog page and using data-turbo-action="advance", Turbo globally preserves scroll positions across all subsequent pages.
- This behavior persists for unrelated pages (e.g., /about), even if they do not use data-turbo-action="advance".
- If the /blog page is not visited, scroll behavior works as expected, resetting to the top for each new page.

### Environment:
Turbo Version: 8.0.11
Browser: Safari 18.1.1, Chrome 131
Operating System: macOS 15.1.1

jsuchal:
Confirming this. This makes the advance action completely unusable since every subsequent visit to any page will "be scrolled randomly". This is very confusing.

@Jamie-Vandenbalck did you around this?

Jamie-Vandenbalck:
@jsuchal 

Just dropping in a quick workaround that might help others until [hotwired/turbo#1344](https://github.com/hotwired/turbo/issues/1344) gets resolved.

In my preloader Stimulus controller (which runs on every page), I added the following snippet:

```
document.addEventListener('turbo:before-render', () => {
    window.scrollTo(0, 0); // Always reset scroll position
});
```

Full context below for reference:
```
import { Controller } from '@hotwired/stimulus';

export default class extends Controller {
    connect() {
        // Bind `setAriaBusy` to the current context
[...]

PR #1417, opened 2025-07-16
Fix scroll preservation issue with advance action

ernsheong:
NOTE: This PR was done by [Jules](https://jules.google/), an AI agent. Please review thoroughly.

When a page uses `data-turbo-action="advance"`, the scroll position is preserved for all subsequent navigations. This is because the `isPageRefresh` method in `src/core/drive/page_view.js` only considers `replace` actions as page refreshes.

This commit fixes the issue by updating the `isPageRefresh` method to also consider `advance` actions as page refreshes when the URL is the same. This ensures that scroll preservation is only applied to the page where `data-turbo-action="advance"` is used.

Fixes: https://github.com/hotwired/turbo/issues/1344


src/core/drive/page_view.js
```diff
@@ -56,7 +56,11 @@ export class PageView extends View {
   }
 
   isPageRefresh(visit) {
-    return !visit || (this.lastRenderedLocation.pathname === visit.location.pathname && visit.action === "replace")
+    return (
+      !visit ||
+      (this.lastRenderedLocation.pathname === visit.location.pathname &&
+        (visit.action === "replace" || visit.action === "advance"))
```
brunoprietog: The `replace` action is intentional. A page refresh should not impact the browsing history.

But even going further, page refreshes do not occur within the context of a frame. There is a confusion of concepts in this implementation. I'm going to close this, feel free to open another solution... (Or maybe don't use `advance` in Turbo frames)

ernsheong: Hi @brunoprietog I'm not surprised this isn't the fix, but can you help on issue #1344 ? It's a blocker for some of us.

Issue #1391, opened 2025-03-31
A full page reload after a redirected fetch request should use the response URL


m-vo:
When a page reload happens after a fetch request including a redirect, the wrong URL might be reloaded. By wrong I mean: As redirects have no body and only the last request triggers the reload, its URL should be used (IMHO).

Assume the following scenario: 
* The current page's head has a content hash `A`.
* Turbo fetches an URL `/foo` , the server redirects to `/bar` (which `fetch` will follow due to `redirect: "follow"`)
* On`/bar` the content hash is now `B`: we need a full page reload.
 
What happens: 
* Turbo makes the browser reload `/foo`.

What should happen: 
* Turbo makes the browser reload `/bar`, .i.e. the fetch response's `url` property.

In our case, `foo` is an endpoint that creates a new record and `bar` is the edit mask. When Turbo reloads the page the "create" endpoint is executed a second time. 🙈

m-vo:
Is there anything what I can do to move this forward? I can try to create a PR, but it would be nice to get some input about the intended behavior from anyone involved in the project first.

brunoprietog:
Yea makes sense, happy to see a PR!

m-vo:
@brunoprietog Please see https://github.com/hotwired/turbo/pull/1420 for my attempt at fixing the issue. 🙂

PR #1420, opened 2025-07-29
Make sure full page reloads use the response URL after a redirect

m-vo:
Fixes https://github.com/hotwired/turbo/issues/1391

With this change, we store the effective location (`visit.redirectedToLocation`)  in `visitRequestCompleted` when the fetch was redirected. This makes sure the redirect is not executed again in case of a forced reload (which might lead to problems, see #1391).

### DIFF
--- a/src/core/native/browser_adapter.js
+++ b/src/core/native/browser_adapter.js
@@ -36,6 +36,10 @@ export class BrowserAdapter {
 
   visitRequestCompleted(visit) {
     visit.loadResponse()
+
+    if(visit.response.redirected) {
+      this.location = visit.redirectedToLocation
+    }
   }
 
   visitRequestFailedWithStatusCode(visit, statusCode) {
--- a/src/core/native/browser_adapter.js
+++ b/src/core/native/browser_adapter.js
@@ -20,6 +20,8 @@ export class BrowserAdapter {
 
   visitStarted(visit) {
     this.location = visit.location
+    this.redirectedToLocation = null
+
     visit.loadCachedSnapshot()
     visit.issueRequest()
     visit.goToSamePageAnchor()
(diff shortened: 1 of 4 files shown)


packagethief (commented):
Thanks @m-vo.

I was surprised that this didn't already work. I expected that since `visit.followRedirect` performs a fresh visit, its location would reflect the redirected location and all would be right with the world! But after tracing the path I see that we render and invalidate the page before that occurs.

I think we could avoid that confusion (and limit any potential fallout) by not reassigning `this.location`. Instead we could record a `redirectedToLocation`, like `Visit` does, and use it only where relevant: when reloading the page.

```diff
diff --git i/src/core/native/browser_adapter.js w/src/core/native/browser_adapter.js
index 2192864..6b34ada 100644
--- i/src/core/native/browser_adapter.js
+++ w/src/core/native/browser_adapter.js
@@ -37,8 +37,8 @@ export class BrowserAdapter {
   visitRequestCompleted(visit) {
     visit.loadResponse()
 
-    if(visit.response.redirected) {
-      this.location = visit.redirectedToLocation
+    if (visit.response.redirected) {
+      this.redirectedToLocation = visit.redirectedToLocation
     }
   }
 
@@ -129,7 +129,7 @@ export class BrowserAdapter {
   reload(reason) {
     dispatch("turbo:reload", { detail: reason })
 
-    window.location.href = this.location?.toString() || window.location.href
+    window.location.href = (this.redirectedToLocation || this.location)?.toString() || window.location.href
   }

```

What do you think?

m-vo:
Thanks for the code suggestion @packagethief. I think this makes perfect sense and is less likely to have side effects than overwriting the location.

packagethief (approved):

PR #1423, opened 2025-08-05
Allow setting turbo-suppress-warning on body element

redross:
The warning previously made it look like placing the data-turbo-suppress-warning attribute on the body element would suppress it, but in fact, only the script tag was being checked for this attribute.

Now, data-turbo-suppress-warning on the body element will also suppress the warning.

The script tag check is kept unchanged to avoid breaking existing code.

The warning message has been updated to mention both possibilities (script tag and body tag) as valid places for
data-turbo-suppress-warning.

Before:
Screenshot from 2025-08-05 10-39-45

_Notice the data-turbo-suppress-warning attribute already present on the body tag, while the message suggests adding it_

After:
Screenshot from 2025-08-05 12-45-18

_The body tag no longer has the data-turbo-suppress-warning attribute, because the warning would no longer appear if it had it_

### DIFF
--- a/src/script_warning.js
+++ b/src/script_warning.js
@@ -1,12 +1,15 @@
 import { unindent } from "./util"
 ;(() => {
   let element = document.currentScript
+  let suppressWarningAttribute = 'data-turbo-suppress-warning'
   if (!element) return
-  if (element.hasAttribute("data-turbo-suppress-warning")) return
+  if (element.hasAttribute(suppressWarningAttribute)) return
 
   element = element.parentElement
   while (element) {
     if (element == document.body) {
+      if (element.hasAttribute(suppressWarningAttribute)) return
+
       return console.warn(
         unindent`
         You are loading Turbo from a <script> element inside the <body> element. This is probably not what you meant to do!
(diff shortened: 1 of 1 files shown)


redross:
This seems like a very minor/tiny PR, would appreciate if someone from the maintainers list could check it out - shouldn't take much of your time, thanks :bow:

Bonus - merging this would also close https://github.com/hotwired/turbo/pull/1424 (or the other way around) - so 2 PRs handled with a single review :smile:

redross:
Gentle ping @packagethief  :wave:

redross:
Another gentle ping @packagethief

redross:
@jorgemanrubia any chance you could take a look?

packagethief:
Thanks @redross. I'd prefer if we only supported `data-turbo-suppress-warning` on `script` elements. On the `body` element the instruction is ambiguous: suppress a warning about what? 

The error message is definitely wrong, though. I'd be happy to accept a change that fixes just that. The bug is that the warning message is assuming `element` is the current `script` element, but it's not. We're overwriting `element` while traversing upward to find the `body`. To get the `script` element reference, we can use `document.currentScript` instead.

Do you want to make that change?

redross:
@packagethief I also made a PR, that only fixes the error message. Maybe I'm missing something, but that seems to match the plan you mentioned above?

The other PR can be found here:
https://github.com/hotwired/turbo/pull/1424

P.S. Though I could also argue that, turbo-supress-warning doesn't give much information about "what it's supressing" even if it's on a script element. It might even be more "descriptive" when it's on the "body" element, since then we kind of have "supress turbo warning related to body" context/hint :smile: 

P.P.S. Sorry about the delay, I rarely use Github, didn't know I had notifications off.

packagethief:
Sorry, I missed the other PR in https://github.com/hotwired/turbo/pull/1424. I'll close this PR in favor of that one.

PR #1425, opened 2025-08-06
Fix UUID generation to include all hex digits

kaisersakhi:
Changed Math.random() * 15 to * 16 to properly generate 0-f range instead of missing the 'f' digit.


skmedix (approved):
Tested and LGTM

PR #1429, opened 2025-08-13
Don't ignore the browser's default for `a[href]` links

stefanvermaas:
Currently, Turbo ignores any link with a target attribute set. However, this means that Turbo ignores the default `target="_self"`.

Some design systems and/or web components set the target to "_self" even though it's technically not required. Because they set this target, Turbo ignores a link it shouldn't be ignoring.

This commit ensures we're including the `target="_self"` on a link element when looking up a link from a click target.

The new Polaris web components are a good example of web components that explicitly set the `target="_self"` even though it's not needed according to the browser specs. 

- The browser specifications on the default target: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/a#target
- Conversation with Shopify team: https://community.shopify.dev/t/pass-all-data-attributes-to-the-shadow-root-elements/19029/15

### DIFF
--- a/src/tests/functional/navigation_tests.js
+++ b/src/tests/functional/navigation_tests.js
@@ -266,10 +266,11 @@ test("following a same-origin [target] link", async ({ page }) => {
 })
 
 test("following a _self [target] link", async ({ page }) => {
-  const [popup] = await Promise.all([page.waitForEvent("popup"), page.click("#self-targeted-link")])
+  await page.click("#self-targeted-link")
+  await nextBody(page)
 
-  assert.equal(pathname(popup.url()), "/src/tests/fixtures/one.html")
-  assert.equal(await visitAction(popup), "load")
+  assert.equal(pathname(page.url()), "/src/tests/fixtures/one.html")
+  assert.equal(await visitAction(page), "advance")
 })
 
 test("following a same-origin [download] link", async ({ page }) => {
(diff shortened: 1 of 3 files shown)


stefanvermaas:
> Thanks @stefanvermaas. Would you mind adding a test for this behavior? We have an [existing test for `[target]` links](https://github.com/hotwired/turbo/blob/a057e6593ec236f02557b062c4f069b937e4719b/src/tests/functional/navigation_tests.js#L261-L266), so something along those lines.

I have added a test case for this.

src/util.js
```diff
@@ -247,7 +247,13 @@ export function doesNotTargetIFrame(name) {
...
 
 export function findLinkFromClickTarget(target) {
-  return findClosestRecursively(target, "a[href]:not([target^=_]):not([download])")
+  const link = findClosestRecursively(target, "a[href], a[xlink\\:href]")
+
+  if (!link) return null
+  if (link.hasAttribute("download")) return null
+  if (link.hasAttribute("target") && link.target !== "_self") return null
+
+  return link
```
stefanvermaas, line 256: Previously, Turbo used a pure CSS selector to find links: `"a[href]:not([target^=_]):not([download])"`

This worked for most standard HTML links, but it caused issues with:
  - Links inside SVG elements
  - Custom elements or shadow DOM links
  - Certain [target] values like "_self"

To fix this, we moved to a small JS-based selection function. It:
  - Walks up the DOM tree recursively
  - Handles shadow DOM hosts and slots
  - Matches both native <a> elements and custom link elements like <s-link> or <s-button>
  - Filters out links targeting iframes or downloads, while respecting "_self"

This approach is more robust for our use case and closely mimics real user clicks, while avoiding brittle CSS selectors.

src/tests/functional/navigation_tests.js
```diff
@@ -277,14 +285,20 @@ test("following a same-origin [download] link", async ({ page }) => {
 })
 
 test("following a same-origin link inside an SVG element", async ({ page }) => {
```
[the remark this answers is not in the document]
packagethief: Looks like it fails because the addition of the self-targeted link to the HTML fixture pushes the SVG elements down the page and outside the viewport. Your change seems fine, and more reliable.

src/util.js
packagethief: There's a subtle difference between this and the original. In the original, we'd only exclude elements that had an underscore-prefixed `target`, like `_blank` or `_top`. Now we'll exclude elements with any target value except `_self`.

As far as I know, this shouldn't matter. A non-keyword `target` should open the link in a new browsing context or specific frame, neither of which should use turbo for navigation. So I think we're good, I just wanted to point it out in case anyone else knows better!

packagethief (approved):

stefanvermaas:
Thanks for merging this @packagethief 🙏

PR #1437, opened 2025-09-12
rootMargin for turbo-frame

michaelkhabarov:
This PR introduces the `data-turbo-lazy-root-margin` attribute for `turbo-frame`.
The value is passed to the `IntersectionObserver` in `AppearanceObserver` as the `rootMargin` option.

The purpose of this option is to trigger lazy frame loading slightly before the user scrolls to the element. This improves user experience by making the site feel more responsive.

Currently, we are using a custom Stimulus controller with its own IntersectionObserver to achieve the same effect.

### DIFF
--- a/src/observers/appearance_observer.js
+++ b/src/observers/appearance_observer.js
@@ -4,12 +4,17 @@ export class AppearanceObserver {
   constructor(delegate, element) {
     this.delegate = delegate
     this.element = element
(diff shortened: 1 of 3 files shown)


michaelkhabarov:
@brunoprietog Hi! Could you please review this when you have a moment? 
It's a fairly small and straightforward change, and would be quite useful for us. Thank you!

qichunren:
This PR is very good!

Issue #1393, opened 2025-04-02
Morphing fails when refreshing after turbo stream


AlphonseSantoro:
Came across a bug when calling the action `turbo_stream.action(:refresh, nil, type: 'morph')`, when the form is submitted and this refresh action is called i get this error:
```
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'id')
    at removeElementFromAncestorsIdMaps
```
Which seems very much like this [issue in idiomorph](https://github.com/bigskysoftware/idiomorph/pull/126), any chance that we can bump idiomorph to 0.7.3 to get that fix inn?

htcarr3:
I think that there are a number of recent updates in idiomorph that solve issues around safe `id`-querying (I'm actually blocked from upgrading to 2.0.12+ due to something related). @botandrose thoughts on pulling in those recent changes to Turbo? I'm also happy to throw together a PR.

botandrose:
@htcarr3 hi Thomas. Yes, I'm getting ready to release a v0.7.4 which should be the last v0.7.x release, and fixes all known bugs. There's just one last pernicious bug I'm trying to nail down first. Once v0.7.4 is out the door, I'll open a PR here to upgrade Turbo to it.

htcarr3:
> [@htcarr3](https://github.com/htcarr3) hi Thomas. Yes, I'm getting ready to release a v0.7.4 which should be the last v0.7.x release, and fixes all known bugs. There's just one last pernicious bug I'm trying to nail down first. Once v0.7.4 is out the door, I'll open a PR here to upgrade Turbo to it.

Amazing! Thank you 🙏

AlphonseSantoro:
@botandrose Do you have an estimate on when you plan to release the fix?

botandrose:
Hi @AlphonseSantoro one month later I'm still trying to capture that bug. I think I should probably just release without it, at this point. I'll post back here when I do

botandrose:
@AlphonseSantoro https://github.com/bigskysoftware/idiomorph/releases/tag/v0.7.4

htcarr3:
@botandrose amazing! Can we cut a new minor version of turbo with the idiomorph version bump?

botandrose:
@htcarr3 Unfortunately, that's not up to me. But you could create a PR that bumps idiomorph and I bet they'd merge it!

AlphonseSantoro:
Lovely, after updating our project with this new release it works perfectly 🎉

PR #1441, opened 2025-09-30
bump idiomorph version to 0.7.4

htcarr3:
Should fix https://github.com/hotwired/turbo/issues/1393 as well as some other idiomorph bugs.

https://github.com/bigskysoftware/idiomorph/releases/tag/v0.7.4

https://github.com/bigskysoftware/idiomorph/compare/v0.7.2...v0.7.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "chai": "~4.3.4",
     "eslint": "^8.13.0",
     "express": "^4.18.2",
-    "idiomorph": "~0.7.2",
+    "idiomorph": "~0.7.4",
     "multer": "^1.4.2",
     "rollup": "^2.35.1"
   },
--- a/yarn.lock
+++ b/yarn.lock
@@ -1931,10 +1931,10 @@ iconv-lite@0.4.24:
   dependencies:
     safer-buffer ">= 2.1.2 < 3"
 
-idiomorph@~0.7.2:
-  version "0.7.2"
-  resolved "https://registry.yarnpkg.com/idiomorph/-/idiomorph-0.7.2.tgz#0c385343991c5ecf75472ecd58a4515fc7cdd47f"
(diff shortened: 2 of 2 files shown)


dvadell:
Hello everybody,

+1 for this please. We are seeing an exception when using turbo because of idiomorph's version. In this function from idiomorph 0.7.2:

```
function moveBeforeById(parentNode, id, after, ctx) {
      const target =
        /** @type {Element} - will always be found */
        (
          ctx.target.querySelector(#${id}) ||
            ctx.pantry.querySelector(#${id})
        );
      removeElementFromAncestorsIdMaps(target, ctx);
      moveBefore(parentNode, target, after);
      return target;
    }
```

Our `id` has square brackets, and `querySelector` needs the `id` escaped ( with backslashes or with `CSS.escape()` ).

In idiomorph's 0.7.4, they changed it to handle `id`s fine:

```
function moveBeforeById(parentNode, id, after, ctx) {
      const target =
        /** @type {Element} - will always be found */
        (     
          // ctx.target.id unsafe because of form input shadowing
          // ctx.target could be a document fragment which doesn't have getAttribute
          (ctx.target.getAttribute?.("id") === id && ctx.target) ||
            ctx.target.querySelector([id="${id}"]) ||
            ctx.pantry.querySelector([id="${id}"])
        );
      removeElementFromAncestorsIdMaps(target, ctx);
      moveBefore(parentNode, target, after);
      return target;
    }
```

Thanks

htcarr3:
@brunoprietog it looks like this is the only commit on main not already released. Thoughts on cutting a minor version release just for this?

jorgemanrubia:
Hey @htcarr3 I need to fix a regression we detected in a recent Turbo version. I will cut a release with that in place. Aiming to do it this week.

htcarr3:
@jorgemanrubia thank you!

jorgemanrubia:
New versions of turbo and turbo-rails released containing this one!

Issue #252, opened 2021-04-20
data-target the nearest frame. Proposal to introduce `data-turbo-target: '_parent'` (or is it `_self` ?)


coorasse:
I have a use case where the wrapping frame has `data-target: '_top'` set as default.
This is good since most of the links within this frame (it's a modal) should simply reload the page.
This works very well for 99% of my links, but (of course!) it happens that I have a modal with a list and pagination links.

These pagination links should of course target the wrapping frame instead of `_top`.

Now, it happens that when I render the pagination, I don't actually know the id of the nearest, wrapping, frame, so I am in the situation of having a `link_to '2', second_page, data: { turbo_target: ? }`.

Could we have a special keyword `_parent` to target the nearest frame? I mean...[we had it 20 years ago 😄 ](https://www.tagindex.net/html/frame/a_target.html) so I guess it makes sense, right?

dhh:
Would be happy to see this 👍

coorasse:
I blocked one day to work on this next week.

PR #1446, opened 2025-10-07
Add support for data-turbo-frame="_parent" in nested frames 

### LINKED ISSUE #1187: Support `data-turbo-frame="_parent"` in nested frames
## Use case

Consider the following HTML:

```html
<turbo-frame id="frame">
  <h2>Frame Navigation</h2>

  <a id="inside" href="/src/tests/fixtures/frame_navigation.html">Inside Frame</a>
  <a id="self" href="/src/tests/fixtures/frame_navigation.html" data-turbo-frame="_self">Self Frame</a>
  <a id="top" href="/src/tests/fixtures/frame_navigation.html" data-turbo-frame="_top">Top</a>
  <turbo-frame id="nested-frame">
    <a id="nested" href="/src/tests/fixtures/frame_navigation.html" data-turbo-frame="_parent">Parent Frame</a>
  </turbo-frame>
</turbo-frame>
```

Clicking "Parent Frame" should trigger the navigation to happen on the top-most `turbo-frame`.

This would be useful in situations where a nested frame might be rendered by a module that is not aware of the turbo-frame ID.

## Proposed implementation

* The `data-turbo-frame` attribute can accepts `_parent` as a value, similar to `_top`
* If the element is has two `turbo-frame` ancestors (first `_self`, then `_parent`) it will target the  `_parent` frame
* If the element only has one `turbo-frame` ancestor it will fall back to the same behaviour as `_top`

anthonyfranco:
## Summary

Implements support for `data-turbo-frame="_parent"` to allow links and forms within nested turbo-frames to target their immediate parent frame without knowing the parent's ID.

## Motivation

This addresses a common use case where:
- Components are rendered within frames but don't know the parent frame's ID
- Nested frames need to update their container without breaking out to `_top`
- Module-based architectures need flexible frame targeting

## Implementation

- Adds `_parent` as a special value alongside existing `_top` and `_self`
- Resolves `_parent` to the immediate parent frame's ID
- Falls back to full page visit when no parent frame exists or it's disabled
- Works for both link clicks and form submissions

### Basic nested frame targeting
```html
<turbo-frame id="modal">
  <turbo-frame id="search-results">
    <a href="/items/123" data-turbo-frame="_parent">
      <!-- Search component doesn't need to know parent's ID -->
    </a>
  </turbo-frame>
</turbo-frame>
```

### Multiple nesting levels
```html
<turbo-frame id="outer">
  <turbo-frame id="middle">
    <turbo-frame id="inner">
      <a href="/path" data-turbo-frame="_parent">Updates middle</a>
    </turbo-frame>
  </turbo-frame>
</turbo-frame>
```

## Related Issues

Resolves #1187
References #252

### DIFF
--- a/src/core/frames/frame_controller.js
+++ b/src/core/frames/frame_controller.js
@@ -10,8 +10,7 @@ import {
   markAsBusy,
   uuid,
   getHistoryMethodForAction,
-  getVisitAction,
-  findClosestRecursively
+  getVisitAction
 } from "../../util"
 import { FormSubmission } from "../drive/form_submission"
 import { Snapshot } from "../snapshot"
@@ -433,13 +432,9 @@ export class FrameController {
 
   #findFrameElement(element, submitter) {
     const id = getAttribute("data-turbo-frame", submitter, element) || this.element.getAttribute("target")
-    if (id === "_parent") {
-      const parentFrame = findClosestRecursively(this.element.parentElement, "turbo-frame")
-      if (parentFrame) {
-        return parentFrame
-      }
-    }
-    return getFrameElementById(id) ?? this.element
+    const target = this.#getFrameElementById(id)
+
+    return target instanceof FrameElement ? target : this.element
   }
 
   async extractForeignFrameElement(container) {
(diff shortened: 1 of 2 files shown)


seanpdoyle:
Thank you for opening this PR! I've pushed some changes to the test coverage to use `expect` rather than `assert`.

I've also incorporated your proposed changes to the implementation into some existing methods.

PR #1461, opened 2025-11-07
Frame Functional Tests: Replace chai with Playwright

seanpdoyle:
Follow-up to https://github.com/hotwired/turbo/pull/1454

This commit continues to migrate from `assert` usage toward Playwright's `expect`. This commit focuses on the
`src/tests/functional/frame_tests.js` file.

Utilize Playwright's built-in [assertions][] so that functional test assertions are consistent. The `chai`-based assertions provide similar coverage, but not much is gained for the suite by having two competing styles of assertion.

The majority of this diff replaces `assert.equal` with `expect(...).toEqual()`. In some cases, text comparison is replaced with Playwright's [toHaveText][], since that assertion bakes in synchronization and waiting support, and has the ability to compare contents in a way that ignores spacing differences.

[assertions]: https://playwright.dev/docs/test-assertions
[toHaveText]: https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-have-text

### DIFF
--- a/src/tests/functional/frame_tests.js
+++ b/src/tests/functional/frame_tests.js
@@ -1,8 +1,6 @@
 import { test, expect } from "@playwright/test"
-import { assert, Assertion } from "chai"
 import {
   attributeForSelector,
-  hasSelector,
   listenForEventOnTarget,
   nextAttributeMutationNamed,
   noNextAttributeMutationNamed,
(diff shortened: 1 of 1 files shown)


packagethief (approved):
Thanks @seanpdoyle!

PR #1475, opened 2025-11-22
Fix noscript style evaluation during navigation (#1464)

### LINKED ISSUE #1464: Rails: Link navigation causes `<style>` inside `<noscript>` to be evaluated
I ran into a strange issue, where a link click causes the `<noscript>` tag and its child `<style>` on my page to be parsed and evaluated.
Adding an `id` and `data-turbo-permanent="true"` is a workaround for this, but only works for successful visits.
On the error page, the content of `<noscript>` remains invisible, but it contains a `<style>` tag, which is evaluated and causes the content of the page to be set to `display: none;`.
The errors on my website are not static; they are rendered within the default layout and attempt to keep navigation/breadcrumbs as intact as possible.
However, when Turbo renders 4xx/5xx errors, it seems to re-evaluate everything, even elements that are marked as permanent.
My browser also warns me that there may be no more than one importmap, which supports my "re-evaluate everything" assumption.

My workaround is this ugly hack:
```
	const noscriptTag = event.detail.newBody.querySelector("noscript");
	if (noscriptTag && noscriptTag.querySelector("style")) noscriptTag.innerText = noscriptTag.innerHTML;
});
```

Also, the warning about a second importmap being loaded remains, which is also not nice.

Surely this can't be the intended behavior? What is causing the duplicate importmap?

yujiteshima:
# Pull Request

**Title:** Fix noscript style evaluation during navigation

## Summary

Fixes #1464

This PR prevents Turbo from incorrectly evaluating `<style>` tags nested within `<noscript>` elements during page navigation.

## Problem

Turbo's page rendering mechanism was evaluating content inside `<noscript>` elements, causing:
- Styles intended only for non-JavaScript environments to be applied
- Content to become hidden when `display: none` styles were applied during navigation

## Solution

The fix deactivates `<noscript>` elements by converting their `innerHTML` to `textContent` in two critical places:

1. **PageRenderer** (`page_renderer.js`): Before adopting the new body element
2. **PageSnapshot** (`page_snapshot.js`): When creating cached snapshots

This conversion prevents browsers from parsing nested HTML/CSS while preserving the content as plain text.

## Testing

Added a new test case that verifies styles inside `<noscript>` elements are not evaluated during Turbo navigation.

## Reproduction Projects

- Unfixed version: https://github.com/yujiteshima/hotwire-turbo-issue-1464-repro
- Fixed version: https://github.com/yujiteshima/hotwire-turbo-issue-1464-repro-fix

### DIFF
--- a/src/tests/fixtures/body_noscript.html
+++ b/src/tests/fixtures/body_noscript.html
@@ -9,13 +9,8 @@
   <body>
     <h1>Body noscript test</h1>
     <p><a id="back-link" href="/src/tests/fixtures/rendering.html">Go back</a></p>
-    <div id="visible-content">This content should be visible</div>
     <noscript>
-      <style>
-        #visible-content {
-          display: none;
-        }
-      </style>
+      <link rel="stylesheet" type="text/css" href="/src/tests/fixtures/noscript.css">
     </noscript>
   </body>
 </html>
--- a/src/tests/functional/rendering_tests.js
+++ b/src/tests/functional/rendering_tests.js
@@ -260,11 +260,11 @@ test("does not evaluate head stylesheet elements inside noscript elements", asyn
 })
 
 test("does not evaluate body stylesheet elements inside noscript elements", async ({ page }) => {
+  expect(await isNoscriptStylesheetEvaluated(page)).toEqual(false)
+
   await page.click("#body-noscript-link")
   await nextEventNamed(page, "turbo:render")
-
-  const isVisible = await page.locator("#visible-content").isVisible()
-  expect(isVisible).toEqual(true)
+  expect(await isNoscriptStylesheetEvaluated(page)).toEqual(false)
 })
 
 test("waits for CSS to be loaded before rendering", async ({ page }) => {


yujiteshima:
## Note on CI Failure

The CI failure appears to be from a pre-existing flaky test, not related to this PR's changes.

**The flaky test:** `navigation_tests.js:451` - "does not fire turbo:load twice after following a redirect"

I tested this on `upstream/main` (before my changes) and got the same intermittent failures - it passes about 80% of the time. The test added in this PR for the noscript fix passes consistently (5/5 runs).

The noscript fix itself is working correctly and the new test coverage is stable. Happy to re-run CI or provide more details if needed!

packagethief (approved):
Thanks @yujiteshima. This looks good. Just one (possibly silly) question. If turbo is controlling navigation we know that JavaScript is enabled, so any `<noscript>` element is irrelevant anyway. Instead of converting the element contents to text, could we just remove the element?

```js
removeNoscriptElements() {
  for (const noscriptElement of this.newElement.querySelectorAll("noscript")) {
    noscriptElement.remove()
  }
}
```

yujiteshima:
@packagethief Thank you for the suggestion—not silly at all! I agree that removing `<noscript>` elements is a better approach.

Since Turbo is running, JavaScript is demonstrably enabled, making `<noscript>` elements unnecessary. There's no practical reason for code to depend on `<noscript>` elements in a JavaScript-enabled environment, so removing them keeps the DOM cleaner and aligns better with Turbo's philosophy of progressive enhancement.

I'll update the implementation to use `removeNoscriptElements()` instead of `deactivateNoscriptElements()`.

yujiteshima:
@packagethief Updated! The implementation now removes `<noscript>` elements entirely using `removeNoscriptElements()`. 

The removal happens after `document.adoptNode()` to ensure consistent behavior across browsers—Firefox requires the element to be adopted into the current document before removal to prevent unintended style evaluation.

All tests are passing on both Chrome and Firefox.

PR #1487, opened 2026-01-18
Reduce flakiness in several tests

radanskoric:
I decided to look at some of the tests that are flaking. The selection of tests which I improved are a bit random because I basically ran the test suite a few times on my machine and observed which ones happened to flake and then looked into which ones I can fix. 

The commit message has all of them listed together with the explanation of the problem and the fix but for easier PR review, I'll copy paste the problem and fix as a comment under each test.

These are not all of the flaky tests, it's just the ones I managed to fix in this coding session. When I'll have more time I'll look at a few more and open a new PR.

### DIFF
--- a/src/tests/functional/form_submission_tests.js
+++ b/src/tests/functional/form_submission_tests.js
@@ -1018,7 +1018,9 @@ test("link method form submission inside frame submits a single request", async 
 
 test("link method form submission targeting frame submits a single request", async ({ page }) => {
   let requestCounter = 0
-  page.on("request", () => requestCounter++)
+  page.on("request", (request) => {
+    if (request.url().includes("/__turbo/redirect")) requestCounter++
+  })
 
   await page.click("#turbo-method-post-to-targeted-frame")
 
@@ -1026,7 +1028,8 @@ test("link method form submission targeting frame submits a single request", asy
 
   await noNextEventNamed(page, "turbo:before-fetch-request")
   expect(fetchOptions.method, "[data-turbo-method] overrides the GET method").toEqual("POST")
-  expect(requestCounter, "submits a single HTTP request then follows a redirect").toEqual(2)
+  await expect(page.locator("#hello h2")).toHaveText("Hello from a frame")
+  expect(requestCounter, "submits a single HTTP request").toEqual(1)
 })
 
 test("link method form submission inside frame", async ({ page }) => {
--- a/src/tests/functional/frame_tests.js
+++ b/src/tests/functional/frame_tests.js
@@ -152,10 +152,10 @@ test("successfully following a link to a page without a matching frame dispatche
 test("successfully following a link to a page without a matching frame shows an error and throws an exception", async ({
   page
 }) => {
-  let error = undefined
-  page.once("pageerror", (e) => (error = e))
-
-  await page.click("#missing-frame-link")
+  const [error] = await Promise.all([
+    page.waitForEvent("pageerror"),
+    page.click("#missing-frame-link")
+  ])
 
   await expect(page.locator("#missing")).toHaveText("Content missing")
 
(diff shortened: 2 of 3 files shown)


src/tests/functional/form_submission_tests.js
```diff
@@ -1018,15 +1018,18 @@ test("link method form submission inside frame submits a single request", async
...
 
   await page.click("#turbo-method-post-to-targeted-frame")
 
   const { fetchOptions } = await nextEventNamed(page, "turbo:before-fetch-request")
 
   await noNextEventNamed(page, "turbo:before-fetch-request")
   expect(fetchOptions.method, "[data-turbo-method] overrides the GET method").toEqual("POST")
-  expect(requestCounter, "submits a single HTTP request then follows a redirect").toEqual(2)
+  await expect(page.locator("#hello h2")).toHaveText("Hello from a frame")
+  expect(requestCounter, "submits a single HTTP request").toEqual(1)
```
radanskoric: PROBLEM: Redirect accounting differs between browsers and Firefox sometimes won't count the redirect as a request, returning just 1.

FIX: Stop checking the redirect by counting requests. Instead assert that redirect happened by checking that the page contains the content from the redirect target. And then directly that we did a single request (as the test describes) by counting only the requests.

src/tests/functional/frame_tests.js
```diff
@@ -152,10 +152,10 @@ test("successfully following a link to a page without a matching frame dispatche
...
   page
 }) => {
-  let error = undefined
-  page.once("pageerror", (e) => (error = e))
-
-  await page.click("#missing-frame-link")
+  const [error] = await Promise.all([
+    page.waitForEvent("pageerror"),
+    page.click("#missing-frame-link")
+  ])
```
radanskoric: PROBLEM: The assertion on the error can run before the error setting callback runs.

FIX: Wait for the error to be set and the page click to finish before continuing with assertions.

src/tests/functional/frame_tests.js
```diff
@@ -184,10 +184,10 @@ test("failing to follow a link to a page without a matching frame dispatches a t
...
   page
 }) => {
-  let error = undefined
-  page.once("pageerror", (e) => (error = e))
-
-  await page.click("#missing-page-link")
+  const [error] = await Promise.all([
+      page.waitForEvent("pageerror"),
+      page.click("#missing-page-link")
+  ])
```
radanskoric: PROBLEM: The assertion on the error can run before the error setting callback runs.

FIX: Wait for the error to be set and the page click to finish before continuing with assertions.

src/tests/functional/navigation_tests.js
```diff
@@ -449,15 +449,21 @@ test("double-clicking on a link", async ({ page }) => {
...
   await nextBeat() // 301 redirect response
 
-  expect(await noNextEventNamed(page, "turbo:load")).toEqual(true)
-
   await nextBeat() // 200 response
   await nextBody(page)
   await nextEventNamed(page, "turbo:load")
+  await nextBeat()
+
+  expect(await page.evaluate(() => window.turboLoadCount)).toEqual(1)
```
radanskoric: PROBLEM: The the event can sometimes arrive slightly before the check.

FIX: By counting the events over the whole scenario is more robust than checking that it happens at a specific moment and more closely matches the purpose of the test.

src/tests/functional/rendering_tests.js
```diff
@@ -57,7 +57,7 @@ test("reloads when tracked elements change", async ({ page }) => {
   )
 
   await page.click("#tracked-asset-change-link")
-  await nextEventNamed(page, "turbo:load")
+  await page.waitForURL("**/tracked_asset_change.html")
```
radanskoric: PROBLEM: Since we're triggering a full page reload it is possible to try to evaluate on the page at the moment of full page navigation. This can then lead to a race condition where you triggera check on the page as it is being navigated away and the actual evaluation happens when its context was already destroyed. This then throws "Execution context was destroyed, most likely because of a navigation". This happens when trying to wait for `turbo:load` event to assert that navigation has completed.

FIX: Use the native playwright `page.waitForURL` to wait for the new page to be loaded. This method doesn't rely on page evaluation and is resistant to navigation issues.

Issue #1479, opened 2025-12-13
Prefetch causes wrong pages to be rendered


fritzmg:
If a user clicks on a link, and while that page is loading clicks on another link and that page happens to load faster than the previous one, the former page will still be loaded in after the latter one has already loaded. You can reproduce it in this example: https://www.inspiredminds.at/turbo.php. "Page 1" has an artificial loading time of 4000ms while "Page 2" only has a loading time of 500ms. To reproduce do the following:

1. Go to https://www.inspiredminds.at/turbo.php
2. Hover & click on _Page 1_.
3. Then immediately click on _Page 2_.
4. It will show the content of "Page 2".
5. Then it will show the content of "Page 1" - but the URL also has changed to `/page1.php`.
6. Click on _Home_.
7. Hover & click on _Page 1_.
8. Then immediately click on _Page 2_.
9. It will show the content of "Page 2".
10. Then it will show the content of "Page 1" - but the URL stays at `/page2.php`.

[Video](https://github.com/user-attachments/assets/b907fb74-52d4-4f52-bdc6-98c460bda694)

This does not happen if you add 

```html
<meta name="turbo-prefetch" content="false">
```

because then when clicking on _Page 2_, the request to _Page 1_ is immediately cancelled by Turbo and thus can never be rendered. But with prefetch enabled, no cancellation ever takes place and pages seem to be simply rendered in the order they are finished instead of the order they were requested.

m-vo:
I think we need to cancel pending fetches and stop active ones from being processed if a newer one was queued. This is somewhat similar to the findings in #1372 (that were only improved but not completely solved).

PR #1490, opened 2026-01-24
Record and cancel pending prefetch requests

### LINKED ISSUE #1244: Turbo does not cancel pending requests that were prefetched on consecutive link clicks
In a situation when page loads or internet connection are slow, clicking multiple links within a short period of time should only load the last link clicked. 
When prefetch is disabled, Turbo works correctly by canceling pending requests issued by previous clicks. 

When prefetch is enabled however, all requests issued by clicks will run to completion. The page will re-render on each completed request resulting in flickering. And the last link clicked is not guaranteed to be the one visible when all requests complete.

In the following example, there are 3 links that take 2s,6s, and 4s to load. Clicking them consecutively in that order will result in the slowest page being opened last, even though it was a second click, not the last one.

<details>
<summary>Screen recording of example</summary>

https://github.com/hotwired/turbo/assets/12209038/285444cb-d201-4d60-bf50-f64bb6f28e8d

</details>

<details>
<summary>Example with Rails</summary>

```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  gem "puma"
  gem "propshaft"
  gem "turbo-rails"
end

  config.load_defaults "7.1"
  
  config.turbo.draw_routes = false
  
  routes.append do
    root to: "application#index"
  end
end

$template = DATA.read

  def index
    sleep(params[:sleep].to_i) if params[:sleep].present?
    render inline: $template, formats: :html
  end
end

App.initialize!

Rack::Server.new(app: App, Port: 3000).start

__END__
<html>
  <head>
    <script type="importmap">
      {
        "imports": {
          "@hotwired/turbo-rails": "<%= asset_path("turbo.js") %>"
        }
      }
    </script>

    <script type="module">
      import "@hotwired/turbo-rails"
    </script>

  </head>

  <body>
    <h1>page <%= (params[:page].present? ? params[:page] : 'root').upcase %></h1>

    <%= link_to "root", root_path %>
    
    <div data-turbo-prefetch="false">
      <h2>without prefetch</h2>

      <ol>
        <li><%= link_to "slow", root_path(page: 'slow', sleep: 2) %></li>
        <li><%= link_to "slowest", root_path(page: 'slowest', sleep: 6) %></li>
        <li><%= link_to "slower", root_path(page: 'slower', sleep: 4) %></li>
      </ol>
    </div>

    <div>
      <h2>with prefetch</h2>

      <ol>
        <li><%= link_to "slow", root_path(page: 'slow', sleep: 2) %></li>
        <li><%= link_to "slowest", root_path(page: 'slowest', sleep: 6) %></li>
        <li><%= link_to "slower", root_path(page: 'slower', sleep: 4) %></li>
      </ol>
    </div>
  </body>
</html>
```

</details>

m-vo:
Fixes #1244, #1372, #1479

Inspired by https://github.com/hotwired/turbo/issues/1244#issuecomment-3756194305 by @SHPmax

### What's this?

Turbo can speed up page loads by prefetching requests. If the connection is slow or the server takes some time to respond, this currently has two side-effects, though:

1) Requesting a resource (that takes longer to load than the prefetch delay) multiple times - e.g. by quickly hovering back and forth over a link - will result in multiple requests to this same resource, that are running in parallel.
    
    https://github.com/user-attachments/assets/591b5667-db79-4479-b7b0-05107163fd50

2) If you request a resource that loads slow, then one that loads fast, the promise of the fast one will fulfill first (displaying a page for instance), then the slow one - that is still running - kicks in and overwrites the result.
   
    https://github.com/user-attachments/assets/526199763-b907fb74-52d4-4f52-bdc6-98c460bda694

    (_taken from #1479_)

### Solution

This PR makes the `link_prefetch_observer` record the last fetch request and cancel the previous one to the same URL. On top of that, during `turbo:visit`, all currently recorded requests but the one for the current URL are cancelled as well. This way, there now can only be one active prefetch request per URL and never a pending one of a different URL a visit happens.

I added test cases to outline the issue. You can see this PR in effect by commenting out the two added `fetchRequest.cancel()` statements in the `link_prefetch_observer`.

/cc @fritzmg, @ilyadh

### DIFF
--- a/src/observers/link_prefetch_observer.js
+++ b/src/observers/link_prefetch_observer.js
@@ -11,6 +11,7 @@ import { prefetchCache, cacheTtl } from "../core/drive/prefetch_cache"
 export class LinkPrefetchObserver {
   started = false
   #prefetchedLink = null
+  #pendingFetchRequests = new Map()
 
   constructor(delegate, eventTarget) {
     this.delegate = delegate
@@ -40,6 +41,7 @@ export class LinkPrefetchObserver {
     })
 
     this.eventTarget.removeEventListener("turbo:before-fetch-request", this.#tryToUsePrefetchedRequest, true)
+    this.eventTarget.removeEventListener("turbo:before-visit", this.#cancelPendingFetchRequests, true)
     this.started = false
   }
 
@@ -54,6 +56,7 @@ export class LinkPrefetchObserver {
     })
 
     this.eventTarget.addEventListener("turbo:before-fetch-request", this.#tryToUsePrefetchedRequest, true)
+    this.eventTarget.addEventListener("turbo:before-visit", this.#cancelPendingFetchRequests, true)
     this.started = true
   }
 
@@ -80,8 +83,29 @@ export class LinkPrefetchObserver {
 
         fetchRequest.fetchOptions.priority = "low"
 
+        // Track pending fetch requests belonging to the current URL
+        const url = location.href
+        if (!this.#pendingFetchRequests.has(url)) {
+          this.#pendingFetchRequests.set(url, [])
+        }
+        this.#pendingFetchRequests.get(url).push(fetchRequest)
+
         prefetchCache.putLater(location, fetchRequest, this.#cacheTtl)
       }
+    }
+  }
+
+  #cancelPendingFetchRequests = (event) => {
+    if (event.detail.fetchRequest) return
+
+    for (const [url, fetchRequests] of this.#pendingFetchRequests.entries()) {
+      const isSameUrl = url === event.detail.url
+
+      for (const fetchRequest of isSameUrl ? fetchRequests.slice(0, -1) : fetchRequests) {
+        fetchRequest.cancel()
+      }
+
+      this.#pendingFetchRequests.set(url, isSameUrl ? [fetchRequests[fetchRequests.length - 1]] : [])
     }
   }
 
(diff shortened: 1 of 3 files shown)


m-vo:
I found a way of further improving this. I will update the PR accordingly.

m-vo:
Just to illustrate the difference. This is the behavior **with** this PR:

https://github.com/user-attachments/assets/7f322bcf-9b40-4979-913f-fab9fae427d4

* In the first part of the video, I hovered over "Page 1" (which issues a prefetch request), clicked on it, then made a visit to "Page 2" which canceled the prefetch to the slow loading resource. 

* In the second part [0:10], I hovered over both links back and forth to trigger multiple prefetch requests. For each resource only one is now running at a time.

fritzmg (approved):
The built resources are also deployed on https://www.inspiredminds.at/turbo.php - and I can confirm that this fixes #1479 and I could not find any other issues 👍

Toflar (approved):
We also had to disable prefetch because of the too many requests and "page result swapping" issues. This PR will allow us to re-enable it! Thank you @m-vo ❤️

m-vo:
Is there anything I could do to move this forward?

/cc @packagethief maybe?

t1ll:
Thank you @m-vo for the fix; can't wait for it to be merged

zoglo:
I've had the same issue happen to me on GitHub quite a bunch of times when my network was slow.
I guess it's negligible that the URL and/or the content were different when not knowing what causes the issue.

It's just more obvious when knowing that it's a possible bug in the Turbo library and does not happen on faster connections / good optimized pages as often.

leofeyer:
@packagethief @jorgemanrubia would you be open to merging this PR? Or should this stay an opt-in solution that users implement themselves when needed, as outlined in https://github.com/hotwired/turbo/issues/1244#issuecomment-3756194305?

zoglo:
@packagethief Is there anything we can do to speed up the process of reviewal and merging this PR?

maikschneider (approved):

TrevorHinesley:
Looks to be approved, is there any reason this can't be moved along? This is a pretty common wrinkle that's encountered, would be great to get it merged in.

Issue #1492, opened 2026-01-27
turbo:load fires very late when Turbo is loaded with defer (misses interactive state)


amohamed97:
When Turbo is loaded using `defer`, the `turbo:load` event can fire **very late** (close to `window.onload`) because Turbo only triggers it in response to `readystatechange` *transitions*.
If Turbo initializes **after** the document has already reached `readyState = "interactive"`, it misses that transition and waits until `"complete"` instead.

This causes JS that depends on `turbo:load` to be delayed until all deferred/async scripts finish loading, which can block page interactivity.

---

### Why this matters

We migrated from **Turbolinks → Turbo**, and this regression started appearing after the migration.

* `turbolinks:load` fired on `DOMContentLoaded`
* It worked reliably even when Turbolinks itself was deferred
* With Turbo, deferring the script can unintentionally tie `turbo:load` to `readyState = "complete"` / `window.onload`

In real production cases, this has caused:

* Frozen or non-interactive UI
* Navigation blocked for users

---

### Root cause (as I understand it)

Turbo triggers `turbo:load` only when it *observes* `readystatechange` changes (`interactive` or `complete`).
It does **not** check the current `document.readyState` when initializing.

So if Turbo loads after `interactive`, it will always wait for `complete`.

---

### Questions

1. Is this behavior intentional / expected when Turbo is deferred?
2. Is Turbo expected to work correctly when loaded using defer, or is synchronous loading assumed?
3. Is there a recommended event for “DOM ready + Turbo-safe initialization” that doesn’t depend on `readyState = "complete"`?
4. Would it make sense for Turbo to trigger `turbo:load` immediately if `document.readyState` is already `"interactive"`?

Thanks for the great work on Turbo 🙏

PR #1493, opened 2026-01-29
Trigger turbo load on domcontentloaded

jjla26:
- Added console logs to track the start, state changes, and interactions within the PageObserver.
- Implemented handling for scripts loaded with the defer attribute, ensuring that DOMContentLoaded is awaited before firing turbo:load.

### DIFF
--- a/src/observers/page_observer.js
+++ b/src/observers/page_observer.js
@@ -39,7 +39,6 @@ export class PageObserver {
 
   interpretReadyState = () => {
     const { readyState } = this
-    
     if (readyState == "interactive") {
       this.pageIsInteractive()
     } else if (readyState == "complete") {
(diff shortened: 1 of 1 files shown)


brunoprietog:
Can you elaborate a bit more on the issue to have a better understanding? Also, I don't think it's appropriate to put so many console logs in production.

amohamed97:
@brunoprietog It's addressing https://github.com/hotwired/turbo/issues/1492

jjla26:
@brunoprietog Apologies for the confusion. This PR was originally intended for our internal fork.

However, since it's here, I'd like to explain the problem we encountered:

When Turbo is loaded with the defer attribute, the turbo:load event doesn't fire at the optimal time.
Current behavior:
Deferred scripts execute when document.readyState is "interactive" (after HTML parsing completes).
The PageObserver.start() method only adds a readystatechange listener for future state changes.
If the script loads when readyState is already "interactive" (which is the normal for a deferred script), it only catches the transition to "complete".
This means turbo:load fires after all assets (CSS, images, fonts, tracking scripts) have loaded causing a significant delay
Real-world impact:
In our application, this delay meant interactive features weren't available until all page assets loaded, creating a poor user experience.

jjla26:
As @amohamed97 mentioned, we're attempting to address https://github.com/hotwired/turbo/issues/1492. However, we'd appreciate your feedback on whether there's an intentional reason for the current behavior, or if perhaps we're misunderstanding how Turbo is meant to be loaded in modern bundler setups.

PR #1495, opened 2026-02-02
Same-page anchor: Account for `<a>` nested in `<svg>`

seanpdoyle:
Follow-up to [#1285][]

While an `<a>` element is typically represented by an [HTMLAnchorElement][], an `<a>` nested inside an `<svg>` element is instead an [SVGAElement][].

Similarly, while [HTMLAnchorElement.href][] returns a `String`, the [SVGAElement.href][] returns an instance of [SVGAnimatedString][], which cannot be supported by neither calls to [String.startsWith][] nor [RegExp.test][].

This commit adds explicit test coverage for a same-page navigation from an `<a>` nested within an `<svg>`. To ensure that the value is a `String`, replace `.href` with `getAttribute("href")`. This restores the implementation to be closer to what was defined in [e591ea9e][].

*Note*

It's worth mentioning that both of these new tests pass _without_ the `page.on("pageerror")` parts are omitted from the coverage. That signals to me that either:

1. the intended behavior still functions when JavaScript errors are raised, 
2. the test coverage does not genuinely reflect real-world use cases

I'm hopeful that it's 1., and that the test coverage _with_ the explicit JavaScript error tracking sufficiently guards against both 1. and 2. being true.

[#1285]: https://github.com/hotwired/turbo/pull/1285
[HTMLAnchorElement]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement
[SVGAElement]: https://developer.mozilla.org/en-US/docs/Web/API/SVGAElement
[HTMLAnchorElement.href]: https://developer.mozilla.org/en-US/docs/Web/API/HTMLAnchorElement/href
[SVGAElement.href]: https://developer.mozilla.org/en-US/docs/Web/API/SVGAElement/href
[SVGAnimatedString]: https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedString
[String.startsWith]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
[RegExp.test]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test
[e591ea9e]: https://github.com/hotwired/turbo/pull/1285/changes/e591ea9eb087a308f8d6997dc3ffe051058e3ec9#diff-9c0ec1b0a889e599f3ff81590864dd9dc65684a86b41f1da75acc53fafed12e3R251-R256

### DIFF
--- a/src/tests/fixtures/navigation.html
+++ b/src/tests/fixtures/navigation.html
@@ -47,6 +47,8 @@
       <p><a id="bare-target-link" href="/src/tests/fixtures/one.html" target>Bare target attribute link</a></p>
       <p><a id="same-origin-download-link" href="/intentionally_missing_fake_download.html" download="x.html">Same-origin download link</a></p>
       <svg width="600" height="100" viewbox="-300 -50 600 100"><text><a id="same-origin-link-inside-svg-element" href="/src/tests/fixtures/one.html">Same-origin link inside SVG element</a></text></svg>
+      <svg width="600" height="100" viewbox="-300 -50 600 100"><text><a id="same-origin-targeted-link-inside-svg-element" href="/src/tests/fixtures/one.html" target="_blank">Same-origin targeted link inside SVG element</a></text></svg>
+      <svg width="600" height="100" viewbox="-300 -50 600 100"><text><a id="same-page-link-inside-svg-element" href="#main">Same-page link inside SVG element</a></text></svg>
       <svg width="600" height="100" viewbox="-300 -50 600 100"><text><a id="cross-origin-link-inside-svg-element" href="about:blank">Cross-origin link inside SVG element</a></text></svg>
       <svg width="600" height="100" viewbox="-300 -50 600 100"><text><a id="same-origin-anchored-svg-link" href="#main">SVG link with hash-only href</a></text></svg>
       <p><a id="same-origin-get-link-form" href="/src/tests/fixtures/navigation.html?a=one&b=two" data-turbo-method="get">Same-origin data-turbo-method=get link</a></p>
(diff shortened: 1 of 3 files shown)


domchristie:
Related: Astro's ClientRouter approach: https://github.com/withastro/astro/blob/dbf71c021e3adf060c34e6f268cf1b5cd480233d/packages/astro/components/ClientRouter.astro#L82-L83

seanpdoyle:
@domchristie interesting! Thanks for sharing that context. I'll copy-paste it here to keep those details as part of this conversation:

```js
if (
  !(link instanceof HTMLAnchorElement) &&
  !(link instanceof SVGAElement) &&
  !(link instanceof HTMLAreaElement)
)
  return

// …

const linkTarget = link instanceof HTMLElement ? link.target : link.target.baseVal;
const href = link instanceof HTMLElement ? link.href : link.href.baseVal;
```

I like that approach in that there are some class-level checks. However, I worry about a future need of adding additional supported classes (in case we miss one now, or in case the "link" is an `<a>`-like custom HTML element of some kind).

For now, I think I prefer the `getAttribute`-based approach for two reasons: 1) it's closer to the original code, 2) it doesn't limit the classes supported.

Am I missing out on something about the alternative?

domchristie:
> Am I missing out on something about the alternative?

No I don't think so. The use of `baseVal` was interesting to me, but I think `getAttribute` works well for both (as it does for the `target` attribute. I feel like some of the Astro code is to satisfy TypeScript, particularly the `instanceof` guards.

packagethief (approved):
Thanks for looking into this @seanpdoyle 🙏

fschwahn:
> I'm hopeful that it's 1., and that the test coverage with the explicit JavaScript error tracking sufficiently guards against both 1. and 2. being true.

I went back and tested this, and the claim that this broke links in SVGs was wrong, I'm sorry about that. In fact, the links still behave as intended, there's just noise from the errors in the browser console.

So I think (at least in my case) it is indeed 1.

PR #1498, opened 2026-02-04
Fix some flaky tests that were failing in recent runs

rosa:
- Fix flaky frame test by using targeted event check: change `noNextEventNamed` to `noNextEventOnTarget` for the frame form navigation test. The previous check would pick up unrelated `turbo:before-fetch-request` events from link prefetching on other elements, causing random test failures.

- Fix rendering test that was failing due to execution context being destroyed during full page reload. Use `waitForURL` instead of `nextEventNamed` since the tracked element mismatch causes a real browser reload.

- Fix `pausable_requests` test race condition by using a single dialog handler that collects all messages, rather than registering handlers sequentially which could miss dialogs on Firefox.

This was mostly Claude work with some direction from me 😅


packagethief (approved):
Thanks @rosa! Confirmed that all these tests pass for me locally ✅

PR #1501, opened 2026-02-10
Preserve noscript elements, remove only stylesheet children (#1497)

### LINKED ISSUE #1497: `noscript` tags get deleted when using turbo
[This PR](https://github.com/hotwired/turbo/issues/1464) introduced a behavior that isn't helpful for everyone. We use noscript tags to define elements that we want to lazy load when the user scrolls down. We use noscript because it should be displayed immediately when JavaScript is disabled. This change completely breaks this.

yujiteshima:
## Summary

Fixes #1497

The previous fix for #1464 (PR #1475) removed all `<noscript>` elements entirely during Turbo navigation to prevent unintended style evaluation. While this solved the style issue, it also destroyed the elements and their DOM structure, breaking JavaScript use cases that rely on `<noscript>` content being present in the DOM.

A common example is lazy loading with `IntersectionObserver`: JavaScript watches elements with a `data-lazy-load` attribute, and when they enter the viewport, reads the `innerHTML` of a child `<noscript>` to extract and insert the actual content (e.g., `` tags). This is a well-established pattern used by libraries such as lazySizes and WordPress plugins like WP Rocket. It works because `<noscript>` content is not rendered when JavaScript is enabled, but remains accessible via `innerHTML` — providing both a progressive enhancement fallback and a mechanism for deferred loading.

### Why full stringification doesn't work either

The initial approach in PR #1475 (before the review change in `86e8efb`) converted `<noscript>` content from DOM to text via `noscriptElement.textContent = noscriptElement.innerHTML`. While this preserves the `<noscript>` element itself, it flattens the entire DOM tree inside it into a plain text node. JavaScript that expects to traverse or query child elements — or that relies on the original HTML structure in `innerHTML` — would break as well.

### How this fix works

When Turbo fetches a page, the HTML is parsed via `new DOMParser().parseFromString()`, which creates a non-browsing context. In this context, `<noscript>` children are parsed as real DOM elements (not hidden text). After `document.adoptNode()` transfers the parsed body into the live document, those child elements become accessible — and problematic ones like `<style>` and `<link rel="stylesheet">` can get evaluated by the browser, which was the original bug in #1464.

This fix takes a targeted approach: instead of removing the entire `<noscript>` element or converting its content to text, it removes only `<style>` and `<link rel="stylesheet">` children from within `<noscript>` elements. Since these are the specific elements that browsers evaluate after adoption, removing them is sufficient to prevent the unintended style application. The `<noscript>` elements themselves and their remaining children (images, etc.) stay intact in the DOM, preserving access for any JavaScript that relies on their content.

### Changes

- **`page_renderer.js`**: Rename `removeNoscriptElements()` → `deactivateNoscriptStylesheetElements()`, only remove style-related children instead of the entire element
- **`page_snapshot.js`**: Apply the same targeted removal in `clone()` for cached snapshots
- **New test fixture and tests**: Verify that `<noscript>` elements with non-style content are preserved after navigation, and that mixed-content `<noscript>` elements have only stylesheets removed

### DIFF
--- a/src/core/drive/head_snapshot.js
+++ b/src/core/drive/head_snapshot.js
@@ -1,3 +1,4 @@
+import { elementIsStylesheet } from "../../util"
 import { Snapshot } from "../snapshot"
 
 export class HeadSnapshot extends Snapshot {
@@ -93,11 +94,6 @@ function elementIsScript(element) {
 function elementIsNoscript(element) {
   const tagName = element.localName
   return tagName == "noscript"
-}
-
-function elementIsStylesheet(element) {
-  const tagName = element.localName
-  return tagName == "style" || (tagName == "link" && element.getAttribute("rel") == "stylesheet")
 }
 
 function elementIsMetaElementWithName(element, name) {
--- a/src/core/drive/page_renderer.js
+++ b/src/core/drive/page_renderer.js
@@ -1,4 +1,4 @@
-import { activateScriptElement, waitForLoad } from "../../util"
+import { activateScriptElement, elementIsStylesheet, waitForLoad } from "../../util"
 import { Renderer } from "../renderer"
 
 export class PageRenderer extends Renderer {
@@ -180,7 +180,7 @@ export class PageRenderer extends Renderer {
   deactivateNoscriptStylesheetElements() {
     for (const noscriptElement of this.newElement.querySelectorAll("noscript")) {
       for (const child of [...noscriptElement.children]) {
-        if (child.localName === "style" || (child.localName === "link" && child.getAttribute("rel") === "stylesheet")) {
+        if (elementIsStylesheet(child)) {
           child.remove()
         }
       }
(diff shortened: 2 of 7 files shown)


yujiteshima:
The frame_tests.js:184 failure on Firefox is a known flaky test caused by a race condition in the page.once("pageerror") pattern — unrelated to this PR's changes. This is already addressed in #1487, which replaces it with Promise.all([page.waitForEvent("pageerror"), page.click(...)]). The two other flaky tests (form_submission_tests.js:1019, link_prefetch_observer_tests.js:221) are also pre-existing timing issues. All rendering_tests.js tests (including the two new ones added in this PR) pass on both Chrome and Firefox.

packagethief (commented):
Thanks for taking another look at this, @yujiteshima.

packagethief (approved):

joker-777:
@packagethief I would be grateful if you could create a new release. This fix is quite important for us.

PR #1503, opened 2026-02-11
Fix activateScriptElement to set the CSP nonce correctly

### LINKED ISSUE #1502: Function activateScriptElement overwrite CSP nonce with empty string causing violation
There is an issue with the function `activateScriptElement` here that I could trace it from a CSP violation in my app. It has been called from `PageRenderer` function `copyNewHeadScriptElements`.

https://github.com/hotwired/turbo/blob/055eb8f3c1b79f1db22d241b6cb84820d15f7662/src/util.js#L1-L15

The issue is that when creating the `script` element, the code correctly assigns the nonce to the script element using the `cspNonce` variable that is correctly computed. However, later on the `copyElementAttributes` overrides the nonce with an empty string because the DOM attribute is an empty string (`element.getAttribute('nonce')` is an empty string).

This is because [accessing the nonce via the DOM returns an empty string for security reasons](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/nonce#accessing_nonces_and_nonce_hiding). But at this point, the HTML has probably passed multiple stages and even `element.nonce` is an empty string.

The solution would be to apply the nonce after copying element attributes:

```diff
 export function activateScriptElement(element) { 
    if (element.getAttribute("data-turbo-eval") == "false") { 
      return element 
    } else { 
      const createdScriptElement = document.createElement("script") 
      const cspNonce = getCspNonce() 
-     if (cspNonce) { 
-       createdScriptElement.nonce = cspNonce 
-     } 
      createdScriptElement.textContent = element.textContent 
      createdScriptElement.async = false 
      copyElementAttributes(createdScriptElement, element) 
+     if (cspNonce) { 
+       createdScriptElement.nonce = cspNonce 
+     } 
      return createdScriptElement 
    } 
  } 
```

mildred:
The `activateScriptElement` function did not correctly set the CSP nonce on occasions. When the element to duplicate has been stripped from its CSP nonce in `elementWithoutNonce`, the nonce attribute is set to an empty string and would overwrite the correct nonce.

Fix #1502

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -4,12 +4,12 @@ export function activateScriptElement(element) {
   } else {
     const createdScriptElement = document.createElement("script")
     const cspNonce = getCspNonce()
-    if (cspNonce) {
-      createdScriptElement.nonce = cspNonce
-    }
     createdScriptElement.textContent = element.textContent
     createdScriptElement.async = false
     copyElementAttributes(createdScriptElement, element)
+    if (cspNonce && !createdScriptElement.nonce) {
+      createdScriptElement.nonce = cspNonce
+    }
     return createdScriptElement
   }
 }


edariedl:
Hi @jorgemanrubia, are there any plans to merge this PR? We have run into the same problem with CSP implementation when using Turbo in our project.

PR #1509, opened 2026-02-25
Do not add post form body to query params if form data is only set via event

lukasbableck:
In a scenario where you have an empty form, so no inputs and no `name` attribute on the submitter, and you're setting the form data via the `turbo:submit-start` event (with `event.detail.formSubmission.fetchRequest.body.append`), you're actually adding the form data to the URL objects `searchParams`, as `resource.searchParams` is used for the body. 
This causes the URL to contain all of the form data, which is especially problematic for larger amounts of data, as the webserver then might just close the connection because the URL is too long. In my case this happened after the URL reached about 7.5KB.
This PR fixes the issue, while also simplifying the logic a bit. 

/cc @m-vo

### DIFF
--- a/src/http/fetch_request.js
+++ b/src/http/fetch_request.js
@@ -213,8 +213,7 @@ export function isSafe(fetchMethod) {
 }
 
 function buildResourceAndBody(resource, method, requestBody, enctype) {
-  const searchParams =
-    Array.from(requestBody).length > 0 ? new URLSearchParams(entriesExcludingFiles(requestBody)) : resource.searchParams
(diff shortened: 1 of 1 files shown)


m-vo (approved):
The fixed bug is really subtle as most of the time you won't realize that you're working with the reference of the actual browser URL in a `turbo:submit-start` event. 👍 

Thanks for the fix @lukasbableck - removing the case for when the body length is 0, also makes the code much easier to understand.

ausi (approved):

fritzmg (approved):

PR #1512, opened 2026-03-03
Fix unhandled promise rejection when prefetch request fails

sskirby:
PrefetchCache.putLater calls request.perform() without catching errors. FetchRequest.perform() re-throws non-AbortErrors after delegating.

In Safari, navigating away from a page while a prefetch is in-flight causes the browser to cancel the fetch with a TypeError instead of an AbortError. Since only AbortErrors are silently handled by FetchRequest, this surfaces as an unhandled rejection in Safari/iOS:

  TypeError: Unhandled Promise Rejection: Load failed

Catch the rejection and evict the stale cache entry so that a fresh request is made when the user clicks the link.

This failure mode is also apparent in Chrome when prefetching while "Offline"

### DIFF
--- a/src/core/drive/prefetch_cache.js
+++ b/src/core/drive/prefetch_cache.js
@@ -14,7 +14,7 @@ class PrefetchCache extends LRUCache {
 
   putLater(url, request, ttl) {
     this.#prefetchTimeout = setTimeout(() => {
-      request.perform()
+      request.perform().catch(() => this.evict(toCacheKey(url)))
       this.put(url, request, ttl)
       this.#prefetchTimeout = null
     }, this.prefetchDelay)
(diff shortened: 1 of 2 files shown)


sskirby:
I am closing this PR because there are many more cases where reloading the page/navigating away will cause a "TypeError: Unhandled Promise Rejection: Load failed" in Safari. Most of which are very difficult to test for even if the project did have a suite of tests for Safari. In addition, the approach taken here has the risk of suppressing many other, unrelated, errors from error reporting tools like Sentry.

Another option would be to ignore `TypeError`s with the message "Load failed" in the `FetchRequest#perform` method, but 1) this is Safari-specific and 2) Safari throws this same error for cases like CORS failures and SSL certificate errors which individual projects may not want suppressed.

formigarafa:
I am looking on a solution along the same lines.
If the event turbo:before-fetch-response were used along the prefetch cache then it would be possible to add a handler for each of the cases.
but even the prefetch cache instance reference is hardcoded which makes it harder to even explore alternatives without a full fork.

PR #1518, opened 2026-03-15
Fix findLinkFromClickTarget for SVG elements (fixes #1510)

### LINKED ISSUE #1510: `findLinkFromClickTarget` breaks for SVG `<a>` elements (SVGAnimatedString has no startsWith)
It looks like PR https://github.com/hotwired/turbo/pull/1285 introduced a regression affecting SVG links (`<a>` elements inside `<svg>`).

Current implementation:
```js
export function findLinkFromClickTarget(target) {
  const link = findClosestRecursively(target, "a[href], a[xlink\\:href]")

  if (!link) return null
  if (link.href.startsWith("#")) return null
  if (link.hasAttribute("download")) return null

  const linkTarget = link.getAttribute("target")
  if (linkTarget && linkTarget !== "_self") return null

  return link
}
```

Introduced change:
```diff
+ if (link.href.startsWith("#")) return null
```

`SVGAElement` does not expose `href` as a string, but as an `SVGAnimatedString`. Which does not implement `startsWith()`.
However, both have access to `getAttribute()`, and `getAttribute('href')` consistently returns a string for both `HTMLAnchorElement` and `SVGAElement`.

Danish-Belal:
SVGAElement exposes href as SVGAnimatedString which does not implement startsWith(). Use getAttribute('href') which returns a string for both HTMLAnchorElement and SVGAElement.

Add tests for clicking SVG links (same-origin and hash-only).

### DIFF
--- a/src/tests/functional/link_prefetch_observer_tests.js
+++ b/src/tests/functional/link_prefetch_observer_tests.js
@@ -121,6 +121,11 @@ test("it doesn't prefetch the page when link has a hash as a href", async ({ pag
   await assertNotPrefetchedOnHover({ page, selector: "#anchor_with_hash" })
 })
 
+test("it doesn't prefetch when hovering SVG link with hash href", async ({ page }) => {
+  await goTo({ page, path: "/hover_to_prefetch.html" })
+  await assertNotPrefetchedOnHover({ page, selector: "#anchor_with_hash_inside_svg" })
+})
+
 test("it doesn't prefetch the page when link has a ftp protocol", async ({ page }) => {
   await goTo({ page, path: "/hover_to_prefetch.html" })
   await assertNotPrefetchedOnHover({ page, selector: "#anchor_with_ftp_protocol" })
--- a/src/tests/functional/link_prefetch_observer_tests.js
+++ b/src/tests/functional/link_prefetch_observer_tests.js
@@ -121,7 +121,7 @@ test("it doesn't prefetch the page when link has a hash as a href", async ({ pag
   await assertNotPrefetchedOnHover({ page, selector: "#anchor_with_hash" })
 })
 
-test("it doesn't prefetch when hovering SVG link with hash href", async ({ page }) => {
+test("it doesn't prefetch the page when SVG link has a hash as a href", async ({ page }) => {
   await goTo({ page, path: "/hover_to_prefetch.html" })
   await assertNotPrefetchedOnHover({ page, selector: "#anchor_with_hash_inside_svg" })
 })
--- a/src/tests/functional/navigation_tests.js
+++ b/src/tests/functional/navigation_tests.js
@@ -310,22 +310,19 @@ test("following a cross-origin link inside an SVG element", async ({ page }) => 
   expect(await visitAction(page)).toEqual("load")
 })
 
-test("clicking a same-origin link inside an SVG element", async ({ page }) => {
+test("following a same-origin SVG link", async ({ page }) => {
   await page.click("#same-origin-link-inside-svg-element")
 
   await expect(page).toHaveURL(withPathname("/src/tests/fixtures/one.html"))
   expect(await visitAction(page)).toEqual("load")
 })
 
-test("clicking an SVG link with hash-only href scrolls to anchor without a visit", async ({ page }) => {
-  expect(
-    await willChangeBody(page, async () => {
-      await page.click("#svg-hash-link")
-    })
-  ).not.toBeTruthy()
+test("following a same-origin SVG anchored link", async ({ page }) => {
+  await page.click("#same-origin-anchored-svg-link")
 
   await expect(page).toHaveURL(withPathname("/src/tests/fixtures/navigation.html"))
   await expect(page).toHaveURL(withHash("#main"))
+  expect(await visitAction(page)).toEqual("load")
   expect(await isScrolledToSelector(page, "#main"), "scrolled to #main").toEqual(true)
 })
 
--- a/src/tests/functional/navigation_tests.js
+++ b/src/tests/functional/navigation_tests.js
@@ -310,6 +310,25 @@ test("following a cross-origin link inside an SVG element", async ({ page }) => 
   expect(await visitAction(page)).toEqual("load")
 })
 
+test("clicking a same-origin link inside an SVG element", async ({ page }) => {
+  await page.click("#same-origin-link-inside-svg-element")
+
+  await expect(page).toHaveURL(withPathname("/src/tests/fixtures/one.html"))
+  expect(await visitAction(page)).toEqual("load")
+})
+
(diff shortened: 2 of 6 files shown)


src/util.js
```diff
@@ -249,11 +249,19 @@ export function doesNotTargetIFrame(name) {
...
 }
 
+/**
+ * Returns href as string for both HTMLAnchorElement and SVGAElement.
+ * SVGAElement exposes href as SVGAnimatedString which has no startsWith().
+ */
+export function getLinkHrefString(link) {
+  return link.getAttribute("href") ?? link.getAttribute("xlink:href") ?? ""
+}
+
```
Samuel-Labagnere, line 261: I feel like the comment could be a little more generic.
It does indeed resolve the consistency problem regarding HTMLAnchorElements and SVGAElements. But it does also fix the simple problem that checking `if (link.href.startsWith("#"))` (as in line 264 of this very file) will always be false because `HTMLAnchorElement#href` returns the final target url rather than what the attribute truly contains.

Would you agree with the following example ?
```js
/**
 * Returns consistently the href attribute value as a string for both HTMLAnchorElement and SVGAElement.
 * HTMLAnchorElement href property returns an absolute URL if the attribute contains a valid relative URL.
 * SVGAElement exposes href as SVGAnimatedString which does not implement String methods.
 * getAttribute() will return the proper value of the attribute in both cases.
 */
export function getLinkHrefString(link) {
  return link.getAttribute("href") ?? link.getAttribute("xlink:href") ?? ""
}
```

> [!NOTE]
> **__Example regarding the difference between `.href` and `.getAttribute("href")`:__**
> If the current website domain is `www.example.org`, you are on the root page `/` and the link is built like `<a id="my-link" href="#some-id">...</a>`.
>
> If you test `document.getElementById("my-link").href`, it will return `https://www.example.org/#some-id` rather than simply `#some-id`. Therefore checking `startsWith("#")` will never work as intended.

src/tests/functional/navigation_tests.js
```diff
@@ -310,6 +310,25 @@ test("following a cross-origin link inside an SVG element", async ({ page }) =>
...
+  expect(
+    await willChangeBody(page, async () => {
+      await page.click("#svg-hash-link")
+    })
+  ).not.toBeTruthy()
+
+  await expect(page).toHaveURL(withPathname("/src/tests/fixtures/navigation.html"))
+  await expect(page).toHaveURL(withHash("#main"))
+  expect(await isScrolledToSelector(page, "#main"), "scrolled to #main").toEqual(true)
+})
```
Samuel-Labagnere: We could try to conventionally stick to how similar tests from this file are written, like `"following a same-origin anchored link"` line 229.

Would this work ?
```js
test("following a same-origin SVG anchored link", async ({ page }) => {
  await page.click("#same-origin-anchored-svg-link") // Implies we should rename the link id as well in the fixture file

  await expect(page).toHaveURL(withPathname("/src/tests/fixtures/navigation.html"))
  await expect(page).toHaveURL(withHash("#main"))
  expect(await visitAction(page)).toEqual("advance")
  expect(await isScrolledToSelector(page, "#main")).toEqual(true)
})
```

src/tests/functional/navigation_tests.js
```diff
@@ -310,6 +310,25 @@ test("following a cross-origin link inside an SVG element", async ({ page }) =>
   expect(await visitAction(page)).toEqual("load")
 })
 
+test("clicking a same-origin link inside an SVG element", async ({ page }) => {
```
Samuel-Labagnere: If we try to stay close to how other tests are written, maybe we should rename it `"following a same-origin SVG link` ?

src/tests/functional/link_prefetch_observer_tests.js
```diff
@@ -121,6 +121,11 @@ test("it doesn't prefetch the page when link has a hash as a href", async ({ pag
   await assertNotPrefetchedOnHover({ page, selector: "#anchor_with_hash" })
 })
 
+test("it doesn't prefetch when hovering SVG link with hash href", async ({ page }) => {
```
Samuel-Labagnere: To reflect on the test right above, should we aim for `"it doesn't prefetch the page when SVG link has a hash as a href"` ?

Samuel-Labagnere (commented):
Great PR overall, lgtm. 👍🏻 
Just added a few comments towards harmonizing your texts with how things are written elsewhere in the project.

Danish-Belal:
Thanks for the review, @Samuel-Labagnere . I’ve addressed all your feedback:

1. getLinkHrefString comment (util.js)
Updated the JSDoc to cover both element types and the difference between .href and getAttribute("href"):

2. Navigation tests

Renamed "clicking a same-origin link inside an SVG element" → "following a same-origin SVG link"
Renamed "clicking an SVG link with hash-only href scrolls to anchor without a visit" → "following a same-origin SVG anchored link"
Renamed fixture link id from #svg-hash-link to #same-origin-anchored-svg-link
Simplified the SVG anchored link test to follow the same pattern as "following a same-origin anchored link"
3. Prefetch test
Renamed "it doesn't prefetch when hovering SVG link with hash href" → "it doesn't prefetch the page when SVG link has a hash as a href" to match the test above.

All changes are pushed. Please let me know if anything else needs adjustment.

src/tests/functional/link_prefetch_observer_tests.js
Danish-Belal: Updated as suggested

src/tests/functional/navigation_tests.js
Danish-Belal: Done

src/tests/functional/navigation_tests.js
Danish-Belal: Done

Danish-Belal:
@Samuel-Labagnere  Looking forward you to review and getting merge my PR.
Thank you.

Samuel-Labagnere (approved):

Samuel-Labagnere:
@Danish-Belal So far so good to me, now it's in the maintainers hands.

Danish-Belal:
@packagethief 
Looking forward for your action

BriceFab:
I've the same issue, thank for the PR. Waiting to be merged

gamecreature:
Same issue here, it seems to be a regression, it worked correctly before.
Please merge this PR. @Danish-Belal Thanks for fixing this!

A workaround that works for me right now, is to replace the A tags with G tags and performing an onclick turbo-frame navigation. (Turbo.visit didn't work for me from the svg, also disable turbo didn't seem to work). 

```svg
<g href="/my-url" data-turbo-frame="my-frame" style="cursor: pointer;"
   onclick="event.preventDefault(); document.querySelector(`#${this.getAttribute('data-turbo-frame')}`).src = this.getAttribute('href'); return false;"  
   >..</g>
```

seanpdoyle (approved):

PR #1546, opened 2026-06-23
Handle `InvalidStateError` during transitions

danielvdao:
## Context

During turbo stream refreshes the `ViewTransition.ready` promise is never handled. Turbo only care about the page itself finishing, hence the `.finished` call that happens. I imagine this can happen with anything that calls transitions though.

In these cases a user can start the transition via refresh, click on another tab, and the transition won't happen as the `ready` promise fails. The page _does_ render correctly though as we call `.finished`. 

However when this occurs, the promise _should_ be handled some way otherwise an error is thrown (`InvalidStateError`). As a result of the unhandled promise, this error gets sent to `console` as one can imagine that users switch tabs.

## Similar issue
React had a similar issue here: https://github.com/react/react/issues/34098 and their fix was to match on the error message here: https://github.com/react/react/pull/34450.

Note: I used a blunt hammer and did a catch-all but open to trying something else too as I'm new to this codebase.

### DIFF
--- a/src/core/drive/view_transitioner.js
+++ b/src/core/drive/view_transitioner.js
@@ -6,7 +6,9 @@ export class ViewTransitioner {
     if (useViewTransition && this.viewTransitionsAvailable && !this.#viewTransitionStarted) {
       this.#viewTransitionStarted = true
       this.#lastOperation = this.#lastOperation.then(async () => {
-        await document.startViewTransition(render).finished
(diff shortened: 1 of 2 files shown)


danielvdao:
Hello @jorgemanrubia I see that you're a maintainer, I was wondering if you are able to review this PR or know if I should wait to get eyes on it? Thank you.

